PR #13, opened 2017-02-24
Update compile scripts to clean lib before building updates

### LINKED ISSUE #10: Add task to remove lib directory before compile
`npm run compile` will transpile our js and copy files into the lib directory. Old files will persist unless we remove them. Before compiling, we should clean out the lib directory.

bjankord:
Resolves #10


Matt-Butler:
+1

PR #18, opened 2017-02-24
Add prepush script

Matt-Butler:
```"prepush": "npm test && npm run compile && git diff --exit-code packages/*/lib/*"```

`npm test` - test and lint all packages.

`npm run compile` - transpile all packages so es5 code is up to date.

`git diff --exit-code packages/*/lib/*"` - check if there is a diff in any of the packages lib directories. If there is an uncommitted change in a package lib directory, then it will exit before pushing. 

The prepush script will automatically work after running `npm install`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "start": "webpack-dev-server --hot --inline --display-error-details --progress --colors --host 0.0.0.0",
     "webpack": "webpack --config webpack.prod.config.js -p --progress --profile --json > build/stats.json",
     "deploy": "npm run webpack && gh-pages -d build",
-    "prepush": "npm test"
+    "prepush": "npm test && npm run compile && git diff --exit-code packages/*/lib/*"
   },
   "devDependencies": {
     "autoprefixer": "^6.6.1",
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by Matt-Butler** [comment]
https://github.com/cerner/terra-ui/issues/5

**Comment by Matt-Butler** [comment]
no_changes

**Comment by Matt-Butler** [comment]
*Change in lib*
change

File: package.json [hunk 1]
```diff
@@ -44,7 +44,8 @@
     "test": "lerna exec --concurrency 1 npm test",
     "start": "webpack-dev-server --hot --inline --display-error-details --progress --colors --host 0.0.0.0",
     "webpack": "webpack --config webpack.prod.config.js -p --progress --profile --json > build/stats.json",
-    "deploy": "npm run webpack && gh-pages -d build"
+    "deploy": "npm run webpack && gh-pages -d build",
+    "prepush": "npm test && npm run compile && git diff --exit-code packages/*/lib/*"
```
Matt-Butler: https://git-scm.com/docs/git-diff#git-diff---exit-code

bjankord:
+1, I think it would be nice to give an error noting the lib files have changed and need to be checked in. Right now you have to be familiar with whats going on with this task to understand the output if there is a diff in the lib file, but this is not a blocker for me. We can look at adding something like that later.

PR #26, opened 2017-02-27
Add postinstall script to boostrap project after npm install is run

bjankord:
### Summary
This addition will run `lerna bootsrap` after `npm install` is run.

When run, this command will:
`npm install` all external dependencies of each package.
Symlink together all Lerna packages that are dependencies of each other.
`npm prepublish` all bootstrapped packages.

[More info about the bootstrap command can be found here](https://github.com/lerna/lerna/#bootstrap)

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "lint:scss": "lerna run lint:scss",
     "jest": "jest",
     "jest:coverage": "jest --coverage",
+    "postinstall": "npm run bootstrap",
     "test": "lerna exec --concurrency 1 npm test",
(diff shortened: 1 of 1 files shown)


Matt-Butler:
+1 

Just a heads up, if another package has terra-ui as a dependency with github reference then this postinstall command will execute in that package.

bjankord:
@Matt-Butler Good point, if this becomes an issue, we can switch the `postinstall` script to something like `post:install` and then manually call that script in our travis.yml file so it doesn't automatically run on `npm install`

PR #38, opened 2017-02-28
Improve night watch test runs.

ryanthemanuel:
### Summary
I'd like to improve how the night watch scripts are run.  This is a first attempt.  Unfortunately, Safari doesn't like being run in parallel at the moment as far as I can tell, so I couldn't extend the pattern to that browser.  When I get sauce labs and IE up and running things will change again.


scripts/run_nightwatch.js
```diff
@@ -0,0 +1,12 @@
+#! /usr/local/bin/node
+
+/* eslint import/no-extraneous-dependencies: ["error", {"devDependencies": true}] */
+const shell = require('shelljs');
+
+const driver = process.argv[2];
```
Matt-Butler: Can we make this
```
const driver = process.argv[2] || 'default';
```

If no param is passed, it assigns default

Matt-Butler:
For the prepush script, I made the directory like this below. The readme has information about the script.
/prepush
  /index.js
  /README.md

I think we should be consistent. How do you guys want to structure this?

scripts/run_nightwatch.js
ryanthemanuel: I like this idea.  Will make this change.

ryanthemanuel:
Would it make sense to have a folder for each script as well as a ReadMe?  I think that's a good start at least.

scripts/run_nightwatch.js
Matt-Butler: Cool, thanks @ryanthemanuel

ryanthemanuel:
I updated the structure of the test script and actually added another for releases.  I added ReadMe's as well.  Let me know what you guys think.

bjankord:
@ryanthemanuel lgtm 👍

scripts/release/release.js
```diff
@@ -0,0 +1,10 @@
+/* eslint import/no-extraneous-dependencies: ["error", {"devDependencies": true}] */
+const shell = require('shelljs');
+
+const releaseType = process.argv[2] || 'patch';
+
+if (shell.exec(`npm test && npm run compile && npm version ${releaseType} -m "Released version %s" && npm publish && git push`).code !== 0) {
```
Matt-Butler: @bjankord found an issue when running `npm test` inside of shell.exec which allows a false positive.  The pretest script will kick off linting. Linting can fail, but npm test still returns a exit-code of 0 from tests. 

If we take npm test out of shelljs, it will fail on linting errors.

ryanthemanuel:
@Matt-Butler I made some changes.  Is that what you were thinking to fix the issue you were mentioning?

PR #69, opened 2017-03-01
add terra-modal package

### LINKED ISSUE #64: Migrate terra-modal into repo
https://github.com/cerner/terra-modal

zhongyn:
### Summary
Added terra-modal package.
After the review of /src is done, will push /test for review and add more examples

This PR resolves #64 

@cerner/terra


packages/terra-modal/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "release:patch": "npm test && node ../../scripts/release/release.js patch",
+    "test": "$(cd ..; npm bin)/jest --config ../../jestconfig.json"
+  },
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "terra-button": "file:../terra-button",
```
Matt-Butler: With lerna, we will have no relative dependencies and no github dependencies. 

Lerna bootstrap will create symlinks between different packages in this repo: https://github.com/lerna/lerna#bootstrap

Matt-Butler: Also, I think terra-button, terra-legacy-theme, and terra-mixins should be a devDependency,

packages/terra-modal/src/modal.scss
```diff
@@ -0,0 +1,81 @@
+@import 'node_modules/terra-legacy-theme/src/terra-legacy-theme';
```
Matt-Butler: We cannot import a theme in any package because we need to be able to switch it out for a different theme. The webpack configuration. 

Currently, in the webpack configuration a consumer will specify the theme: https://github.com/cerner/terra-ui/blob/master/webpack.config.js#L42-L44

This theme file will be loaded before any other scss file.

Matt-Butler:
rm `packages/terra-modal/yarn.lock`
rm `yarn.lock`

packages/terra-modal/examples/ModalWrapper.jsx
```diff
@@ -0,0 +1,75 @@
...
+  header: PropTypes.node,
+  body: PropTypes.node,
+  footer: PropTypes.node,
+  canCloseOnBackdropClick: PropTypes.bool,
+  size: PropTypes.string,
+  hasCloseButton: PropTypes.bool,
+};
+
+const defaultProps = {
+  body,
```
Matt-Butler: Just curious, what does this do?

packages/terra-modal/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "terra-modal",
+  "main": "lib/Modal.js",
+  "private": true,
+  "version": "0.1.0",
+  "description": "Modals are dialog prompts containing additional information to be revealed to the user. Modals consists of an backdrop, a modal dialog with a header and body, and a close button.",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.cerner.com/terra/terra-ui.git"
```
Matt-Butler: `git+https://github.cerner.com/terra/terra-ui.git`

`git+https://github.com/cerner/terra-ui.git`

packages/terra-modal/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "Cerner",
+    "Terra",
+    "terra-modal",
+    "Modal",
+    "UI"
+  ],
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.cerner.com/terra/terra-ui/issues"
```
Matt-Butler: `"https://github.com/cerner/terra-ui/issues"`

packages/terra-modal/examples/ModalWrapper.jsx
zhongyn: just es6 syntax sugar of property shorthand: http://es6-features.org/#PropertyShorthand

Matt-Butler: Cool, I didn't know you could do that.

packages/terra-modal/src/modal.scss
```diff
@@ -0,0 +1,79 @@
...
+  background-color: #1c1f21;
+  border: 0;
+  height: 100%;
+  left: 0;
+  opacity: 0.8;
+  padding: 0;
+  position: fixed;
+  top: 0;
+  width: 100%;
+  z-index: 1000;
```
zhongyn: will update this when this pull request is merged: https://github.com/cerner/terra-legacy-theme/pull/12.
or if the <Overlay /> component is completed, we can swap the backdrop with overlay

packages/terra-modal/src/modal.scss
```diff
@@ -0,0 +1,79 @@
...
+  background-color: #fff;
+  border-radius: 0.25em;
+  left: 0;
+  margin: 0 auto;
+  outline: 0;
+  position: fixed;
+  right: 0;
+  top: 5%;
+  width: 85%;
+  z-index: 1001;
```
zhongyn: will update this when this pull request is merged: cerner/terra-legacy-theme#12.

packages/terra-modal/src/Portal.jsx
```diff
@@ -0,0 +1,51 @@
...
+   * Whether or not the children of <Portal/> should be rendered
+   */
+  isOpen: PropTypes.bool,
+};
+
+const defaultProps = {
+  isOpen: false,
+};
+
+class Portal extends React.Component {
```
tbiethman: Can we investigate the use of something like `react-portal` directly? https://github.com/tajo/react-portal

I'd be happy to maintain less code if possible.

packages/terra-modal/src/Dialog.jsx
```diff
@@ -0,0 +1,119 @@
...
+      this.setState({ bodyHeight: window.innerHeight - headerHeight - footerHeight });
+    } else {
+      this.setState({ bodyMaxHeight: (window.innerHeight * 0.85) - headerHeight - footerHeight });
+    }
+  }
+
+  render() {
+    const bodyStyle = {
+      maxHeight: this.state.bodyMaxHeight,
+      height: this.state.bodyHeight,
```
tbiethman: Are maxHeight and height mutually exclusive?

tbiethman: Is there any way for a consumer to provide a different modal height?

zhongyn: currently the height can't be customized. We can add a height prop and enable that

tbiethman: Since setState does a shallow merge of the new state object with the previous state object, I think that if for whatever reason the isFullscreen prop were to be updated for the same instance, both bodyHeight and bodyMaxHeight would be present in the state object. We could either clear the state in the componentWillReceiveProps function and then call this function, or explicitly remove the necessary value when this function is called.

zhongyn: good catch, will update

zhongyn: @tbiethman currently if using the portal to render a modal, once a modal prop is updated in whatever way, it will create a new instance and mount it to body. In that case, is it necessary to remove the value? Or do you have any example that the prop of the same instance will be updated?

packages/terra-modal/src/Portal.jsx
zhongyn: the custom portal here handles open/close by a prop isOpen, while react-portal is more comprehensive, it supports other props besides:
- openByClickOn (e.g. <button/>, so user don't need to assign a callback to the button), 
- closeOnEsc (which is handled by <Modal/> in this PR)
- closeOnOutsideClick (which is handled by <Backdrop/> in this PR)
- callbacks: onOpen/onClose/onUpdate/beforeClose

Basically it depends on in which level we want to handle onRequestClose event and do we need callbacks in the portal lifecycle

packages/terra-modal/src/Portal.jsx
```diff
@@ -0,0 +1,51 @@
...
+      this.renderContent(this.props);
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (!this.props.isOpen && nextProps.isOpen) {
+      this.renderContent(nextProps);
+    } else if (this.props.isOpen && !nextProps.isOpen) {
+      this.unrenderContent();
+    }
```
zhongyn: need to add this to handle any prop update
```
    } else if (this.props.isOpen && nextProps.isOpen) {
      this.unrenderContent();
      this.renderContent(nextProps);
    }
```

packages/terra-modal/src/Modal.scss
```diff
@@ -0,0 +1,81 @@
+@import '~terra-mixins';
+
+.terra-Modal-backdrop {
+  background-color: #1c1f21;
+  border: 0;
+  height: 100%;
+  left: 0;
```
Matt-Butler: Is this bidirectional?

zhongyn: The backdrop cover the full screen so it's bidi here

Matt-Butler:
@zhongyn, When you have time, can you fix the lint errors?

zhongyn:
@Matt-Butler yeah, will fix the lint error. Also, the CI report includes Button nightwatch error. Is it still in progress?

packages/terra-modal/src/Dialog.jsx
zhongyn: @tbiethman updated with height prop and explicitly set the maxHeight and height mutually exclusive

packages/terra-modal/.babelrc
```diff
@@ -0,0 +1,4 @@
+{
+  "presets": ["es2015", "react"],
+  "plugins": ["transform-object-assign", "transform-object-rest-spread", "transform-runtime"],
```
ryanthemanuel: Transform-runtime is necessary to support Set.  I tried to scope this plugin just to terra-modal but I couldn't scope it just to transform Set.  Unfortunately it does a few other things as well.  Are we OK with this solution or should we investigate an alternative to Set?

zhongyn: @ryanthemanuel Don't worry. The config is migrated from the old module repo. Since terra-modal will depend on terra-overlay. I will close this PR and send a new one for modal after overlay is available in master.

zhongyn:
Close this PR. Will create a new one after overlay is available.

PR #93, opened 2017-03-07
Sauce labs

ryanthemanuel:
### Summary
These are the necessary changes to run tests against sauce labs.  I've also refactored the night watch scripts a little bit to be more efficient and logical location-wise.  This might require a change in the generator as well unfortunately, but I think it's for the best in the long run.  I'm willing to make any changes that are necessary for existing projects myself.  To actually run against sauce labs you will need to have my access key which I would prefer not to make public.  We'll have to discuss a way to share it amongst the team.


packages/terra-toolkit/src/scripts/README.md
```diff
@@ -0,0 +1,15 @@
+#nightwatch
```
ryanthemanuel: I will move this documentation into the README in the docs directory.  As well as add information about update-sauce.js

ryanthemanuel:
In case anyone is curious, here is the output of what gets generated by running against sauce labs for all drivers for terra-button: [nightwatch.zip](https://github.com/cerner/terra-ui/files/826083/nightwatch.zip)

.travis.yml
```diff
@@ -11,3 +11,7 @@ branches:
 cache:
   directories:
     - node_modules
+env:
+  global:
+    - SAUCE_USERNAME=cerner
```
Matt-Butler: Should this be an env variable?

ryanthemanuel: I'm not quite sure what you are asking. This is defining SAUCE_USERNAME as an environment variable with respect to the Travis build.

Matt-Butler: Should this information be secret/internal?

ryanthemanuel: I'm not sure the username needs to (especially since it's so easy to guess). The key on the next line is encrypted with this repo's private key so that will keep anyone from using it outside this repo.

PR #102, opened 2017-03-09
Preparing for the release of terra-toolkit.

ryanthemanuel:
### Summary
I would like to release terra-toolkit.  These changes will support that.


Matt-Butler:
Before we release, we should decouple this from terra-site: https://github.com/cerner/terra-ui/blob/master/packages/terra-toolkit/tests/test.config.js#L5

Matt-Butler:
Should terra-toolkit still work outside the monorepo?

ryanthemanuel:
This package should definitely exist outside of the mono-repo.  SMART is wanting to consume this.  I was giving preferential treatment to the mono repo with the default webpack config.  I'm ok with that being explicitly set per project if we want to.

Matt-Butler:
Thanks @ryanthemanuel, we need to go through and remove all of the relative references for this to work outside the monorepo:

```
"server_path" : "../../node_modules/selenium-server-standalone-jar/jar/selenium-server-standalone-3.2.0.jar",
```

bjankord:
I just noticed this package's README.md is a little out of sync. We should update the main README.md to match the other repos:
* Remove [contributing section](https://github.com/cerner/terra-ui/tree/master/packages/terra-toolkit#contributing) or at minimum update link in this section so it doesn't 404.
* If we remove this section, remove link to [contributing section](https://github.com/cerner/terra-ui/tree/master/packages/terra-toolkit#contributing) in table of contents
* Update [getting started section](https://github.com/cerner/terra-ui/tree/master/packages/terra-toolkit#getting-started) to match other repos [getting started section](https://github.com/cerner/terra-ui/tree/master/packages/terra-button#getting-started)

package.json
```diff
@@ -77,7 +77,7 @@
     "react-dom": "^15.4.2",
     "react-router": "^3.0.2",
     "sauce-connect-launcher": "^1.2.0",
-    "selenium-server-standalone-jar": "3.1.0",
+    "selenium-server-standalone-jar": "3.2.0",
```
Matt-Butler: Since this package is used in `/src` it should be moved to a dependency in the terra-toolkit package. This will ensure that this package is installed with terra-toolkit.

ryanthemanuel: This already is a dependency for terra-toolkit.

ryanthemanuel:
I made the ReadMe changes and made changes to not have hard coded paths for the selenium standalone server jar and phantomjs.  I actually think I'm ok with the tests for terra-toolkit mooching off of terra-site since it is only a testing concern for terra-toolkit and not a run time concern.  If people are super opposed to that though, I can make that change.

package.json
Matt-Butler: Cool, can we remove it from the root package.json then?

Matt-Butler:
Thanks for making those updates @ryanthemanuel 

So is this file only needed for testing?
https://github.com/cerner/terra-ui/blob/master/packages/terra-toolkit/src/server-launcher.js#L3

ryanthemanuel:
Good point.  Removed.

PR #104, opened 2017-03-09
Attempt performance improvements

ryanthemanuel:
### Summary

Don't merge this in.  It's a Travis CI test


packages/terra-toolkit/src/test-settings.js
```diff
@@ -96,6 +96,7 @@ module.exports = (testConfigPath, settings) => {
       testingConfiguration[key].globals = {};
       testingConfiguration[key].globals.width = windowSizes[windowSizeKey][0];
       testingConfiguration[key].globals.height = windowSizes[windowSizeKey][1];
+      testingConfiguration[key].globals.windowSizeKey = windowSizeKey;
```
StephenEsser: +1 on exposing this

dkasper-was-taken:
+1

nightwatch.conf.js
```diff
@@ -0,0 +1,18 @@
+/* eslint import/no-extraneous-dependencies: ["error", {"devDependencies": true}] */
+
+const testSettings = require('./packages/terra-toolkit/lib/index').testSettings;
```
Matt-Butler: We should probably stop using relative paths in the monorepo.

This code _should_ work the same:
```
const testSettings = require('terra-toolkit').testSettings;
```

Matt-Butler: Change if you want, we can always come back and fix this.

ryanthemanuel: Talked with @Matt-Butler offline about this and since this is at the root level, it needs to have a path unfortunately.

PR #120, opened 2017-03-13
Issue101 remove progress bar focus

sid141288:
### Summary
Adding a style to not display the outline when the ProgressBar is clicked on.
Issue: https://github.com/cerner/terra-ui/issues/101
focusissueprogressbarie10

focusissueprogressbarchrome

focusissueprogressbarfirefox

focusissueprogressbaredge

focusissueprogressbarie11


Matt-Butler:
+1

zhongyn:
By removing focus, i think it means non-reachable via sequential keyboard navigation. The change here removes the focus style. But i think it still can be focused. @Matt-Butler @mmalaker  what's the expected behavior?

sid141288:
@zhongyn the tabIndex was set to -1 earlier; that disabled keyboard navigation on the terra-progress-bar. https://github.com/cerner/terra-ui/blob/master/packages/terra-progress-bar/src/ProgressBar.jsx#L55

zhongyn:
So MDN suggest that progress bar should be accessible: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role. But that's the case when using a div with role="progressbar"

zhongyn:
i tested it with chrome and mac native screen reader. Both of them can't recognize the <progress/> element

PR #129, opened 2017-03-16
Initial ResponsiveElement Implementation

StephenEsser:
### Summary
The ResponsiveElement provides a way to conditionally render different elements based on the immediate parent viewport size or the window. 

#### Breakpoints
There is currently no means of obtaining the themed breakpoints into Javascript. This pull request includes a possible solution that follows very closely to the current implementation in legacy Terra. This would likely move to another package, but has been embedded here so the ResponsiveElement can utilize it. The themed breakpoints are added into a psudo SCSS class and loaded into the dom, read back into javascript, and then removed from the dom. This only happens once.

#### Name
The current name is ResponsiveElement, but I catch myself calling it a ResponsiveComponent pretty often. Votes for name are welcome.

#### Resize Observer 
When responding to a parent viewport a resize observer polyfill is used to observe any resize events on that element: https://github.com/que-etc/resize-observer-polyfill
This could easily be swapped out for another observer like: https://github.com/wnr/element-resize-detector

When bound to the window traditional resize events are added. 

#### Awkwardness 
With the current breakpoints there is an awkward gap that makes sense in CSS but not in a React Component. When an element width is between 0px - 543px it considered "default". In CSS this makes sense because these would be the standard styles that get applied to an element, and theming would start at tiny. But in React this forces another size to be added as a prop. A "default" sized prop to be rendered at breakpoints less than tiny. 
```
.class {
  padding: 5px;

  @media tiny {
    // CSS for tiny
  }
}
```

|--default--|--Tiny--|--Small--|--Medium--|--Large--|--Huge--|
0 -------- 544 ---- 768 ---- 992 -------- 1216 ---- 1440

https://github.com/cerner/terra-ui/blob/master/packages/terra-legacy-theme/src/_breakpoints.scss

@cerner/terra


packages/terra-responsive-element/src/ResponsiveElement.scss
```diff
@@ -0,0 +1,3 @@
+.terra-ResponsiveElement {
+  width: 100%;
```
mhemesath: why is this needed on a block level element?

packages/terra-responsive-element/src/ResponsiveElement.jsx
```diff
@@ -0,0 +1,122 @@
...
+    /* eslint-disable no-unused-vars */
+    const { defaultElement, tiny, small, medium, large, huge, responsiveTo, ...customProps } = this.props;
+
+    const className = classNames(
+      'terra-ResponsiveElement',
+      { [`${customProps.className}`]: customProps.className },
+    );
+
+    return (
+      <div {...customProps} ref={this.setContainer} className={className}>
```
mhemesath: NIT comment.. is there a known use case for needing to allow custom class names on this element? Wondering if we don't just limit this to responsive element switching.

StephenEsser: There is no known use case I am aware of. I added in the custom props to follow the conventional standard of our other components to make it consistent. 

Do anyone have thoughts about keeping it / removing it?

packages/terra-responsive-element/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "UI"
+  ],
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-ui/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-ui#readme",
+  "devDependencies": {
+    "resize-observer-polyfill": "^1.4.1",
```
Matt-Butler: `resize-observer-polyfill` is imported in `src`. We need to move this to a dependency so it is installed when an application consumes responsiveElement.

packages/terra-responsive-element/src/ResponsiveElement.jsx
```diff
@@ -0,0 +1,122 @@
...
+import ResizeObserver from 'resize-observer-polyfill';
+import classNames from 'classnames';
+import getBreakpoints from './breakpoints';
+import './ResponsiveElement.scss';
+
+const propTypes = {
+  /**
+   * An element to be displayed at default breakpoints ( Between 0 and tiny )
+  */
+  defaultElement: PropTypes.element,
```
Matt-Butler: Should this be required?

packages/terra-responsive-element/src/breakpoints.js
```diff
@@ -0,0 +1,30 @@
+import '../src/breakpoints.scss';
```
Matt-Butler: `import './breakpoints.scss';`

packages/terra-responsive-element/src/ResponsiveElement.scss
StephenEsser: This was added as a promise / safeguard, mainly for issues we have seen in the past with flexbox. I went ahead and removed the class and CSS. It can be added back if it ever becomes necessary.

packages/terra-responsive-element/src/ResponsiveElement.jsx
bjankord: Let's remove it. We can add it when we have a known need for it.

packages/terra-responsive-element/src/ResponsiveElement.jsx
StephenEsser: I went back and forth on this. This component could be used to hide content at breakpoints smaller than a specific size. If we decide this component forces content to be displayed we can make this required.

packages/terra-responsive-element/src/breakpoints.js
StephenEsser: Nice catch, this may or may not have been initially written a long time ago in a galaxy where scss wasn't yet compiled into the lib folder.

packages/terra-responsive-element/src/ResponsiveElement.jsx
mjhenkes: I like the flexibility of being able to hide content at smaller break points.

packages/terra-responsive-element/src/breakpoints.js
```diff
@@ -0,0 +1,30 @@
...
+  data = window.getComputedStyle(datasource, ':before').getPropertyValue('content');
+  data = data.replace(/^['"]+|\s+|\\|(;\s?})+|['"]$/g, '');
+
+  breakpoints = JSON.parse(data);
+
+  document.body.removeChild(datasource);
+
+  return breakpoints;
+};
+
```
zhongyn: is it possible to import the breakpoints.scss file from terra-legacy-theme as a raw string and then parse that string:
`import breakpoints from 'raw-loader!terra-legacy-theme/breakpoints.scss';`

StephenEsser: We want to read the variables from outside of terra-legacy-theme to account for any theming incase the breakpoints are redefined. I can't think of a way to parse a scss file that reads non-static variables.

zhongyn: +1

packages/terra-responsive-element/src/ResponsiveElement.jsx
mjhenkes: My opinion is that the standard should be to keep custom elements on anything that adds an element to the dom. I vote to keep it in.

packages/terra-responsive-element/src/ResponsiveElement.jsx
bjankord: Same, let's keep this optional.

PR #137, opened 2017-03-16
Update image component

aganapuram:
Image react component

Tech Design: https://github.cerner.com/terra/tech-designs/blob/terra-image/terra-image.md

@cerner/terra


packages/terra-image/src/Image.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+const defaultProps = {
+  variant: 'default',
+  isFluid: false,
+  alt: ' ',
+};
+
+const Image = ({
+  src,
+  variant,
```
Matt-Butler: Can you please rename `variant` to `intent`?

bjankord: I think 'rounded', 'circle', 'thumbnail' fall inline better with how we've using variant.
Intent is used for things like primary, secondary, warning, negative, positive. I'd recommend we keep it as variant.

packages/terra-image/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "terra-image",
-  "main": "lib/Image.scss",
+  "main": "lib/Image.js",
   "private": true,
   "version": "0.1.0",
```
bjankord: Let's roll this back to `0.0.0`

packages/terra-image/examples/ImageVariant.jsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import Image from '../src/Image';
+
+const ImageVariant = () => (
+  <div>
+    <h2>Default Image Style</h2>
+    <Image src="http://lorempixel.com/320/320/nature/" alt="example of defualt styles" />
+    <br />
```
JakeLaCombe: Maybe perhaps we can push the image in a paragraph instead of using a bunch of brs? One thing I found with blue_steel is that if you use patterns that are not recommended, then that behavior gets copied into the consumer applications very easily.

PR #146, opened 2017-03-17
Update terra-mixins version number

bjankord:
### Summary
terra-mixins v1.0.1 was recently released. This update bumps all dependents to new version. Pinning version to v1.0.1 to ensure we and consumers don't end up with multiple versions of this package.

### DIFF
--- a/packages/terra-arrange/package.json
+++ b/packages/terra-arrange/package.json
@@ -36,11 +36,11 @@
   "peerDependencies": {
     "react": "^15.4.2",
     "react-dom": "^15.4.2",
-    "terra-mixins": "1.0.1"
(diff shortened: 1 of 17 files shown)


windse7en:
Why do we keep `terra-mixins` twice in dependency and peerDependency?

Matt-Butler:
+1

mjhenkes:
Does npm allow you to set a version to something like 1.0.* ? Seems like we should allow minor versions without having to update all packages.

bjankord:
@mjhenkes we should be able to use `^1.0.0`. This will include everything greater than a particular version in the same major range. I'll test updating to that.

PR #149, opened 2017-03-20
Terra overlay

zhongyn:
### Summary
Overlay is an element that overlays the viewport or a DOM element usually to mark it as disabled.
Common usage:
- Shade out main content while mobile menus are open
- Shade out main content while modals are open
- Shade out a panel and display a loading indicator while ajax call is running

Accessibility:
- Focus stays inside of overlay children when overlay is open
- Allow closing the overlay on ESC key

| Prop     | Type | Default | Description  |
|----------|------|---------|--------------|
| isOpen | bool | false | Whether or not <Overlay/> should be rendered |
| isFullScreen | bool | true | Display full screen overlay|
| onRequestClose | func | null | Callback triggered on overlay click or ESC key |
| container | DOM element | | Element that will be overlaid |
| canClickOutsideClose | bool | true | Allow closing overlay by clicking backdrop|
| centerContent | bool | true | Center the content horizontally and vertically |
| children | node | null | VisIble overlay content when overlay is open |

overlay
Bidi:
overlay-rtl

@cerner/terra


packages/terra-overlay/docs/README.md
```diff
@@ -0,0 +1,77 @@
...
+          className='my-class'
+        >
+          {<p>I am the content</p>}
+        </Overlay>
+      </div>
+    );
+  }
+}
+
+export default class OverlayOnElement extends Component {
```
Matt-Butler: nit, export at bottom of file

packages/terra-overlay/examples/OverlayDefault.jsx
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import React, { Component } from 'react';
+import Button from 'terra-button';
+import Overlay from '../src/Overlay';
+
+export default class OverlayDefault extends Component {
```
Matt-Butler: nit, export at bottom of file

packages/terra-overlay/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "terra-overlay",
+  "main": "lib/Overlay.js",
+  "private": true,
+  "version": "0.1.0",
```
Matt-Butler: 0.0.0

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -0,0 +1,167 @@
...
+
+  restoreFocus() {
+    if (this.focusBeforeOpen && this.focusBeforeOpen.focus) {
+      this.focusBeforeOpen.focus();
+      this.focusBeforeOpen = null;
+    }
+  }
+
+  handleEscKeyup(e) {
+    if (e.keyCode === 27) {
```
Matt-Butler: Can you please make 27 a variable?

packages/terra-overlay/src/Overlay.scss
```diff
@@ -0,0 +1,30 @@
+@import '~terra-mixins';
+
+.terra-Overlay {
+  background-color: #1c1f21;
```
Matt-Butler: Can we use a color from terra-legacy-theme?

packages/terra-overlay/src/Portal.jsx
```diff
@@ -0,0 +1,60 @@
...
+  renderContent(props) {
+    if (!this.container) {
+      this.container = document.createElement('div');
+      props.target.appendChild(this.container);
+    }
+    ReactDOM.unstable_renderSubtreeIntoContainer(this, props.children, this.container);
+  }
+
+  render() {
+    return null;
```
Matt-Butler: Something about this seems wrong. Did you see this somewhere else?

packages/terra-overlay/docs/README.md
```diff
@@ -0,0 +1,77 @@
...
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-overlay`
+  - `yarn add terra-overlay`
+
+## Usage
+
+```jsx
+import React, { Component } from 'react';
+import Button from 'terra-button';
+import Overlay from '../src/Overlay';
```
Matt-Butler: Should this be `import Overlay from 'terra-overlay`?

packages/terra-overlay/src/Overlay.scss
bjankord: You can use `$terra-dark` here. It has a [default value](https://github.com/cerner/terra-core/blob/master/packages/terra-legacy-theme/src/_palette.scss#L90) of `#1c1f21`background-color: $terra-dark;`

packages/terra-overlay/docs/README.md
zhongyn: updated

packages/terra-overlay/examples/OverlayDefault.jsx
zhongyn: updated

packages/terra-overlay/package.json
zhongyn: updated

packages/terra-overlay/src/Overlay.jsx
zhongyn: updated

packages/terra-overlay/src/Overlay.scss
zhongyn: good catch, updated

packages/terra-overlay/docs/README.md
zhongyn: updated

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -0,0 +1,168 @@
...
+      centerContent,
+      children,
+      ...customProps
+    } = this.props;
+    const overlayClass = classNames(
+      'terra-Overlay',
+      { 'terra-Overlay--center': centerContent },
+      customProps.className,
+    );
+    const onELementPosition = isFullScreen ? null : {
```
tbiethman: Nitpicking: capital L should be lowercase

packages/terra-overlay/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "terra-overlay",
+  "main": "lib/Overlay.js",
+  "private": true,
+  "version": "0.0.0",
+  "description": "terra-overlay",
```
StephenEsser: I would copy the ReadMe description over to here

packages/terra-overlay/package.json
```diff
@@ -0,0 +1,58 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/cerner/terra-core.git"
+  },
+  "keywords": [
+    "Cerner",
+    "Terra",
+    "terra-overlay",
+    "ResponsiveElement",
```
StephenEsser: Does this keyword belong here?

packages/terra-overlay/src/Portal.jsx
zhongyn: `<Portal />` itself is an empty react component. In its lifecycle it mounts/unmounts its children to other node in the DOM tree instead of the intrinsic react tree. It is like a bus. [React-bootstrap](https://github.com/react-bootstrap/react-overlays/blob/master/src/Portal.js#L92), [material-ui](https://github.com/callemall/material-ui/blob/master/src/internal/RenderToLayer.js#L120) and [react-portal](https://github.com/tajo/react-portal/blob/master/lib/portal.js#L150) have the similar implementations.

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -0,0 +1,168 @@
...
+  isFullScreen: true,
+  container: document.body,
+  centerContent: true,
+  canClickOutsideClose: true,
+};
+
+class Overlay extends React.Component {
+  constructor() {
+    super();
+    this.state = {};
```
StephenEsser: I don't see state referenced anywhere in this file, this can probably be removed

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -0,0 +1,168 @@
...
+  restoreFocus() {
+    if (this.focusBeforeOpen && this.focusBeforeOpen.focus) {
+      this.focusBeforeOpen.focus();
+      this.focusBeforeOpen = null;
+    }
+  }
+
+  handleEscKeyup(e) {
+    if (e.keyCode === ESCAPE) {
+      this.props.onRequestClose();
```
StephenEsser: There is no validation this function was passed in, it should probably be declared as required in propTypes

packages/terra-overlay/src/Overlay.scss
```diff
@@ -0,0 +1,28 @@
+.terra-Overlay {
+  background-color: $terra-dark;
+  height: 100%;
+  left: 0;
```
StephenEsser: Will this work in bi-di?

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -0,0 +1,168 @@
...
+    };
+    const overlayStyle = {
+      ...onELementPosition,
+      ...customProps.style,
+    };
+    /* eslint-disable jsx-a11y/no-static-element-interactions */
+    return (
+      <Portal isOpen={isOpen} target={container}>
+        <div
+          ref={(overlay) => { this.overlay = overlay; }}
```
StephenEsser: I think custom props should be spread onto this to remain consistent with the other components.

This does introduce a weird situation with ref and onClick. These will cancel out any custom ref and onClick as long as they're added after spreading customProps. Normally these functions would attempt to call the custom prop if one existed after performing the required actions on the component. So ref would call the custom prop ref as a callback, and onClick would call props.onClick as a callback (if they existed). 

This might be a situation where this component owns those actions specifically and doesn't callback, but I still think custom props should be used here. The callbacks probably should happen, but they can also always be opened up in the future.

packages/terra-overlay/src/Overlay.jsx
zhongyn: good catch, updated

packages/terra-overlay/package.json
zhongyn: updated

packages/terra-overlay/package.json
zhongyn: good catch, copied from ResponsiveElement, updated

packages/terra-overlay/src/Overlay.jsx
zhongyn: removed

packages/terra-overlay/src/Overlay.jsx
zhongyn: yes, updated it to be required

packages/terra-overlay/src/Overlay.scss
zhongyn: overlay will be either fullscreen or fill the entirety of a DOM element (width and height = 100%), so it work in bi-di. Attached a gif above of rtl

packages/terra-overlay/src/Overlay.jsx
zhongyn: yes, custom props should be there. I updated the props here such that custom props spreads at the beginning of the div to avoid overriding the existing onClick and ref as ResponsiveElement does.

packages/terra-overlay/src/Portal.jsx
bjankord: My thought is portal should probably be its own stand-alone component. This way it can be used not only with overlay/modal but any component that needs to render children to other node in the DOM tree.

bjankord:
Functionally and visually, this looks to be doing everything we want. I have concerns about [overlaymanager.js](https://github.com/cerner/terra-core/blob/31a5929ec139416a58ae0a305e3930c57bab5700/packages/terra-overlay/src/OverlayManager.js). This seems out of scope from what was in the tech design and seems like a higher level concern than terra-core. I also have a concern about nesting the portal component in the overlay package. I think we should have a separate package for the portal component.

packages/terra-overlay/src/Portal.jsx
zhongyn: agree. Portal can be moved to its own component, such as rending pop-up

bjankord:
Closing this until we have capacity to continue development on this component.

Issue #151, opened 2017-03-20
`npm install` removes dependencies from node_modules in root


Matt-Butler:
This issue can be resolved be removing the preinstall script.
https://github.com/npm/npm/issues/10379

When executing `npm install` when node_modules dependencies are present in the root directory, all of the dependencies are deleted.
![npm_install](https://cloud.githubusercontent.com/assets/8146159/24107662/5c93a0e8-0d59-11e7-8b8e-90585024988e.gif)

bjankord:
Related https://github.com/npm/npm/issues/10379

PR #155, opened 2017-03-20
Remove preinstall script

Matt-Butler:
Remove preinstall script

@cerner/terra

Details:
https://github.com/cerner/terra-core/issues/151


package.json
```diff
@@ -45,7 +45,6 @@
     "jest": "jest",
     "jest:coverage": "jest --coverage",
     "pretest": "npm run lint",
-    "preinstall": "npm run clean:all",
```
StephenEsser: I was under the impression this was intentional, so that `npm install` would always remove old modules and generate fresh dependencies. I know a few people have run into issues with `npm install` and old modules that cause the command to fail or behave in an odd way.

bjankord: > I've also hooked into the npm install lifecycle (using preinstall lifecycle hook). When a user runs `npm install`, the script will clean all the `node_modules` directories, then npm will install new modules into node_modules directories.

I missed that this leaves an empty `node_modules` directory in the root directory and that only node_modules in the individual terra-component package directories remain. 

This is due to [an issue](npm/npm#10379) with how `preinstall` works since npm v3. scripts defined in `preinstall` run after node_modules are installed. So what actually happens when running `npm install` in the root directory is:
* node_modules installed in root
* (preinstall scripts run) clean:all runs and removes root node_modules and individual package's node_modules
* (postinstall script runs) lerna bootstraps packages by npm linking them and running npm install for each package.

Unfortunately, it doesn't look like the npm team will revert `preinstall` behavior to match how it worked with npm v2. So its better to drop our clean task from being called on `preinstall`. I do think it would be nice to set up a task that could be run from root directory that does the following in order:
* npm run clean:all
* npm install
* npm run bootstrap

Matt-Butler: I will add another script that executes 
```
npm run clean:all
npm install
npm run bootstrap
```

Matt-Butler: https://github.com/cerner/terra-core/pull/155/commits/5a9f2d5550f205b4474792608719cc4766b02556

dkasper-was-taken:
+1

PR #158, opened 2017-03-21
Update i18n nightwatch test to fix the random failure

windse7en:
### Summary

There are some test failures randomly happening in travis CI. The root cause is nightwatch assertions comes before all required data loaded in i18n.

So add `waitForElementVisible()` to wait for content loading and add `pause(time)` to pause and wait for the locale switch.


packages/terra-i18n/tests/nightwatch/i18n-spec.js
```diff
@@ -20,62 +22,79 @@ module.exports = {
   'Displays ajax error message in default': (browser) => {
     browser
       .url(`http://localhost:${browser.globals.webpackDevServerPort}/#/tests/i18n-tests/default`)
+      .waitForElementVisible('select', 1000)
```
windse7en: This `1000` is the maximum time in millisecond to wait for `select` element to be visible.

packages/terra-i18n/tests/nightwatch/i18n-spec.js
```diff
@@ -12,6 +12,8 @@ const defaultTranslations = enTransaltions;
 
 const screenshot = require('terra-toolkit').screenshot;
 
+const pauseInms = 1000;
```
windse7en: The `pauseInms` is pause in millisecond. The `1000ms` value is common used in nightwatch test examples. It works in all 10 tests in local.

There is no API like `waitForElmenentUpdate`. So use `pause()` function in here.

ryanthemanuel: If we start using pause all over the place it is going to make our functional tests get slower and slower.  Is there a way to wait for a new element to be visible in this case?

ryanthemanuel: Also have you tried playing around with the callback from the click command to see if that will accomplish the same thing in a more asynchronous fashion?

packages/terra-i18n/tests/nightwatch/i18n-spec.js
```diff
@@ -20,62 +22,79 @@ module.exports = {
...
       .assert.containsText('div', esTransaltions['Terra.ajax.error']);
   },
 
   'Displays ajax error message in fi-FI': (browser) => {
     browser
       .url(`http://localhost:${browser.globals.webpackDevServerPort}/#/tests/i18n-tests/default`)
+      .waitForElementVisible('select', 1000)
       .click('select option[value="fi-FI"]')
+      .pause(pauseInms)
       .assert.containsText('div', fiFITransaltions['Terra.ajax.error']);
```
ryanthemanuel: Just noticing that all of these Translations are misspelled.  Transaltions should be Translations.

packages/terra-i18n/tests/nightwatch/i18n-spec.js
ryanthemanuel: Take a look at this to see what I mean: https://github.com/nightwatchjs/nightwatch/wiki/Understanding-the-Command-Queue

packages/terra-i18n/tests/nightwatch/i18n-spec.js
windse7en: Nice catch. That is a typo by copy and paste. fix it in [commit](https://github.com/cerner/terra-core/pull/158/commits/2934eb8dddbc14562edb7a6698eb7707ad94a8a4)

packages/terra-i18n/tests/nightwatch/i18n-spec.js
windse7en: Yeah. That is the discussion I really want to have in this PR. I don't try to use click callback. I will try to see what we can get from that.

windse7en: We can't use callback in this case.
> The primary purpose for callbacks is to allow you to capture data from the test while its running, but they also provide an additional opportunity to add commands to the command queue

The benefits of callback: 
1. access to the result data.
2. add commands to the command queue.

We can not access the `browser.url('....')` object. We can only use `browser` object and `result` data object. And it adds command to commands queue, like:
```javascript
    browser
      .url(`http://url`)
      .waitForElementVisible('select', 1000)
      .click('select option[value="es"]')
      .assert.containsText('div', esTranslations['Terra.ajax.error']);
```
We still can't tell if all required data is loaded.

An alternative approach is using custom commands assertions, like `waitForText`, in [here](https://github.com/maxgalbu/nightwatch-custom-commands-assertions/blob/master/docs/waitForText.md). But I don't want to add such dependency. 

I also dislike the `pause(1000)`, it will add 1s latency to each test.

bjankord: I'm open to adding custom commands assertions, like `waitForText`. What are your thoughts @ryanthemanuel?

windse7en: Assert failure before required locale data, it can randomly happen, like in [here](https://travis-ci.org/cerner/terra-core/builds/213563117#L2394)

windse7en: Thanks, Ryan's help. Update to use `expect ... before` syntax to get rid of the `pause` function.

windse7en: Expect before docs: http://nightwatchjs.org/api#expect-before

PR #163, opened 2017-03-22
Update terra-legacy-theme version number

bjankord:
@cerner/terra


bjankord:
This is currently failing on random nightwatch tests.

Issue #162, opened 2017-03-21
Can only update a mounted or mounting component.


Matt-Butler:
![console_error](https://cloud.githubusercontent.com/assets/8146159/24174507/e70babc0-0e5e-11e7-9d8c-49c97f5866e7.gif)

I am seeing this issue when revisiting the i18n component. @StephenEsser what did you do to resolve this with the responsive component?

PR #166, opened 2017-03-23
Set state componentDidMount

Matt-Butler:
Fixes this issue. State will only be set after the component did mount.
https://github.com/cerner/terra-core/issues/162

@cerner/terra

### DIFF
--- a/packages/terra-i18n/tests/nightwatch/DefaultI18n.jsx
+++ b/packages/terra-i18n/tests/nightwatch/DefaultI18n.jsx
@@ -11,8 +11,11 @@ class Base extends React.Component {
       locale: props.locale,
       messages: {},
     };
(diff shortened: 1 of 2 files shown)


windse7en:
Nice catch. Could you also update it in the test code? https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/tests/nightwatch/DefaultI18n.jsx#L14

Matt-Butler:
> Nice catch. Could you also update it in the test code? https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/tests/nightwatch/DefaultI18n.jsx#L14


https://github.com/cerner/terra-core/pull/166/commits/444f0f2b3c09ef5f68cd2bca0987cfd55dc3c950

PR #167, opened 2017-03-23
Spectre changes

ryanthemanuel:
### Summary
These are changes to (when enabled) submit screenshots to http://monstrous-screen-grabber.herokuapp.com/  It is a work in progress, and there are many improvements to be made, but this is a first pass.  Unfortunately, there are some minor non-passive changes involved, but I've updated the appropriate packages and will update the generator as well.


packages/terra-toolkit/src/spectre.js
```diff
@@ -0,0 +1,73 @@
...
+
+  if (process.env.SAVE_TO_SPECTRE === 'true') {
+    const data = JSON.stringify({
+      project: 'terra-ui',
+      suite: suiteName,
+    });
+
+    try {
+      const req = http.request({
+        hostname: 'monstrous-screen-grabber.herokuapp.com',
```
Matt-Butler: Some of this configuration is specific to us. Should this be broken out into a config file?

ryanthemanuel: Yes it should be.  But I'm trying to do iterative improvements.  I need to move on in the interim and I think this is a good temporary solution for what we're dealing with.

Issue #65, opened 2017-03-01
Migrate terra-button-group into repo


bjankord:
https://github.com/cerner/terra-button-group

PR #171, opened 2017-03-27
Terra button group

jmsv6d:
### Summary
Button Group React Component

@cerner/terra

https://github.com/cerner/terra-core/issues/65


packages/terra-button-group/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Button Group
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-button-group.svg)](https://www.npmjs.org/package/terra-button-group)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+{insert description}
+
+- [Getting Started](#getting-started)
+- [Documentation](#documentation)
```
Matt-Butler:
```
- [Documentation](https://github.com/cerner/terra-core/tree/master/packages/terra-button-group/docs)
```
We want to use the absolute path here so the path works when we publish it to npm.

packages/terra-button-group/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Button Group
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-button-group.svg)](https://www.npmjs.org/package/terra-button-group)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+{insert description}
```
Matt-Butler: Please add a description

packages/terra-button-group/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "terra-button-group",
+  "main": "lib/ButtonGroup.js",
+  "private": true,
+  "version": "0.1.0",
```
Matt-Butler: 0.0.0

packages/terra-button-group/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
+    "terra-toolkit": "^0.1.0",
+    "terra-props-table": "^0.0.0",
+    "terra-markdown": "^0.0.0"
```
Matt-Butler:
```
"terra-props-table": "^0.0.0",
"terra-markdown": "^0.0.0"
```

We can remove these two devDependencies. These are needed for the examples only. Since we moved the examples over to terra-site, we now only need these in terra-site.

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -0,0 +1,143 @@
...
+      this.handleOnClick(event, index);
+
+      if (onClick) {
+        onClick(event);
+      }
+    };
+  }
+
+  render() {
+    const { onChange, variant, size, isCompact, isSelectable, buttons, children, ...extraProps } = this.props;
```
Matt-Butler: `extraProps` => `customProps`

This variable is named customProps in the other components

packages/terra-site/src/App.jsx
```diff
@@ -32,6 +33,8 @@ const App = props => (
       <li><Link to="/site/standout">Standout</Link></li>
       <li><Link to="/site/status">Status</Link></li>
       <li><Link to="/site/title">Title</Link></li>
+      <li><Link to="/site/slide-panel">Slide Panel</Link></li>
+      <li><Link to="/site/i18n">I18n</Link></li>
```
Matt-Butler: Lines 36/37 are duplicate code.

Matt-Butler:
For selectable button groups, we currently have option selection behavior. Do we have a need checkbox selection behavior?

jmsv6d:
With the feedback from UX we only need the radio style selection: https://wiki.ucern.com/pages/viewpage.action?spaceKey=UserExperience&title=Terra+UI%3A+Approved+Patterns

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -0,0 +1,143 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+
+import ButtonGroupButton from './ButtonGroupButton';
+import './ButtonGroup.scss';
+
+const propTypes = {
+  /**
```
windse7en: Should we have props in some order in here? As the order in here will decide the final order in docs, top ones will get more consumers attention.

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -0,0 +1,143 @@
...
+  constructor(props) {
+    super(props);
+    this.handleOnClick = this.handleOnClick.bind(this);
+    this.state = {
+      selectedIndex: ButtonGroup.getInitialState(this.props.buttons.concat(this.props.children), this.props.isSelectable),
+    };
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const newState = ButtonGroup.getInitialState(nextProps.buttons.concat(nextProps.children), nextProps.isSelectable);
```
windse7en: This constant name is a little confusing. Is this `newSelectedIndex`?

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    return null;
+  }
+
+  constructor(props) {
+    super(props);
+    this.handleOnClick = this.handleOnClick.bind(this);
+    this.state = {
+      selectedIndex: ButtonGroup.getInitialState(this.props.buttons.concat(this.props.children), this.props.isSelectable),
```
windse7en: What's the data type of `selectedIndex`? Could you explain why we need to have all buttions and `isSelectable` as params in here?

packages/terra-button-group/src/ButtonGroup.jsx
jmsv6d: We could. But besides buttons and/or children I'm not sure if any of the other props have higher importance than the others. Do you have a suggested order?

packages/terra-button-group/src/ButtonGroup.jsx
jmsv6d: selectedIndex is an integer. The initial selected state is set by setting isSelected prop on ButtonGroup.Button, so I need to loop through all the buttons to determine the selected button. `isSelectable` is sent because there shouldn't be any selected buttons if the button group isn't selectable. Setting the selectedIndex to null when the button group isn't selectable would overwrite any ButtonGroup.Button that was set to selected when the button group isn't selectable when the buttons are cloned on render here: https://github.com/cerner/terra-core/pull/171/files#diff-abb2a53429231a34e97fc4fbb1d7d9dcR122

packages/terra-button-group/src/ButtonGroup.jsx
windse7en: I think we can discuss this in war room. It's good to have other teammates' opinions in here. My initial thought is like
1. required props come first(should get most attention), boolean props come together(has the same type and follow the same naming convention), functions come together(has the same type and follow the same naming convention).
2. `defaultProps` should be in the same order in the `propTypes`.

bjankord: Lets group required / non-required and group those 2 alphabetically.

mjhenkes: This seems pretty picky to me. I'm all for keeping code uniform but my vote is keeping that uniformity to be the responsibility of the linter. If you guys feel strongly about ordering props, update the linter to enforce that prop order instead of documenting the standard wiki. That way anyone who checks in will have the correct format and we won't have to look for it in code reviews.

packages/terra-button-group/docs/themeable-variables.md
```diff
@@ -0,0 +1,3 @@
+# Themeable Variables
+
+Variables to `theme` the terra-button-group component.
```
bjankord: If there are no themeable variables, we can delete this file.

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -0,0 +1,143 @@
...
+const propTypes = {
+  /**
+   * Callback function when the state changes
+   **/
+  onChange: PropTypes.func,
+
+  /**
+   * Sets the button group style variation. One of `defaut` or `secondary`
+   **/
+  variant: PropTypes.oneOf(['default', 'secondary']),
```
bjankord: Should primary be added to this list as well?

packages/terra-button-group/src/ButtonGroupButton.jsx
```diff
@@ -0,0 +1,59 @@
+import React, { PropTypes } from 'react';
```
bjankord: I think this will be easier to maintain if we treat this component as an h.o.c. that wraps the `<Button />` component. This way we don't have to maintain duplicate "button" components / styles.

packages/terra-button-group/src/ButtonGroup.jsx
jmsv6d: No, the UX standards only wanted default and secondary to allowed styles for the button group

packages/terra-button-group/src/ButtonGroupButton.jsx
jmsv6d: My main concern with that is that there are some button styles that we don't want the consumers to be able to set on a per button basis, but rather be set on the button group as a whole (size, isCompact, variant). We could pass those props down from the ButtonGroup to the ButtonGroup.Button to be set on Button, but those shouldn't be exposed to the consumer in the api for the ButtonGroup.Button.

mjhenkes: The button group was originally just that, but as it became apparent that UX did not want all the available styles on button to be available for the button group, we ended up just creating a button group button with only those styles.

bjankord: Yeah, my concern is with maintaining styles / react implementation of buttons in two places, `<Button />` and `<ButtonGroup.Button />`. I understand the UX concern of not wanting button link buttongroups and the concern of not wanting people to pass in props like href to `<ButtonGroup.Button />` and then having it render as an anchor.

I think we can avoid these with a h.o.c.  around `<Button />`. We can limit what variants are passed to it via the oneOfType propType check to ensure people can't use primary or link buttons. And for first class props button supports like `href` that we don't want to support in `<ButtonGroup.Button>`, we can intercept those props and set them to null so they don't get passed down to the wrapped `<Button />` component.

packages/terra-button-group/src/ButtonGroup.jsx
windse7en: Actually, I'm not strongly on the props order in `jsx` file. Just throw the concern on our doc site. Currently, we will generate the propType docs as the same order as in the source file.
screen shot 2017-03-30 at 2 56 35 pm

So reviewing the propTypes likes a kind of reviewing docs.

I'm also fine that we don't concern this now. We can add an enhancement to `terra-propTypes` to make final docs in desired order(this order should be consistent across all components).

+1

packages/terra-button-group/src/ButtonGroup.jsx
windse7en: Thanks so much for your clear explanation.
+1

packages/terra-button-group/src/ButtonGroupButton.jsx
mjhenkes: Are you suggesting that `<ButtonGroup.Button />` simply wrap a `<Button />` and offer a scoped down api? Assuming the api of button group is a subset I'd be fine with that implementation.(@jmsv6d, could you verify?) Though i don't think we have to go so far as nulling custom props if people send them in. At that point you're going beyond the supported api at your own risk. Secondly would you mind if we pushed that change in as a separate pull request? It should be a passive api change and I'd like to break the changes up into smaller pull requests.

bjankord: > Are you suggesting that `<ButtonGroup.Button />` simply wrap a `<Button />` and offer a scoped down api?

Yeah.

> Secondly would you mind if we pushed that change in as a separate pull request?

+1, sounds good to me.

mjhenkes: I've added https://github.com/cerner/terra-core/issues/188 to track this work.

PR #175, opened 2017-03-27
Prepare to publish terra-markdown

Matt-Butler:
### Summary
Preparing to publish terra-markdown.

@cerner/terra


packages/terra-markdown/package.json
```diff
@@ -21,17 +20,13 @@
...
     "react": "^15.4.2",
-    "react-dom": "^15.4.2",
-    "terra-mixins": "^1.0.0"
+    "react-dom": "^15.4.2"
   },
   "dependencies": {
-    "classnames": "^2.2.5",
-    "terra-mixins": "^1.0.0"
+    "marked": "^0.3.6",
+    "github-markdown-css": "^2.4.1"
```
Matt-Butler: I moved these 2 dependencies from the root package.json into the terra-markdown package.json.  Both of these dependencies are imported in the `src` files so they need to be in this package.json. 

https://github.com/cerner/terra-core/blob/master/packages/terra-markdown/src/Markdown.jsx#L3-L4

packages/terra-markdown/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "terra-markdown",
   "main": "lib/Markdown.js",
-  "private": true,
   "version": "0.0.0",
```
Matt-Butler: I will update the version in the next commit

packages/terra-markdown/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "terra-markdown",
   "main": "lib/Markdown.js",
-  "private": true,
```
Matt-Butler: Removing the private attribute. https://docs.npmjs.com/files/package.json#private

packages/terra-markdown/src/Markdown.jsx
```diff
@@ -1,7 +1,7 @@
-/* eslint-disable import/no-extraneous-dependencies */
 import React, { PropTypes } from 'react';
 import marked from 'marked';
-import 'github-markdown-css/github-markdown.css';
+// eslint-disable-next-line import/extensions
+import 'github-markdown-css';
```
Matt-Butler: The main attribute points to the css we want to import so we can update the syntax to this.
https://github.com/sindresorhus/github-markdown-css/blob/gh-pages/package.json#L12

Issue #57, opened 2017-03-01
Migrate terra-table into repo


bjankord:
https://github.com/cerner/terra-table

bjankord:
Resolved with #177

PR #177, opened 2017-03-28
Terra Table

nagrawal3:
### Summary
Terra Table React Component

@cerner/terra

https://github.com/cerner/terra-core/issues/57


packages/terra-site/src/examples/table/Index.jsx
```diff
@@ -0,0 +1,70 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import React from 'react';
+import PropsTable from 'terra-props-table';
+import Markdown from 'terra-markdown';
+import ReadMe from 'terra-table/docs/README.md';
+// Component Source
+// eslint-disable-next-line import/no-webpack-loader-syntax, import/first, import/no-unresolved, import/extensions
```
Matt-Butler: Instead of disabling just the next-line, eslint allows you to disable blocks of code.
```
/* eslint-disable import/no-webpack-loader-syntax, import/first, import/no-unresolved, import/extensions */
import TableSrc from '!raw-loader!terra-table/src/Table';
import SingleSelectableRowsSrc from '!raw-loader!terra-table/src/SingleSelectableRows';
import TableHeaderSrc from '!raw-loader!terra-table/src/TableHeader';
....
/* eslint-enable import/no-webpack-loader-syntax, import/first, import/no-unresolved, import/extensions */
```

packages/terra-table/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Table
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-table.svg)](https://www.npmjs.org/package/terra-table)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+{insert description}
```
Matt-Butler: Nit: please update

packages/terra-table/lib/Table.scss
```diff
@@ -0,0 +1,71 @@
...
+.terra-Table--isSelected {
+  background-color: $terra-blue-20 !important;
+}
+
+.terra-Table-hidden {
+  visibility: hidden;
+}
+
+.terra-Table-sort-indicator {
+  float: right;
```
Matt-Butler: bidi

https://github.com/cerner/terra-core/blob/master/packages/terra-mixins/src/Mixins.scss#L304

packages/terra-table/src/Table.scss
```diff
@@ -0,0 +1,71 @@
+@import '~terra-mixins';
+@import './variables';
+@import './mixins';
+@import './terra-table';
+
+[data-column-min-width='tiny'] {
```
Matt-Butler: Can we make these css classes instead of using attribute selectors?

packages/terra-table/src/TableHeaderContent.jsx
```diff
@@ -0,0 +1,76 @@
...
+  const dataSort = {
+    'data-sort': sort,
+  };
+
+  return (
+    <th {...customProps} className={contentClassName} data-column-min-width={minWidth} {...heightProperty} {...dataSort}>
+      <div>
+        {display}
+        <span className={ascSortInd}>{iconUp}</span>
+        <span className={descSortInd}>{iconDown}</span>
```
Matt-Butler: Instead of updating css visibility, can we render only the sort/icon we need?

packages/terra-table/src/TableRow.jsx
```diff
@@ -0,0 +1,67 @@
...
+  height: PropTypes.oneOf(['tiny', 'small', 'medium', 'large', 'huge']),
+  /**
+   * Whether or not row is selectable
+   */
+  isSelectable: PropTypes.bool,
+};
+
+const defaultProps = {
+  isSelected: false,
+  isSelectable: undefined,
```
Matt-Butler: false?

Matt-Butler:
Selectable table rows are not keyboard accessible.

Matt-Butler:
We probably want to update the cursor to pointer for selectable rows: 
selectable_rows

packages/terra-table/docs/themeable-variables.md
```diff
@@ -0,0 +1,3 @@
+# Themeable Variables
+
```
bjankord: If there are no themeable sass variables, this file can be removed.

packages/terra-table/src/TableRowContent.jsx
```diff
@@ -0,0 +1,31 @@
+import React, { PropTypes } from 'react';
+
+const propTypes = {
+  /**
+   * Content to be displayed for the row cell
+   */
+  display: PropTypes.any.isRequired,
```
bjankord: Nit: Change this to `content`. It is easier to reason about what this is an matches with other react components we have.

packages/terra-table/src/_variables.scss
```diff
@@ -0,0 +1,35 @@
...
+$terra-table-row-striped-th-background-color: $terra-grey-20 !default;
+$terra-table-row-striped-background-color: $terra-grey-10 !default;
+$terra-table-tfoot-border-color: $terra-grey-30 !default;
+$terra-table-tfoot-border-width: 1px !default;
+$terra-table-thead-background-color: $terra-grey-20 !default;
+$terra-table-thead-border-color: $terra-grey-30 !default;
+$terra-table-thead-border-width: 1px !default;
+$terra-table-tr-border-color: $terra-grey-30 !default;
+$terra-table-tr-border-width: 1px !default;
+$terra-table-cell-max-height-tiny: 26px;
```
bjankord: The `max-height` and `min-width` values should be set in rem units.

packages/terra-table/src/TableRow.jsx
nagrawal3: I keep it undefined because, if Table is non-selectable, then there is no need to define this attribute for every row.

tbiethman: You can just remove it from the defaultProps altogether. It'll be undefined by default.

packages/terra-table/docs/themeable-variables.md
nagrawal3: Added variables in doc https://github.com/cerner/terra-core/pull/177/commits/c9b2d6834df2d0749688c1158f19abec55eb5a0e#diff-a2f8fe2900b7404b7406ee8301ba4387R4

packages/terra-table/src/TableRow.jsx
nagrawal3: removed the default for isSelectable. Cannot add default for height as it will set the maximum height on the row and by default we are not setting restrictions on height

nagrawal3:
Updated the code with changes as per above comments

packages/terra-table/src/TableRowContent.jsx
```diff
@@ -0,0 +1,35 @@
...
+  }) => {
+  const contentClassName = classNames([
+    { [`terra-Table-row-max-height--${height}`]: height },
+    'terra-Table-cell',
+    customProps.className,
+  ]);
+
+  return (
+    <td {...customProps} className={contentClassName}>
+      <div>{content}</div>
```
tbiethman: Do we need to wrap the content in an extra <div>?

nagrawal3: By default, the minimum height set on the table cell is the minimum height required by the content in the cell. Therefore in order to restrict the maximum height of the cell I have to add extra div tag.

packages/terra-site/src/examples/table/NoStripedTable.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+const NoStripedTable = () => (
+  <Table isStriped={false}>
+    <Table.Header>
+      <Table.HeaderContent content={'Column Heading 1'} key={1} minWidth={'small'} />
+      <Table.HeaderContent content={'Column Heading 2'} key={2} minWidth={'medium'} />
+      <Table.HeaderContent content={'Column Heading 3'} key={3} minWidth={'large'} />
+    </Table.Header>
+    <Table.Rows>
+      <Table.Row key={0}>
```
JakeLaCombe: I know that it's not good practice to set the key equal to the index of the array, and this section of code feels like it's doing just that, even though there isn't a map or forEach function being called. Could we generate unique ids for these keys instead of using the numbers 1, 2, and 3? I used [shortid](https://github.com/dylang/shortid) to generate it in one of my code changes for example.

mjhenkes: We should leave it up to the user to set keys. Setting a random key is actually worse than just using the index since keys help React identify which items have changed, are added, or are removed.
Check this thread for some more information.
https://github.com/facebook/react/issues/1342

JakeLaCombe: Totally agree on that. For this example, it feels too closely like it's setting the key equal to the value of it's array index, which also is not a good practice for assigning keys to react components. I feel like a developer would look at this example, and think that it could use the map function for example, and assign the key equal to the numerical index of that given element.

mjhenkes: Ah, I didn't pay enough attention to see that this is an example. +1 to using non-index keys in the example.

nagrawal3: changed in commit bb5f8be

packages/terra-site/src/examples/table/NoStripedTable.jsx
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import React from 'react';
+import Table from 'terra-table';
+
+const shortid = require('shortid');
```
Matt-Butler: Should consumers be providing these keys rather than us generating them?

packages/terra-site/package.json
```diff
@@ -26,6 +26,8 @@
     "react-dom": "^15.4.2",
     "react-intl": "^2.2.3",
     "react-router": "^3.0.2",
+    "shortid": "^2.2.8",
+    "terra-application": "^0.x",
```
Matt-Butler: We are deprecating `terra-application`. Please remove this line.

packages/terra-site/src/examples/table/NoStripedTable.jsx
Matt-Butler: Sorry, I know there was some discussion about this before. I just want to understand why we do this.

nagrawal3: We are generating the keys in example files only, we are not generating the keys in the code.

tbiethman: I still hesitate doing random key generation, even in examples, because we know this is a pattern we don't want to perpetuate. Can we make the examples slightly more meaningful, allowing us to make use keys that'd make sense? Something like:

```
<Table isStriped={false}>
  <Table.Header>
    <Table.HeaderContent content={'Name'} key="NAME" minWidth={'small'} />
    <Table.HeaderContent content={'Address'} key="ADDRESS" minWidth={'medium'} />
    <Table.HeaderContent content={'Phone'} key="PHONE_NUMBER" minWidth={'large'} />
  </Table.Header>
  <Table.Rows>
    <Table.Row key="PERSON_0">
      <Table.RowContent content="John Smith" key="NAME" />
      <Table.RowContent content="123 Adams Drive" key="ADDRESS" />
      <Table.RowContent content="111-222-3333" key="PHONE_NUMBER" />
    </Table.Row>
    <Table.Row key="PERSON_1">
      <Table.RowContent content="Jane Smith" key="NAME" />
      <Table.RowContent content="321 Drive Street" key="ADDRESS" />
      <Table.RowContent content="111-222-3333" key="PHONE_NUMBER" />
    </Table.Row>
    <Table.Row key="PERSON_2">
      <Table.RowContent content="Dave Smith" key="NAME" />
      <Table.RowContent content="213 Raymond Road" key="ADDRESS" />
      <Table.RowContent content="111-222-3333" key="PHONE_NUMBER" />
    </Table.Row>
  </Table.Rows>
</Table>
```

We want to get the point across that the key structure should match the schema of whatever data structure is being presented.

packages/terra-site/package.json
```diff
@@ -26,6 +26,7 @@
     "react-dom": "^15.4.2",
     "react-intl": "^2.2.3",
     "react-router": "^3.0.2",
+    "shortid": "^2.2.8",
```
ryanthemanuel: Have we done some analysis about adding this as a dependency?

packages/terra-site/src/examples/table/NoStripedTable.jsx
ryanthemanuel: +1 to Tyler's comment.  That can allow us to get rid of the additional dependency.

packages/terra-table/docs/README.md
```diff
@@ -0,0 +1,72 @@
...
+## Usage
+
+### Table
+```jsx
+import React from 'react';
+import Table from 'terra-table';
+
+<Table isStriped={false}>
+  <Table.Header>
+    <Table.HeaderContent display={'Column Heading 1'} key={1} minWidth={'small'} />
```
bjankord: The examples are using a `content` prop instead of `display`. Should this be updated to switch `display` prop to `content`

packages/terra-site/src/examples/table/NoStripedTable.jsx
JakeLaCombe: Let's move this to the war room tomorrow. In my component, I'm finding a need to generate random keys, and so far, this is the best solution I found.

PR #178, opened 2017-03-28
Migrate terra date picker

benbcai:
### Summary
Date Picker React Component

### Additional Details
Known issues:
* I will spend some time tomorrow to continue to look at the react warning that I get when opening the picker. This warning doesn't appear to have any functional impact and only occurs when rendering the picker in terra-site. It does not occur when rendered directly in terra-date-picker.

Todos in next iteration:
* Use date_util for formats
* Use forms for date input
* Update picker styles
* Add date validation/helper text

@cerner/terra


packages/terra-date-picker/docs/README.md
```diff
@@ -0,0 +1,61 @@
...
+  customInput={<CustomInputWithButton />}
+/>
+
+//Renders a date range
+<DateRange
+  startDate={moment().add(1, 'days')}
+  endDate={moment().add(7, 'days')}
+/>
+
+const isWeekday = (date) => {
```
Matt-Butler: I would suggest not using this isWeekday function.  Based on the locale, the first day of the week can be either Sunday or Monday. 

https://momentjs.com/docs/#/get-set/weekday/

Matt-Butler: Actually, nevermind. It looks like the `day()` method is not locale aware.

packages/terra-date-picker/lib/DatePicker.js
```diff
@@ -0,0 +1,231 @@
...
+          withPortal = _props.withPortal,
+          customProps = _objectWithoutProperties(_props, ['customInput', 'endDate', 'excludeDates', 'filterDate', 'includeDates', 'maxDate', 'minDate', 'isEndDateRange', 'isStartDateRange', 'selectedDate', 'startDate', 'withPortal']);
+
+      var classes = (0, _classnames2.default)(['terra-DatePicker']);
+
+      // TODO: Need translation from date_util
+      var todayString = 'Today';
+
+      // TODO: Get the locale from date_util
+      var userLocale = window.navigator.language;
```
Matt-Butler: We should probably discuss how locale should be set for an application. 

We need to ensure react-intl and momentjs are always on the same locale.

windse7en: Agree. I will add my comment into the source code.

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -0,0 +1,187 @@
...
+      withPortal,
+      ...customProps
+    } = this.props;
+    const classes = classNames(['terra-DatePicker']);
+
+    // TODO: Need translation from date_util
+    const todayString = 'Today';
+
+    // TODO: Get the locale from date_util
+    const userLocale = window.navigator.language;
```
windse7en: To hard code this locale value. We can add default `locale` prop to `en` in `defaultProps`. The ideal workflow will be
```
getUsersLocaleByApp(may from window, request header, Redux store) 
  --> I18nProvider 
  --> date_util 
  --> DatePicker props
```

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -0,0 +1,187 @@
...
+      isStartDateRange,
+      selectedDate,
+      startDate,
+      withPortal,
+      ...customProps
+    } = this.props;
+    const classes = classNames(['terra-DatePicker']);
+
+    // TODO: Need translation from date_util
+    const todayString = 'Today';
```
windse7en: This hard code can be set in `default props`. We can update it later.

packages/terra-date-picker/src/DatePicker.jsx
bjankord: We should look at how we got this value with rails / jquery globalize.

packages/terra-date-picker/src/DatePicker.jsx
benbcai: I didn't set it in defaultProps because we don't want to give the ability to change the label on the Today button so I'm not exposing the todayButton prop.

packages/terra-date-picker/docs/README.md
benbcai: Also, this is only used in the example to show how the filterDate prop works.

packages/terra-date-picker/lib/DatePicker.scss
```diff
@@ -0,0 +1,54 @@
...
+  display: inline-block;
+  min-width: 14.643em; 
+}
+
+.terra-DatePicker-input {
+  border: 0.071em solid $terra-grey-40;
+  box-shadow: inset 0 0.143em 0.143em $terra-grey-20;
+  font-size: 1em;
+  height: 1.143em;
+  line-height: 1.143em;
```
StephenEsser: This came up a few weeks ago, line height should be unitless. 

Here is a short article Matt Butler posted on a different pull request: https://css-tricks.com/almanac/properties/l/line-height/#article-header-id-0

packages/terra-date-picker/src/DatePicker.jsx
benbcai: I did some investigation on how ion_common and blue_steel get/set the locale in rails and Globalize. I'll add what I found to tomorrow's war room agenda and we can talk about it.

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -0,0 +1,38 @@
+import React, { PropTypes } from 'react';
+import Button from 'terra-button';
+import IconCalendar from 'terra-icon/lib/icon/IconCalendar';
+
+const Icon = <IconCalendar />;
+
+// eslint-disable-next-line react/prefer-stateless-function
+class DatePickerInput extends React.Component {
```
StephenEsser: Does this need to be a class or can it be functional? 

```
const DatePickerInput = ({ value, onChange, onClick, onKeyDown, placeholder }) => (
  <div className="terra-DatePicker-customInput">
    <input
      className="terra-DatePicker-input"
      type="text"
      value={value}
      onChange={onChange}
      placeholder={placeholder}
    />
    <Button
      className="terra-DatePicker-button"
      onClick={onClick}
      onKeyDown={onKeyDown}
      icon={Icon}
    />
  </div>
);
```

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -0,0 +1,38 @@
...
+  }
+}
+
+DatePickerInput.propTypes = {
+  onChange: PropTypes.func,
+  onClick: PropTypes.func,
+  onKeyDown: PropTypes.func,
+  placeholder: PropTypes.string,
+  value: PropTypes.string,
+};
```
StephenEsser: I think these should be documented like the Component below.

To be consistent propTypes should also move to the top of the file.
https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L5

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -0,0 +1,182 @@
...
+  /**
+   * A callback function to execute when a valid date is selected or entered.
+   */
+  onChange: PropTypes.func,
+};
+
+const defaultProps = {
+  isEndDateRange: false,
+  isStartDateRange: false,
+  withPortal: undefined,
```
StephenEsser: Why does this boolean default to undefined instead of true or false?

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -0,0 +1,182 @@
...
+   */
+  isStartDateRange: PropTypes.bool,
+  /**
+   * The default start date for a date range.
+   */
+  startDate: PropTypes.object,
+  /**
+   * Indicates to display the picker in a full screen overlay.
+   */
+  withPortal: PropTypes.bool,
```
StephenEsser: Should this boolean following the is standard other booleans follow? isWithPortal

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -0,0 +1,182 @@
...
+      maxDate,
+      minDate,
+      isEndDateRange,
+      isStartDateRange,
+      selectedDate,
+      startDate,
+      withPortal,
+      ...customProps
+    } = this.props;
+    const classes = classNames(['terra-DatePicker']);
```
StephenEsser: You might just use this string inline with the ReponsiveElement below since classNames isn't needed here

```
      <ResponsiveElement
         className="terra-DatePicker"
         responsiveTo="window"
         defaultElement={portalPicker}
         medium={popupPicker}
       />
```

packages/terra-date-picker/src/DateInput.jsx
benbcai: I initialize made this functional but I started seeing this warning in the console
> Warning: Stateless function components cannot be given refs (See ref "input" in 
> DatePickerInput created by DateInput). Attempts to access this ref will fail.
I think it is referring to this [ref](https://github.com/Hacker0x01/react-datepicker/blob/master/src/datepicker.jsx#L353) in react-datepicker. So I made it a class to get rid of the warning.

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -0,0 +1,182 @@
+import React, { PropTypes } from 'react';
+import classNames from 'classnames';
+import ReactDatePicker from 'react-datepicker';
+import moment from 'moment';
+import ResponsiveElement from 'terra-responsive-element';
+import DateInput from '../lib/DateInput';
+import '../lib/DatePicker.scss';
```
StephenEsser: I think all these imports should be relative:
```
import DateInput from './DateInput';
import './DatePicker.scss';
```
These paths will work from src and lib

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -0,0 +1,182 @@
...
+        withPortal={withPortal === undefined ? false : withPortal}
+        dateFormatCalendar=" "
+        dateFormat={momentDateFormat}
+        fixedHeight
+        locale={userLocale}
+        placeholderText={momentDateFormat}
+        dropdownMode={'select'}
+        showMonthDropdown
+        showYearDropdown
+      />);
```
StephenEsser: I'm a little on the fence if this is how the ResponsiveElement should be used, I think for right now this is what we have to work with, but this seems like a case where the same element just needs to be rendered differently, rather than swapping out an entire element for another. It looks like only one parameter is different between these huge blocks.

I'm wondering if it would be possible to wire in a onChange event that can be passed into the ResponsiveElement or maybe a new component that is called back with the current breakpoint size so an Element can render itself differently.

packages/terra-date-picker/src/DatePicker.scss
```diff
@@ -0,0 +1,54 @@
...
+  font-size: 1em;
+  height: 1.143em;
+  line-height: 1.143em;
+  margin: 0.357em 0;
+  padding: 0.429em 0.714em;
+  vertical-align: middle;
+
+  @if $terra-bidi {
+    [dir='ltr'] & {
+      border-radius: 0.286em 0 0 0.286em;
```
Matt-Butler: In terra mixins, we have some mixins that will handle ltr logic:
[border-radius start](https://github.com/cerner/terra-core/blob/master/packages/terra-mixins/src/Mixins.scss#L217)
[border-radius end](https://github.com/cerner/terra-core/blob/master/packages/terra-mixins/src/Mixins.scss#L234)

packages/terra-date-picker/src/DatePicker.scss
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    [dir='rtl'] & {
+      border-radius: 0 0.286em 0.286em 0;
+      margin-left: -0.071em;
```
Matt-Butler: [terra-margin-start](https://github.com/cerner/terra-core/blob/master/packages/terra-mixins/src/Mixins.scss#L329)

packages/terra-date-picker/src/DatePicker.scss
```diff
@@ -0,0 +1,54 @@
+@import '~terra-mixins';
+@import '../node_modules/react-datepicker/dist/react-datepicker.min';
```
Matt-Butler: We can't always guarantee that this relative path will work. I would update the syntax to 

@import '~react-datepicker/dist/react-datepicker.min';

This syntax is from the sass-loader we use with webpack: https://github.com/webpack-contrib/sass-loader#imports

packages/terra-date-picker/src/DateRange.jsx
```diff
@@ -0,0 +1,71 @@
...
+    super(props);
+    this.state = {
+      startDate: props.startDate,
+      endDate: props.endDate,
+    };
+    this.handleChangeStart = this.handleChangeStart.bind(this);
+    this.handleChangeEnd = this.handleChangeEnd.bind(this);
+  }
+
+  handleChange({ startDate, endDate }) {
```
Matt-Butler: If you feel like it, this function can use default props: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

packages/terra-date-picker/lib/DatePicker.scss
benbcai: Good to know. https://github.com/cerner/terra-core/commit/d1272119a4c8555ed3db41f88ef190d8e0002666

packages/terra-date-picker/src/DateRange.jsx
```diff
@@ -0,0 +1,71 @@
...
+    let startDateForRange = startDate || this.state.startDate;
+    let endDateForRange = endDate || this.state.endDate;
+
+    if (startDateForRange.isAfter(endDateForRange)) {
+      const tempDate = startDateForRange;
```
Matt-Butler: If you feel like it, this can use destructuring assignment: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

```
let startDateForRange = "march 20th";
let endDateForRange = "march 5th";

//  print initial values
console.log([startDateForRange, endDateForRange])

// swap values
[startDateForRange, endDateForRange] = [endDateForRange, startDateForRange]

//  print swapped values
console.log([startDateForRange, endDateForRange])
```

Up to you if you want to use it.

packages/terra-date-picker/src/DateInput.jsx
benbcai: Added prop documentation to this class and DateRange and moved propTypes to the top. https://github.com/cerner/terra-core/commit/d1272119a4c8555ed3db41f88ef190d8e0002666

packages/terra-date-picker/src/DatePicker.jsx
benbcai: withPortal is now removed since we are abstracting this using ResponsiveElement. https://github.com/cerner/terra-core/commit/d1272119a4c8555ed3db41f88ef190d8e0002666

packages/terra-date-picker/src/DatePicker.jsx
benbcai: withPortal is now removed. The props are now alphabetized: https://github.com/cerner/terra-core/commit/d1272119a4c8555ed3db41f88ef190d8e0002666

packages/terra-date-picker/src/DatePicker.jsx
benbcai: Updated to use inline class string and removed the classNames dependency. https://github.com/cerner/terra-core/commit/d1272119a4c8555ed3db41f88ef190d8e0002666

packages/terra-date-picker/src/DateRange.jsx
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  handleChangeEnd(endDate) {
+    this.handleChange({ endDate });
+  }
+
+  render() {
+    return (<div className="terra-DatePicker-range">
+      <DatePicker
+        {...this.props}
```
Matt-Butler: For our other components, we have been putting customProps (`...this.props`) last. I think this was to allow customProps to be last and override attributes. Can someone else confirm this is correct?

packages/terra-date-picker/src/DatePicker.jsx
benbcai: https://github.com/cerner/terra-core/commit/d1272119a4c8555ed3db41f88ef190d8e0002666

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -0,0 +1,173 @@
...
+    const userLocale = window.navigator.language;
+
+    // TODO: Get date format from date_util
+    const localMoment = moment();
+    localMoment.locale(userLocale);
+    const momentDateFormat = localMoment.localeData().longDateFormat('L');
+
+    const portalPicker =
+      (<ReactDatePicker
+        {...customProps}
```
Matt-Butler: `...customProps` should be the last prop

packages/terra-date-picker/src/DatePicker.scss
benbcai: Updated to use the mixins: https://github.com/cerner/terra-core/commit/c599197256ce3c511ea206ca5e87c4015f4f6fee

packages/terra-date-picker/src/DatePicker.scss
benbcai: I don't think the terra-margin-start mixin will work for this scenario. The reason being that I want to set margin-left on the input only in 'rtl' and don't set margin-right in 'ltr'. Basically what I want is set margin-left only in 'rtl' for the input and set margin-left only in 'ltr' for the calendar button.

packages/terra-date-picker/src/DatePicker.scss
benbcai: Updated: https://github.com/cerner/terra-core/commit/c599197256ce3c511ea206ca5e87c4015f4f6fee

packages/terra-date-picker/src/DateRange.jsx
benbcai: Updated: https://github.com/cerner/terra-core/commit/c599197256ce3c511ea206ca5e87c4015f4f6fee

packages/terra-date-picker/src/DateRange.jsx
benbcai: This is cool and condense the code to 1 line: https://github.com/cerner/terra-core/commit/c599197256ce3c511ea206ca5e87c4015f4f6fee

packages/terra-date-picker/src/DateRange.jsx
benbcai: Looks like there are some packages that put customProps first and some put it last. I'll bring this up in the war room so they can be consistent. I agree it should be last and updated it. https://github.com/cerner/terra-core/commit/c599197256ce3c511ea206ca5e87c4015f4f6fee

First:
https://github.com/cerner/terra-core/blob/master/packages/terra-progress-bar/src/ProgressBar.jsx#L47
https://github.com/cerner/terra-core/blob/master/packages/terra-slide-panel/src/SlidePanel.jsx#L74

Last:
https://github.com/cerner/terra-core/blob/master/packages/terra-clinical-item-view/src/Comment.jsx#L41
https://github.com/cerner/terra-core/blob/master/packages/terra-image/src/Image.jsx#L60

packages/terra-date-picker/src/DatePicker.jsx
benbcai: Updated: https://github.com/cerner/terra-core/commit/c599197256ce3c511ea206ca5e87c4015f4f6fee

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -0,0 +1,173 @@
...
+
+const defaultProps = {
+  isEndDateRange: false,
+  isStartDateRange: false,
+};
+
+class DatePicker extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
```
JakeLaCombe: Minor: Can you put a few line breaks in this area? It feels very cluttered, and a new line on lines 63 and 67 can help with some of the formatting.

benbcai: Added line breaks to clean up formatting. https://github.com/cerner/terra-core/commit/7870ad670ebe55044db8273441da4c5c5b8f95e5

packages/terra-date-picker/src/DatePicker.jsx
benbcai: I'm going to keep this as is until we have a better way for the Responsive Element to handle this.

packages/terra-date-picker/tests/nightwatch/components/DatePickerOnChange.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import DatePicker from '../../../lib/DatePicker';
+
+const handleOnChange = (selectedDate) => {
+  window.console.log('**handleOnChange** The selected date is: ', selectedDate);
+};
+
+const DatePickerOnChange = () => (
+  <div>
```
JakeLaCombe: Are the extra divs needed on these tests?

benbcai: Nope, I think it was copied/pasted from another component. Removed: https://github.com/cerner/terra-core/commit/6f069c2d15d4fef6e983c0e62b91a8b053041e48

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -0,0 +1,55 @@
+import React, { PropTypes } from 'react';
+import Button from 'terra-button';
+import IconCalendar from 'terra-icon/lib/icon/IconCalendar';
+
+const Icon = <IconCalendar />;
+
+const propTypes = {
```
emilyrohrbough: Because you do not provide default propTypes, should these be required?

benbcai: None of these are really required. If no props are provided, it would just render a date picker with no default date.

Issue #4, opened 2017-02-24
Investigate grid styles triggering horizontal scrollbar


bjankord:
# Issue Description
When viewing the grid demos in storybook and prior to storybook, a horizontal scrollbar displays on the window.

## Additional Details (optional)
The current implementation adds "grid gutters" via padding on both the left and right of columns. The outer padding on the first and last column should be canceled out by a negative left and right margin on the grid row. This is a popular approach taken by other frameworks. All of the other frameworks that use this approach are impacted by this issue as far as I could tell. 

This issue is usually hidden by a container element which adds padding equal to or greater than the amount of the grid gutter. The container essentially consumes the space of the negative margins, eliminating the need for a horizontal scrollbar.

In legacy terra, we apply grid gutter spacing by applying it as left padding on columns. The left padding on the first column is then canceled out by a negative left margin on the row. This approach seems to achieve the same visual without triggering a horizontal scrollbar. This link describes some of the logic behind the legacy terra grid styling around this behavior: https://csswizardry.com/2011/08/building-better-grid-systems/

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
No horizontal bar appears on window when using grid

## Current Behavior
Horizontal bar appears on window when using grid when there is no margin/padding on the html or body element.

## Steps to Reproduce

1. Navigate to page to view grid examples

## Environment

* Component Version: v1.0.0
* Browser Name and Version: All browsers
* Operating System and version (desktop or mobile):

PR #185, opened 2017-03-30
Release updates for grid: Update version and dependents

bjankord:
@cerner/terra

### DIFF
--- a/packages/terra-site/package.json
+++ b/packages/terra-site/package.json
@@ -33,7 +33,7 @@
     "terra-clinical-item-view": "^0.x",
     "terra-content": "^0.x",
     "terra-detail-view": "^0.x",
-    "terra-grid": "^1.0.0",
+    "terra-grid": "^2.0.0",
     "terra-i18n": "^0.x",
     "terra-icon": "^0.x",
     "terra-image": "^0.x",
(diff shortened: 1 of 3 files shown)


packages/terra-site/package.json
```diff
@@ -33,7 +33,7 @@
     "terra-clinical-item-view": "^0.x",
     "terra-content": "^0.x",
     "terra-detail-view": "^0.x",
-    "terra-grid": "^1.0.0",
+    "terra-grid": "^2.0.0",
```
StephenEsser: The terra-detail-view still has a DevDependency on terra-grid, this was used for examples before they moved to terra-site, but I think there are still tests utilizing the grid. Should terra-detail-view also be bumped, or is it something we can just update in the future?

bjankord: Good call. Versioning is a bit odd here. `terra-grid v1.0.0` does not contain react components. So while `terra-detail-view` denotes a dev dependency on `terra-grid: ^1.0.0` it was really referencing the code in terra-grid that would become `terra-grid v2.0.0` via npm-link. I'm open to talking about this more in our bi-weekly meetings. There is some functionality around our release script that led us to using this strategy.

Matt-Butler:
Just curious if this is an issue yet? https://github.com/cerner/terra-core/issues/4

bjankord:
@Matt-Butler #4 is still an issue. Fill land bug fix for it in a future release.

PR #186, opened 2017-03-30
Update ResponsiveElement examples to be interactive

StephenEsser:
### Summary
The previous ResponsiveElement examples animated the container size automatically so the user could see the different elements swapping out.

This made it difficult for verification and general behavior testing.

I updated the examples to have controls for changing the size of the container manually to get a better feel of how the component behaves.

screen shot 2017-03-30 at 3 04 35 pm

@cerner/terra


packages/terra-site/src/examples/responsive-element/Index.jsx
```diff
@@ -9,20 +9,44 @@ import { version } from 'terra-responsive-element/package.json';
...
 
 const ResponsiveElementExamples = () => (
   <div>
     <div id="version">Version: {version}</div>
     <Markdown id="readme" src={ReadMe} />
     <PropsTable id="props" src={ResponsiveElementSrc} />
     <h2>Responsive to Parent</h2>
-    <AnimatedResponsiveElement />
-    <Responsive />
+    <h4>All Breakpoints</h4>
```
Matt-Butler: nit: h3

Matt-Butler: same for the other headers please :)

packages/terra-site/src/examples/responsive-element/Index.jsx
```diff
@@ -9,20 +9,44 @@ import { version } from 'terra-responsive-element/package.json';
...
+    />
+    <br />
+    <h3>Default and Medium Breakpoints</h3>
+    <ResponsiveExample
+      defaultElement={<div>Default</div>}
+      medium={<div>Medium</div>}
+    />
+    <br />
+    <h3>Only Medium</h3>
+    <h5>Note: Does not render elements when smaller than medium.</h5>
```
Matt-Butler: Sorry, these need to be h4. The only reason I care is when we use accessibility tools this will be an issue. We need to use heading tags for semantic structuring, not styling. h6 must be used after h5, h5 must be used after h4, etc.

http://webaim.org/techniques/semanticstructure/#correctly

StephenEsser: You're killing me smalls

Matt-Butler: haha, sorry

StephenEsser: I removed the heading tag from these notes and used a basic paragraph, let me know if there is a more appropriate tag to use

Issue #188, opened 2017-03-31
ButtonGroup.Button should use Button


mjhenkes:
Update  `<ButtonGroup.Button />` to wrap a `<Button />` and offer a scoped down api. This will help alleviate concerns with maintaining styles / react implementation of buttons in two places, but still allow `<ButtonGroup.Button />` to be scoped down to an api that makes sense for a button within a button group. This should be a passive change.

PR #183, opened 2017-03-30
Terra base

bjankord:
### Summary
Adding `terra-base` component.

Thanks for contributing to Terra. 
@cerner/terra

[CONTRIBUTORS.md]: ../blob/master/CONTRIBUTORS.md
[License]: ../blob/master/LICENSE

PR #190, opened 2017-03-31
Make ButtonGroup.Button a wrapper for Button

jmsv6d:
### Summary
ButtonGroup.Button should be a higher order component that wraps terra-button with a restricted API.  #188 

@cerner/terra


packages/terra-button-group/package.json
```diff
@@ -27,11 +27,13 @@
   "peerDependencies": {
     "react": "^15.4.2",
     "react-dom": "^15.4.2",
-    "terra-mixins": "^1.0.0"
+    "terra-mixins": "^1.0.0",
+    "terra-button": "^0.1.0"
```
Matt-Butler: For components < 1.x, we have been versioning it like this:
```
"terra-button": "^0.x"
```
Which allows terra-button 0.1, 0.2.

This site is really useful for learning more about npm versioning:
https://semver.npmjs.com/

packages/terra-button-group/package.json
```diff
@@ -27,11 +27,13 @@
   "peerDependencies": {
     "react": "^15.4.2",
     "react-dom": "^15.4.2",
-    "terra-mixins": "^1.0.0"
+    "terra-mixins": "^1.0.0",
+    "terra-button": "^0.x"
```
StephenEsser: It looks like this missed was missed during the terra-base PR: https://github.com/cerner/terra-core/pull/183

I'd sync up with @bjankord to see if this should have a peerDependency of terra-base, it can probably be added in a different pull request if it doesn't really relate to this one

PR #191, opened 2017-03-31
Terra demographics banner

JakeLaCombe:
### Summary
This is the demographics banner for Terra.

### Additional Details
This is a barebones demographics banner without tags or link support. These components will be added in the future.

@cerner/terra


package.json
```diff
@@ -60,6 +60,7 @@
     "babel-plugin-transform-object-rest-spread": "^6.22.0",
     "babel-preset-es2015": "^6.18.0",
     "babel-preset-react": "^6.16.0",
+    "css-loader": "^0.27.3",
```
Matt-Butler: We can remove `css-loader` from the root package.json. This dependency is only used in webpack so we should only need this dependency in terra-site.

package.json
```diff
@@ -78,6 +79,7 @@
     "react-dom": "^15.4.2",
     "react-router": "^3.0.2",
     "shelljs": "^0.7.6",
+    "style-loader": "^0.16.0",
```
Matt-Butler: rm. same comment as above.

packages/terra-demographics-banner/jest/jest-transform-cache--Users-jl021230-Documents-PopulationHealth-terra-terra-core-packages-terra-demographics-banner-45d11b25ca1731bb00daaa0f4666f685/0e/DemographicsBannerDisplay_0eea0e4c2ad7fc5977abdd7d30346168
```diff
@@ -0,0 +1,169 @@
+/* istanbul ignore next */'use strict';var _createClass = function () {function defineProperties(target, props) {for (var i = 0; i < props.length; i++) {var descriptor = props[i];descriptor.enumerable = descriptor.enumerable || false;descriptor.configurable = true;if ("value" in descriptor) descriptor.writable = true;Object.defineProperty(target, descriptor.key, descriptor);}}return function (Constructor, protoProps, staticProps) {if (protoProps) defineProperties(Constructor.prototype, protoProps);if (staticProps) defineProperties(Constructor, staticProps);return Constructor;};}();var /* istanbul ignore next */_react = require('react'); /* istanbul ignore next */var _react2 = _interopRequireDefault(_react);
```
Matt-Butler: I think we can remove a couple of these files. Is this from a merge tool?

JakeLaCombe: Looks like I had a few cached jest tests that got pulled into the pull request. I removed them all here. https://github.com/cerner/terra-core/pull/191/commits/2aeae872d5e6f1851cdb0c14cca7c9706349bb2a

packages/terra-demographics-banner/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Demographics Banner
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-demographics-banner.svg)](https://www.npmjs.org/package/terra-demographics-banner)
+[![Build Status](https://travis-ci.org/cerner/terra-ui.svg?branch=master)](https://travis-ci.org/cerner/terra-ui)
+
+{insert description}
```
bjankord: Can you paste the description from `docs/README.md` here. "The demographics component is used to display demographic information about a person in a condensed, easy to read format."

packages/terra-demographics-banner/package.json
```diff
@@ -0,0 +1,61 @@
...
+  "bugs": {
+    "url": "https://github.com/cerner/terra-ui/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-ui#readme",
+  "devDependencies": {
+    "terra-props-table": "^0.1.0"
+  },
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
```
bjankord: Add `"terra-base": "^0.x",` to peerDependencies

packages/terra-demographics-banner/README.md
JakeLaCombe: Sure. Done here. https://github.com/cerner/terra-core/pull/191/commits/03c4ff1e008c487db836b804241264f278ae7453

packages/terra-demographics-banner/package.json
JakeLaCombe: I can do that, but it looks like the other packages are not doing at the moment. Should we do a global update of those as well?

packages/terra-demographics-banner/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "terra-mixins": "^1.0.0"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "react-intl": "^2.2.3",
+    "shortid": "^2.2.8",
+    "terra-i18n": "^0.0.0",
```
bjankord: npm does not treat matching of `^0.0.0` and `^1.0.0` versions the same when using the caret character.

This can be tested with https://semver.npmjs.com/.

Using the caret will update the package to the most recent major version. `^1.2.3` will match any `1.x.x` release matching version `1.2.3` or higher including `1.2.4`, `1.3.0`, `1.40.1`, but will hold off on `2.0.0`.

Though using the caret with a `0` based major version, the caret doesn't have any impact.

npm will only match `^0.0.0` to version `0.0.0`. 
npm will only match `^0.1.0` to version `0.1.0`.

To match all minor and patch releases under a `0` major release there are a couple of options. We can use `0.x` or `<1.0.0`.

We've been using `"^0.x"` for components on the `0` major version.

**tl;dr Can you update the terra-i18n,  terra-icon, and terra-toolkit packages to reference** `"^0.x"`

packages/terra-demographics-banner/package.json
bjankord: Lets do it just for this package now and we'll get another PR out to do it globally and get the generator updated.

packages/terra-demographics-banner/src/DemographicsBanner.scss
```diff
@@ -0,0 +1,71 @@
+/* stylelint-disable  */
+
+@import '~terra-mixins';
+@import './variables';
+@import './mixins';
+
+%terra-demographics-banner {
+  background: $demographics-background;
+  color: #fff;
+  display: -ms-flex;
```
bjankord: We've been working on the idea the consumers will have [autoprefixer configured](https://github.com/cerner/terra-core/blob/master/packages/terra-site/webpack.config.js#L49-L57) in their build tools. With this, I believe autoprefixer will handle adding this.

packages/terra-demographics-banner/src/DemographicsBanner.scss
```diff
@@ -0,0 +1,71 @@
...
+  display: flex;
+  margin-bottom: 10px;
```
bjankord: We want to use rem values for units most of the time. We currently don't have a mixin for rems so calculations will need to be manual. For conversion, `14px` is assumed to be computed font size on the html element.

Here's a quick conversion for 10px when root font size is computed 14px.
`10px` = `0.714rem`

packages/terra-demographics-banner/src/DemographicsBanner.scss
```diff
@@ -0,0 +1,71 @@
...
+    }
+  }
+}
+
+.terra-DemographicsBanner {
+  @extend %terra-demographics-banner;
+}
+
+.terra-DemographicsBanner--deceased {
+  @extend %terra-demographics-banner;
```
bjankord: Rather than extension of styles via Sass extend, prefer extension of styles through a base class and additional modifier classes which can be composed in the html. This is to help reduce total file size of CSS.

e.g.
```html
<!-- Base Demographics Banner -->
<div class="terra-DemographicsBanner">...</div>

<!-- Deceased Demographics Banner -->
<div class="terra-DemographicsBanner terra-DemographicsBanner--deceased">...</div>
```

```css
.terra-DemographicsBanner {
  base styles ......
}

.terra-DemographicsBanner--deceased {
  background-color: $demographics-background-deceased;
}
```

packages/terra-demographics-banner/src/DemographicsBanner.scss
```diff
@@ -0,0 +1,71 @@
...
+  background: $demographics-background-deceased;
+}
+
+.terra-DemographicsBanner--row {
+  display: flex;
+
+  span + span {
+    display: inline-block;
+    margin-left: 5px;
```
bjankord: For directionality based styles, like margin-left, use the directionality based mixin:

```css
span + span {
  @include terra-margin-start(0.357rem); // 5px
  display: inline-block;
}
```

packages/terra-demographics-banner/src/DemographicsBanner.scss
```diff
@@ -0,0 +1,71 @@
...
+}
+
+.terra-DemographicsBanner--content {
+  flex-grow: 1;
+  margin: auto 10px;
+}
+
+.terra-DemographicsBanner--application-content,
+.terra-DemographicsBanner--identifiers {
+  margin-left: auto;
```
bjankord: For directionality based styles, like margin-left, use the directionality based mixin

packages/terra-demographics-banner/src/DemographicsBanner.scss
```diff
@@ -0,0 +1,71 @@
...
+}
+
+[dir='rtl'] {
```
bjankord: Using the directionality based mixins should allow you to delete the styles below.

packages/terra-demographics-banner/package.json
```diff
@@ -0,0 +1,61 @@
...
+    "release:major": "npm test && node ../../scripts/release/release.js major",
+    "release:minor": "npm test && node ../../scripts/release/release.js minor",
+    "release:patch": "npm test && node ../../scripts/release/release.js patch",
+    "test": "npm run test:spec && npm run test:nightwatch-default",
+    "test:spec": "$(cd ..; npm bin)/jest --config ../../jestconfig.json",
+    "test:all": "npm run test:nightwatch-default && npm run test:nightwatch-chrome && npm run test:nightwatch-firefox && npm run test:nightwatch-safari",
+    "test:nightwatch-default": "node ./node_modules/terra-toolkit/lib/scripts/nightwatch.js",
+    "test:nightwatch-chrome": "node ./node_modules/terra-toolkit/lib/scripts/nightwatch.js chrome",
+    "test:nightwatch-firefox": "node ./node_modules/terra-toolkit/lib/scripts/nightwatch.js firefox",
+    "test:nightwatch-safari": "node ./node_modules/terra-toolkit/lib/scripts/nightwatch-non-parallel.js safari"
```
StephenEsser: I don't know if the generator is updated to include some of the recent changes to testing harness Ryan implemented. The syntax for these scripts has changed a little bit.

https://github.com/cerner/terra-core/blob/master/packages/terra-button/package.json#L37

Here are pull request implementing the differences: 
https://github.com/cerner/terra-core/pull/180
https://github.com/cerner/terra-core/pull/167

@ryanthemanuel

packages/terra-demographics-banner/src/DemographicsBanner.jsx
```diff
@@ -0,0 +1,122 @@
...
+  deceasedDate: PropTypes.string,
+  /**
+   * Additional Details to display in the banner.
+   */
+  additionalDetails: PropTypes.node,
+  /**
+   * Additional rows of key value pairs to add to the demographics banner. The rows will be added after the person rows.
+   */
+  applicationRows: PropTypes.array,
+};
```
StephenEsser: We had talked about sorting these first by required / non-required and then again alphabetically. Is this a standard we want to start enforcing?

packages/terra-demographics-banner/src/DemographicsBanner.jsx
```diff
@@ -0,0 +1,122 @@
...
+      demographicsMessages: require(`./translations/${props.locale}.js`),
+    };
+  }
+
+  componentDidMount() {
+    i18nLoader(this.props.locale, this.setState, this);
+  }
+
+  render() {
+    if (this.state.load) {
```
StephenEsser: Is this to account for the double render caused by the i18n loader setting state after mounting?

packages/terra-demographics-banner/src/DemographicsBanner.jsx
```diff
@@ -0,0 +1,122 @@
...
+      return (
+        <I18nProvider locale={this.state.locale} messages={Object.assign(this.state.messages, this.state.demographicsMessages)}>
```
StephenEsser: Something about this seems really awkward. If Object.assign is modifying `this.state.messages` by copying values over from demoGraphicsMessages I think this is dangerous. Where is messages being declared? Or is state being set initialized and set here without calling set state? 

https://facebook.github.io/react/docs/react-component.html
```
Never mutate this.state directly, as calling setState() afterwards may replace the mutation you made. Treat this.state as if it were immutable.
```

I might be misinterpreting what this is doing

packages/terra-demographics-banner/src/DemographicsBanner.jsx
```diff
@@ -0,0 +1,122 @@
...
+            gender={this.props.gender}
+            locale={this.props.locale}
+            identifiers={this.props.identifiers}
+            dateOfBirth={this.props.dateOfBirth}
+            gestationalAge={this.props.gestationalAge}
+            postMenstrualAge={this.props.postMenstrualAge}
+            deceasedDate={this.props.deceasedDate}
+            additionalDetails={this.props.additionalDetails}
+            applicationRows={this.props.applicationRows}
+          />
```
StephenEsser: It might be a little messy deconstructing all of these but I think this banner should also pass custom props down to the display to be consistent with the other components. I think it's likely a user will want to add custom classes / ids / events to the banner. Opening up custom props will let them bind an onclick and navigate, I think a link was something the original tech design called out, and an onclick would open the door at least temporarily for consumers to navigate themselves.

packages/terra-demographics-banner/src/DemographicsBanner.scss
```diff
@@ -0,0 +1,71 @@
...
+
+%terra-demographics-banner {
+  background: $demographics-background;
+  color: #fff;
+  display: -ms-flex;
+  display: flex;
+  margin-bottom: 10px;
+  padding: 10px;
+
+  h1 {
```
StephenEsser: Should we add a suit css class for this?

packages/terra-demographics-banner/tests/nightwatch/demographics-banner-spec.js
```diff
@@ -0,0 +1,76 @@
+/* eslint-disable no-unused-expressions */
+
+const screenshot = require('terra-toolkit').screenshot;
+
+module.exports = {
+  afterEach: (browser, done) => {
```
StephenEsser: I don't know why the build is still passing without these changes, but if you inspect the output of the travis CI build, it's throwing a bunch of errors near the bottom ( I think this branch doesn't have the changes to the button ). These tests will throw the same errors without these changes.

( Copy pasted from above )

I don't know if the generator is updated to include some of the recent changes to testing harness Ryan implemented. The syntax for these scripts has changed a little bit.

https://github.com/cerner/terra-core/blob/master/packages/terra-button/package.json#L37

Here are pull request implementing the differences: 
https://github.com/cerner/terra-core/pull/180
https://github.com/cerner/terra-core/pull/167

packages/terra-demographics-banner/src/DemographicsBanner.jsx
StephenEsser: You might consider just spreading props here actually.

```
<DemographicsBannerDisplay {...this.props} />
```

packages/terra-demographics-banner/src/DemographicsBanner.jsx
bjankord: To enforce it, we need the ability to lint it. It looks like the eslint react plugin [we use](https://github.com/cerner/terra-core/blob/master/.eslintrc#L4) has a rule for this: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-prop-types.md

packages/terra-demographics-banner/src/DemographicsBanner.jsx
bjankord: We are exploring different solutions to solve this issue.

packages/terra-demographics-banner/src/DemographicsBanner.scss
JakeLaCombe: Fixed it here. https://github.com/cerner/terra-core/pull/191/commits/ed668501d50053087fff3d92cfc4675fb7f487ee

packages/terra-demographics-banner/src/DemographicsBanner.jsx
JakeLaCombe: That sounds correct. From talking with @windse7en, this is needed for that purpose I believe.

packages/terra-demographics-banner/src/DemographicsBannerDisplay.jsx
```diff
@@ -0,0 +1,168 @@
...
+
+    return null;
+  }
+
+  render() {
+    let mainClass = 'terra-DemographicsBanner';
+
+    if (this.props.deceasedDate) {
+      mainClass += '--deceased';
+    }
```
JakeLaCombe: Fixed right here. https://github.com/cerner/terra-core/pull/191/commits/ed668501d50053087fff3d92cfc4675fb7f487ee

packages/terra-demographics-banner/tests/nightwatch/demographics-banner-spec.js
JakeLaCombe: I think we may need to update the generator. I checked out master, and I see the same errors occuring inside terra-toolkit.

StephenEsser: It looks like these updates were checked into the generator. When you checked out the new generator did you first uninstall the old one?
https://github.com/cerner/generator-terra-module/pull/30
```
npm uninstall -g generator-terra-module
```
If this isn't run before installing a new version it won't pull the updates

StephenEsser: The error being thrown is actually the default nightwatch test added by an older version of the generator.

terra-base also has this issue: https://github.com/cerner/terra-core/blob/master/packages/terra-base/tests/nightwatch/base-spec.js

packages/terra-demographics-banner/package.json
JakeLaCombe: https://github.com/cerner/terra-core/pull/191/commits/88e82fd67f12c4a9251bd1e5d0c4af87012faa0f

packages/terra-demographics-banner/package.json
JakeLaCombe: Updated here. https://github.com/cerner/terra-core/pull/191/commits/88e82fd67f12c4a9251bd1e5d0c4af87012faa0f

packages/terra-demographics-banner/package.json
JakeLaCombe: Looks like I had a code conflict on my branch. I fixed it here. https://github.com/cerner/terra-core/pull/191/commits/de319ec334c14f16107f08769e49696070cdc4d3

packages/terra-demographics-banner/src/DemographicsBanner.scss
JakeLaCombe: Removed here. https://github.com/cerner/terra-core/pull/191/commits/4e58dc27a98a4200169e20de81c5ce5068a209ed

packages/terra-demographics-banner/src/DemographicsBanner.scss
JakeLaCombe: https://github.com/cerner/terra-core/pull/191/commits/886657980c81a1f4e37e8f089ecb95d880166791

packages/terra-demographics-banner/src/DemographicsBanner.scss
JakeLaCombe: https://github.com/cerner/terra-core/pull/191/commits/886657980c81a1f4e37e8f089ecb95d880166791

packages/terra-demographics-banner/src/DemographicsBanner.scss
JakeLaCombe: https://github.com/cerner/terra-core/pull/191/commits/886657980c81a1f4e37e8f089ecb95d880166791

packages/terra-demographics-banner/src/DemographicsBanner.scss
JakeLaCombe: Fixed Here - https://github.com/cerner/terra-core/pull/191/commits/8b47699ff8a23cf0ffaf17d4406306a8ab0596ef

packages/terra-demographics-banner/src/DemographicsBanner.scss
```diff
@@ -0,0 +1,71 @@
+/* stylelint-disable  */
+
+@import '~terra-mixins';
+@import './variables';
+@import './mixins';
+
+%terra-demographics-banner {
+  background: $demographics-background;
+  color: #fff;
```
JakeLaCombe: Fixed Here. Made themeable as well. https://github.com/cerner/terra-core/pull/191/commits/8b47699ff8a23cf0ffaf17d4406306a8ab0596ef

packages/terra-demographics-banner/src/DemographicsBanner.jsx
JakeLaCombe: I mainly didn't do that, because the linter wanted to ensure all props are used. Given that we have another class checking that, I'll just lint the other file instead.

https://github.com/cerner/terra-core/pull/191/commits/539802755f0da2e861c744c8ad3d4f3299bf0f71

packages/terra-demographics-banner/tests/nightwatch/demographics-banner-spec.js
JakeLaCombe: Fixed it here. https://github.com/cerner/terra-core/pull/191/commits/b94f4efae5adff93b8c7fcf4356ced63480f3c47

packages/terra-demographics-banner/package.json
JakeLaCombe: Updated the syntax here: https://github.com/cerner/terra-core/pull/191/commits/b94f4efae5adff93b8c7fcf4356ced63480f3c47

StephenEsser: I think you might have accidentally updated the wrong package json

JakeLaCombe: Yup. That was a sublime 2 view layout error I made :(

https://github.com/cerner/terra-core/pull/191/commits/cf5123d5d9621267803b73686ef98a257442c19b

packages/terra-demographics-banner/src/DemographicsBanner.scss
JakeLaCombe: Agreed.

https://github.com/cerner/terra-core/pull/191/commits/580347b7cdbcab20a2987db336175b5a0d1f5fed

packages/terra-demographics-banner/src/DemographicsBanner.jsx
JakeLaCombe: Came across some warnings. I fixed them here. https://github.com/cerner/terra-core/pull/191/commits/5f5ac87e1d8e11d7072fa7c1196dff0eae9bd9a3

packages/terra-demographics-banner/src/DemographicsBanner.jsx
JakeLaCombe: I did fix the issue of updating the state in this commit. https://github.com/cerner/terra-core/pull/191/commits/405669b4fc219012cb27d3362d5908639263da35.

Still discussing i18n approach.

package.json
JakeLaCombe: https://github.com/cerner/terra-core/pull/191/commits/1eb41ad91917c5cd4583a264c8e9b688bba97d44

package.json
JakeLaCombe: https://github.com/cerner/terra-core/pull/191/commits/1eb41ad91917c5cd4583a264c8e9b688bba97d44

packages/terra-demographics-banner/src/DemographicsBanner.jsx
JakeLaCombe: Alphabetized here. https://github.com/cerner/terra-core/pull/191/commits/84bfd2759b32558a6f825be281af43d7339931ea

packages/terra-demographics-banner/docs/themeable-variables.md
```diff
@@ -0,0 +1,8 @@
+# Themeable Variables
+
+Variables to `theme` the terra-demographics-banner component.
+
+| Variable Name                       | Default value   | Usage                                                        |
+|-------------------------------------|-----------------|--------------------------------------------------------------|
+| $demographics-background            | $terra-blue-100 | Background of the Demographics Banner.                       |
+| $demographics-background-deceased   | #000            | Background of the Demographics Banner for a deceased person. |
```
Matt-Butler: $terra-dark

packages/terra-demographics-banner/lib/DemographicsBanner.scss
```diff
@@ -0,0 +1,49 @@
+/* stylelint-disable  */
```
Matt-Butler: remove

packages/terra-demographics-banner/lib/DemographicsBanner.scss
```diff
@@ -0,0 +1,49 @@
...
+
+.terra-DemographicsBanner {
+  background: $demographics-background;
+  color: $demographics-foreground;
+  display: flex;
+  margin-bottom: 0.714em;
+  padding: 0.714em;
+}
+
+.terra-DemographicsBanner--person-name {
```
Matt-Butler: nit: I think these components are decendants for the demographics banner: https://github.cerner.com/standards/rfc/blob/master/published/0004-css-naming-convention.md#descendent
```
.terra-DemographicsBanner--person-name
```
to
``` 
terra-DemographicsBanner-person-name
```

packages/terra-demographics-banner/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "UI"
+  ],
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-ui/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-ui#readme",
+  "devDependencies": {
+    "terra-props-table": "^0.1.0"
```
Matt-Butler: I don't think this component will need this as a devDependency.  This devDependency will exist in `terra-site`

packages/terra-demographics-banner/src/DemographicsBanner.jsx
```diff
@@ -0,0 +1,112 @@
+/* eslint-disable global-require, import/no-dynamic-require, react/no-unused-prop-types */
```
Matt-Butler: Why do we need to disable all these linting rule?

packages/terra-demographics-banner/src/DemographicsBanner.jsx
```diff
@@ -0,0 +1,112 @@
...
+const defaultProps = {
+  additionalDetails: null,
+  age: null,
+  applicationRows: [],
+  dateOfBirth: null,
+  deceasedDate: null,
+  gender: null,
+  gestationalAge: null,
+  identifiers: {},
+  locale: 'en',
```
Matt-Butler: This is dangerous to include as a default.  If a locale is provided in the application, but not provided for this component, this component will default to `en` and we may have a mismatch in locale.

packages/terra-demographics-banner/src/DemographicsBanner.scss
```diff
@@ -0,0 +1,49 @@
+/* stylelint-disable  */
```
Matt-Butler: rm

packages/terra-demographics-banner/src/DemographicsBannerDisplay.jsx
```diff
@@ -0,0 +1,188 @@
+import React, { PropTypes } from 'react';
+import { FormattedHTMLMessage, injectIntl, intlShape } from 'react-intl';
+import shortid from 'shortid';
```
Matt-Butler: Relevant conversation about using generated keys here: https://github.com/cerner/terra-core/pull/177#discussion_r109933410

packages/terra-demographics-banner/src/DemographicsBanner.jsx
JakeLaCombe: no-dynamic-require and global-require causes lint here, but is needed to load the appropriate translations. https://github.com/cerner/terra-core/pull/191/files#diff-c554f9bb4d56617bb2a5b86d37b215f2R85

Since I am passing all the props straight to the banner, react/no-unused-prop-types is throwing a lint warning.

packages/terra-demographics-banner/lib/DemographicsBanner.scss
```diff
@@ -0,0 +1,49 @@
+/* stylelint-disable  */
+
+@import '~terra-mixins';
+@import './variables';
+@import './mixins';
```
Matt-Butler: We can remove mixins import and file if it is not used

packages/terra-demographics-banner/src/DemographicsBanner.scss
JakeLaCombe: https://github.com/cerner/terra-core/pull/191/commits/75c7f95b4cd3393c9dcfe84ea2c669f21035589c

packages/terra-demographics-banner/docs/themeable-variables.md
Matt-Butler: Please add this line too : https://github.com/cerner/terra-core/pull/191/files#diff-31058e8b65ca289134f485c9efa0ed3dR2

packages/terra-demographics-banner/src/DemographicsBannerDisplay.jsx
JakeLaCombe: I'll bring this up in war room tomorrow. I'm not sure how to handle this scenario, and so far this is the best solution I found for generating a unique key for this element.

packages/terra-demographics-banner/src/DemographicsBanner.jsx
Matt-Butler: We can do [eslint-disable-next-line](https://github.com/cerner/terra-core/blob/7f1d3732290e7335506c5a630e2ac616361cd9ff/packages/terra-site/src/examples/button/ButtonOnClick.jsx#L5) on line 84 so it will continue to lint the rest of the file.

Matt-Butler:
How is content supposed to wrap for this component? Should we enforce some seperation from the left and right columns?
@mmalaker 

screen shot 2017-04-06 at 5 15 24 pm

JakeLaCombe:
I had to make some changes to get rid of the I18n stuff, so I opened up a new pull request. https://github.com/cerner/terra-core/pull/204

PR #202, opened 2017-04-10
Lock react 15.4.2

Matt-Butler:
### Summary
The upgrade to React 15.5 is non-passive. I am locking react down at 15.4.2 until we update our packages to be compatible with 15.5.

@cerner/terra


packages/terra-base/package.json
```diff
@@ -28,8 +28,7 @@
     "react-dom": "^15.4.2"
   },
   "dependencies": {
-    "classnames": "^2.2.5",
-    "react": "^15.4.2"
```
Matt-Butler: I am removing `react` as a dependency for several of our packages. I am noticing this is forcing multiple nested versions of `react` to be installed.

JakeLaCombe:
+1

packages/terra-base/package.json
bjankord: I have some reservations about removing react as a dependency. Let's talk this some more tomorrow

Matt-Butler: Here is the issue I am seeing:
1) lock react at 15.4.2 in the root directory and in terra-site.
2) `npm run clean:install` 
3) Inspect node_modules in any package that has a dependency on react. You will see a react directory with react 15.5.x.  This gets confusing quickly because I am unsure which version of react each of these packages is using during testing / runtime.

We can discuss more tomorrow.

Matt-Butler: This is also the same approach cloudflare-ui is taking: https://github.com/cloudflare/cf-ui/blob/master/packages/cf-component-callout/package.json#L14

emilyrohrbough:
+1

PR #212, opened 2017-04-12
Remove detail view from the core repo.

dkasper-was-taken:
### Summary
With the detail view moving to the clinical repo, we can remove it from the core repo.


emilyrohrbough:
+1

Issue #205, opened 2017-04-11
terra-table selected rows accessibility


Matt-Butler:
Selectable table rows are not keyboard accessible.

bjankord:
On #214 it was mentioned, "We might also need to add an aria attribute so screen readers can determine if a row is selected."

I believe adding `aria-selected="true"` to the selectable table row element, `<tr>`, is how we want to convey this to screen-readers. We'll want to toggle that to `aria-selected="false"` when the row is unselected.

More info:
* https://www.w3.org/TR/wai-aria/states_and_properties#aria-selected
* http://www.ssbbartgroup.com/blog/how-not-to-misuse-aria-states-properties-and-roles/

Matt-Butler:
I started this branch which is adding aria-selected to the table row. Voice-over is not reading the selected state. I'll continue to look into the issue.

https://github.com/cerner/terra-core/tree/accessibility-attributes

PR #214, opened 2017-04-13
Table keyboard selection

emilyrohrbough:
### Summary
Added onKeyDown prop to terra-table to navigate rows via 'tab' and select rows via 'enter'. Brett Jankord wanted the same key interactions as on button. 

References issue #205.


tbiethman:
Can we test these changes using nightwatch?

bjankord:
I just noticed that row selection is happening on any key down, like `tab`, `up arrow`, `down arrow`. We should scope this to just the `enter key`.

I believe adding the following if condition around [these lines](https://github.com/cerner/terra-core/blob/ea1e5226ae2a8460ee8870479695797939654241/packages/terra-table/src/SingleSelectableRows.jsx#L72-L78) will make it so only `enter key` makes a selection:

```
if (event.nativeEvent.keyCode === Number(13)) {
  ...
}
```
We'll want to test this in our supported browsers.

ryanthemanuel:
I agree with Tyler.  Let's see if we can nightwatch test this.  Take a look at this API: http://nightwatchjs.org/api/keys.html

I haven't used it yet, but it might lead in the right direction.

Matt-Butler:
We might also need to add an aria attribute so screen readers can determine if a row is selected.

bjankord:
@Matt-Butler Let's tackle that in a separate PR.

packages/terra-table/src/SingleSelectableRows.jsx
```diff
@@ -60,10 +65,26 @@ class SingleSelectableRows extends React.Component {
...
+
+    return (event) => {
+      if (event.nativeEvent.keyCode === Number(13)) {
+        if (row.props.isSelectable !== false && this.shouldHandleSelection(index)) {
+          this.handleSelection(event, index);
+        }
+
+        if (initialOnKeyDown) {
+          initialOnKeyDown(event);
+        }
```
tbiethman: We always want to execute the initialOnKeyDown function, even if the key pressed isn't Enter. We need to break this if-block out of the key check.

Matt-Butler: react-portal has the same sort of logic here: https://github.com/tajo/react-portal/blob/master/lib/portal.js#L116-L120

It might be a good example to reference.

emilyrohrbough: [Fixed Here](https://github.com/cerner/terra-core/pull/214/commits/6ea79e3907715a2787754f89d01e54e23d263d66#diff-7e5169a43d403abe170ac96e59213836R123)

PR #216, opened 2017-04-14
fix image examples

Matt-Butler:
### Summary
Updating image examples so the demo matches the description.
@cerner/terra


Matt-Butler:
image_fluid_after_update
image_after_update

These are the two updated examples

mmalaker:
Thanks Matt!

PR #219, opened 2017-04-14
Added customProps support and translations for demographic banner

windse7en:
### Summary
Add `customPorps` support in `I18nProvider` and add translations for the demographic banner.


windse7en:
Adding `customProps` to I18nProvider will help us to solve this issue in `react-intl` : https://github.com/yahoo/react-intl/wiki/Components#using-react-intl-with-react-native

packages/terra-i18n/lib/translations/en-US.js
```diff
@@ -11,7 +11,12 @@ function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { de
 (0, _reactIntl.addLocaleData)(_en2.default);
 
 var messages = {
-  'Terra.ajax.error': 'This content failed to load.'
+  'Terra.ajax.error': 'This content failed to load.',
+  'Terra.demographicsBanner.dateOfBirth': 'DOB: ',
```
Matt-Butler: Most of these translations will be concatenated with a value.  It is bad practice to concatenate translations: 
http://www.emreakkas.com/internationalization/10-internationalization-tips-for-developers-i18n-checklist

Instead, most of these translations should have a variable for value.

windse7en: I fully agree with the idea mentioned in that blog. It's a different case in here. In here, this translation is just the translation string value for the label. A good example in demographics banner is in this link: https://github.com/cerner/terra-core/blob/Terra-Demographics-Banner/packages/terra-demographics-banner/src/translations/en-US.js#L2
```
'DOB: ' + DOBValueComponent
```

Matt-Butler: +1

packages/terra-i18n/src/I18nProvider.jsx
```diff
@@ -18,13 +18,26 @@ const propTypes = {
...
+    messages,
+    ...customProps
+  } = props;
+
+  return (
+    <IntlProvider
+      {...customProps}
+      locale={props.locale}
+      key={props.locale}
+      messages={props.messages}
```
benbcai: nit: Could just use the const `locale` and `messages` instead of `props.locale` and `props.messages`.

packages/terra-i18n/lib/translations/en-US.js
JakeLaCombe: I wasn't 100% certain on our best solution. Here is how blue_steel handles those concerns

    %b= dob
    - if deceased
      = "#{t('blue_steel.demographics.deceased')}:"
      %b= dod
    - if ga
      = "#{t('blue_steel.demographics.ga')}:"

Not sure 100% if it's best to keep the : as part of the message, or to include it in the markup of the application.

packages/terra-i18n/src/I18nProvider.jsx
windse7en: Nice catch: fix the props usage in this commit: https://github.com/cerner/terra-core/pull/219/commits/af32ffbfa09fda6d00b31cac83a5872477459641#diff-65d50c29354b491f8aa1ed537bde19baR23

packages/terra-i18n/src/I18nProvider.jsx
```diff
@@ -18,13 +18,27 @@ const propTypes = {
...
-    </div>
-  </IntlProvider>
-);
+const I18nProvider = (props) => {
+  const {
+    children,
+    locale,
+    messages,
+    ...customProps
+  } = props;
```
StephenEsser: These props could be spread in the definition above.

```
const I18nProvider = ({ children, locale, messages, ...customProps }) => (
  <IntlProvider
    {...customProps}
    locale={locale}
    key={locale}
    messages={messages}
  >
    <div>
      {children}
    </div>
  </IntlProvider>
);
```
I believe this is how most of our functional components are handling props

PR #221, opened 2017-04-17
Selectable table aria attributes

Matt-Butler:
### Summary
Adding `aria-selected` attribute to selectable rows.

If a row is selected, aria-selected = true
If a row is not selected, aria-selected = false

@cerner/terra

### DIFF
--- a/packages/terra-table/src/TableRow.jsx
+++ b/packages/terra-table/src/TableRow.jsx
@@ -53,6 +53,8 @@ const TableRow = ({
     customProps.className,
   ]);
 
+  customProps['aria-selected'] = isSelected ? true : null;
+
   const cloneChildren = cloneChildItems(children, height);
   return (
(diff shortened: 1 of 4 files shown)


packages/terra-table/src/TableRow.jsx
```diff
@@ -53,6 +53,8 @@ const TableRow = ({
     customProps.className,
   ]);
 
+  customProps['aria-selected'] = isSelected ? true : false;
+
   const cloneChildren = cloneChildItems(children, height);
   return (
     <tr {...customProps} className={rowClassNames}>
```
tbiethman: I know the linter threw a fit about this, but I too would prefer to not mutating customProps and just inlining the attribute like `<tr {...customProps} aria-selected={isSelected} className={rowClassNames}>`

Matt-Butler: +1

Issue #164, opened 2017-03-22
Add px to rem sass mixin.


bjankord:
# Issue Description
Most designers / developers think in pixels, however we want to use rems to ensure our UI is scalable. Adding a mixin for converting pixels to rems allows people to think and work in pixels, but output rems.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
- Ability to use rem values and know their computed pixel value

## Current Behavior
Using rem values, we have no idea of their computed value unless we actually compute them.

bjankord:
There are some complexities around how this mixin would be used in relation to bi-directional styles. Here are some ideas that come to mind for how we can handle bidi styling with the rem mixin.

1. Idea 1: Make the rem mixin take a single property (font-size, padding-left, right, border-left, etc) and a single value (10px, 20px, etc). With this approach, the bidi mixins would wrap the rem mixin. So using margin-start(30px) would generate the ltr and rtl styles in rem units. This works for the majority of our use-cases, but [there are times](https://github.com/cerner/terra-core/pull/224#discussion_r112778189) where we may want to use bidi mixins, but not use rem values. 

We could add an additional property to the bidi mixins to control if they output rem units or output the unit passed to it. This might look like `margin-start(3em, false)` which would output `margin-left: 3em;` instead of rem units.

2. Idea 2: This idea takes the opposite approach, rather than the bidi mixins wrapping the rem mixin, the rem mixin would wrap the bidi-mixins. This adds a lot of complexity as we would want to maintain the current bidi mixins for when we dont want rem units as noted above. We would need bidi logic inside of the rem mixin. This means checks to see what css property was passed in to determine what to return. For example, with `rem(margin-left, 30px)`, we'd need to check the property, see which direction it specified and output the correct bidi styles.

3. Idea 3: Seeing the complexity associated between between bidi mixins and rem mixins, I prefer a much more simple solution compared to the two ideas above. This idea is we set the html font-size to `font-size: 62.5%`. This allows us to use rem units but thinking in pixels trivial. `1.6rem` = `16px`, `2.4rem` = `24px`. This also avoids having to pass a variable to mixins to [determine the root font element size](https://github.com/cerner/terra-core/blob/9dcf5c29a4b4cbd450f49b892dfd5f6e12335b70/packages/terra-mixins/src/Mixins.scss#L548-L550).

bjankord:
Here is the associated code change needed for idea 3:

[...]

PR #224, opened 2017-04-17
Modal

Matt-Butler:
### Summary
Code Review for Modal component

I wanted to get this out so people can start looking at it. I am still going to add documentation to this code review.

I am also seeing this warning in the console when viewing nightwatch tests. I will fix that as well.
```
Warning: Exception thrown by hook while handling onSetChildren: Invariant Violation: Item must have been set
```

@cerner/terra


packages/terra-modal/lib/Modal.scss
```diff
@@ -0,0 +1,31 @@
...
+.terra-Modal {
+  background-color: #fff;
+  border-radius: 5px;
+  bottom: 10px;
+  display: block;
+  left: 10px;
+  max-height: 100%;
+  outline: none;
+  overflow-x: hidden;
+  overflow-y: auto;
```
Matt-Butler: screen shot 2017-04-17 at 1 28 26 pm

I am on the fence if we should include this for the modal. If a HOC modal dialog does not exist, this will enable overflow to scroll. However, if a HOC does exist, there is the possibility of having 2 scroll bars.

bjankord: Let's add an `isScrollable` prop that adds the `overflow-y: auto` style when set to true. This allows us to handle both cases described above.

Matt-Butler: +1, I'll add that

packages/terra-modal/src/ModalOverlay.jsx
```diff
@@ -0,0 +1,17 @@
...
+
+const propTypes = {
+  classNameOverlay: PropTypes.string,
+};
+
+const defaultProps = {
+  classNameOverlay: null,
+};
+
+const ModalOverlay = ({ classNameOverlay, ...customProps }) => <div className={classNameOverlay} {...customProps} />;
```
jmsv6d: This component seems unnecessary.

Matt-Butler:
```
Warning: Exception thrown by hook while handling onSetChildren: Invariant Violation: Item must have been set
```
I think this issue is resolved in react 15.5: https://github.com/facebook/react/issues/8487

packages/terra-modal/src/ModalOverlay.jsx
Matt-Butler: I kept this logic separate in case we want to create an overlay component. That way it would be easy to switch out.

bjankord:
@Matt-Butler In regards to the warning thats being logged, do we know what code / commit is causing this. I'm curious how the CI tests pass if this is being logged, unless its a console.warn statement that is triggered instead of a console.error statement.

Are we thinking React 15.5 is a hard dependency for modal to fix this issue?

packages/terra-modal/src/Modal.jsx
```diff
@@ -0,0 +1,142 @@
...
+  /**
+   * If set to true, the modal will be fullscreen on all breakpoint sizes
+   **/
+  isFullscreen: PropTypes.bool,
+  /**
+   * If set to true, the modal will rendered as opened
+   **/
+  isOpened: PropTypes.bool,
+  /**
+   * String that labels the modal for screen readers
```
tbiethman: I think this comment is out of date.

packages/terra-modal/lib/Modal.scss
```diff
@@ -0,0 +1,34 @@
+@import '~terra-mixins';
+
+.terra-Modal {
+  background-color: #fff;
```
StephenEsser: Should we enforce the use of terra-legacy-theme colors? 
https://github.com/cerner/terra-core/blob/master/packages/terra-legacy-theme/src/_palette.scss#L67

packages/terra-modal/lib/Modal.scss
```diff
@@ -0,0 +1,34 @@
...
+  border-radius: 5px;
+  bottom: 10px;
+  display: block;
+  left: 10px;
+  max-height: 100%;
+  outline: none;
+  overflow: hidden;
+  position: fixed;
+  right: 10px;
+  top: 10px;
```
StephenEsser: Was a decision made about using px units vs rem? Will we just swing back through and use the new mixin for converting these when it's available? 
https://github.com/cerner/terra-core/issues/164

packages/terra-modal/package.json
```diff
@@ -0,0 +1,56 @@
...
+  },
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "focus-trap-react": "^3.0.2",
+    "terra-mixins": "^1.0.0",
+    "react-portal": "^3.0.0"
```
StephenEsser: Does this component need an explicit terra-base dependency? 
https://github.com/cerner/terra-core/pull/251

packages/terra-modal/src/Modal.jsx
```diff
@@ -0,0 +1,142 @@
...
+   **/
+  ariaLabel: PropTypes.string.isRequired,
+  /**
+   * This callback is called when the closing event is triggered but it prevents normal removal from the DOM. So, you can do some DOMNode animation first and then call removeFromDOM() that removes the modal from DOM.
+   **/
+  beforeClose: PropTypes.func,
+  /**
+   * Content inside the modal dialog
+   **/
+  children: PropTypes.element.isRequired,
```
StephenEsser: Similar to how we standardized the height / width properties I think children is also something we should standardize.

I believe most components specify the child type as node so that any renderable object can be passed into children. I think it becomes a little confusing when some components can render anything while other components will only accept element types.

packages/terra-modal/src/ModalContent.jsx
```diff
@@ -0,0 +1,71 @@
...
+  closeOnOutsideClick: true,
+  role: 'dialog',
+};
+
+
+class ModalContent extends React.Component {
+
+  componentDidMount() {
+    // Disable background scrolling
+    document.body.style.overflow = 'hidden';
```
StephenEsser: Would the Overlay prevent any scroll events? Many Orion applications utilize a SlidePanel as a base component that manages its' own overflow, without the overlay I think these would still scroll. I think the current implementation has the full screen overlay prevent any scrolling.

packages/terra-modal/src/ModalContent.jsx
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  componentWillUnmount() {
+    // Enable background scrolling
+    document.body.style.overflow = null;
```
StephenEsser: This could potentially cause some thrashing if the consuming application has styles set on the body

packages/terra-modal/src/ModalOverlay.jsx
```diff
@@ -0,0 +1,17 @@
+import React, { PropTypes } from 'react';
+import './ModalOverlay.scss';
+
+const propTypes = {
+  classNameOverlay: PropTypes.string,
```
StephenEsser: Is there a good reason to keep this as an explicit prop rather than just spreading custom props onto the overlay?

I think you're using className above to pass the class, and not actually using this prop.

packages/terra-modal/src/ModalContent.jsx
Matt-Butler: I'm not sure I quite follow. Are you proposing that when the modal is active, content in the background can still scroll? 

Also, the current implementation enforces that there is always an overlay when a modal is active.

packages/terra-modal/src/ModalContent.jsx
Matt-Butler: This should only affect the overflow style property on the body. When do you think this would be an issue?

packages/terra-modal/src/ModalContent.jsx
StephenEsser: I think content can still scroll even if you set this on body because that's not the only element that can have an overflow set, so I don't think this is an effective way to disable scrolling. But the Overlay can stop scrolling, does that make any sense?

packages/terra-modal/src/ModalContent.jsx
StephenEsser: I think could cause unwanted side effects. The base layout every Orion application uses currently has body overflow set to hidden. This will result in removing that style and potentially causing a lot of issues with the layout. 

https://github.cerner.com/orion/orion-rails-layouts/blob/778b1c4973bfbeab416e228f66f71ab75c737d8e/app/views/orion/rails/layouts/_base.html.haml#L25

https://github.cerner.com/orion/orion-rails/blob/c26aa2eb622708549f70bcd84a9f95adf3713d7a/app/assets/stylesheets/orion/layouts/base.less#L15

packages/terra-modal/src/ModalOverlay.jsx
Matt-Butler: I'm going to move the overlay class logic into overlay

packages/terra-modal/src/Modal.jsx
Matt-Butler: updated

packages/terra-modal/lib/Modal.scss
bjankord: We should be using rem units. When the rem mixin is available we can start using pixel units in the mixin and output rems.

packages/terra-modal/docs/themeable-variables.md
```diff
@@ -0,0 +1,3 @@
+# Themeable Variables
+
```
bjankord: If there are no themeable variables, this file can be removed.

packages/terra-modal/src/Modal.scss
```diff
@@ -0,0 +1,34 @@
+@import '~terra-mixins';
+
+.terra-Modal {
+  background-color: $terra-white;
```
bjankord: Use rem values for these pixel values. 
With the 14px root font, 10px = 0.714rem. 5px = 0.357rem.

packages/terra-modal/src/Modal.scss
```diff
@@ -0,0 +1,34 @@
...
+  border-radius: 5px;
+  bottom: 10px;
+  display: block;
+  left: 10px;
```
bjankord: Update this to use the position-start mixin: https://github.com/cerner/terra-core/tree/master/packages/terra-mixins/docs#terra-position-start--terra-position-end

packages/terra-modal/src/Modal.scss
```diff
@@ -0,0 +1,34 @@
...
+  max-height: 100%;
+  outline: none;
+  overflow: hidden;
+  position: fixed;
+  right: 10px;
```
bjankord: Update this to use to position-end mixin: Update this to use the position-start mixin: https://github.com/cerner/terra-core/tree/master/packages/terra-mixins/docs#terra-position-start--terra-position-end

packages/terra-modal/src/Modal.scss
```diff
@@ -0,0 +1,34 @@
...
+  top: 10px;
+  z-index: $terra-z-index-6 + 1;
+
+  @media screen and (min-width: 48em) {
+    &:not(.terra-Modal--fullscreen) {
+      bottom: auto;
+      box-shadow: 0 3px 7px rgba(0, 0, 0, 0.3);
```
bjankord: These can stay as pixel values.

packages/terra-modal/src/Modal.scss
Matt-Butler: I think we want to keep left, right, top, and bottom as 10px instead of rem. If the user updates the font size in their browser, we probably want the area around the modal dialog to stay consistent.
**Large Font**
largefont
**Small Font**
smallfont

packages/terra-modal/docs/themeable-variables.md
Matt-Butler: removed file

packages/terra-modal/src/Modal.scss
Matt-Butler: updated to use mixin

packages/terra-modal/src/Modal.scss
Matt-Butler: updated to use mixin

packages/terra-modal/src/Modal.scss
bjankord: Good point, yeah lets leave these position properties as PX units.

PR #231, opened 2017-04-17
Base enhancement

### LINKED ISSUE #222: Base Enhancement, add i18n in Base component
## Current Behavior
Currently, we only have one BaseStyles( [code link](https://github.com/cerner/terra-core/commit/6b2fe080356bc04661eb0a0dea6db27d324b2133#diff-32b703eb84aad852d11b77791c8ff8e0R11) ), which contains the validation of HTML tag and pass all `customProps` down to the children.

## Expected Behavior after the enhancement
The Base component will have an I18nProvider to wrap the current <BaseStyles />. In this way, we will guarantee:
1. Consumers must provide I18nProvider as the ancestor of the all React components. All the translations will work correctly.
2. We don't provide IntlBase and BaseStyles as two separate components. Consumers just can easily consume it by `import Base from "terra-base";`.
3. This I18nProvider is also the first place to merge customized translations from App with terra-i18n default translations and pass down to make it available to all children.

## What does include in this enhancement?
API:
1. `locale` is used for setting locale and load all required locale data.
2. `children` wrap all children in new Base.
3. `customMessages`, all required translations from App level.

React Lifecycle functions update:
1. `componentDidMound` to guarantee the component will be rendered after `i18nLoader` load all I18n required files.
2. `componentWillReceiveProps` to guarantee if the locale prop or customMessage prop update, the Base will be rerendered.

## Influence of this enhancement
1. `terra-i18n` will be a dependency for `terra-base`.

windse7en:
### Summary
Fixes #222 in this PR.
1. Add `terra-i18n` as dependency to wrap `BaseStyles` with I18nProvider.
2. Update jest unit test with new Base.
3. Add nightwatch test for locale switch and customMessage display.
4. Update the example in terra-site.
5. Update the doc.

### Discussion on exporting
The discussion on exporting only `Base` or exporting `IntlBase & BaseStyles` is still open in this PR.
Pros of only one Base:
Consumers can import as simple as possible:
```
import Base from 'terra-base';
<Base>
...
</Base>
```
instead of
```
import { IntlBase, BaseStyles } from 'terra-base';
<IntlBase>
  <BaseStyles>
   ...
  </BaseStyles>
</IntlBase>
```

Pros of exporting `IntlBase & BaseStyles`:
If there is bugs in `react-intl` or `intl` or `terra-i18n`, the `BaseStyles` will still work, if there is no any transltions in the consuming applicaion.


packages/terra-base/src/Base.jsx
```diff
@@ -1,38 +1,74 @@
...
    */
   children: PropTypes.node.isRequired,
+  /**
+   * Customized translations provided by consuming application.
+   */
+  customMessages: PropTypes.object,
+  /**
+   * The locale name.
+   */
+  locale: PropTypes.string,
```
Matt-Butler: required

packages/terra-base/src/Base.jsx
```diff
@@ -1,38 +1,74 @@
...
+};
+
+const defaultProps = {
+  customMessages: {},
+  locale: 'en',
```
Matt-Butler: I think we should remove the default locale.

If a user defines the locale as fi-FI on the server and it not configured to update on the client, then we can can multiple locales on the same page.

Matt-Butler: If we force a consumer to define the react-intl locale, then it will be more transparent to them how it works.

packages/terra-base/src/Base.jsx
```diff
@@ -1,38 +1,74 @@
...
 
 const propTypes = {
   /**
    * The component(s) that will be wrapped by `<Base />`.
    */
   children: PropTypes.node.isRequired,
+  /**
+   * Customized translations provided by consuming application.
+   */
+  customMessages: PropTypes.object,
```
Matt-Butler: Is this for a single locale at a time?

windse7en: Yes. This is only for the single locale. `messages` in i18n is also only for single locale. I will update the comments to make it clear.

packages/terra-base/src/Base.jsx
windse7en: I thought it would be convenient for the application without translations. I will update locale to require.

packages/terra-base/src/Base.jsx
windse7en: Updated in https://github.com/cerner/terra-core/pull/231/commits/8ac922bcb942e78cf1747a2af0d185f7ea7cbb69

packages/terra-base/src/Base.jsx
bjankord: I agree with removing the default prop for locale, with requiring it, we will ensure one is set and should help wtih sync issues related to locale on server and client-side locale.

packages/terra-base/src/Base.jsx
```diff
@@ -1,38 +1,76 @@
...
+  locale: PropTypes.string.isRequired,
+  /**
+   * Customized translations provided by consuming application
+   * only for current locale.
+   */
+  customMessages: PropTypes.object,
+};
+
+const defaultProps = {
+  locale: 'en',
```
bjankord: We should remove locale from being a default prop.

packages/terra-site/src/index.html
```diff
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="x-ua-compatible" content="ie=edge">
+    <meta http-equiv="Content-Language" content="en-US">
```
bjankord: What's the logic for setting this in the meta tag? This is set on the html element via the lang attribute.

windse7en: This is to avoid the French translation popup. In chrome, if the page contains not enough information, chrome may consider the application is in French. Other information is included in your post link.

packages/terra-base/src/Base.jsx
```diff
@@ -1,38 +1,75 @@
...
   }
 
   render() {
-    const { children, ...customProps } = this.props;
-    return <div {...customProps}>{children}</div>;
+    if (!this.state.load) return null;
+
+    const {
+      children,
+      locale,
```
Matt-Butler: This is sort of confusing to me.... we have `locale` here, but below we use `this.state.locale`. Is this correct?

packages/terra-base/src/Base.jsx
```diff
@@ -1,38 +1,75 @@
...
+      customMessages,
+      ...customProps
+    } = this.props;
+
+    const messages = Object.assign({}, this.state.messages, customMessages);
+
+    return (
+      <I18nProvider
+        locale={this.state.locale}
```
ryanthemanuel: Just so I understand what's happening here, does the messages object only contain the messages for the given locale?  And if so what is the purpose of locale?  Is it to format things like numbers, dates, etc.?

packages/terra-base/src/BaseStyles.jsx
```diff
@@ -0,0 +1,38 @@
...
+  children: PropTypes.node.isRequired,
+};
+
+class BaseStyles extends React.Component {
+
+  componentDidMount() {
+    // Checks to run when not in production
+    if (process.env.NODE_ENV !== 'production') {
+      // Check to ensure terra-Base class is set on html element
+      if (!new RegExp('(^|\\s)terra-Base(\\s|$)').test(document.documentElement.className)) {
```
ryanthemanuel: Given that we're assuming that this is the root element, I'm curious about some scenarios we're trying to support.  We were going to potentially (in the interim phases for Orion) have multiple React components spread throughout the page amidst other non React elements.  Would terra-base still need to be at the root level in this scenario?  Or would it make sense to be at the root of each component.  I'm trying to figure out how this might work in our temporary hybrid world.

packages/terra-base/src/Base.jsx
windse7en: Yeah. The props locale is used to initialize the `state.locale`. `state.locale` is used to guarantee all required locale date has loaded before the Base component can be rendered. So in [constructor](https://github.com/cerner/terra-core/pull/231/files/1a59225243ab3e8053b088359f6b5b397f6aa398#diff-32b703eb84aad852d11b77791c8ff8e0R32), we use props.locale and in the render, we use `state.locale`. 

To explain the difference between `props.locale` and `state.locale`, this code may be helpful [en locale data](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/src/translations/en.js#L12). The `state.locale` is coming from the translations data. That will update the `state.locale` as `state.messages`.

packages/terra-base/src/Base.jsx
windse7en: Yeah. messages object only contains the translations for given locale. It's just a translation mapping hash. [code link](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/src/translations/en.js#L6). 

`locale` is part of the required value for `IntlProvider(react-intl)`. It will load other locale data like 'react-intl/locale-data/en'(like CLDR data), which is used for format dates, numbers.

PR #244, opened 2017-04-19
Terra time input

benbcai:
### Summary
The terra-time-input component is an input for entering time. During time entry the value is masked to conform to the time format.

Translations are currently hard-coded. Will update translations once we have a strategy to get the translated strings (i.e. using react context).


packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -0,0 +1,113 @@
...
+    // TODO: Get the time format translation.
+    return 'HH:mm';
+  }
+
+  static formattedTime(timeMoment) {
+    if (timeMoment) {
+      return timeMoment.clone().locale(TimeInput.userLocale()).format(TimeInput.timeFormat());
+    }
+
+    return '';
```
emilyrohrbough: So if no default time is given, an empty string is displayed until a time is selected?

benbcai: If no default time is passed in the prop or there is no value in the input (the user manually backspaced to delete the time), then it would display the placeholder text (i.e.; HH:mm) in the input.

packages/terra-time-input/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "terra-base": "^0.x",
+    "terra-mixins": "^1.0.0"
+  },
+  "dependencies": {
+    "moment": "^2.18.1",
+    "react-text-mask": "^4.1.0",
+    "terra-mixins": "^1.0.0"
```
StephenEsser: There is a pull request out now that add terra-base as an explicit dependency, you might check out this PR for more information: https://github.com/cerner/terra-core/pull/251

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      defaultTime: TimeInput.formattedTime(this.props.defaultTime),
+      value: TimeInput.formattedTime(this.props.defaultTime),
```
StephenEsser: nit: I think you can drop `this` from `this.props.defaultTime`

benbcai: Yup, dropped 'this'. https://github.com/cerner/terra-core/commit/eefcee225cc9d65e5c1891c4e3958eea9382a8a5

packages/terra-time-input/package.json
benbcai: Added terra-base as a dependency. https://github.com/cerner/terra-core/commit/eefcee225cc9d65e5c1891c4e3958eea9382a8a5

I didn't import `terra-base/lib/baseStyles` since it is not yet in master. I'll update to import it once https://github.com/cerner/terra-core/pull/251 is merged.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -0,0 +1,113 @@
+import React, { PropTypes } from 'react';
+import moment from 'moment';
+import TextMask from 'react-text-mask';
```
bjankord: Add `import 'terra-base/lib/baseStyles';` here.

benbcai: Done.

packages/terra-time-input/src/TimePipe.js
```diff
@@ -0,0 +1,26 @@
+export default function TimePipe(timeFormat = 'HH:mm') {
```
Matt-Butler: I think we want to remove the default format. It will not always be the same default based on the locale.

Matt-Butler: nit: export at the bottom of the file

packages/terra-time-input/package.json
bjankord: #251 has been merged.

packages/terra-time-input/lib/TimeInput.js
```diff
@@ -0,0 +1,160 @@
...
+    key: 'userLocale',
+    value: function userLocale() {
+      // TODO: Get the locale from i18nProvider
+      return 'en-US';
+    }
+  }, {
+    key: 'timeFormat',
+    value: function timeFormat() {
+      // TODO: Get the time format translation.
+      return 'HH:mm';
```
JakeLaCombe: How well does this format work with times such as 3:29? When we uplifted moment in blue_steel, we had to change to h:mm format as hh:mm would look specifically for a two digit hour.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -0,0 +1,113 @@
...
+    this.state = {
+      defaultTime: TimeInput.formattedTime(this.props.defaultTime),
+      value: TimeInput.formattedTime(this.props.defaultTime),
+    };
+
+    this.handleChange = this.handleChange.bind(this);
+    this.handleInputKeyDown = this.handleInputKeyDown.bind(this);
+  }
+
+  handleChange(event) {
```
JakeLaCombe: When I talked to nick smith about form designs, he suggested not attaching the input value to the state, and instead just allowing the consumer to pass in either a value and onChange function, or pass in a defaultValue for an uncontrolled value. Perhaps we should do that here as well?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -0,0 +1,113 @@
...
+      ...customProps
+    } = this.props;
+
+    return (
+      <div className="terra-TimeInput">
+        <TextMask
+          {...customProps}
+          className="terra-TimeInput-input"
+          type="text"
+          value={this.state.value}
```
JakeLaCombe: What happens if the developer passes in a null value, but a non null defaultTime? When I did forms, I had to remove the value key from the hash of props if they provided a default value, but an actual value.

packages/terra-site/package.json
```diff
@@ -22,7 +22,7 @@
     "test:spec": "echo \"Warning: no test written\" && exit 0"
   },
   "dependencies": {
-    "moment": "^2.17.1",
+    "moment": "^2.18.1",
```
Matt-Butler: https://github.com/Hacker0x01/react-datepicker/blob/master/package.json#L94
I would recommend we match the same version of moment that react-datepicker has.

```
"moment": "^2.17.1",
```

packages/terra-time-input/lib/TimeInput.js
benbcai: When I met with Neil a while back, he said it should always use a 2-digit hour (HH) and should not allow a single digit.

packages/terra-time-input/src/TimeInput.jsx
benbcai: I would agree with that for a normal input but I have to manage the state of the input value because I have to increment/decrement the minute when the user hits the up/down arrow keys. For that reason, I think I have to attach the value to the state.

packages/terra-time-input/src/TimeInput.jsx
benbcai: I updated to mimic forms by defaulting `undefined` to value and defaultValue.

packages/terra-time-input/src/TimePipe.js
benbcai: Dropped the default time format.

benbcai: Moved the export to the bottom.

packages/terra-site/src/examples/time-input/Index.jsx
```diff
@@ -0,0 +1,30 @@
...
+import React from 'react';
+import moment from 'moment';
+import PropsTable from 'terra-props-table';
+import Markdown from 'terra-markdown';
+import ReadMe from 'terra-time-input/docs/README.md';
+import { version } from 'terra-time-input/package.json';
+import TimeInput from 'terra-time-input/src/TimeInput';
+
+// Component Source
+// eslint-disable-next-line import/no-webpack-loader-syntax, import/first, import/no-unresolved, import/extensions
```
StephenEsser: Woah

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -0,0 +1,113 @@
...
+   */
+  defaultTime: PropTypes.object,
+    /**
+   * A callback function to execute when a time value is entered.
+   */
+  onChange: PropTypes.func,
+};
+
+const defaultProps = {
+  defaultTime: null,
```
benbcai: Yup, these props have been updated to default to `undefined`
https://github.com/cerner/terra-core/blob/terra_time_input/packages/terra-time-input/src/TimeInput.jsx#L30-L32

bjankord:

Issue #206, opened 2017-04-11
Deprecate terra-menu


Matt-Butler:
We are deprecating terra-menu and creating tabs. We need to remove the terra-menu code.

PR #245, opened 2017-04-19
Remove menu component

Matt-Butler:
### Summary
We are no longer going to support the terra-menu component. Instead, we will be creating a tabs component in the near future.

@cerner/terra

https://github.com/cerner/terra-core/issues/206


mjhenkes:
We're actually thinking our collapsible button view may actually be a menu component. We'll share what we're thinking with you guys, but we can remove this now.

Issue #225, opened 2017-04-17
Standardize height prop


Matt-Butler:
We have several components that have the prop `height` but different implementations. This property should be more predictable across all components.

[string('tiny', 'small', 'medium', 'large', 'huge')](https://github.com/cerner/terra-core/blob/7f1d3732290e7335506c5a630e2ac616361cd9ff/packages/terra-progress-bar/src/ProgressBar.jsx#L13)
```
<ProgressBar height="tiny" value={15} />
```

[string](https://github.com/cerner/terra-core/blob/7f1d3732290e7335506c5a630e2ac616361cd9ff/packages/terra-icon/src/IconBase.jsx#L22)
```
<Icon height="50" width="50" />
```

[number](https://github.com/cerner/terra-core/blob/master/packages/terra-image/src/Image.jsx#L25)
```
<Image height="50" />
```

Matt-Butler:
Move forward with height as html attribute usage
If we come to it, explore per component height props, e.g. iconHeight=”small”

bjankord:
Fixed in #247

PR #247, opened 2017-04-19
Standardize height prop

Matt-Butler:
### Summary
Updating the height PropType of image and progress bar to promote consistency.

@cerner/terra

https://github.com/cerner/terra-core/issues/225


packages/terra-progress-bar/src/ProgressBar.jsx
```diff
@@ -10,7 +10,7 @@ const propTypes = {
   /**
    * Sets the size of the progress-bar from the following values; `tiny`, `small`, `medium`, `large` and `huge`
    */
-  height: PropTypes.oneOf(['tiny', 'small', 'medium', 'large', 'huge']),
+  progressbarHeight: PropTypes.oneOf(['tiny', 'small', 'medium', 'large', 'huge']),
```
bjankord: Lets update this prop name to `heightSize`

Matt-Butler: +1 I agree
https://github.com/cerner/terra-core/pull/247/commits/3000669a7d9045d4624df03a05e91803ac0cff14

mjhenkes:
Correct me if i'm wrong, but this would be major version bump for image and progress bar? Are we saying that pre 1.0 versions don't need to be passive?

Matt-Butler:
> Correct me if i'm wrong, but this would be major version bump for image and progress bar? Are we saying that pre 1.0 versions don't need to be passive?
At this stage I'm ok either way, I just want to be able to state our strategy.


My 2 cents, anything < 1.0 has the possibility to be non-passive. I would do a minor release for progress bar and image.

PR #251, opened 2017-04-20
Make base dependency explicit

bjankord:
### Summary
Currently our `<Base />` component is used to set minimal global styles for an application. This includes CSS to help normalize box-sizing, reset margins/paddings, and define global font styles. For it to work as intended, it requires consumers to import `terra-base` before the other terra components in their app. This is to help ensure that these styles are defined at the beginning of the extracted stylesheet and are inherited down to the other terra components correctly. 

This is an implicit dependency that is hard to define and communicate to consumers. We've used peerDependencies to make consumers aware when they download a component like terra-button, it has a theres a relation to terra-base. But it requires them to go read the [terra-base doc](http://engineering.cerner.com/terra-core/#/site/base) and have some understanding of how CSS inheritance works (setting fonts on html to set fonts for all elements) and webpack's text extract plugin behavior. We talked about using react's context API to warn users if terra components are rendered outside of `<Base>` but quickly realized how much of pain this would be to maintain / write up for each component.

This code change in this PR works to make the dependency more explicit and allow components to be rendered independently of being within `<Base>`. This is achieved by adding terra-base as a dependency to all terra component's package.json file. In addition, a JS file has been imported into all terra components that will load the base styles / check HTML element for proper attribute (terra-Base class and dir attribute). This file contains a check to ensure that global styles are only loaded once in an application when a consumer uses a terra component.

The impact of this change is a consumer can use any terra component, button, form, table and the base styles will be loaded correctly in the CSS file coming before the component styles. And these base styles will only be loaded once per application. The consumer no longer needs to be as aware of the dependency of the base styles and the component styles.

This change makes the react `<Base />` component of little value, however we've discussed using as a entry point for setting locale, translations, and possibly themes. For the time being, I'm keeping it around, mainly to maintain passivity and to hold on to it until we decide more of what we are gonna do with i18n.

If you pull down this branch and run the webpack dev server, note that none of the components are rendered within `<Base>`, though base styles are style added to the CSS file.

@cerner/terra

### DIFF
--- a/packages/terra-base/src/baseStyles.js
+++ b/packages/terra-base/src/baseStyles.js
@@ -1,0 +1,23 @@
+if (!global.baseStylesLoaded) {
+  console.log('base styles');
+
+  // eslint-disable-next-line
+  require('./Base.scss');
+}
+
+global.baseStylesLoaded = true;
+
+// Checks to run when not in production
+if (process.env.NODE_ENV !== 'production') {
+  // Check to ensure terra-Base class is set on html element
+  if (!new RegExp('(^|\\s)terra-Base(\\s|$)').test(document.documentElement.className)) {
+    // eslint-disable-next-line
+    console.warn('The html element is missing the terra-Base class.');
+  }
+
+  // Check to ensure dir attribute is set on html element
+  if (!document.documentElement.hasAttribute('dir')) {
+    // eslint-disable-next-line
+    console.warn('The html element is missing the dir attribute. For terra directionality based styles to render correctly, add dir="ltr" or dir="rtl" to the html element.');
+  }
+}
--- a/packages/terra-base/src/baseStyles.js
+++ b/packages/terra-base/src/baseStyles.js
@@ -1,10 +1,15 @@
+/**
+ * The following code is used to ensure global styles are loaded only once in an app
+ * that consumes terra components.
+ */
 if (!global.baseStylesLoaded) {
-  console.log('base styles');
-
-  // eslint-disable-next-line
+  /* eslint-disable global-require */
   require('./Base.scss');
 }
 
+/**
+ * Once global styles are loaded, set global to prevent duplication of global styles
+ */
 global.baseStylesLoaded = true;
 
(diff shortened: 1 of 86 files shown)


bjankord:
@mjhenkes I thought about this some. Here are the considerations that came to mind, the `Base.jsx` file for the react component gets compiled down to `Base.js`. Depending on if we decided to keep `<Base>` we may want to keep this new file named as `baseStyles`. 

Though if we are thinking we want this file that is being loaded into all components to do more than just load global styles, I'm open to a different name, like `base-helper.js`, `base-util.js`, something else, etc.

packages/terra-base/src/baseStyles.js
```diff
@@ -0,0 +1,28 @@
+/**
+ * The following code is used to ensure global styles are loaded only once in an app
+ * that consumes terra components.
+ */
+if (!global.baseStylesLoaded) {
```
Matt-Butler: Do we need this check yet?

bjankord: Yes. This check is run in every terra-component that has styles. 

The first component which loads will run this condition and it will evaluate that `global.baseStylesLoaded` is `false`. The `!` operator will cast `global.baseStylesLoaded` to `true` causing the code inside the if condition to run.

Then on line 13, we set `global.baseStylesLoaded` to `true`. The next terra component which loads will run this condition and it will evaluate that `global.baseStylesLoaded` is `true`. The `!` operator will cast `global.baseStylesLoaded` to `false` causing the code inside the if condition to be skipped preventing the duplication of the global base styles.

bjankord: Looks like webpack will automatically dedupe this and the webpack extract text plugin will only include this once. I removed the if condition: https://github.com/cerner/terra-core/pull/251/commits/6b700b9a8487dc9a937a2451ccdabc75df6714ed

Matt-Butler: Thank you for checking on this

Issue #173, opened 2017-03-27
Selectable List keyboard accessibility


Matt-Butler:
Selectable Lists are inaccesibe with a keyboad. We need to update selectable list items to have a tab-index="0". http://webaim.org/techniques/keyboard/tabindex

![selectable_list_keyboard accessibility](https://cloud.githubusercontent.com/assets/8146159/24371595/e795067c-12f0-11e7-80a2-a45a2315d3a8.gif)

Matt-Butler:
Relates to this issue: https://github.com/cerner/terra-core/issues/205

bjankord:
See these commits for what was done in table:
* https://github.com/cerner/terra-core/commit/e899b2aac2c87adf67ed6db648f97bb55b4da7a3

PR #266, opened 2017-04-26
Selectable List keyboard accessibility

emilyrohrbough:
### Summary
Added keyboard accessibility to single- and multiple-selectable lists. Addresses Issue #173.

### Note
The aria-selected attribute was not include on the ListItems; aria-selected is used for gridcell, option, row and tab roles. The current ListItem component uses [listitem roles]( https://github.com/cerner/terra-core/blob/master/packages/terra-list/src/ListItem.jsx#L54) and listitem roles do not have a aria state attribute that is appropriate for indicating a selection. 

[Info on Aria-selected state](https://www.w3.org/TR/wai-aria/states_and_properties#aria-selected)
[Info on Aria listitem role](https://www.w3.org/TR/wai-aria/roles#listitem)

What do we want to do about this? Change the role of the ListItem component? I assume we want to support aria accessibility for this component.

@cerner/terra


bjankord:
@emilyrohrbough Let's hold off on adding aria-selected to the ListItems, it doesn't seem like it fits within the intended use of that attribute. The keyboard interactions are a nice accessibility improvement for this functionality. We can tackle communicating selected state to screen readers in a future enhancement.

packages/terra-list/src/SingleSelectList.jsx
```diff
@@ -32,27 +36,17 @@ class SingleSelectList extends React.Component {
 
   static selectedIndexFromItems(items) {
     for (let i = 0; i < items.length; i += 1) {
-      if (items[i].props.isSelected === true) {
+      if (items[i].props.isSelected === true && items[i].props.isSelectable !== false) {
```
Matt-Butler: nit: we should change this sometime. I'll log an issue for this.

Setting boolean variables to undefined is making us write some weird logic like this: 
```
items[i].props.isSelectable !== false
```
Boolean default props should always be either true of false, not undefined:
https://github.com/cerner/terra-core/blob/master/packages/terra-list/src/ListItem.jsx#L27-L30

Matt-Butler: https://github.com/cerner/terra-core/issues/267

Matt-Butler: We can fix this in a different code review

packages/terra-list/src/MultiSelectList.jsx
```diff
@@ -154,20 +189,19 @@ class MultiSelectList extends React.Component {
     return this.props.children.length;
   }
 
-  unusedVariables(variable) {
-    return variable === this;
-  }
-
   render() {
-    const { children, isDivided, onChange, maxSelectionCount, ...customProps } = this.props;
+    const { children, isDivided, ...customProps } = this.props;
```
Matt-Butler: @StephenEsser has this same sort of logic in the responsive element. Instead of removing these params, what do you think about just ignoring the eslint rule?

https://github.com/cerner/terra-core/blob/master/packages/terra-responsive-element/src/ResponsiveElement.jsx#L101-L102

emilyrohrbough: I think it would be good to decide a consistent approach to handling params in this situation. This logic occurs in a few places and are handled a bit differently each time.

Matt-Butler: Sounds good. I'll log a separate issue for this so we can move this pr along.

Matt-Butler: https://github.com/cerner/terra-core/issues/268

Issue #209, opened 2017-04-12
Update to webpack 2


bjankord:
### Description of Issue
To move us to a more modern build tool, we should update to webpack 2. In order to migrate to webpack 2, we need to ensure all of our webpack plugins / dependencies we use today are compatible with webpack 2.

Work on this has started in this branch:
https://github.com/cerner/terra-core/tree/webpack2

### Expected Outcomes
- webpack config is updated to be compatible with webpack 2
- webpack plugins are updated to be compatible with webpack 2

Matt-Butler:
I merged the webpack 2 branch with master today.

I am getting errors when running nightwatch tests after upgrading.

windse7en:
1. Update to translation chunk name in webpack 2, fix in `terra-site/webpack.config.js`:
```
const webpack = require('webpack');
plugins:...
    new webpack.NamedChunksPlugin(),
```

2. Add `es6-promise` to the terra-site. `require.ensure` is based on that. Fix in `terra-site/src/index.jsx`:
```
import 'es6-promise/auto';
```

bjankord:
Let's use [babel-polyfill](https://babeljs.io/docs/usage/polyfill/) instead of [es6-promise](https://www.npmjs.com/package/es6-promise) as it seems to be used by more of our consumers.

bjankord:
Fixed in #270

PR #270, opened 2017-04-27
Webpack2 Integration

windse7en:
### Summary
Update monorepo to integrate with webpack 2.
1. Bump up `terra-site` and `terra-toolkit` webpack version to 2 in devDependencies and upgrade relative loaders, plugins, and script.
2. Add `babel-polyfill` in the `terra-site` Index.jsx file. As webpack dynamic import relies on consuming app providing the promise polyfill, webpack doc [link](https://webpack.js.org/guides/code-splitting-async/#promise-polyfill).
3. Update `server-launcher.js` in`terra-toolkit`. Create the WebpackDevServer need to switch current directory to `terra-site` to find all required node_modules.

### Additional Details
1. Increase the `waitInms` for I18n nightwatch test. It will make us confident on test failure. This won't increase the test time.


packages/terra-site/webpack.prod.config.js
```diff
@@ -16,9 +17,13 @@ config.plugins.push(new webpack.DefinePlugin({
 // Clean build before running
 config.plugins.push(new CleanPlugin('build', { exclude: ['stats.json'] }));
 
+// Minify css and js
+config.plugins.push(new webpack.LoaderOptionsPlugin({ minimize: true }));
+
```
Matt-Butler: I think we also want to add the UglifyJsPlugin.

https://webpack.js.org/guides/production-build/#simple-approach

Matt-Butler:
https://github.com/cerner/terra-core/issues/209

packages/terra-site/webpack.prod.config.js
Matt-Butler: @bjankord mentioned that we cannot do tree shaking with the UglifyJsPlugin. He recommended that we check out https://github.com/babel/babili

bjankord: Full tree-shaking support with webpack 2 is a bit involved, and there are some issues we need to consider around using UglifyJsPlugin vs babili plugin. With using the UglifyJsPlugin, we could do partial tree shaking, (unused functions are removed, but not unused classes). This requires a config change in the [babelrc file](https://gist.github.com/blacksonic/e7799a87dda003c5f4da4f6eae85336c/raw/32a618c4d8af5602729bb616b33b984d8413c5d4/.babelrc) and the [webpack config](https://gist.github.com/blacksonic/70078c33e3dd3a25615c9236932c9fb2/raw/fe195b0845d75ff7ca333a08cd134c42486ff7f0/webpack.prod.config.js) to work. 

I'd like to explore using the babili plugin instead of UglifyJsPlugin so we can also remove unused classes trim the size even more.

This is a really good write up on webpack 2 tree-shaking and whats all involved to get full tree-shaking support: https://blog.craftlab.hu/how-to-do-proper-tree-shaking-in-webpack-2-e27852af8b21

More info on babili: https://babeljs.io/blog/2016/08/30/babili

I'm also ok if we want to merge this down as is now and explore tree-shaking functionality in a separate PR or handle it in this PR.

windse7en: To explain tree-shaking: 
> Javascript bundling will only include code that is necessary to run your application. It has been made available by the static nature of ES2015 modules (exports and imports can’t be modified at runtime), which lets us detect unused code at bundle time. refer to [this blog](https://blog.craftlab.hu/how-to-do-proper-tree-shaking-in-webpack-2-e27852af8b21)

To add UglifyJsPlugin or bablili in thie PR, I have three concerns: 
1. I'm not sure their pros and cons. We may need to create another PR to document those concerns. It helps us to track the decision and convince consumers with our final choice. 
2. And in this PR, we add `cheap-source-map` into the webpack config. We may need to test that with minified files by UglifyJsPlugin or bablili.
3. As this will be our recommendation across all consumers, we may want to compare the compression ratio to achieve the best performance.

bjankord: Yeah, let's tackle adding UglifyJsPlugin or bablili in another PR. I agree with the concerns you listed @windse7en. I'll log an issue to explore this in a separate PR.

PR #272, opened 2017-04-27
I18n webpack plugin

ryanthemanuel:
## Summary

Below is a description of how the I18nAggregatorPlugin will be utilized.

### I18nAggregatorPlugin

This plugin needs to be added to the list of plugins available in your webpack config with a required option of baseDirectory:

```js
new I18nAggregatorPlugin({
  baseDirectory: __dirname,
}
```

By including this plugin, all dependencies will be examined for translation files in a translation folder at the root of the directory for supported locales (e.g. de.js or en.js) formatted as follows:

```js
const messages = {
  'Terra.ajax.error': 'Inhalt konnte nicht geladen werden.',
};

module.exports = {
  messages,
};
```

Then all of the messages for a given language across all dependencies will be aggregated into an overall translation file for that language and placed into an aggregated-translations directory.  In order for these resulting files to be able to be loaded by terra-i18n, that directory then needs to be added to the list of directories that the webpack resolver looks into.


packages/terra-i18n/package.json
```diff
@@ -27,12 +27,13 @@
...
     "react": "^15.4.2",
-    "react-dom": "^15.4.2"
+    "react-dom": "^15.4.2",
+    "react-intl": "2.2.3"
   },
   "dependencies": {
     "classnames": "^2.2.5",
     "intl": "^1.2.5",
-    "react-intl": "^2.2.3",
+    "mkdirp": "^0.5.1",
```
ryanthemanuel: I think mkdirp might be able to be a development dependency.

windse7en:
Nice work. 

We can just use `json` file for all the translations file, as we already extract import `react-intl/local...` into the plugin. We can just use `packages/terra-i18n/translations/en-GB.json`.

packages/terra-site/webpack.config.js
```diff
@@ -6,6 +6,7 @@ const path = require('path');
 const autoprefixer = require('autoprefixer');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
+const I18nAggregatorPlugin = require('./node_modules/terra-i18n/lib/I18nAggregatorPlugin');
```
windse7en: In here, it's a little hard to separate the plugin and I18n. In general, I18n component is a dependency and Plugin should be a devDependency. I don't have some ideas in here, but we may want to have some discussions on it.

ryanthemanuel: Yeah I agree there's a weird mix of development dependencies and runtime dependencies going on here.  We can discuss it tomorrow in the war room.

windse7en:
There is a potential problem with loading all translations by running plugin at built time. For example:
We have translations in terra components:
```
  terra-i18n/translations
  terra-datepicker/translations
```
We also have translations in orion components:
```
  orion-component/translations
  orion-component-with-lots-of-translations/translations
```
All the translations in `orion-component-with-lots-of-translations` need to be lazy-loaded. Application don't want to have all translations loaded at first. But all translations will be aggregated into one file by the plugin. We may add an regex argument to the `I18nAggregatorPlugin.js` function to filter `node_muldes/components` map.

packages/terra-site/webpack.config.js
Matt-Butler: I would also recommend updating the syntax to
```
const I18nAggregatorPlugin = require('terra-i18n/lib/I18nAggregatorPlugin');
```
in case the node_modules directory is not relative

packages/terra-i18n/src/translationLoaders.js
```diff
@@ -1,55 +1,71 @@
...
+    const i18n = require('es.js');
+    /* eslint-enable import/no-extraneous-dependencies, import/no-unresolved */
     callback.call(scope, i18n);
   }, 'es-translations');
 };
 
 const loadFiFITranslation = (callback, scope) => {
   require.ensure([], (require) => {
-    const i18n = require('./translations/fi-FI.js');
+    /* eslint-disable import/no-extraneous-dependencies, import/no-unresolved */
```
Matt-Butler: nit: we can use `eslint-disable-next-line`

packages/terra-i18n/src/i18nLoader.js
```diff
@@ -2,10 +2,12 @@
 import intlLoaders from './intlLoaders';
 import translationLoaders from './translationLoaders';
 
+import i18nSupportedLanguages from './i18nSupportedLanguages';
+
 const hasIntl = typeof (Intl) !== 'undefined';
 
 const permitParams = (locale, callback) => {
-  const supportedLocales = ['en', 'en-US', 'en-GB', 'es', 'de', 'fi-FI', 'fr', 'pt'];
+  const supportedLocales = i18nSupportedLanguages;
```
Matt-Butler: Can we just rename the import supportedLocales?
```
import supportedLocales from './i18nSupportedLanguages';
```

packages/terra-i18n/tests/nightwatch/i18n-spec.js
```diff
@@ -1,17 +1,19 @@
...
-const enUSTranslations = require('../../lib/translations/en-US.js').messages;
-const esTranslations = require('../../lib/translations/es.js').messages;
-const fiFITranslations = require('../../lib/translations/fi-FI.js').messages;
-const frTranslations = require('../../lib/translations/fr.js').messages;
-const ptTranslations = require('../../lib/translations/pt.js').messages;
+const screenshot = require('terra-toolkit').screenshot;
+const fs = require('fs');
+const path = require('path');
+
+const deTranslations = JSON.parse(fs.readFileSync(path.resolve(__dirname, '..', '..', 'translations', 'de.json'), 'utf8'));
```
Matt-Butler: nit, can we dry up some of this?

packages/terra-site/webpack.config.js
ryanthemanuel: Good call

packages/terra-i18n/src/i18nLoader.js
ryanthemanuel: Ha.  Good call.

packages/terra-i18n/tests/nightwatch/i18n-spec.js
ryanthemanuel: Good call

packages/terra-i18n/src/I18nAggregatorPlugin.js
```diff
@@ -0,0 +1,56 @@
...
+      const translationFile = path.resolve(options.baseDirectory, 'node_modules', module, 'translations', `${language}.json`);
+      if (fs.existsSync(translationFile)) {
+        Object.assign(currentLanguageMessages, JSON.parse(fs.readFileSync(translationFile, 'utf8')));
+      }
+    });
+
+    if (currentLanguageMessages !== {}) {
+      mkdirp(path.resolve(options.baseDirectory, 'aggregated-translations'));
+      fs.writeFileSync(path.resolve(options.baseDirectory, 'aggregated-translations', `${language}.js`),
+`import { addLocaleData } from 'react-intl';
```
windse7en: Could we extract this string interpolation into a function? Like:
```
const generateFileTitle = (language, messages) => (
`import { addLocaleData } from 'react-intl';
import localeData from 'react-intl/locale-data/${language.split('-')[0]}';

addLocaleData(localeData);

const messages = ${JSON.stringify(messages, null, 2)};

module.exports = {
  areTranslationsLoaded: true,
  locale: '${language}',
  messages,
};`
);
```

packages/terra-i18n/src/I18nAggregatorPlugin.js
```diff
@@ -0,0 +1,56 @@
+import fs from 'fs';
+import path from 'path';
+import mkdirp from 'mkdirp';
+
+import i18nSupportedLanguages from './i18nSupportedLanguages';
+
+function apply(options) {
+  function getDirectories(srcPath) {
```
windse7en: How about extracting this function out from apply function?

packages/terra-i18n/src/I18nAggregatorPlugin.js
```diff
@@ -0,0 +1,56 @@
...
+  messages,
+};`,
+      );
+    }
+  });
+}
+
+module.exports = (options) => {
+  let updatedOptions = options;
+  if (updatedOptions instanceof Array) {
```
windse7en: Just curious, why we want to check the options is Array or not?

ryanthemanuel: This format was recommended on the webpack website: https://webpack.github.io/docs/plugins.html  I'm cool with simplifying it if we think that we can count on options always being passed in a certain way.

packages/terra-i18n/src/I18nAggregatorPlugin.js
ryanthemanuel: I will make this change.

packages/terra-i18n/src/I18nAggregatorPlugin.js
ryanthemanuel: I will make this change.

packages/terra-site/webpack.config.js
bjankord: Let's look into the feasibility of splitting this into 2 components.

packages/terra-i18n/src/I18nAggregatorPlugin.js
windse7en: Cool. I'm good with following the webpack plugin conventions.

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -0,0 +1,59 @@
...
+  supportedLocales.forEach((language) => {
+    const currentLanguageMessages = {};
+    getDirectories(path.resolve(options.baseDirectory, 'node_modules')).forEach((module) => {
+      const translationFile = path.resolve(options.baseDirectory, 'node_modules', module, 'translations', `${language}.json`);
+      if (fs.existsSync(translationFile)) {
+        Object.assign(currentLanguageMessages, JSON.parse(fs.readFileSync(translationFile, 'utf8')));
+      }
+    });
+
+    if (currentLanguageMessages !== {}) {
```
JakeLaCombe: I know it may seem weird, but I think it's still worth generating the file for a supported translation even if the messages are empty. Some applications have a hard time dealing with determining how to handle the case when a requested file is not found. Anyone else have thoughts on it?

windse7en: There is an assumption: we will always provide translations on supported languages. So if there is an empty object, it means the translated modules are not installed correctly. We may throw some error in here or add an enhancement to do such validation later.

packages/terra-i18n-plugin/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "babel-jest": "^19.0.0",
+    "terra-toolkit": "^0.x"
+  },
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2"
+  },
+  "dependencies": {
+    "mkdirp": "^0.5.1",
+    "terra-i18n": "^0.x"
```
windse7en: Currently, the only cause of this dependency(`terra-i18n`) is caused by `terra-i18n/lib/i18nSupportedLocales`. I wonder if there is some way we can get rid of this dependency. Or just duplicate SupportedLocales to get rid of dependency.

emilyrohrbough: I added the `i18nSuppportedLocales` file. I was a bit hesitant to make this change so would like to know others opinions on this [change](https://github.com/cerner/terra-core/pull/272/commits/ec68c739248e90704dbd53760e99077915938e57#diff-c21518f2edbdb19b046e55f1fd4d42d1R32).

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
emilyrohrbough: I added in this [error message](https://github.com/cerner/terra-core/pull/272/commits/ec68c739248e90704dbd53760e99077915938e57#diff-4417b6331e7583b396c941b65e5e4246R41) to indicate if a supported translation was not supplied correctly.

package.json
```diff
@@ -76,6 +76,7 @@
     "react": "15.4.2",
     "react-addons-test-utils": "15.4.2",
     "react-dom": "15.4.2",
+    "react-intl": "2.2.3",
```
Matt-Butler: Why do we need this in the root? I think this would be part of terra-i18n, the webpack plugin, or terra-site

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
Matt-Butler: What happens if a user needs i18n support but does not use any of our components with i18n messages?

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -0,0 +1,61 @@
...
+  locale: '${language}',
+  messages,
+};`;
+}
+
+function getDirectories(srcPath) {
+  return fs.readdirSync(srcPath).filter(file => fs.statSync(path.join(srcPath, file)).isDirectory());
+}
+
+function apply(options) {
```
Matt-Butler: Nit: can we rename this? js already has the popular function apply

ryanthemanuel: Again, this was kind of how the example was written in webpack.  I believe we'd be fine renaming it though.

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
ryanthemanuel: I'm wondering if we could check the current base directory for a translations directory as well.  Similar to the logic on line 29, but looking inside of options.baseDirectory/translations directly.  That way you can have translations at the top level of the project as well.

ryanthemanuel:
Ultimately I'm fine with duplicating the supported locales since it's not a ton of code, but I do worry about them being out of sync.  Maybe we could add it as a parameter to the plugin?  That way there wouldn't be a direct dependency on terra-i18n anymore.

package.json
emilyrohrbough: [Removed](https://github.com/cerner/terra-core/pull/272/commits/b478c09c858a6e60c1a28bea149790077150ffaa#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L79)

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
emilyrohrbough: Renamed to aggregateTranslations [here](https://github.com/cerner/terra-core/pull/272/commits/b478c09c858a6e60c1a28bea149790077150ffaa#diff-4417b6331e7583b396c941b65e5e4246R67)

packages/terra-i18n-plugin/package.json
emilyrohrbough: `i18nSupportedLocales` file is now passed into the i18n-plugin as per @ryanthemanuel's suggestion. This will keep the supported languages file within i18n and ensure the files stay in sync. This also removed the terra-i18n dependency from the i18n-plugin. Change made [here](https://github.com/cerner/terra-core/pull/272/commits/b478c09c858a6e60c1a28bea149790077150ffaa#diff-9999c132bd53c2a36d76ea10d837dd57R11).

emilyrohrbough:
`i18nSupportedLocales` file is now passed into the i18n-plugin as per @ryanthemanuel's suggestion. This will keep the supported languages file within i18n and ensure the files stay in sync. This also removed the terra-i18n dependency from the i18n-plugin. Change made [here](https://github.com/cerner/terra-core/pull/272/commits/b478c09c858a6e60c1a28bea149790077150ffaa#diff-9999c132bd53c2a36d76ea10d837dd57R11).

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
emilyrohrbough: Updated the plugin to search for translations in options.baseDirectory/translation and included any found translations in the currentLanguageMessages object. View change [here](https://github.com/cerner/terra-core/pull/272/commits/b478c09c858a6e60c1a28bea149790077150ffaa#diff-4417b6331e7583b396c941b65e5e4246R29)

Issue #273, opened 2017-04-27
rm -rf does not work on windows machines


Matt-Butler:
https://github.com/isaacs/rimraf

rimraf is a popular library that replaces `rm -rf` and is OS agnostic.

Matt-Butler:
https://github.com/cerner/terra-core/pull/281

PR #274, opened 2017-04-27
rm -rf does not work on Windows

Matt-Butler:
https://github.com/cerner/terra-core/issues/273


Matt-Butler:
Several open soure projects use rimraf. 

Cloudflare-ui uses rimraf: https://github.com/cloudflare/cf-ui/blob/master/package.json#L59

Issue #264, opened 2017-04-26
Update props in terra-date-picker and terra-time-input from a moment object to an ISO 8601 string


benbcai:
The type for the following props in terra-date-picker and terra-time-input is a moment object. The type should be updated to take an ISO 8601 string for specifying the date and time.

terra-date-picker:
* `endDate`
* `maxDate`
* `minDate`
* `selectedDate`
* `startDate`

terra-time-input:
* `defaultTime`

See https://github.cerner.com/terra/tech-designs/pull/22#pullrequestreview-5665 for reason for making this change.

PR #275, opened 2017-04-27
terra-date-picker - Update props type from moment object to ISO-8601 …

benbcai:
…string representation. Issue 264.

Updating the type of props in terra-date-picker from an object (moment) to string (ISO 8601 string) for representing the date. 

Resolves https://github.com/cerner/terra-core/issues/264

@cerner/terra


packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -32,39 +36,64 @@ const propTypes = {
...
+      let index = 0;
+      for (index = 0; index < dates.length; index += 1) {
+        const momentDate = moment.utc(dates[index], dateFormat);
+        if (momentDate.isValid()) {
+          momentDates.push(momentDate);
+        }
+      }
+    }
+
+    return momentDates.length > 0 ? momentDates : dates;
```
JakeLaCombe: What if some dates are valid, and other dates are not? Should we perhaps return some kind of mapping indicating the successful dates, and another indicating the failure ones?

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -32,39 +36,64 @@ const propTypes = {
...
 const defaultProps = {
   isEndDateRange: false,
   isStartDateRange: false,
 };
 
 class DatePicker extends React.Component {
+
+  static safeMoment(date) {
+    if (date && dateFormat) {
+      const momentDate = moment.utc(date, dateFormat);
```
JakeLaCombe: What happens if you pass in date and dateFormat when they are empty? If it returns an invalid date, you can make the code a little cleaner by not needing the if statement.

However, I also understand that if you know they are both empty, then no need to interact with moment for a small performance boost.

packages/terra-date-picker/src/DateRange.jsx
```diff
@@ -1,27 +1,34 @@
...
+  static safeMoment(dateISO8601) {
+    const momentDate = moment.utc(dateISO8601);
+
+    return momentDate.isValid() ? momentDate : null;
+  }
 
   constructor(props) {
     super(props);
     this.state = {
+      format: 'MM/DD/YYYY', // TODO: Get the format from i18n
```
JakeLaCombe: One concern I remember we had was different applications having different formats. We may have discussed the issue before (I'm sure we did), but I unfortunately don't remember our solution. What was our final conclusion on that issue?

packages/terra-date-picker/src/DatePicker.jsx
benbcai: If date is empty, it would return an invalid moment object. If format is empty but with a valid date, it would still create the moment object but it may not inconsistent across browsers. Moment recommends to always create using both a date + format. See http://momentjs.com/docs/#/parsing/string/

packages/terra-date-picker/src/DatePicker.jsx
benbcai: If a date is invalid, it will be filtered on line 82. So it will only use valid dates. We can store the invalid dates but what should we do with them? Perhaps, log a warning message.

packages/terra-date-picker/src/DateRange.jsx
benbcai: The format is temporarily hard-coded for now and will be updated to use the format based on the locale using i18n. This format is needed to create a moment object from the ISO 8601 string on line 43.

benbcai:
I'm going to close this PR and issue a new one with some of the more recent changes.

Issue #254, opened 2017-04-21
Table.Rows Fails When Given One Child Row


emilyrohrbough:
### Description of Issue
The current version of terra-table does not allow for a table with one row to be generated and results in webpage being unrended. The Table.Rows component throws the following error when it is given one Table.Row child.  

> Uncaught TypeError: children.map is not a function 

Although most people will want more than one row in a table, the component should still render when given one row.

#### Component Version
terra-table 0.2.0

### Steps to Reproduce the Issue
The following is an example of when the error is created

  `<Table.Rows>`
       ` <Table.Row key={'row'}> <Table.Cell content={row.displays} key="cell" />
        </Table.Row> `
    `</Table.Rows>`

### Expected Outcomes
- The Table.Rows component should render when given one row or many rows.

tbiethman:
We need to use React.Children.map(props.children, ()=>{}) to safely deal with the opaque children object.

PR #282, opened 2017-04-28
Fix Table.Rows Rendering Error for One Child

emilyrohrbough:
### Summary
Updated Rows to handle zero-many children by using React.Children.map.
Addresses #254 

@cerner/terra

### DIFF
--- a/packages/terra-table/src/TableRows.jsx
+++ b/packages/terra-table/src/TableRows.jsx
@@ -23,7 +23,7 @@ const defaultProps = {
 };
 
 function cloneChildItems(children, onClick, onKeyDown) {
-  return children.map((child) => {
+  return React.Children.map(children, (child) => {
     const newProps = {};
     if (onClick) {
       newProps.onClick = onClick;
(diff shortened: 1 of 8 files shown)


packages/terra-table/src/TableRows.jsx
```diff
@@ -23,7 +23,7 @@ const defaultProps = {
 };
 
 function cloneChildItems(children, onClick, onKeyDown) {
-  return children.map((child) => {
+  return React.Children.map(children, (child) => {
```
StephenEsser: An alternative is to mark children as required and or default children to an empty array
```
const defaultProps = {
  children: [],
  onClick: undefined,
  onKeyDown: undefined,
};
```

Matt-Butler:
@emilyrohrbough If you have time, can you add a test so we can verify this change fixed the issue?

emilyrohrbough:
@Matt-Butler Tests included.

Matt-Butler:
@emilyrohrbough Thank you!

PR #290, opened 2017-05-01
Issue 263 date time prop type

### LINKED ISSUE #264: Update props in terra-date-picker and terra-time-input from a moment object to an ISO 8601 string

benbcai:
Updating the type of props in terra-date-picker from an object (moment) to string (ISO 8601 string) for representing the date.

Resolves #264

@cerner/terra


packages/terra-date-picker/README.md
```diff
@@ -3,7 +3,7 @@
 [![NPM version](http://img.shields.io/npm/v/terra-date-picker.svg)](https://www.npmjs.org/package/terra-date-picker)
 [![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
 
-The terra-date-picker component provides users a way to enter or select a date from the date picker.
+The terra-date-picker component provides users a way to enter or select a date from the date picker. terra-date-picker is essentailly a wrapper for [react-datepicker](https://github.com/Hacker0x01/react-datepicker) and leverages many of its props. One important difference between terra-date-picker and [react-datepicker](https://github.com/Hacker0x01/react-datepicker) is that all of the date props in [react-datepicker](https://github.com/Hacker0x01/react-datepicker) must be a [moment](http://momentjs.com/docs/) object whereas the date props in terra-date-picker are ISO 8601 string representation of the date.
```
JakeLaCombe: essentially*

benbcai: updated.

packages/terra-date-picker/src/DateUtil.jsx
```diff
@@ -0,0 +1,30 @@
+import moment from 'moment';
+
+class DateUtil {
+  static createSafeMoment(date, format) {
```
JakeLaCombe: This function feels misnamed. When I think createSafeMoment, it makes me think that it's returning a moment object, whereas if the moment object isn't valid, it returns the value that was passed in. Perhaps createSafeDate would be a better option than createSafeMoment?

packages/terra-date-picker/src/DateUtil.jsx
```diff
@@ -0,0 +1,33 @@
+import moment from 'moment';
```
mjhenkes: should this file be js instead of jsx? Since it doesn't have any react components in it.

benbcai: Updated the file from .jsx to .js. https://github.com/cerner/terra-core/pull/290/commits/ea493813ebb2ba2c613dd5152cf3fa7d224a11f3

bjankord:

PR #292, opened 2017-05-01
terra-date-picker - Issue 276 - Allow passing additional attributes t…

### LINKED ISSUE #276: Allow passing additional attributes to the datepicker input element
We need to be able to add additional attributes to the input tag of the datepicker. For example, the name attribute.

benbcai:
Allow passing additional attributes to the date-picker input element.

Resolves #276

@cerner/terra


packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -8,6 +8,10 @@ import './DatePicker.scss';
 
 const propTypes = {
   /**
+   * Custom attributes to apply to the date input.
+   */
+  attrs: PropTypes.object,
```
StephenEsser: My personal preference would be to make this prop name a little more descriptive. I spoke with Jake and Matt Butler a little bit about this and he thought something like inputAttributes might be more clear.

I don't necessarily think this prop name and the forms prop name need to match, but if there was a generic name that fit all cases it would be more consistent. `attrs` feels a little too generic but I don't have any good alternatives.

benbcai: Agree. Renamed prop attrs to inputAttributes https://github.com/cerner/terra-core/pull/292/commits/ed6a858697a6cc5ee834047a74ad225eeca4db75

bjankord:

Issue #267, opened 2017-04-27
Default props for boolean props should only be true or false


Matt-Butler:
# Issue Description
Default props for boolean variables should be either true or false.

One example:
https://github.com/cerner/terra-core/blob/master/packages/terra-list/src/ListItem.jsx#L27-L30

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Any boolean default prop should be set to either true or false.

## Current Behavior
Not all boolean props are set as true or false.

Matt-Butler:
https://github.com/cerner/terra-core/pull/266/files#r113695096

emilyrohrbough:
Defining an item as isSelectable has selection indication via hover highlighting/mouse indication.  
If default prop is true, all nonselectable list will need to pass isSelectable=false for each item to prevent hover highlighting/mouse pointer.

Issue #268, opened 2017-04-27
Handling unused props in the render method


Matt-Butler:
# Issue Description

We have some components that have props not used inside the render method. We should be consistent on how we handle them.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior

[Method 1: delete the props so they are not spread](https://github.com/cerner/terra-core/blob/81337a98119f2fc51c22cd0ce0a471c66400c78e/packages/terra-table/src/SingleSelectableRows.jsx#L121-L135)
```
  render() {
    const { children, ...customProps } = this.props;
    const clonedChilItems = this.clonedChildItems(children);
    if ('onClick' in customProps) {
      delete customProps.onClick;
    }
    if ('onKeyDown' in customProps) {
      delete customProps.onKeyDown;
    }
    return (
      <TableRows {...customProps}>
        {clonedChilItems}
      </TableRows>
    );
  }

}
```

[Method 2: ignore the eslint rule](https://github.com/cerner/terra-core/blob/57baaf99ebb86e32cf39e6c979932eb3311e5e47/packages/terra-responsive-element/src/ResponsiveElement.jsx#L100-L110)

```
  render() {
    /* eslint-disable no-unused-vars */
    const { defaultElement, tiny, small, medium, large, huge, responsiveTo, ...customProps } = this.props;

    return (
      <div {...customProps} ref={this.setContainer}>
        {this.props[this.state.element]}
      </div>
    );
  }
}
```

## Current Behavior

emilyrohrbough:
Here is some interesting [discussion](https://github.com/babel/babel-eslint/issues/95) on an issue logged in the babel-eslint repo. One comment referenced to this [proposal](https://github.com/babel/babel-eslint/issues/95) which suggested the use of `_` for unused variables.

windse7en:
It seems [eslint #7968 PR](https://github.com/eslint/eslint/pull/7968) fixes this issue. If there are unused props and `...customProps(remainder props)`, it's not considered to be unused props. As those props will affect the value in remainder props. So it won't get lint error.

bjankord:
Let's go with option 1 but take a look at the following

Need to:
* Check what version this fix landed in eslint
* Check what airbnb config to see if it took this rule

Matt-Butler:
[...]

PR #294, opened 2017-05-02
Updated Default Bool Props and Fixed List MulitSelection Bug

emilyrohrbough:
### Summary
1. The default boolean props for ListItem and TableRow were updated from undefined to a true or false value. Due to this change, the selection implementation was cleaned up and snapshots and examples were updated to reflect this change.
This address issue #267.

2. Multiple Selection had a bug that allowed you to select more rows than the indicated maximum selection. This was fixed and the examples and tests were updated to reflect this was the intended behavior. Additional tests were included for when no maximum selection was placed on the list.

@cerner/terra


packages/terra-list/src/MultiSelectList.jsx
```diff
@@ -163,16 +163,10 @@ class MultiSelectList extends React.Component {
       newProps.isSelected = isSelected;
     }
```
dkasper-was-taken: Looks like it was removed in a previous commit, but if the item isn't selectable, I don't believe we should be wrapping its onClick or KeyDown as we have no interaction with the element.

dkasper-was-taken: This would also mean we don't need to check if the element is selectable in the click event.

packages/terra-list/lib/SingleSelectList.js
```diff
@@ -65,7 +65,7 @@ var SingleSelectList = function (_React$Component) {
     key: 'selectedIndexFromItems',
     value: function selectedIndexFromItems(items) {
       for (var i = 0; i < items.length; i += 1) {
-        if (items[i].props.isSelected === true && items[i].props.isSelectable !== false) {
+        if (items[i].props.isSelected === true && items[i].props.isSelectable) {
```
Matt-Butler: I think we can make
```
items[i].props.isSelected === true
```
just
```
items[i].props.isSelected
```

packages/terra-list/src/SingleSelectList.jsx
```diff
@@ -36,7 +36,7 @@ class SingleSelectList extends React.Component {
 
   static selectedIndexFromItems(items) {
     for (let i = 0; i < items.length; i += 1) {
-      if (items[i].props.isSelected === true && items[i].props.isSelectable !== false) {
+      if (items[i].props.isSelected === true && items[i].props.isSelectable) {
```
Matt-Butler: I think we can make
```
items[i].props.isSelected === true
```
just
```
items[i].props.isSelected
```

emilyrohrbough: Yes. Fixed [here](https://github.com/cerner/terra-core/pull/294/commits/df5e2565500f08672ccad1c1bbc518d47a0888c8#diff-0e18bc4abc1b29779e0a97b763acd271R68).

packages/terra-list/src/MultiSelectList.jsx
emilyrohrbough: That is a good point. Fixed [here](https://github.com/cerner/terra-core/pull/294/commits/df5e2565500f08672ccad1c1bbc518d47a0888c8#diff-0e18bc4abc1b29779e0a97b763acd271R180).

packages/terra-list/src/SingleSelectList.jsx
```diff
@@ -123,22 +123,13 @@ class SingleSelectList extends React.Component {
...
+    if (newProps.isSelectable) {
       newProps.tabIndex = '0';
     }
 
-    if (item.props.hasChevron === undefined) {
-      newProps.hasChevron = this.props.hasChevrons;
-    }
+    newProps.hasChevron = this.props.hasChevrons;
 
     return newProps;
```
dkasper-was-taken: I believe the change to delete the "onChange" and "maxSelectionCount" in mutli select, and 'onChange' and 'hasChevrons' in single select should not have been made. That is directly manipulation the props and we are losing the onChange event passed into the component by the user. A better solution would be to use:
// eslint-disable-line no-unused-vars
when we are unpacking the customProps.

dkasper-was-taken: These values are used elsewhere in the code, just not within render.

packages/terra-list/src/MultiSelectList.jsx
dkasper-was-taken: We can also update the wrap functions to remove isSelectable from this check:
if (item.props.isSelectable && this.shouldHandleSelection(index)) {

packages/terra-list/src/SingleSelectList.jsx
emilyrohrbough: Issue #268 has been made regarding this implementation. No decision has been made at this time how how to handle theses unused props.

packages/terra-list/src/MultiSelectList.jsx
```diff
@@ -138,7 +138,7 @@ class MultiSelectList extends React.Component {
 
     return (event) => {
       if (event.nativeEvent.keyCode === KEYCODES.ENTER) {
-        if (item.props.isSelectable !== false && this.shouldHandleSelection(index)) {
+        if (item.props.isSelectable && this.shouldHandleSelection(index)) {
           this.handleSelection(event, index);
         }
```
dkasper-was-taken: The onChange event for the wrappedKeydown should be located inside the same if check as the handle selection similar to the wrappedOnClick above. Otherwise we will always call the onChange.

dkasper-was-taken: Also we may need to break out the logic inside handleSelection. Set state isn't occuring synchronously, so we need to calculate the desired value for the onChange notifcation.

In single select, it will just be the new index.

Multi-select will need additional logic.

packages/terra-list/src/SingleSelectList.jsx
dkasper-was-taken: In either case, we need to not delete the values, even if it's temporary. Deleting them introduces a bug.

dkasper-was-taken: Actually looks like my bad. 'this.props.onChange' should be the correct usage, not sure if removing the delete is necessary.

packages/terra-list/src/MultiSelectList.jsx
dkasper-was-taken: Or move onChange into handleSelection.

packages/terra-list/lib/MultiSelectList.js
```diff
@@ -130,6 +130,9 @@ var MultiSelectList = function (_React$Component) {
       }
 
       this.setState({ selectedIndexes: newIndexes });
+      if (this.props.onChange) {
+        this.onChange(event, newIndexes);
```
dkasper-was-taken: this.props.onChange(event, newIndexes);

packages/terra-list/src/MultiSelectList.jsx
```diff
@@ -92,6 +92,9 @@ class MultiSelectList extends React.Component {
     }
 
     this.setState({ selectedIndexes: newIndexes });
+    if (this.props.onChange) {
+      this.onChange(event, newIndexes);
```
dkasper-was-taken: this.props.onChange(event, newIndexes);

packages/terra-list/src/MultiSelectList.jsx
```diff
@@ -186,6 +182,7 @@ class MultiSelectList extends React.Component {
   }
 
   render() {
+    /* eslint-disable no-unused-vars */
```
dkasper-was-taken: you may not need this now.

PR #297, opened 2017-05-02
Alias moment and react-dom

Matt-Butler:
### Summary
- Alias moment and react-dom.  There were multiple copies of moment being pulled into our bundle. This is due to react-datepicker having moment as a dependency instead of a peerDependency. React-datepicker has an outstanding issue for this here: https://github.com/Hacker0x01/react-datepicker/issues/122

After adding the alias for moment and react-dom, we can see that only one copy is pulled into our bundle. 

**Before:**
before
**After**
after

These images were generated using https://github.com/th0r/webpack-bundle-analyzer

Making this update
terra-core.js    5.62 MB
to
terra-core.js    4.61 MB

@cerner/terra


packages/terra-site/webpack.config.js
```diff
@@ -70,7 +70,9 @@ module.exports = {
   ],
   resolve: {
     extensions: ['.js', '.jsx'],
-    alias: { react: path.resolve(__dirname, 'node_modules', 'react') },
+    alias: { 'react$': path.resolve(__dirname, 'node_modules', 'react') },
+    alias: { 'react-dom$': path.resolve(__dirname, 'node_modules', 'react') },
+    alias: { 'moment$': path.resolve(__dirname, 'node_modules', 'moment') },
```
Matt-Butler: https://webpack.js.org/configuration/resolve/#resolve-alias

PR #301, opened 2017-05-03
Implement terra-form and related components

JakeLaCombe:
### Summary
These are the base components needs to build forms using the Terra Framework

### Additional Details
This is barebones components. Validations will come at a later time.

@cerner/terra


packages/terra-form/src/components/Field.jsx
```diff
@@ -0,0 +1,61 @@
...
+const defaultProps = {
+  children: null,
+  error: null,
+  help: null,
+  isInline: false,
+  label: null,
+  required: false,
+};
+
+const Field = ({ children, error, help, isInline, label, required, ...customProps }) => {
```
Matt-Butler: The  required prop is unused here

JakeLaCombe:
I moved this to a different pull request due to build issues. I also used a different generator as well on the other pull request, one that is more upto date than the one used here I believe.

PR #305, opened 2017-05-04
Implement Terra Forms

JakeLaCombe:
### Summary
There are the base components used for building forms with Terra.

### Additional Details
This is bare bones forms. Features such as validation will come at a later date.


packages/terra-form/docs/ChoiceField.md
```diff
@@ -0,0 +1,22 @@
+# Choice Field
```
StephenEsser: What was the reasoning for naming this a ChoiceField, as opposed to something like RadioGroup?

packages/terra-form/src/components/ChoiceField.jsx
```diff
@@ -0,0 +1,103 @@
+/* eslint-disable jsx-a11y/label-has-for */
+
+import React, { PropTypes } from 'react';
+
+import Control from './Control';
+import Field from './Field';
+
+const propTypes = {
+  /**
+   * Choices to populate radio buttosn for
```
StephenEsser: Typo on buttons

packages/terra-form/src/components/ChoiceField.jsx
```diff
@@ -0,0 +1,103 @@
...
+  /**
+   * Initial Value of the input
+   */
+  value: PropTypes.string,
+};
+
+const defaultProps = {
+  choices: [],
+  choiceName: 'name',
+  choiceValue: 'value',
```
StephenEsser: From a form submission standpoint I think defaulting these values could be really dangerous, accidentally submitting incorrect information to a server.

packages/terra-form/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
+    "terra-toolkit": "^0.x"
+  },
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2"
```
bjankord: Need to add `"terra-base": "^0.x",` to peerDependencies and dependencies.

packages/terra-form/src/components/ChoiceField.jsx
```diff
@@ -0,0 +1,103 @@
...
+import React, { PropTypes } from 'react';
+
+import Control from './Control';
+import Field from './Field';
+
+const propTypes = {
+  /**
+   * Choices to populate radio buttosn for
+   */
+  choices: PropTypes.array,
```
StephenEsser: I believe Brett called out on the original Tech Design that this prop might make more sense being called `options`.

I personally don't think we should use a prop API that consists of an Array of hashes with key value pairs. The React form libraries I've worked with use children and I think it makes it a lot easier for the consumer. 

```
      <RadioGroup onChange={this.onChange} value={this.state.value}>
        <Radio value={1}>A</Radio>
        <Radio value={2}>B</Radio>
        <Radio value={3}>C</Radio>
        <Radio value={4}>D</Radio>
      </RadioGroup>
```
https://ant.design/components/radio/

packages/terra-form/docs/ChoiceField.md
```diff
@@ -0,0 +1,22 @@
...
+import { ChoiceField } from 'terra-form';
+
+<ChoiceField
+  label="What State do you live in?"
+  name="children_present"
+  value="children_present"
+  error="This field is required"
+  help="State Locations help determine what campus to place you at"
+  required={true}
+  attrs={{ className: 'healtheintent-application' }}
```
StephenEsser: Looking at the ChoiceField.jsx below I don't think attrs is every actually used.

packages/terra-form/src/components/ChoiceField.jsx
JakeLaCombe: +1 to that idea. I think that looks more clear than passing in choiceName and choiceValue as well.

packages/terra-form/src/components/ChoiceField.jsx
JakeLaCombe: correct. the choiceName is the label of the input, and choiceValue is the value of the input. I like the idea you posted down with the children. I'll see what other developers think of it as well

packages/terra-form/src/components/ChoiceField.jsx
```diff
@@ -0,0 +1,103 @@
+/* eslint-disable jsx-a11y/label-has-for */
```
StephenEsser: What is the error being thrown for on this?

packages/terra-form/src/components/Control.jsx
```diff
@@ -0,0 +1,73 @@
+// eslint-disable react/jsx-boolean-value
```
StephenEsser: What is this masking?

packages/terra-form/docs/ChoiceField.md
JakeLaCombe: I think the main decision was to keep in consistent naming with the textareaField and textField components. I'm not opposed to renaming it radio group though. Perhaps it should be RadioGroupField to indicate that it utilizes a field component as well?

packages/terra-form/src/components/Control.jsx
```diff
@@ -0,0 +1,73 @@
...
+
+const propTypes = {
+  /**
+   * The input type
+   */
+  type: PropTypes.oneOf(['checkbox', 'radio']).isRequired,
+  /**
+   * Additional attributes for the input object
+   */
+  attrs: PropTypes.object,
```
StephenEsser: Does anyone have strong opinions on this? Something like inputAttributes seems more clear to me

packages/terra-form/docs/Field.md
```diff
@@ -0,0 +1,24 @@
...
+import { Field } from 'terra-form';
+
+<Field
+  type="checkbox"
+  label="Do you have any Children?"
+  name="children_present"
+  value="children_present"
+  error="This field is required"
+  help="Families are eligible for family package plans"
+  required={true}
```
StephenEsser: Does the Field accept this prop?

packages/terra-form/src/components/Input.jsx
```diff
@@ -0,0 +1,71 @@
...
+    PropTypes.number,
+  ]),
+  /**
+   * Name of the input
+   */
+  name: PropTypes.string,
+  /**
+   * Functional to be called when the input is changed.
+   */
+  onChange: PropTypes.func,
```
StephenEsser: I think the description meant to have `function`

packages/terra-form/docs/Input.md
```diff
@@ -0,0 +1,16 @@
...
+Generic input which represents an HTML input element directly.
+
+## Usage
+
+```jsx
+import React from 'react';
+import { Input } from 'terra-form';
+
+<Input
+  required={false}
```
StephenEsser: Nit: Our Booleans are getting a little awkward. This is kind of up to preference, this value defaults to false, so passing in an explicit false does the same thing. There are also examples in these docs that use `required={true}` when just passing required in should do the same thing. Do we have a preference or standard on how we recommend using booleans? The two generic paths being:
```
<Input required />
```
or
```
<Input required={true} />
```
The current prop `required` also breaks our `is` prefix, but it matches the html API, so I don't know which one is more user friendly

JakeLaCombe: Our linter prefers us to use the top approach you have when the boolean value is known to be true. Looks like I forgot to update the documentation to reflect the change. I'll get the MD fixes.

As for the is prefix, I feel like the solution is situational. required is a standard html attribute, so for a variable like required, I think keeping it as it would be the ideal approach. For something like isInline, inline isn't a mapped html attribute, so I feel that the prefix would be needed there.

packages/terra-form/src/components/Control.jsx
JakeLaCombe: I think some of us agreed to switch it to that. I'll make the change

packages/terra-form/src/components/Textarea.jsx
```diff
@@ -0,0 +1,76 @@
+import React, { PropTypes } from 'react';
```
StephenEsser: Is this the file that was causing all the case sensitive issues? Should this be TextArea?

packages/terra-form/docs/Input.md
StephenEsser: In our early meetings I thought we had discussed not using html attributes as prop names, but I could be remembering that conversation incorrectly, it was a while back.

packages/terra-form/docs/ChoiceField.md
```diff
@@ -0,0 +1,22 @@
...
+## Usage
+
+```jsx
+import React from 'react';
+import { ChoiceField } from 'terra-form';
+
+<ChoiceField
+  label="What State do you live in?"
+  name="children_present"
+  value="children_present"
```
bjankord: I think the values used in the `name` and `value` props need updated, they don't seem to make sense in the ChoiceField example.

packages/terra-form/docs/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+[Textarea](https://github.com/cerner/terra-core/tree/master/packages/terra-button/docs/Textarea.md)
+
+[TextareaField](https://github.com/cerner/terra-core/tree/master/packages/terra-button/docs/TextareaField.md)
+
+[TextField](https://github.com/cerner/terra-core/tree/master/packages/terra-button/docs/TextField.md)
+
+To use these components, create a React component that contains a form element, and then insert the Terra form components into that form.
+
+The recommended approach for building forms is to stick with the field components that are build for you (ChoiceField, MultiChoiceField, NumberField, Textarea, TextareaField, Field). If you need to create a custom field, the Field, Control, Input, and Textarea components are available for use.
```
bjankord: typo: `build` => `built`

packages/terra-form/docs/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+When using input fields, React has two ways of using inputs: controlled or uncontrolled. Controlled maps the value of the input to the state of the component, while uncontrolled does not (https://facebook.github.io/react/docs/uncontrolled-components.html). If you would like to use a controlled input (ideal for forms that need to do validations with JavaScript and send ajax requests), provide a value and an onChange function to the component. If you would like to use an uncontrolled input (ideal for forms that do straight backend requests), provide just a defaultValue into the input **Note: If you provide both a value and defaultValue to the input, React will be confused as to whether you are working with an uncontrolled input or a controlled one. Provide either a value or a defaultValue to the input, but not both.**
```
bjankord: Let's link the first reference of controlled in this paragraph to https://facebook.github.io/react/docs/forms.html#controlled-components and link the first reference on uncontrolled to https://facebook.github.io/react/docs/uncontrolled-components.html

packages/terra-form/src/components/Field.jsx
```diff
@@ -0,0 +1,56 @@
...
+
+const Field = ({ children, error, help, isInline, label, ...customProps }) => {
+  const fieldClasses = classNames(
+    'terra-Form-field',
+    { 'terra-Form-field--inline': isInline },
+    customProps.className,
+  );
+
+  return (
+    <p {...customProps} className={fieldClasses}>
```
bjankord: nit: change `p` element to `div` element.

packages/terra-form/docs/README.md
```diff
@@ -0,0 +1,37 @@
...
+  - `npm install terra-form`
+  - `yarn add terra-form`
+
+## Usage
+
+Terra has several different form components for building forms.
+
+[ChoiceField](https://github.com/cerner/terra-core/tree/master/packages/terra-button/docs/ChoiceField.md)
+
+[Control](https://github.com/cerner/terra-core/tree/master/packages/terra-button/docs/Control.md)
```
bjankord: we'll want to update these links, currently they all point to the button package docs.

`https://github.com/cerner/terra-core/tree/master/packages/terra-button/docs/Control.md` => `https://github.com/cerner/terra-core/tree/master/packages/terra-form/docs/Control.md`

packages/terra-form/src/components/Field.jsx
```diff
@@ -0,0 +1,56 @@
...
+  help: PropTypes.node,
+  /**
+   * Determines whether the field in an inline field
+   */
+  isInline: PropTypes.bool,
+  /**
+   * Label of the input
+   */
+  label: PropTypes.node,
+};
```
bjankord: I believe we will also want to add a prop for `required` to this component. In the prior code base, we used a the presence of a class, `.field-required`, to alter the label display to help indicate that a field was a required input. 

With the required prop, we could add a class `terra-Form-field--required` when it is set to true to alter the styling.
Here is a sample of the css code from the prior code base:

```css
.field-required > label:first-child {
  &:before{
    .padding-end(5px);
    color: @text-error;
    content: '*';
    display: inline-block;
    font-weight: @bold-font-weight;
  }
}
```

packages/terra-form/src/components/Textarea.jsx
JakeLaCombe: I debated that myself. I mainly stick to Textarea to keep the same naming convention of the html node textarea. I'm not picky on this though, and if you'd like it changed, I have no problem doing so.

packages/terra-form/src/components/ChoiceField.jsx
JakeLaCombe: https://github.com/cerner/terra-core/pull/305/commits/ca52f87cdfff47ee4868e83427dcf21fc7786de5

packages/terra-form/docs/Input.md
JakeLaCombe: I did see these standards where they wanted to mirror the DOM whenever it was possible. https://github.cerner.com/terra/conventions/blob/master/react.md#mirror-dom-wherever-possible

I'll look through war room agendas to see if we had updates.

packages/terra-form/src/components/Field.jsx
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable react/jsx-boolean-value, jsx-a11y/label-has-for */
```
Matt-Butler: There are a few places where eslint is disabled for the entire file. Can we update it to just ignore specific lines so it is easier to what is being ignored and why it is being ignored?

packages/terra-form/docs/Input.md
StephenEsser: It was most likely when there were conflicting props, like using title as a prop to display text, when title would otherwise be spread onto the dom element and perform a different role. When the roles are the same it probably makes more sense to mirror.

packages/terra-form/src/components/TextareaField.jsx
```diff
@@ -0,0 +1,126 @@
...
+   */
+  required: PropTypes.bool,
+  /**
+   * How many rows the textarea should have.
+   */
+  rows: PropTypes.number,
+  /**
+   * Initial Value of the input
+   */
+  value: PropTypes.string,
```
Matt-Butler: nit: this has the same description as defaultValue

packages/terra-form/lib/Form.scss
```diff
@@ -0,0 +1,60 @@
+@import '~terra-mixins';
+
+.terra-Form-input,
+.terra-Form-textarea {
+  border: $terra-grey-15 solid 1px;
+  border-radius: 0.25em;
+  font-size: 16px;
```
Matt-Butler: Do we need this? Should it flex with the base font-size?

packages/terra-form/lib/Form.scss
```diff
@@ -0,0 +1,60 @@
...
+.terra-Form-error {
+  color: $terra-red-70;
+  display: block;
+  margin-top: 0.357em;
+}
+
+@media screen and (min-width: $terra-small-breakpoint) {
+  .terra-Form-field--inline {
+    display: inline-block;
+    margin-right: 0.714em;
```
Matt-Butler: Should we use a mixin here? https://github.com/cerner/terra-core/blob/db2fec58904dd16ccab541dd5d575eb1fc10284e/packages/terra-mixins/src/Mixins.scss#L343

packages/terra-form/src/components/ChoiceField.jsx
```diff
@@ -0,0 +1,103 @@
...
+
```
Matt-Butler: We probably  need to import baseStyles in all of these components

```
```

bjankord: Yes, this allows them to be loaded independently without needing to be within the `<Base>` component.

packages/terra-form/docs/README.md
Matt-Butler: Do we need to support both controlled and uncontrolled forms? If we don't need to, I would prefer to go with a single approach so it is easier for a consumer to understand how to use it.

packages/terra-site/src/examples/form/examples/Field.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { Field } from 'terra-form';
+
+const FieldExamples = () => (
+  <form>
```
Matt-Butler: Do we provide a form component?

packages/terra-site/src/examples/form/navigation.md
```diff
@@ -0,0 +1,23 @@
+## Components
+
+[ChoiceField](#choice-field)
```
Matt-Butler: The urls seem wrong here. 
```
http://localhost:8080/#/choice-field
```

It should be 
```
http://localhost:8080/#/form/choice-field
```
The routes also don't exist.
screen shot 2017-05-08 at 5 06 04 pm

packages/terra-site/src/examples/form/examples/FullForm.jsx
```diff
@@ -0,0 +1,201 @@
+import React from 'react';
```
Matt-Butler: submit_issue

This example seems a little weird. The data is being updated without hitting submit

Matt-Butler: Only changes to the checkboxes update the values immediately.

packages/terra-form/src/Form.jsx
```diff
@@ -0,0 +1,23 @@
...
+import Control from './components/Control';
+import Field from './components/Field';
+import Input from './components/Input';
+import MultiChoiceField from './components/MultiChoiceField';
+import NumberField from './components/NumberField';
+import Textarea from './components/Textarea';
+import TextareaField from './components/TextareaField';
+import TextField from './components/TextField';
+
+export {
```
Matt-Butler: I don't think we want to export these components from a single file.   Consumers should  import these from the file they are exported from.

```
import { ChoiceField } from 'terra-form';
```
^ This will include all of the form components in our bundle.

```
import ChoiceField from 'terra-form/lib/choicefield';
```
^ This will only include the choicefield in our bundle.

This example is where tree shaking can help, but I don't think we should assume consumers have that ability.

StephenEsser:
reeqkt3bmo76w

packages/terra-site/src/examples/form/examples/Field.jsx
JakeLaCombe: Not for Terra form. The consumer will need to build the form theirselves, and then provide the terra form components they wish to use.

packages/terra-form/src/Form.jsx
JakeLaCombe: I'm fine with the suggestion.

To allow for future additions to be added with ease, I've structured the code to have src/components, which allows us to easily integrate enhancements such as src/validations.

https://github.com/cerner/terra-core/pull/305/commits/71c6a527fd474ebaff4a30b9f3f8d3316d36512f

packages/terra-site/src/examples/form/examples/FullForm.jsx
JakeLaCombe: That was a mistake with how I cloned the multi-choice data. I didn't do enough of a clone to prevent this from occurring. It should be fixed now. https://github.com/cerner/terra-core/pull/305/commits/6fa0cb4882422c2a20378ef7ad938c27a571bef1

packages/terra-form/docs/README.md
JakeLaCombe: We will. For some forms, it's been requested that they have no interactions with JavaScript outside of having React render the form. For this case, it's easiest to leave the input as uncontrolled.

packages/terra-site/src/examples/form/navigation.md
JakeLaCombe: I was going to hash links here. It's odd, as this is the best way I'm seeing to provide a hash link, but the router keeps trying to redirect me. If I can't provide hash links in an efficient way, I'll look into just making new url's for all the components.

JakeLaCombe: I decided to use the terra buttons, and do the scrollIntoView approach. From looking at https://github.com/ReactTraining/react-router/issues/394, it looks like that's the recommended approach.

https://github.com/cerner/terra-core/pull/305/commits/916aa0d7b5e2b244c2598284bb757011425af412

packages/terra-form/src/components/ChoiceField.jsx
JakeLaCombe: https://github.com/cerner/terra-core/pull/305/commits/3aa7abe917c13b32018425d8311b7b3f1097c823

packages/terra-form/lib/Form.scss
JakeLaCombe: https://github.com/cerner/terra-core/pull/305/commits/ad5b73249724de39d108d2ed8f54e25281f04dbc

packages/terra-form/lib/Form.scss
JakeLaCombe: We do need to set the font-size, as the default font size is too small.

However, it does need to flex with the base font-size. Converted to em.

https://github.com/cerner/terra-core/pull/305/commits/ea8cbabd8d698303db9e476bc73d56dc79e9a75d

packages/terra-form/docs/README.md
mhemesath: > For some forms, it's been requested that they have no interactions with JavaScript outside of having React render the form.

Who requested this? I think I agree with @Matt-Butler  in that we only support controlled forms.

packages/terra-form/src/components/Control.jsx
mhemesath: `inputAttributes` seems extremely verbose.... FWIW, this follows the pattern of what Django does which uses `attrs` and has worked there. I'd rather have to check the documentation once, learn what it is and then use that going forward over having to type `inputAttributes` every time I set this... 

https://docs.djangoproject.com/en/1.11/ref/forms/widgets/#django.forms.Widget.attrs

packages/terra-form/src/components/ChoiceField.jsx
mhemesath: > The React form libraries I've worked with use children and I think it makes it a lot easier for the consumer

I think this depends on if you are dealing with static choices that are coded or dynamic choices populated from the server.

These options aren't mutually exclusive. Technically what you're describing can already be done use the current design.. ChoiceField is just a higher order component that does that for you. It prevents consumers from having to do the boilerplate code of listing our the radio fields or looping over them and creating them.

mhemesath: I wish these conversations would have happened on the design as it would have been easier to follow... but the <X>Field components are convenience molecules to make it easy to create the proper inputs to capture the proper data. Nothing prevents anyone from generating forms in a style that leaks the implementation details of  HTML

mhemesath: Ultimately, OCS from what I hear is going to specify rules that dictate when you use radios/checkboxes vs select vs search. That will rule out using a static approach of rendering radios such as this because it should flex depending on how many choices there are. 

I propose enable what you're suggesting, but also create choiceField which will be the abstraction to flip impl between select and radios depending on how many choices there are.

StephenEsser: @mhemesath 

I've added this topic to the War Room agenda, will you be able to attend on Friday?

packages/terra-form/docs/README.md
nickpith: If you are by default passing down props to the input on the *Field components, you will support controlled and uncontrolled with little to no extra work. By the fact, that all of these are dealing with inputs. Consumers need to be be aware of how React handles inputs and that means them being aware of controlled vs uncontrolled inputs. 

I seems like the concern here has more to do with the README being confusing.

JakeLaCombe: @mhemesath It was from Longitudinal plan. They just want to generate forms that they send straight to the backend server.

packages/terra-form/src/components/Input.jsx
JakeLaCombe: In my opinion, I think we should keep just onChange as first class, as that's the most critical one for dealing with controlled inputs. There are several others we could add, but that could make this list very large.

packages/terra-form/src/components/Field.jsx
JakeLaCombe: Good idea! Added here. https://github.com/cerner/terra-core/pull/305/commits/14a9635cfe793da32e115c71c71cd2b17458c483

packages/terra-form/docs/README.md
JakeLaCombe: Here is the issue that prompted this request. https://github.com/cerner/terra-core/issues/276

packages/terra-form/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Form
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-form.svg)](https://www.npmjs.org/package/terra-form)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+Components for building forms
```
emilyrohrbough: Should we have a more detailed description of what components are available for building a form?

packages/terra-form/package.json
```diff
@@ -0,0 +1,53 @@
...
+    "terra-base": "^0.x"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "terra-base": "^0.x",
+    "terra-mixins": "^1.0.0"
+  },
+  "scripts": {
+    "compile": "npm run compile:clean && npm run compile:build",
+    "compile:clean": "rm -rf lib",
```
emilyrohrbough: "compile:clean": "$(cd ..; npm bin)/rimraf lib",

packages/terra-form/src/components/ChoiceField.jsx
```diff
@@ -0,0 +1,106 @@
...
+import React, { PropTypes } from 'react';
```
emilyrohrbough: Once Matt's PR for React 15.5 compatibility is merged in, PropTypes will need imported from the prop-types package. Should this be done before his is merged in so there are less changes in the future?

packages/terra-form/src/components/Input.jsx
```diff
@@ -0,0 +1,74 @@
...
+  value: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+  ]),
+};
+
+const defaultProps = {
+  defaultValue: undefined,
+  name: null,
+  onChange: () => {},
```
emilyrohrbough: We are inconsistent between packages on how to handle default onChange/onClick/onKeyDown functions. Button has no default defined, ActionHeader defaults to null, Table & List defaults to undefined, and here it defaults to an empty function. Do we need to use a standard?

packages/terra-form/src/components/MultiChoiceField.jsx
```diff
@@ -0,0 +1,96 @@
+// eslint-disable react/jsx-boolean-value
+
+import React, { PropTypes } from 'react';
+
```
emilyrohrbough: Missing `import 'terra-base/lib/baseStyles';`

emilyrohrbough: Or did you not include it because its using the Field Component? If that is the case, then you added `import 'terra-base/lib/baseStyles'` to the ChoiceFeild component

packages/terra-form/src/components/Textarea.jsx
```diff
@@ -0,0 +1,79 @@
...
+import classNames from 'classnames';
+import 'terra-base/lib/baseStyles';
+
+import './Textarea.scss';
+
+const propTypes = {
+  /**
+   * How many columns the textarea should have
+   */
+  cols: PropTypes.number,
```
emilyrohrbough: Do you think the names cols and rows will confuse the consumer to think they should be inputing arrays instead of number values?

packages/terra-form/docs/components/MultiChoiceField.md
```diff
@@ -0,0 +1,21 @@
...
+<MultiChoicefield
+  label='Select the programming languages you have experience with'
+  name="language_experience"
+  error="This field is required"
+  help="Note: All checked languages should have at least 3 months experience"
+  required={true}
+  choices=[{ name: 'C', value: 'C' }, { name: 'java', value: 'Java' }]
+  isInline={false}
+  required={false}
+/>
```
StephenEsser: Is it possible to pre select choices with this API?

emilyrohrbough:
Will there be any css styling on ChoiceField, MultiChoiceField, NumberField,TextField, or TextAreaField? There was nothing specific to these components in Field.scss so just wondering how that worked or if they weren't special.

packages/terra-site/src/TestLinks.jsx
```diff
@@ -11,6 +11,7 @@ const TestLinks = () => (
       <li><Link to="/tests/button-tests">Button Tests</Link></li>
       <li><Link to="/tests/button-group-tests">Button Group Tests</Link></li>
       <li><Link to="/tests/date-picker-tests">DatePicker Tests</Link></li>
+      <li><Link to="/tests/form-tests">DatePicker Tests</Link></li>
```
emilyrohrbough: Oops, you forgot to change this to Form Tests!

emilyrohrbough:
After looking at the site examples, it appears the consumer needs to declare `<form>` before using the form components. 

Would it be reasonable to create a wrapper component called `Form` that the user can drop these form components into as children?Then each component could be aliased via the `Form` component so the consumer would only need to import `terra-form` and use dot notation.

JakeLaCombe:
Terra form actually will not be providing a base component. Consumers would have to provide the form node theirselves, and then use the components listed here to build that form. This is similar to the approach react-form uses. You can view more details here. https://jira2.cerner.com/browse/TERRAUI-573?focusedCommentId=2634665&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-2634665

JakeLaCombe:
@emilyrohrbough The styling for components such as TextField and NumberField will come from the base field. They will use the styles that a Field uses for it's components.

packages/terra-form/README.md
JakeLaCombe: In my opinion, we should provide links to the packages that this module contains as opposed to describing them here. Otherwise, we could risk this page getting rather large when folders like validations are later included.

packages/terra-form/docs/components/MultiChoiceField.md
JakeLaCombe: I feel like the original tech design called in for passing those in through a values array, but I don't feel like that's the best API. The place that would belong most likely is on this hash

packages/terra-form/src/components/Input.jsx
JakeLaCombe: Let's bring that up on the war room as well today

packages/terra-form/docs/components/MultiChoiceField.md
mhemesath: @StephenEsser thats what the values is for.

packages/terra-form/src/components/Input.jsx
mhemesath: why is onChange defaulted?

packages/terra-form/src/components/ChoiceField.jsx
bjankord: War room discussion output was that we create the `<RadioGroup />` component that uses children API and a `<ChoiceField />` that uses a hash API.

packages/terra-form/src/components/Input.jsx
JakeLaCombe: We have the need to support both controlled and uncontrolled inputs, so we can't always expect the developer to provide an onChange function.

One option is to just put it on the inputAttrs attribute for forms, and delegate it that way to the input element.

packages/terra-form/src/components/ChoiceField.jsx
JakeLaCombe: I don't know if that's better though. I see a need to pass in additional attributes to the individual nodes as well. I think we should just require the same api structure as that of the actual choice input, meaning we just rely on the name and value attributes straight up. Otherwise, this api will be very confusing for users to consume.

packages/terra-form/src/components/Control.jsx
JakeLaCombe: In the war room, we came to a census for inputAttrs

JakeLaCombe:
This pull request is too lengthy. I'll split up the forms into multiple pull requests.

Issue #293, opened 2017-05-02
Add onRequestClose prop to Modal to ensure state is driven purely by consuming components


tbiethman:
### Description of Issue

The primary issue is the calling of `closePortal` here: https://github.com/cerner/terra-core/blob/master/packages/terra-modal/src/ModalContent.jsx#L54

The Modal component itself should be stateless; it has an isOpened prop that is used to open and close the modal.  However, the `closePortal` function that is provided by the Portal (and called in the linked code above) closes the Portal (and therefore Modal) without notifying the component that is rendering the Modal of that closure. This leaves the consuming component's state in a weird spot.

Instead of using `closePortal` for our escape/outside click handling, we should have an `onRequestClose` prop that we call instead that consumers must set.  With this, the Modal consumer is in complete control of the Modal state. This is similar to how the `react-modal` works: https://reactcommunity.org/react-modal/examples/on_request_close.html

```
class ModalPresenter extends React.Component {
  openModal() {
    this.setState({ isOpen: true });
  }

  closeModal() {
    this.setState({ isOpen: false });
  }

  render () {
    return (
      <div>
        <button onClick={this.openModal}>Open Modal</button>
        <Modal
          isOpened={this.state.isOpen}
          onRequestClose={this.closeModal}
        >
          <div>Stuff</div>
          <button onClick={this.closeModal}>Close Modal</button>
        </Modal>
      </div>
    )
  }
}
```

### Expected Outcomes
- The Modal and internal components are completely stateless.

Matt-Butler:
Good catch on this @tbiethman 

Here is a demonstration of the issue:
![modalstate](https://cloud.githubusercontent.com/assets/8146159/25663348/0a24a13a-2fdd-11e7-934b-695c52c872b9.gif)

In this example, the user provides methods to handle the isOpened state.
```
  handleOpenModal() {
    this.setState({ isOpened: true });
  }

  handleCloseModal() {
    this.setState({ isOpened: false });
  }

<button onClick={this.handleOpenModal}>Open Modal</button>
<button onClick={this.handleCloseModal}>Close Modal</button>
```
When we use the open and close buttons, the behavior is as expected.
![expected](https://cloud.githubusercontent.com/assets/8146159/25663472/6953bea2-2fdd-11e7-907c-4b29b7f7ad5b.gif)


However, when using the props `closeOnEsc` and `closeOnOutsideClick`, they do not modify the isOpened state.
[...]

PR #321, opened 2017-05-05
Add onRequestClose prop to Modal to ensure state is driven purely by consuming components

Matt-Butler:
### Summary
For dismissing the modal with clicking on the overlay or pressing the esc key, we were using react-portal. This was not updating our isOpened state. This PR is correcting the value of the isOpened state.

https://github.com/cerner/terra-core/issues/293

After these changes, the isOpened state is accurate.
isopened_state

@cerner/terra

### DIFF
--- a/packages/terra-modal/src/Modal.jsx
+++ b/packages/terra-modal/src/Modal.jsx
@@ -94,19 +94,15 @@ class Modal extends React.Component {
   }
 
   componentDidMount() {
-    if (this.props.closeOnEsc) {
-      document.addEventListener('keydown', this.handleKeydown);
-    }
+    document.addEventListener('keydown', this.handleKeydown);
   }
 
   componentWillUnmount() {
-    if (this.props.closeOnEsc) {
-      document.removeEventListener('keydown', this.handleKeydown);
-    }
+    document.removeEventListener('keydown', this.handleKeydown);
   }
 
   handleKeydown(e) {
-    if (e.keyCode === KEYCODES.ESCAPE && this.props.isOpened) {
+    if (e.keyCode === KEYCODES.ESCAPE && this.props.isOpened && this.props.closeOnEsc) {
       this.props.onRequestClose();
     }
   }
--- a/packages/terra-modal/src/Modal.jsx
+++ b/packages/terra-modal/src/Modal.jsx
@@ -41,7 +41,7 @@ const propTypes = {
   /**
    * If set to true, the modal will rendered as opened
    **/
-  isOpened: PropTypes.bool,
+  isOpened: PropTypes.bool.isRequired,
   /**
    * If set to true, the modal dialog with have overflow-y set to scroll.
    * It is recommended not to use this prop and instead create a HOC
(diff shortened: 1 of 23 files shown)


packages/terra-modal/src/Modal.jsx
```diff
@@ -79,8 +83,28 @@ const defaultProps = {
...
+
 class Modal extends React.Component {
+  constructor() {
+    super();
+    this.handleKeydown = this.handleKeydown.bind(this);
+  }
+
+  componentDidMount() {
+    if (this.props.closeOnEsc) {
+      document.addEventListener('keydown', this.handleKeydown);
```
tbiethman: We probably want to remove this on componentWillUnmount

packages/terra-modal/src/Modal.jsx
```diff
@@ -97,13 +121,14 @@ class Modal extends React.Component {
...
           role,
+          onRequestClose,
            ...customProps } = this.props;
 
     return (
       <Portal
         isOpened={isOpened}
-        closeOnEsc={closeOnEsc}
-        closeOnOutsideClick={closeOnOutsideClick}
+        closeOnEsc={false}
```
tbiethman: Could we just remove these guys completely? Or do they default the values to true?

packages/terra-modal/src/Modal.jsx
Matt-Butler: +1

packages/terra-modal/src/Modal.jsx
Matt-Butler: I'll test that right now. react-portal does not provide default props for either of them.

packages/terra-modal/src/Modal.jsx
```diff
@@ -79,8 +83,28 @@ const defaultProps = {
...
+};
+
 class Modal extends React.Component {
+  constructor() {
+    super();
+    this.handleKeydown = this.handleKeydown.bind(this);
+  }
+
+  componentDidMount() {
+    if (this.props.closeOnEsc) {
```
tbiethman: If, for whatever reason, someone passed the closeOnEsc prop to the Modal component after its initial mounting, we wouldn't add the event handler.

I'm not sure how realistic a scenario that is, but we could always add the event handler and check the prop in the  handler to cover that case.

Matt-Butler: So you recommend always adding the event handler and only trigger onRequestClose when the prop is set?

packages/terra-modal/src/Modal.jsx
Matt-Butler: https://github.com/cerner/terra-core/pull/321/commits/c171e322755b7e4616000fc0b6715820927dd9d0

packages/terra-modal/src/Modal.jsx
tbiethman: I think so. It'd be easier than checking the props on didUpdate every time.

packages/terra-modal/src/Modal.jsx
Matt-Butler: https://github.com/cerner/terra-core/pull/321/commits/de085f27a1f398ca35f224f1e84353e61cb69c4d
Remove the props set to false. We get the same behavior without them set.

packages/terra-modal/src/Modal.jsx
Matt-Butler: https://github.com/cerner/terra-core/pull/321/commits/a0e51cb62cb98405696bb575141f7b7f8bb12a37

Issue #349, opened 2017-05-15
Make terra-toolkit's server-launcher Compatible Across All Repos


emilyrohrbough:
The Webpack 2 Integration pull request had the issue where webpack can't find the correct devDependencies modules directory. The quick solution implemented was to to switch working directory to site and make all modules available.  This fix was in terra-toolkit's `server-launcher.js` file [here](https://github.com/cerner/terra-core/pull/270/files#diff-e1ecc7a7ac21bfa2834574ebe1915fd4R14).

Terra-toolkit is a dependency in terra-clinical, so a different solution needs to be implemented to handle both terra-core and terra-clinical devDependencies modules.

mjhenkes:
This is currently breaking the build in the terra-clinical repo

Matt-Butler:
https://github.com/cerner/terra-core/pull/270/files#diff-9999c132bd53c2a36d76ea10d837dd57L14
This section of the webpack config was resolving root to `terra-site` 

Resolve.root was updated to resolve.modules in webpack 2.
https://webpack.js.org/configuration/resolve/#resolve-modules

I think adding resolve.modules should fix the issue. I'll look into it.

Matt-Butler:
Resolved in https://github.com/cerner/terra-core/issues/356.

bjankord:
@emilyrohrbough @mjhenkes This fix has been released in terra-toolkit@0.3.1

PR #356, opened 2017-05-16
Remove terra-site reference from toolkit

Matt-Butler:
### Summary
https://github.com/cerner/terra-core/issues/349

https://github.com/cerner/terra-core/pull/270/files#diff-9999c132bd53c2a36d76ea10d837dd57L14
Adding back in the resolve-loader.  root has been updated to modules for webpack 2.

@cerner/terra

### DIFF
--- a/packages/terra-site/webpack.config.js
+++ b/packages/terra-site/webpack.config.js
@@ -109,4 +109,7 @@ module.exports = {
       errors: true,
     },
   },
+  resolveLoader: {
+    modules: [path.resolve(path.join(__dirname, 'node_modules'))],
+  },
 };
--- a/packages/terra-toolkit/lib/server-launcher.js
+++ b/packages/terra-toolkit/lib/server-launcher.js
@@ -22,14 +22,6 @@ exports.launchServer = function () {
   return new Promise(function (resolve) {
     var compiler = void 0;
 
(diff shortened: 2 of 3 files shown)


packages/terra-site/webpack.config.js
```diff
@@ -109,4 +109,7 @@ module.exports = {
       errors: true,
     },
   },
+  resolveLoader: {
```
bjankord: How does this differ from
```js
resolve: {
  modules: [path.resolve(path.join(__dirname, 'node_modules'))],
}
```
We use something like that for translations:

https://github.com/cerner/terra-core/blob/4a8c7bff4223a28aa016686a82e1fb7fe34d9b49/packages/terra-site/webpack.config.js#L76-L78

Can you move `resolveLoaders` up so it follows the `resolve` key.

Matt-Butler: I tried that initially but it didn't work. It sounds like the resolve loader is used to locate webpack loaders where resolve is used to locate package dependencies.

bjankord: Ok cool, good to know 👍

windse7en: Useful issue on resolveLoader modules: https://github.com/webpack/webpack/issues/3066#issuecomment-262671479

PR #362, opened 2017-05-17
Update import from react-docgen

Matt-Butler:
### Summary
Small change in the way we import parse from react-docgen. This takes our bundle from 5.59 MB -> 5.32 MB. 

@cerner/terra


Matt-Butler:
I just realized that lib/parse doesn't include another part of the library we need. Closing

Issue #340, opened 2017-05-11
Remove text-mask as a dependency in terra-time-input


benbcai:
Pull out terra-time-input due to an issue in [text-mask] (https://github.com/text-mask/text-mask/issues/510). It can be added back once the issue is resolved.

PR #365, opened 2017-05-18
terra-time-input - the return of the terra-time-input.

benbcai:
There was a prop-type error in react-text-mask (https://github.com/cerner/terra-core/issues/340) causing us to have to pull out terra-time-input. The error is now fixed in react-text-mask 5.0.2. This pull request is to pull in react-text-mask 5.0.2 and merge terra-time-input back in terra-core.

Resolves https://github.com/cerner/terra-core/issues/424

@cerner/terra

Deployment Link: https://terra-core-deployed-pr-365.herokuapp.com/


packages/terra-time-input/package.json
```diff
@@ -0,0 +1,55 @@
...
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "terra-base": "^0.4.1",
+    "terra-mixins": "^1.1.1"
+  },
+  "dependencies": {
+    "moment": "^2.17.1",
+    "prop-types": "^15.5.8",
+    "react-text-mask": "^5.0.2",
```
Matt-Butler: https://github.com/text-mask/text-mask/issues/510#issuecomment-301098488
^ This describes the warning we were seeing before and how it was fixed.

Matt-Butler:
Our other issue with text-mask was prop-types comes pre-bundled through webpack.  https://github.com/text-mask/text-mask/issues/529

Since prop-types comes pre-bundled with text-mask, it will be duplicated in our bundle. However, the duplication should be only a couple KBs.
http://stackoverflow.com/questions/43981041/webpack-dedupe-webpack-bundle

To fix the duplication, we would need to go contribute to text-mask.

Matt-Butler:
This is the same review that we approved earlier. This is probably good to get functionally and UX reviewed.

mjhenkes:
@mmalaker @neilpfeiffer This PR is ready for functional and UX Review. :)

benbcai:
@mmalaker @neilpfeiffer. terra-time-input has been deployed to http://engineering.cerner.com/terra-core/#/site/time-input. You may so functional and UX review.

mmalaker:
Functional verification completed

neilpfeiffer:
Notes from 5/24/17 Terra-core meeting:
- Change the hours & minutes nodes to independent inputs so they are easier to interact with
- Use default placeholder attribute for "hh" and "mm" text-mask
- For 24 time, if user enters 3-9 for hours: auto-rewrite to two digits ("03") on deblur, if user enters 6-9 for minutes: auto-rewrite to ("07") on deblur
- Listen for arrow-left and arrow-right to move between the independent hour and time inputs

packages/terra-time-input/tests/nightwatch/component/TimeInputDefaultTime.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import TimeInput from '../../../lib/TimeInput';
+
+const TimeInputDefault = () => (
+  <div>
```
JakeLaCombe: Is this div needed?

JakeLaCombe: Can you check the other divs in your nightwatch examples as well?

packages/terra-time-input/src/TimePipe.js
```diff
@@ -0,0 +1,33 @@
+function TimePipe(timeFormat) {
+  return function processConformedValue(conformedValue) {
+    if (!timeFormat || timeFormat.length < 0) {
+      return false;
```
JakeLaCombe: We should return null or undefined as opposed to false for this function. If feels like you can get into maintenance issues if you allow for different return types from a function

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -0,0 +1,114 @@
...
+      inputAttributes,
+      onChange,
+      value,
+      ...customProps
+    } = this.props;
+
+    const attributes = { ...customProps, ...inputAttributes };
+
+    return (
+      <div className="terra-TimeInput">
```
JakeLaCombe: For terra-form, I'll be attaching the customProps to the outside container, the div, and input attributes to the input, the textmask in your case. I think we want to do that here as well.

bjankord:
@benbcai  With #414 merged down to master, we'll want to delete the `lib` directory in `packages/terra-time-input`.

packages/terra-time-input/tests/nightwatch/component/TimeInputDefaultTime.jsx
benbcai: Removed: https://github.com/cerner/terra-core/pull/365/commits/464ec5e88746b1b3153189df18e404eef2e671f5

packages/terra-time-input/src/TimePipe.js
benbcai: This file is no longer needed because the time input no longer depends on text-mask.

packages/terra-time-input/src/TimeInput.jsx
benbcai: Good point. Updated https://github.com/cerner/terra-core/pull/365/commits/464ec5e88746b1b3153189df18e404eef2e671f5

packages/terra-form/src/Input.jsx
```diff
@@ -37,34 +37,38 @@ const defaultProps = {
...
-      required={required}
-      {...additionalInputProps}
-      className={classNames('terra-Form-input', additionalInputProps.className)}
-    />
-  );
-};
+    return (
+      <input
+        name={name}
+        ref={(input) => { this.textInput = input; }}
```
benbcai: The only change made in this file is adding a ref on line 64 and changing Input from a functional component to a class component. The reason for this change is because the time-input needs a reference to the hour Input and minute Input so that it can move focus between them. The ref attribute only works on a class component and does not work on a functional because there is no instance of it. See https://facebook.github.io/react/docs/refs-and-the-dom.html#refs-and-functional-components

bjankord:
These updates look good to me @benbcai. Let's run this through another round of functional verification and UX verification since we changed the time input a bit based on @neilpfeiffer's original UX verification. cc/ @mmalaker

mmalaker:
Keyboard navigation issues found during functional verification:
- Chrome and Firefox both work as expected.
- Desktop Safari, IE, and Edge do not allow navigation between the hour and minute time fields with backspace/delete or the left and right arrow keys.
- Mobile Safari on iPad does not allow navigation with delete key. Also, you have to switch to the numeric keyboard manually, both when initially entering values and when switching between the hour field and the minute field.
- Android on the MC40 allows entry of time values but also does not allow navigation between fields with backspace.
- Most surprising, the latest Android version on Galaxy S6 does not allow time entry into the field at all, either through touch or on-screen keyboard actions.

benbcai:
@mmalaker The issue in Safari/IE/Mobile where using backspace/delete does not navigate from the minute input back to the hour input should now be fixed. Mobile device should now show the numeric keyboard. I didn't test on an Android on the MC40 because I don't have one. I tested on a Galaxy S6 with Andriod 7.0 and I was able to enter values in the hour and minute inputs before the code change. Is your S6 also on Andriod 7.0? 

screenshot_20170602-150121 2

mmalaker:
Functional verification completed. All time examples and test pages display and function correctly. Keyboard access issues fixed. Note that there is a minor, cosmetic, character alignment issue present in IE10 and 11. That issue will be fixed in the next forms PR.

neilpfeiffer:
Tested: PR [#365]( https://github.com/cerner/terra-core/pull/365)

**Reviewed:**
- Follows pattern considerations documented in the current Cerner Design Standard: 
[Date/Time Format](https://design.cerner.com/styleguide/patterns/datetime-format/) and  [Time Input](https://design.cerner.com/styleguide/patterns/data-pickers/)
- Follows UX defined Terra initial design implementation for Time Input.


**Pass** |
-----------|
UX Review|

\
\
**Suggestion** (for Future Enhancement)
- Have text entry overwrite current populated values when cursor is in the first or second position (before the “tens” or “ones” position) inside of the input element, so user doesn’t have to forcibly delete or remove any existing values.

mjhenkes:
🎉 🎈

Issue #366, opened 2017-05-18
Status: update color prop to not be required.


Matt-Butler:
Currently, the color prop is required for status. 
https://github.com/cerner/terra-core/blob/master/packages/terra-status/src/Status.jsx#L15

We have consumers that are instead using css classes to color the status component:
```
<Status className=“my-concept-color” color=“” />
```

We should update this prop to no longer be required.

Matt-Butler:
Resolved in #367

PR #367, opened 2017-05-18
Status color prop is no longer required

Matt-Butler:
### Summary
https://github.com/cerner/terra-core/issues/366

### DIFF
--- a/packages/terra-status/lib/Status.js
+++ b/packages/terra-status/lib/Status.js
@@ -36,7 +36,7 @@ var propTypes = {
   /**
    * The color of the status indicator. Accepts a CSS color value.
    */
-  color: _propTypes2.default.string.isRequired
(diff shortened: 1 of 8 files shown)


alex-bezek:
On wellness we use a set of generated classes that provide themed background color, text color, etc for different feature pages. Because of this we would like to use the className instead of color. 
Currently color is required making us do something like
`<Status className="wellness-CustomClass-feature1" color=""/>`

Matt-Butler:
There was a question about what happens when color is undefined here:
https://github.com/cerner/terra-core/blob/master/packages/terra-status/src/Status.jsx#L24

When color is undefined, this is how it will render in the DOM:
https://github.com/cerner/terra-core/pull/367/files#diff-e21d1754b07d3e37c23273eac79d1379R92

PR #380, opened 2017-05-19
Terra search field initial implementation

dkschoonover:
### Summary
Initial implementation of the Terra Search Field.  Supports executing a search callback after a provided delay or a manually triggered search.

Deployment Link: https://terra-core-deployed-pr-380.herokuapp.com/


packages/terra-search-field/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "terra-search-field",
+  "main": "lib/SearchField.js",
+  "private": true,
+  "version": "0.0.0",
+  "description": "{insert description}",
```
emilyrohrbough: Add description

packages/terra-search-field/docs/README.md
```diff
@@ -0,0 +1,17 @@
...
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-search-field`
+
+## Usage
+
+```jsx
+import React from 'react';
+import SearchField from 'terra-search-field';
+
+<SearchField placeholder="Search" />
```
emilyrohrbough: Nit: Add onSearch in example since no default search function is provided by the component.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -0,0 +1,94 @@
...
+  render() {
+    const { placeholder, ...customProps } = this.props;
+    const searchFieldClassNames = classNames(['terra-SearchField',
+      customProps.className,
+    ]);
+
+    return (
+      <div className={searchFieldClassNames}>
+        <input className="terra-SearchField-input" type="search" placeholder={placeholder} value={this.state.searchText} onChange={this.handleTextChange} />
+        <Button className="terra-SearchField-button" onClick={this.handleSearch}>
```
emilyrohrbough: I am curious, why is the search field within a button?

packages/terra-search-field/tests/jest/SearchField.test.jsx
```diff
@@ -0,0 +1,123 @@
...
+  it('renders a search field with a value', () => {
+    const searchField = shallow(<SearchField />);
+    searchField.setState({ searchText: 'Test' });
+
+    expect(searchField).toMatchSnapshot();
+  });
+});
+
+describe('Manual Search', () => {
+  it('triggers search on button press', () => {
```
emilyrohrbough: nit: 'triggers search on click'

packages/terra-search-field/tests/jest/SearchField.test.jsx
```diff
@@ -0,0 +1,123 @@
...
+
+  it('renders a search field with a value', () => {
+    const searchField = shallow(<SearchField />);
+    searchField.setState({ searchText: 'Test' });
+
+    expect(searchField).toMatchSnapshot();
+  });
+});
+
+describe('Manual Search', () => {
```
emilyrohrbough: Does the search ever occur on input change or on 'enter'?

packages/terra-search-field/tests/nightwatch/search-field-spec.js
```diff
@@ -0,0 +1,30 @@
...
+
+module.exports = {
+  before: (browser, done) => {
+    browser.resizeWindow(browser.globals.width, browser.globals.height, done);
+  },
+
+  afterEach: (browser, done) => {
+    screenshot(browser, 'terra-search-field', done);
+  },
+
```
emilyrohrbough: Add examples & nightwatch tests for `minimumSearchTextLength`, `searchDelay` and `onSearch` props.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -0,0 +1,94 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Button from 'terra-button';
+import IconSearch from 'terra-icon/lib/icon/IconSearch';
+import classNames from 'classnames';
```
StephenEsser: Each component should import the base styles:
https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L4

Here's a little more information on why: https://github.com/cerner/terra-core/pull/251

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  render() {
+    const { placeholder, ...customProps } = this.props;
+    const searchFieldClassNames = classNames(['terra-SearchField',
+      customProps.className,
+    ]);
+
+    return (
+      <div className={searchFieldClassNames}>
```
StephenEsser: Custom props should be spread onto the container here, before className. This allows consumers to pass through any extra html properties that aren't supplied first class through the prop API. ( ID, className, data attributes, etc.. )

Examples:
https://github.com/cerner/terra-core/blob/master/packages/terra-progress-bar/src/ProgressBar.jsx#L49
https://github.com/cerner/terra-core/blob/master/packages/terra-status/src/Status.jsx#L24

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * How long the component should wait after input before performing an automatic search.
+   */
+  searchDelay: PropTypes.number,
+
+  /**
+   * A callback to perform search. Sends parameter {String} searchText.
+   */
+  onSearch: PropTypes.func,
+};
```
StephenEsser: I would get a few other opinions on this before implementing anything, but we might need to supply an inputAttributes prop to spread onto the input field. The date picker does something similar: https://github.com/cerner/terra-core/blob/master/packages/terra-date-picker/src/DateInput.jsx#L13

There's a lot of use cases where consumers want/need to pass down a name or id to the input field. (or their own custom events, onBlur, onFocus etc.. )

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -0,0 +1,94 @@
...
+    this.state = {
+      searchText: '',
+    };
+  }
+
+  handleTextChange(event) {
+    this.setState({ searchText: event.target.value });
+
+    if (!this.searchTimeout) {
+      this.searchTimeout = setTimeout(this.handleSearch, this.props.searchDelay);
```
StephenEsser: I ran into a few issues with timeouts and un mounting where a timeout would be initialized but the component unmounted before the callback was fired and it would throw some errors.

It might be worth it to clear this timeout when the component unmounts

dkschoonover: +1, this should definitely clear the timeout on unmount.

packages/terra-search-field/tests/nightwatch/search-field-spec.js
dkschoonover: Will do, I was a little bit confused about what to test with jest vs nightwatch.  I will add these in.

packages/terra-search-field/tests/jest/SearchField.test.jsx
dkschoonover: Search will occur on input after the provided delay. It is does not currently trigger a search on "enter" press.

packages/terra-search-field/tests/nightwatch/search-field-spec.js
dkschoonover: Added more nightwatch tests and expanded examples: 58e41c3

packages/terra-search-field/lib/SearchField.scss
```diff
@@ -0,0 +1,46 @@
...
+    padding: 0.286rem;
+    width: 100%;
+
+    &:focus {
+      border-color: $terra-blue-60;
+      outline: none;
+    }
+  }
+
+  .terra-SearchField-button {
```
dkschoonover: I think we need to have our own button class here because of how the search field and button need to kind of flow together (see screenshot).
screen shot 2017-05-23 at 9 47 36 am

packages/terra-search-field/src/SearchField.jsx
dkschoonover: There was talk about removing the button in the future, but I think for now the plan is to leave it.  We will go ahead and remove the "bar" part of search though and just have it be a field.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -0,0 +1,98 @@
...
+    const { placeholder, ...customProps } = this.props;
+    const searchFieldClassNames = classNames(['terra-SearchField',
+      customProps.className,
+    ]);
+
+    return (
+      <div {...customProps} className={searchFieldClassNames}>
+        <input className="terra-SearchField-input" type="search" placeholder={placeholder} value={this.state.searchText} onChange={this.handleTextChange} />
+        <Button className="terra-SearchField-button" onClick={this.handleSearch}>
+          <IconSearch />
```
dkschoonover: I can definitely see this being valuable, it all depends on how much customization should be allowed for here.  I talked to Neil a little bit about whether the icon was enough context and he said in conjunction with the placeholder text it should be fine.  

Maybe the placeholder text should have a default if not provided to ensure this context is always available.

@neilpfeiffer, do you have any thoughts on this?

packages/terra-search-field/lib/SearchField.scss
```diff
@@ -0,0 +1,46 @@
+@import '~terra-mixins';
+
+.terra-SearchField {
+  display: flex;
```
dkschoonover: This has been updated.  I also removed the 100% width style since this is no longer a bar and just a field.

packages/terra-search-field/src/SearchField.jsx
dkschoonover: I'd like to get some other thoughts on this before making the change.  Is there a concern about exposing too much about the implementation through props?  Would it be better to selectively expose functionality as needed?

I can see the value in making this change,  but I worry about locking in the implementation if props are exposed for each specific child component.

StephenEsser: This might be a case where we only add it in the future if it becomes needed.

Unless anyone knows of an immediate need for either one I would just move forward without it. We can passively add it as an additional prop in the future if it ever becomes necessary.

packages/terra-search-field/lib/SearchField.scss
```diff
@@ -0,0 +1,45 @@
...
+.terra-SearchField {
+  display: inline-flex;
+
+  .terra-SearchField-input {
+    @include terra-border-radius-start(0.286em);
+    @include terra-border-radius-end(0);
+
+    appearance: textfield;
+    border: 0.071em solid $terra-grey-15;
+    border-right: 0;
```
StephenEsser: I would check if there is a terra-mixin to handle bidi on this border.

Matt-Butler: https://github.com/cerner/terra-core/blob/master/packages/terra-mixins/src/Mixins.scss#L121

Matt-Butler: Here is an example of one place it is used:
https://github.com/cerner/terra-core/blob/d7e750f7dbc1b4b409e1a4ea08e5d8064db0ee36/packages/terra-progress-bar/src/ProgressBar.scss#L35


```
@include terra-border-end(0);
```

bjankord:
@dkschoonover  With #414 merged down to master, we'll want to delete the `lib` directory in `packages/terra-search-field`.

dkschoonover:
@bjankord Thanks for the heads up.  Lib files have been removed in 6bced11.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -0,0 +1,100 @@
...
+   */
+  placeholder: PropTypes.string,
+
+  /**
+   * The minimum number of characters to perform a search.
+   */
+  minimumSearchTextLength: PropTypes.number,
+
+  /**
+   * How long the component should wait after input before performing an automatic search.
```
StephenEsser: Might be worth documenting what time units this is expecting

packages/terra-search-field/tests/jest/SearchField.test.jsx
emilyrohrbough: @mmalaker Is there a need for search on 'enter'?

packages/terra-search-field/tests/nightwatch.conf.js
```diff
@@ -0,0 +1,9 @@
+/* eslint import/no-extraneous-dependencies: ["error", {"devDependencies": true}] */
+
+const testSettings = require('terra-toolkit').testSettings;
+const resolve = require('path').resolve;
+const nightwatchConfiguration = require('../node_modules/terra-toolkit/lib/nightwatch.json');
```
emilyrohrbough: Update to:
`require('terra-toolkit/lib/nightwatch.json')`

emilyrohrbough:
It might be good to add some context to your nightwatch examples so they can be quickly varied. For example, the minimum length search example doesn't specify what minimum length was used.

dkschoonover:
@emilyrohrbough Added in c532799

emilyrohrbough:
@dkschoonover Those look good. Context in the examples themselves would also be useful for people viewing the test examples on site so they know the intended behavior. there as well

mmalaker:
Functional verification completed. All search field examples and test pages display and function correctly.

dkschoonover:
@emilyrohrbough How does this look?  Added more description in the links for the test site: bae3ae0

packages/terra-search-field/.npmignore
```diff
@@ -0,0 +1,3 @@
+*.log
+node_modules
+src
```
Matt-Butler: Since this review started, we added a couple other things to this file.
```
*.log
node_modules
src
target
reports
tests
```

Matt-Butler:
I am getting a couple warnings in the js console: 
screen shot 2017-06-02 at 2 32 05 pm

packages/terra-search-field/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
+    "terra-toolkit": "^0.x"
+  },
+  "peerDependencies": {
+    "react": "^15.5.4",
```
Matt-Butler: We have a couple consumers currently using react 15.4.2. Can you update this to 
```
    "react": "^15.4.2",
    "react-dom": "^15.4.2",
```

dkschoonover:
@Matt-Butler Thanks for catching the warnings - that was a late change and I completely missed it.  .npmignore file was also updated in this commit.

1ac694a105583d56177555f54ffedb9aeecee957

packages/terra-search-field/src/SearchField.jsx
bjankord: > This might be a case where we only add it in the future if it becomes needed.
This sounds like a good option to me.

neilpfeiffer:
Tested: PR [#380](https://github.com/cerner/terra-core/pull/380)

**Reviewed:**
- Follows pattern considerations documented in the current Cerner Design Standard: [PowerChart Touch - Search](https://design.cerner.com/styleguide/powerchart-touch/search/)
- Follows UX defined Terra design implementation for Search.



**Pass** |
-----------|
UX Review|

PR #395, opened 2017-05-25
fix: Add type to datepicker button to prevent unexpected form submits

bjankord:
### Summary
Users have reported that the small calendar icon that is present on the datepicker is a button which has a type submit on it which causes my form to submit when they try to open the calendar.

This fix is to explicitly set the button to type `button` to prevent unexpected form submissions.

@cerner/terra


bjankord:
Fixes #396

PR #400, opened 2017-05-25
Base enhancement

windse7en:
### Summary
Base enhancement:
1. Add optional locale and customMessages to support Base with translation.
2. Add error message if there is customMessages but no locale value.
3. Add loadingPlaceHolder prop to handle


packages/terra-base/src/Base.jsx
```diff
@@ -1,20 +1,92 @@
...
+    const childComponent = (
+      <div {...customProps}>
+        {children}
+      </div>
+    );
+
+    const messages = Object.assign({}, this.state.messages, customMessages);
+
+    if (locale === undefined) return childComponent;
+    if (!this.state.areTranslationsLoaded) return loadingPlaceholder;
```
Matt-Butler: If you have time, can you add a test/example for the loadingPlaceholder?

windse7en: As we currently don't have a loading spinner to indicate the loading, I can only test by faked component(like label). After a discussion with Brett, I will remove the `loadingPlaceholder` for  now. As it's a quite easy prop, I will add it back when we have the loading indicator, or some consumers really require this feature.

Issue #403, opened 2017-05-26
Remove terra-mixins/src/terra-mixins.scss


bjankord:
## Progress
- [ ] Design Input
- [x] Expected Behaviour
- [ ] Tech Design
- [ ] Development
- [ ] Functional Validation
- [ ] UX Validation
- [ ] Released

------------------------------------------------------------

# Feature / Enhancement Request
When we originally created the terra-mixins package, we had a file for the mixins located at `packages/terra-mixins/src/terra-mixins.scss`. We started pointing the `package.json` main key to `.scss` files for packages that only contained `.scss` as well as renaming the `.scss` files.
For mixins, we added `terra-mixins/src/Mixins.scss` which is a copy of `/terra-mixins/src/terra-mixins.scss` and pointed the package.json main key to ` "main": "lib/Mixins.scss",` We kept the `/terra-mixins/src/terra-mixins.scss` file for passivity for any packages or projects that were referencing that file. Now all of the terra packages reference terra mixins via the main key in the package.json. We can safely delete the `/terra-mixins/src/terra-mixins.scss` file.

## Expected Behavior
`/terra-mixins/src/terra-mixins.scss` is removed from terra-mixins package.

PR #404, opened 2017-05-26
Update issue template

### LINKED ISSUE #398: Update issue template
Recently we updated the issue template to make it easier to capture information for different types of issues, new features, bugs, enhancements, etc. In the current template we have a section for new features / enhancements and one for bugs. We've found that this is a bit confusing for users as they were filling out both sections when reporting a bug, duplicating Expected Behavior in both the Feature / Enhancement Request section and the Bug Report section.

A proposed change is to group all the sections for all issues to the top, and put sections only needed for bug reports at the bottom. This way we can add a note that if the user is not reporting a bug, they can delete the sections at the bottom.

bjankord:
### Summary
Resolves #398 

@cerner/terra


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,3 +1,4 @@
+<!-- Keep the progress checklist.  -->
 ## Progress
```
bjankord: I'm on the fence about having this progress section, I lean towards removing it. Curious what others think about removing it?

Matt-Butler: I like it, maybe move it to the bottom so it isn't the first thing someone sees.

bjankord: We have a comment that reads, "If you are reporting a bug, please fill out the sections below. Otherwise, the sections below can be deleted." My concern with moving it to the bottom is people would delete it if they see that and are not reporting a bug.

bjankord: What makes me lean towards removing it is that for issues like this [one](https://github.com/cerner/terra-core/issues/403), most of the checkboxes are not applicable or can be identified by other ways. I feel like setting up columns in our project board or using labels we can communicate the progress status.

Matt-Butler: +1 sounds good to me

JakeLaCombe: I'm curious on who would be responsible for updating it. Would it be the contributor of the request, or a terra core contributor? It could be hard for a contributor to constantly have to keep the checklist up to date I envision.

bjankord: My thought was that the terra-core team would maintain the progress checklist. I'm thinking now thought that we maintain this via moving issues around on our project board / adding labels and not burden non-core contributors with this.

emilyrohrbough: I thought @mjhenkes was in favor of requiring all three- The progress checklist for quick reference, the project board we would maintain, and the labels for quick filtering and tracking purposes?

.github/ISSUE_TEMPLATE.md
```diff
@@ -6,42 +7,43 @@
 - [ ] Functional Validation
 - [ ] UX Validation
 - [ ] Released
-<!--- Please fill out the appropriate template below as best you can. Keep the Progress checklist. -->
 
-------------------------------------------------------------
+<!------------------------------------------------------------------->
+<!-- Please fill out the appropriate template below as best you can. -->
```
emilyrohrbough: nit: change template to sections since there is only one template with this change

.github/ISSUE_TEMPLATE.md
bjankord: Discussed with team this morning more about this issue. Current plan is to use the [GitHub project board](https://github.com/cerner/terra-core/projects/1) as status tracker. Ideally we find a way to automate adding labels to issues based on what column the issue card is in. Feel free to add the labels as we see fit, e.g. "Needs UX Review", "Need functional review" but we should be tracking this by having columns for these on our project board.

The board serves as a overall status of the project, while labels on an issue would serve as individual status when you don't want to view the project board. Both are valuable. However, I lean towards project board being more valuable than labels with the idea that we should automate labelling via the project board in the future.

> In response to who would track it, would it make sense that the individual who completed their relative piece update the progress status and move it to the next step?

Thats what I'm thinking, each core contributor manages their issues across the board. The core team should help manage this for outside contributions as well. I don't expect outside contributors to update our project board.

This process is not set in stone, but just a point we are at as we explore using GitHub to manage the project. If we find we want to add the Progress checklist back to the issue template, we can add it back.

Issue #363, opened 2017-05-17
Add localization to demographics banner.


JakeLaCombe:
# Issue Description
The Demographics Banner right now does not have it's messages localized, as the component wasn't ready for it. Now that the solution for translations is coming through, we need to remove the props that take messages for labels, and replace them with translated messages.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
- Props for labels have been removed, and replaced with formatted messages inside the component.

## Current Behavior
Labels for Demographics Banner are only provided in English language.

bjankord:
Resolved in #431

PR #411, opened 2017-05-30
Add default labels to Demographics Banner

JakeLaCombe:
### Summary
The Demographics Banner needs default values for gestational and post menstrual age labels.


packages/terra-demographics-banner/src/DemographicsBanner.jsx
```diff
@@ -79,12 +79,12 @@ const defaultProps = {
...
   gender: '--',
   gestationalAge: null,
-  gestationalAgeLabel: null,
+  gestationalAgeLabel: 'GA',
   identifiers: {},
   personName: '--',
   photo: null,
   postMenstrualAge: null,
-  postMenstrualAgeLabel: null,
+  postMenstrualAgeLabel: 'PMA',
```
StephenEsser: Is there any concern about locale for these defaults?

JakeLaCombe: We do have another issue logged that will create a localized version of this banner. https://github.com/cerner/terra-core/issues/363

As for the current banner, since it's not localized, we decided to default it to the english translation.

Issue #384, opened 2017-05-23
.npmignore ignore target / reports


Matt-Butler:
When installing our components I noticed that a couple packages also installed target, tests, and reports.

We should add these to the .npmginore file.
Before adding tests to the npmignore, we want to make sure we don't need any of the nightwatch test routes to be available.

bjankord:
Resolved in #415

PR #415, opened 2017-05-31
Update npmignore

Matt-Butler:
### Summary
Updating .npmignore. Adding:
- target
- reports
- tests

We shouldn't need any of the examples/routes from the tests directory for terra-ui.com. The tests directory only contains examples intended for functional verification.
 
https://github.com/cerner/terra-core/issues/384

@cerner/terra


packages/terra-site/.npmignore
```diff
@@ -0,0 +1,11 @@
...
+node_modules
+src
+target
+reports
+tests
+build
+aggregated-translations
+stats.json
+webpack.config.js
+webpack.prod.config.js
```
Matt-Butler: Site has a couple additional files that the other packages don't ignore

ryanthemanuel:
Will the test links be broken in the deployed site if we don't include those directories?

Matt-Butler:
> Will the test links be broken in the deployed site if we don't include those directories?

For GitHub pages, we generate the bundle for the deployed site locally. It works on my local machine.

For the review apps, it clones the repo, installs dependencies, and compiles src to lib before generating the bundle. It is working on this PR.

bjankord:
@ryanthemanuel, I can confirm what @Matt-Butler said. I checked out this branch and `npm run clean:install`. Then I ran `npm run deploy` to deploy to gh-pages from this branch. It looks like the deployment successfully kept the test links: http://engineering.cerner.com/terra-core/#/tests/icon-tests/bidi

It also looks like test links are available in the Heroku reviewable apps deployment:
https://terra-core-deployed-pr-415.herokuapp.com/static/#/tests/icon-tests/height-width

Issue #428, opened 2017-06-02
Update Demographics examples to use the Image component for the patient photo


ryanthemanuel:
# Issue Description
It would be a more helpful example for our consumers to use the Image component for the patient photo.  That way they can get the bigger picture of the entire ecosystem

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [X] Other

Matt-Butler:
https://github.com/cerner/terra-core/pull/431

windse7en:
Close this issue, as it's merged into master by #431 PR.

PR #431, opened 2017-06-02
Demographic banner localization

windse7en:
### Summary
1. Add locale switch to the App in `terra-site`
2. Add translations into DemographicsBanner
3. Update DemographicsBanner labels to translations
4. Extract small components in DemographicsBanner into eacth file
5. Add intl validation into DemographicsBanner
6. icon lib js update (by compiler)
7. Remove unused eslint rule

### Deployment Link
https://terra-core-deployed-pr-431.herokuapp.com/


packages/terra-demographics-banner/src/_utils.jsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
```
JakeLaCombe: When I think of utilities, I think of methods that manipulate arrays and strings as opposed to creating UI widgets. I almost feel like this should be called something like shared objects maybe?

packages/terra-demographics-banner/tests/jest/DemographicsBannerDisplay.test.jsx
```diff
@@ -0,0 +1,103 @@
...
+it('renders the banner that contains additional information', () => {
+  const banner = shallow(
+    <DemographicsBannerDisplay
+      applicationContent={<span className="risk-score">5%</span>}
+      age="25 Years"
+      applicationRows={[{ Address: '1501 Walnut St. Apt #123', City: 'Kansas City MO' }, { ZIP: 64108 }]}
+      dateOfBirth="May 9, 1993"
+      gender="Male"
+      identifiers={{ MRN: 12343, REA: '3JSDA' }}
+      photo={<img alt="My Cat" src="" />}
```
JakeLaCombe: It's been requested to use the terra image component instead of a standard img tag. Could we perhaps add that change in the examples, as it's not too large?

packages/terra-demographics-banner/src/_utils.jsx
windse7en: Usually, for the shared component, we can create a private component file to contain it. 
But dues to the array of React elements, `personDetails applicationIdentifiers` have to be the method. To avoid that, I may need to add a div to wrap it. But I'm not sure we want to that.

The reason to use `utils` is in general, these two methods are reused by other code.

packages/terra-demographics-banner/tests/jest/DemographicsBannerDisplay.test.jsx
windse7en: Sure. I will update it.

windse7en: Update in the new commit. Instead of adding `terra-image` into the devDependency, use relative path to import the image component code.

packages/terra-demographics-banner/src/DemographicsBannerDisplay.jsx
```diff
@@ -0,0 +1,102 @@
...
+   * The persons preferred first name if they have one.
+   */
+  preferredFirstName: PropTypes.string,
+};
+
+const defaultProps = {
+  applicationContent: null,
+  age: '--',
+  dateOfBirth: '--',
+  dateOfBirthLabel: 'DOB',
```
emilyrohrbough: Did we want to localize the labels?

packages/terra-demographics-banner/src/DemographicsBanner.jsx
```diff
@@ -61,221 +45,64 @@ const propTypes = {
...
-
-    if (this.props.deceasedDate) {
-      elements.push(
-        <DemographicsBannerValue
-          key="deceased"
-          label={this.props.deceasedDateLabel}
-          value={this.props.deceasedDate}
-        />,
-      );
+const contextTypes = {
```
emilyrohrbough: I am curious, If we require a local, do we need to provide default label values in english? Does providing them give a fallback if there are no translations for the labels?

windse7en: The designed behavior is: if the DemoBanner isn't inside the Base with locale(which provides context.intl), it needs to throw the friendly error message to tell consumers to do that. The reason behind that is, as we support the consumer, who doesn't want to have translations, to use Base without locale, we need to throw such error message if they try to use translated component(like DemoBanner) inside untranslated Base.

emilyrohrbough: Thanks for the explanation! Very thorough!

windse7en: No problem.

packages/terra-demographics-banner/src/DemographicsBannerDisplay.jsx
```diff
@@ -0,0 +1,102 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import ResponsiveElement from 'terra-responsive-element';
+import 'terra-base/lib/baseStyles';
+
+import './DemographicsBanner.scss';
+import SmallDemographicsBannerDisplay from './_SmallDemographicsBannerDisplay';
```
ryanthemanuel: What's the reason for these being underscored?

packages/terra-demographics-banner/src/DemographicsBanner.jsx
```diff
@@ -61,221 +45,64 @@ const propTypes = {
...
-}
+const DemographicsBanner = (
+  {
+    age,
+    dateOfBirth,
+    gender,
+    personName,
+    ...customProps
+  }, {
+    intl,
```
ryanthemanuel: What's the difference between adding intl here vs. explicitly calling injectIntl with the exported component?

packages/terra-demographics-banner/src/DemographicsBannerDisplay.jsx
windse7en: I want to use this convention to show this is the private component. Private components are just used inside the module and never should be exported.

packages/terra-demographics-banner/src/DemographicsBanner.jsx
windse7en: `injectIntl` add a wrapper around `DemographicsBanner` to extract intl from context and pass by props.

By using injectIntl, react-intl will throw the `intl` can't be found error message at first. We can't thow our friendly message. And that error isn't catchable.

ryanthemanuel: Makes sense.

packages/terra-demographics-banner/src/DemographicsBannerDisplay.jsx
ryanthemanuel: Has this been decided in terms of conventions?  I'm not opposed to it, but if this is the direction we're going I'd like to document it.  I think there are other components being worked on that could benefit from this strategy as well.  I'll add this to the war room.

packages/terra-demographics-banner/package.json
```diff
@@ -21,7 +21,8 @@
   },
   "homepage": "https://github.com/cerner/terra-core#readme",
   "devDependencies": {
-    "terra-toolkit": "^0.4.0"
+    "terra-toolkit": "^0.4.0",
+    "react-intl": "^2.3.0"
```
bjankord: Curious why react-intl needs to be added as a devDependency, is this for testing?

windse7en: Yeap. It's for the jest testing. To test, it should throw expected error message, if there is no intl in context. [code link](https://github.com/cerner/terra-core/pull/431/files#diff-b906926b8eeeef355e489c2e716c19b0R3)

packages/terra-demographics-banner/tests/jest/_LargeDemographicsBannerDisplay.test.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+import Image from '../../../terra-image/lib/Image';
+import LargeDemographicsBannerDisplay from '../../src/_LargeDemographicsBannerDisplay';
+
+it('renders small banner that contains all valid information', () => {
```
bjankord: Should this be renders `large` banner.

packages/terra-demographics-banner/src/_utils.jsx
bjankord: Lets discuss in our meeting later today, I lean towards `_renderUtils.jsx` but understand the logic for calling it `_utils.jsx`

JakeLaCombe: Discussed today. _sharedObjects should be a good file name.

packages/terra-demographics-banner/tests/jest/DemographicsBannerDisplay.test.jsx
JakeLaCombe: Thanks!

packages/terra-demographics-banner/src/_utils.jsx
windse7en: Update the _utils file name to _sharedObjects in [this commit](https://github.com/cerner/terra-core/pull/431/commits/f78d603898418000f4d27811e99d80a4951ed18c)

mmalaker:
Functional verification completed. Demographics banner labels now flex by locale. Discussed adding localization to date of birth format. If we decide to include it, there is a separate issue logged: [#446](https://github.com/cerner/terra-core/issues/446)

windse7en:
Fixed #428.

Issue #432, opened 2017-06-02
Selectable Lists need the ability to selectively enabled/disabled chevrons and selectability.


dkasper-was-taken:
# Issue Description
Previously SingleSelectList and MultiSelectList defaulted their children to isSelectable true, and would set the value of it's children's hasChevron according to it's own hasChevrons prop. This provided that the child items hadn't explicitly set their own isSelectable and hasChevron prop, in which case it would use that value.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Given the intention was to prevent an unset bool, the best course might be for ListItem to not default boolean values, but have the HOC of multi and single select be defaulted.  As the 4 checks in the ListItem are looking for a positive value, this shouldn't hurt the ListItem's functionality within the basic List implementation.

emilyrohrbough:
It was decided an exception would be made in regards to defining default boolean props of false. The exception is if a subcomponent is designed to be used by multiple higher order components with different behaviors and these HOC expect varying default properties of the subcomponent, the defaultProps value could be defaulted to undefined. Thus, the HOC is responsible for the default properties. This allows for a cleaner API.

This addresses the issue #397 

Documentation updates have been made in this [pull request](https://github.cerner.com/terra/conventions/pull/23).

emilyrohrbough:
Resolved in #439.

bjankord:
This has been released in terra-list@v0.5.0

Issue #421, opened 2017-05-31
Triggering select actions with spacebar


mmalaker:
# Issue Description
Currently, selectable lists and tables can be triggered through keyboard navigation using the tab key to set focus and the enter key to select. For accessibility, we may want to include spacebar as a selection option.

## Issue Type
Spike and minor enhancement
- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
- Research usage of spacebar as a keyboard selection option
- Utilize spacebar select consistently where appropriate

## Current Behavior
Spacebar does not trigger select on selectable lists or tables.

Matt-Butler:
This should be a pretty quick fix:
List
https://github.com/cerner/terra-core/blob/master/packages/terra-list/src/SingleSelectList.jsx#L6
https://github.com/cerner/terra-core/blob/master/packages/terra-list/src/MultiSelectList.jsx#L6

Table
https://github.com/cerner/terra-core/blob/master/packages/terra-table/src/SingleSelectableRows.jsx#L7

emilyrohrbough:
PR #436 added spacebar selection to tables.

emilyrohrbough:
PR #439 added spacebar selection to lists.

bjankord:
This has been released in terra-list@v0.5.0 and terra-table@v0.6.0

Issue #397, opened 2017-05-25
Easier MultiSelectList API


JakeLaCombe:
# Issue Description
Our current multiselect list api looks as followed

    <MultiSelectList maxSelectionCount={2}>
        <MultiSelectList.Item content={<p>test</p>} key="123" isSelectable />
        <MultiSelectList.Item content={<p>test</p>} key="124" isSelectable />
        <MultiSelectList.Item content={<p>test</p>} key="125" isSelectable />
    </MultiSelectList>

If a user is trying to make all of the items in the list selectable, I find it rather tedious to have to put isSelectable on every single Item.

I propose we offer an api where you can pass a single boolean attribute isWholeListSelectable, or maybe even isSelectable to the parent component that will make all the items selectable by default.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior

Pass this in to make the entire list selectable

    <MultiSelectList maxSelectionCount={2} isWholeListSelectable>
        <MultiSelectList.Item content={<p>test</p>} key="123" />
        <MultiSelectList.Item content={<p>test</p>} key="124" />
        <MultiSelectList.Item content={<p>test</p>} key="125" />
    </MultiSelectList>

## Current Behavior
If a user is trying to make all of the items in the list selectable, I find it rather tedious to have to put isSelectable on every single Item

bjankord:
My thoughts are a component named `<MultiSelectList>` shouldn't need to set a prop to set all ListItems within it to be selectable. isSelectable should be the default behavior of ListItems within a MultiSelectList. Users would have the option to set `isSelectable={false}` to opt out of the implicit selectability of ListItems within MultiSelectList.

Implementation wise, I see this being clone the children, add `isSelectable` to all `MultiSelectList.Item`s unless`isSelectable={false}` is already set on a `MultiSelectList.Item`

dkasper-was-taken:
This issue is related to #432

emilyrohrbough:
It was decided an exception would be made in regards to defining default boolean props of false. The exception is if a subcomponent is designed to be used by multiple higher order components with different behaviors and these HOC expect varying default properties of the subcomponent, the defaultProps value could be defaulted to undefined. Thus, the HOC is responsible for the default properties. This allows for a cleaner API.

These changes reflect the suggestions made in #439.

Documentation updates have been made in this [pull request](https://github.cerner.com/terra/conventions/pull/23).

[...]

PR #439, opened 2017-06-05
Update selectable list

emilyrohrbough:
### Summary
In this PR:

1. Changed ListItem's `isSelectable` and `hasChevrons` defaults to undefined. (Issues #397 and #432)
2. Added spacebar selection (Issue #421)
3. Cleaned up nightwatch tests and clarified tests for trigging onChange.

@cerner/terra

### Deployment Link
https://terra-core-deployed-pr-439.herokuapp.com/static/#/site/list


packages/terra-list/src/SingleSelectList.jsx
```diff
@@ -129,7 +136,11 @@ class SingleSelectList extends React.Component {
       newProps.onKeyDown = onKeyDown;
     }
 
+    // Uses the props.hasChevron value, unless the consumer specifies the item hasChevron attribute as false.
     newProps.hasChevron = this.props.hasChevrons;
+    if (item.props.hasChevron === false) {
```
dkasper-was-taken: For chevron we might need to check for undefined, since the value for hasChevrons can be true and false, rather than the isSelectable which is always true.

emilyrohrbough: The default prop for `hasChevrons` is false, so a Single Select List will always give a list item a `hasChevron` value of false. When a Single Select List is passed hasChevrons, it will give every list item a `hasChevron` value of true, unless a specific item was given `hasChevron={false}`. I am not sure undefined needs checked. Can you think of an instance where it would?

dkasper-was-taken: I'm referring to the scenario where the value for hasChevrons is false, as the default, and the value for hasChevron is true. Since the item check is only for === false, then true would never be registered.

emilyrohrbough: Ah I see what you are saying. Very good point, this does need to check for true or false. I was not thinking about it in the opposite way. Added an example & test.

mmalaker:
Functional verification completed. List examples and test pages display and function correctly.

Issue #374, opened 2017-05-19
Popup Component


dkasper-was-taken:
# Issue Description
Create a popup component capable of displaying with and without an arrow indicator.

## Issue Type

- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The Terra Popup should allow the dynamic presentation of children elements aligned with a presenting target. The option for an arrow that maintains a correct attachment, regardless of flipping and pinning.

Open questions: 
- What should the prescribed set of sizes be?
- What visual design should be placed on the popup content and arrow?

## Responsiveness:
The Terra popup should be either responsive to a boundingRef container or default to the window for it's bounds and edge pinning. When both vertical and horizontal bounds are breached, a close button should be injected.  Popup dimensions are set heights and widths, from a given subset of prescribed dimensions. The popup will attempt to flip it's positioning should it's initial positioning not allow it's presentation between the bounds and the target element.

## React Props - Popup:
| Prop | Type | Default | Description |
|-------|-------|---------|---------------|
| boundingRef | Function | undefined | |
| children  | node  | undefined     | Required content to display as popup content.|
| classNameArrow | String | null | Class name to append to the arrow element. |
| classNameContent | String | null | Class name to append to the content element inner. |
| classNameOverlay | String | null | Class name to append to the overlay element. |
| contentAttachment | String | 'top center' | String pair of top, middle, bottom, and left, center, right. |
| contentDimensions | String | '25x 25x' | Modifier values based on a 16 x 9 base size. |
| isArrowDisplayed |Bool | false | Should an arrow be presented at the calculated attachment. |
| isHeaderDisabled |Bool | false | Should the default header be disabled for bounds breach.|
| isOpen | Bool | false | Should the popup be displayed.|
| onRequestClose | Function | undefined | Callback function when popup close conditions are met. |
| targetRef | Function | undefined | Required element ref callback to be presented. |

## React Props - _PopupContent:
| Prop | Type | Default | Description |
|-------|-------|---------|---------------|
| arrow | Element | undefined | The arrow to be presented with the frame. |
| arrowPosition | String | 'top' | Margin direction to allow for arrow, 'top', 'bottom', 'left', and 'right'. |
| children | node  | [] | Content to display within the popup content. |
| classNameInner | String | null | Class name to append to the inner content div. |
| closeOnEsc | Bool | false | Should the popup trigger a close event on esc keydown. |
[...]

PR #444, opened 2017-06-08
Terra Popup

dkasper-was-taken:
### Summary
Terra Popup, new feature, a tether enabled implementation.

Addressed Issue: #374 

### Deployment Link
https://terra-core-deployed-pr-444.herokuapp.com/static/#/site/popup


Matt-Butler:
https://cl.ly/1O0q373H1Q1T
The middle-center example throws a warning

packages/terra-popup/src/_PopupOverlay.jsx
```diff
@@ -0,0 +1,53 @@
...
+  children: PropTypes.node,
+};
+
+const defaultProps = {
+  children: [],
+};
+
+class PopupOverlay extends React.Component {
+
+  componentWillMount() {
```
Matt-Butler: I have read a few different places that using componentDidMount is preferable over componentWillMount. What issue did you see when using componentDidMount?

There has also been some discussion about deprecating componentWillMount 
https://github.com/facebook/react/issues/7671

Matt-Butler:
popup-tab
When opening a popup, the tabIndex of the content ends up being behind the rest of the focusable elements on the page. We might want to log an enhancement to make the next tab focus the popup content.

packages/terra-popup/src/_PopupOverlay.jsx
dkasper-was-taken: On windows firefox, the their scrollbar changes the width of the page, it's not an overlay. So when setting document to overflow: hidden, the width changes on the page by the width of the scrollbar. So my component would layout to a specific point, and then the scrollbar would dissapear, leaving the popup at an incorrect position.

packages/terra-popup/README.md
```diff
@@ -0,0 +1,32 @@
+# Terra Popup
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-popup.svg)](https://www.npmjs.org/package/terra-popup)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+The Terra Popup is high order component, that launches tether positioned content, with the ability to display a dynamic arrow.
```
Matt-Butler: nit: higher-order component

packages/terra-popup/tests/nightwatch/components/BoundedHeightPopup.jsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import Popup from '../../../lib/Popup';
+
+class BoundedPopup extends React.Component {
```
Matt-Butler: popup-bounded
The bounded popup will move when tabbing to an element outside the viewable portion of the page.

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -0,0 +1,281 @@
...
+
+class PopupContent extends React.Component {
+  static getContentStyle(height, maxHeight, width, maxWidth) {
+    const validHeight = maxHeight <= 0 || height <= maxHeight ? height : maxHeight;
+    const validWidth = maxWidth <= 0 || width <= maxWidth ? width : maxWidth;
+    return { height: `${validHeight.toString()}px`, width: `${validWidth.toString()}px` };
+  }
+
+  static addPopupHeader(children, onRequestClose) {
+    const icon = <IconClose className="terra-PopupContent-closeButton" onClick={onRequestClose} height="30" width="30" style={{ float: 'right' }} />;
```
Matt-Butler: We should add tabIndex="0" to the close button so it is focusable.

Matt-Butler: or use a button component

packages/terra-popup/tests/nightwatch/components/BoundedHeightPopup.jsx
dkasper-was-taken: I might have to add a focus trap to the element.

packages/terra-popup/src/_PopupContent.jsx
dkasper-was-taken: I can do that, although I'm waiting on ux visuals, so this structure might change.

dkasper-was-taken:
Fixed the warning.

dkasper-was-taken:
Addressed tabbing and focus issues, by introducing a focus trap and tabIndex to the close button.

packages/terra-popup/src/_PopupOverlay.jsx
dkasper-was-taken: This scenario is even worse on Edge, as it scrolls vertically as well.

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -0,0 +1,284 @@
...
+
+class PopupContent extends React.Component {
+  static getContentStyle(height, maxHeight, width, maxWidth) {
+    const validHeight = maxHeight <= 0 || height <= maxHeight ? height : maxHeight;
+    const validWidth = maxWidth <= 0 || width <= maxWidth ? width : maxWidth;
+    return { height: `${validHeight.toString()}px`, width: `${validWidth.toString()}px` };
+  }
+
+  static addPopupHeader(children, onRequestClose) {
+    const icon = <IconClose tabIndex="0" className="terra-PopupContent-closeButton" onClick={onRequestClose} height="30" width="30" style={{ float: 'right' }} />;
```
Matt-Butler: What do you think about adding an event to close the popup with the enter / space key when the close icon is focused?

dkasper-was-taken: Good idea, I'll try to get that setup.

packages/terra-popup/src/_PopupContent.jsx
bjankord: Yeah, let's [mark up the close button similar to how we did in modal dialog](https://github.com/cerner/terra-core/blob/master/packages/terra-modal/tests/nightwatch/components/ModalDialog.jsx#L41-L43) test page:
http://engineering.cerner.com/terra-core/#/tests/modal-tests/dialog

packages/terra-popup/src/_PopupContent.jsx
bjankord: If we switch the markup to use a button element like the [close button here](https://github.com/cerner/terra-core/blob/master/packages/terra-modal/tests/nightwatch/components/ModalDialog.jsx#L41-L43), the enter/space key actions to interact with the button should come natively without needing to code up an event handler.

dkasper-was-taken: That's true, I've been holding off since we don't have any ux input on it yet, but was going to try with the icon, just in case.

bjankord: If you use a button with the variant set to `link`, I believe it should look the same as you have it now. Might need some CSS to adjust the color.

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -0,0 +1,286 @@
...
+
+class PopupContent extends React.Component {
+  static getContentStyle(height, maxHeight, width, maxWidth) {
+    const validHeight = maxHeight <= 0 || height <= maxHeight ? height : maxHeight;
+    const validWidth = maxWidth <= 0 || width <= maxWidth ? width : maxWidth;
+    return { height: `${validHeight.toString()}px`, width: `${validWidth.toString()}px` };
+  }
+
+  static addPopupHeader(children, onRequestClose) {
+    const icon = <IconClose tabIndex="0" height="30" width="30" />;
```
Matt-Butler: closebutton

hover/focus: #004c76
default: #1c1f21


When I spoke with Neil earlier about the modal dismiss button he instructed me to match the example above. @neilpfeiffer can you confirm this?

Here is the code I have started for the modal dialog dismiss:
https://github.com/cerner/terra-core/blob/modal-dialog/packages/terra-modal-dialog/src/ModalDialogHeaderDismiss.jsx

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -0,0 +1,286 @@
...
+      customProps.className,
+    ]);
+
+    const innerClassNames = classNames([
+      'terra-PopupContent-inner',
+      { 'terra-PopupContent-inner--isFullScreen': isFullScreen },
+      classNameInner,
+    ]);
+
+    // Delete the disableOnClickOutside and disableOnClickOutside prop that comes from react-onclickoutside.
```
emilyrohrbough: listed `disabledOnClickOutside and disabledOnClickOutside` instead of `disabledOnClickOutside and enabledOnClickOutside`

packages/terra-popup/src/Popup.jsx
```diff
@@ -0,0 +1,370 @@
...
+  /**
+   * CSS classnames that are append to the overlay.
+   */
+  classNameOverlay: PropTypes.string,
+  /**
+   * Attachment point for the popup, this will be mirrored to the target.
+   */
+  contentAttachment: PropTypes.oneOf(TetherComponent.attachmentPositions),
+  /**
+   * A string pair of breakpoint sizes, ('10x', '25x', '50x', '75x', '10x'), i.e. 'vertical horizontal'
```
emilyrohrbough: This comment seems a bit confusing, we might want to expand on the description to make it clearer for a new consumer.

packages/terra-popup/src/Popup.jsx
```diff
@@ -0,0 +1,370 @@
...
+};
+
+class Popup extends React.Component {
+
+  static mirrorAttachment(attachment) {
+    const parsedValue = Popup.parseStringPair(attachment);
+    let horizontal = parsedValue.horizontal;
+    let vertical = parsedValue.vertical;
+
+    if (parsedValue.vertical === 'middle') {
```
emilyrohrbough: What is the reasoning for only handling the `middle` case?

dkasper-was-taken: Middle's attachment point is mirrored along the x-axis, top and bottom are mirrored along the y-axis, so a a different value from the attachment string pair is used.

packages/terra-popup/src/Popup.jsx
```diff
@@ -0,0 +1,370 @@
+import React from 'react';
```
emilyrohrbough: I am wondering if it would be good to move the positioning/offset logic into its own `popupUtils` file.

packages/terra-popup/src/Popup.jsx
```diff
@@ -0,0 +1,370 @@
...
+    if (parsedValue.vertical === 'middle') {
+      horizontal = MIRROR_LR[parsedValue.horizontal];
+    } else {
+      vertical = MIRROR_TB[parsedValue.vertical];
+    }
+
+    return `${vertical} ${horizontal}`;
+  }
+
+  static getContentOffset(attachment, targetNode, arrowOffset) {
```
emilyrohrbough: I think documentation of these methods that are doing the positions and offset logic is needed.

packages/terra-popup/src/Popup.jsx
dkasper-was-taken: Potentially could be, I had considered it, but I don't really feel strong either way about it. Does anyone have a strong preference?

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -0,0 +1,286 @@
...
+class PopupContent extends React.Component {
+  static getContentStyle(height, maxHeight, width, maxWidth) {
+    const validHeight = maxHeight <= 0 || height <= maxHeight ? height : maxHeight;
+    const validWidth = maxWidth <= 0 || width <= maxWidth ? width : maxWidth;
+    return { height: `${validHeight.toString()}px`, width: `${validWidth.toString()}px` };
+  }
+
+  static addPopupHeader(children, onRequestClose) {
+    const icon = <IconClose tabIndex="0" height="30" width="30" />;
+    const button = <button className="terra-PopupContent-close" onClick={onRequestClose} style={{ float: 'right' }}>{icon}</button>;
```
emilyrohrbough: Should this styling be pulled into the scss file instead of hardcoded?

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -0,0 +1,286 @@
...
+      contentHeightMax,
+      contentWidth,
+      contentWidthMax,
+      isHeaderDisabled,
+      onRequestClose,
+      refCallback,
+      ...customProps
+    } = this.props;
+
+    const showArrow = PopupContent.shouldShowArrow(arrow, arrowPosition, contentHeight, contentHeightMax, contentWidth, contentWidthMax);
```
emilyrohrbough: Did you make `shouldShowArrow()` static to eliminate the need to bind it in the constructor? Just curious, haven't really seen this.

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -0,0 +1,286 @@
...
+    const contentStyle = PopupContent.getContentStyle(contentHeight, contentHeightMax, contentWidth, contentWidthMax);
+    const isFullScreen = PopupContent.isFullScreen(contentHeight, contentHeightMax, contentWidth, contentWidthMax);
+
+    let content = children;
+    if (isFullScreen && !isHeaderDisabled) {
+      content = PopupContent.addPopupHeader(children, onRequestClose);
```
emilyrohrbough: Should there be a check to ensure onRequestClose is defined before creating a header?

packages/terra-popup/src/_TetherComponent.jsx
```diff
@@ -0,0 +1,251 @@
...
+  /**
+   * Should tethering be disabled following the initial positioning.
+   */
+  disableOnPosition: PropTypes.bool,
+  /**
+   * Should element be tethered to the page.
+   */
+  isEnabled: PropTypes.bool,
+  /**
+   * The ability to disable "gpu" and "moveElement" optimizations.
```
emilyrohrbough: Does this need to be of type `PropType.shape` to help guide the user of the 'gpu' and 'moveElement' structures? Also, does this prop give the ability to enable certain optimizations as well? If no, possibly renaming prop to `disableOptimizations` might be clearer.

packages/terra-popup/src/_PopupContent.jsx
dkasper-was-taken: This is kind of placeholder until we get the ux design, so this will likely change.

packages/terra-popup/tests/jest/Popup.test.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import PopupDefaultExample from './PopupDefaultExample';
+
+// Snapshot Tests
+it('should shallow a default component', () => {
+  const popup = shallow(<PopupDefaultExample />);
+  expect(popup).toMatchSnapshot();
+});
+
+it('should render a default component', () => {
```
emilyrohrbough: Same test is here twice :)

packages/terra-popup/tests/jest/Popup.test.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import PopupDefaultExample from './PopupDefaultExample';
+
```
emilyrohrbough: Looks like you have the same test twice, but the snapshot is different.?

packages/terra-popup/src/_TetherComponent.jsx
dkasper-was-taken: Tether doesn't document all the optimization values that possible, as they left it open ended, I wasn't sure if I wanted to lock the value for this down. I could if we are comfortable with it, as well as I could remove it, if we don't think anyone will use it.

packages/terra-popup/tests/jest/Popup.test.jsx
dkasper-was-taken: it's the different between shallow and render modifiers in the test.

packages/terra-popup/src/Popup.jsx
```diff
@@ -0,0 +1,370 @@
...
+      onRequestClose,
+      targetRef,
+      ...customProps
+    } = this.props; // eslint-disable-line no-unused-vars
+
+    let portalContent = children;
+    if (isOpen) {
+      this.attachment = Popup.parseStringPair(contentAttachment);
+
+      const boundingFrame = boundingRef ? boundingRef() : undefined;
```
emilyrohrbough: I am curious. In you're example, the `boundingRef` gets the parentNode. Are there instances where that the boundingRef will not be the parentNode?

packages/terra-popup/src/_PopupContent.jsx
dkasper-was-taken: It's more a style choice, if a method is purely functional, with no transformative behavior, I tend to prefer static methods.

packages/terra-popup/src/Popup.jsx
dkasper-was-taken: They are not directly associated with parent, it would be odd for a bounding ref not associated to the presenting target element, but it's possible.  The more likely scenario is, parent of parent of parent, etc. Where the presenting location could be several divs deep, and we didn't want to take just the first parent.

packages/terra-popup/src/Popup.jsx
```diff
@@ -0,0 +1,370 @@
...
+  const combinations = [];
+  for (let i = 0; i < DIMENSIONS.length; i += 1) {
+    for (let j = 0; j < DIMENSIONS.length; j += 1) {
+      combinations.push(`${DIMENSIONS[i]} ${DIMENSIONS[j]}`);
+    }
+  }
+  return combinations;
+};
+
+const DIMENSION_COMBINATIONS = COMBINE();
```
emilyrohrbough: @JakeLaCombe  ^That's cool.

packages/terra-popup/src/PopupContent.scss
```diff
@@ -0,0 +1,54 @@
...
+  border: 0;
+  color: #999;
+  cursor: pointer;
+  font-size: 1.45rem;
+  font-weight: 700;
+  line-height: 1;
+  padding: 0;
+
+  &:hover {
+    color: #111;
```
dkasper-was-taken: It was just matching the current pattern for the modal dialog, I am still waiting on final ux.

packages/terra-popup/src/Popup.jsx
Matt-Butler: +1 This is a pretty big file. Separating the logic out would make it more readable

packages/terra-popup/src/Popup.jsx
```diff
@@ -0,0 +1,368 @@
...
+
+const BASE_WIDTH = 16;
+const BASE_HEIGHT = 9;
+
+const DIMENSIONS = [
+  '10x',
+  '25x',
+  '50x',
+  '75x',
+  '100x',
```
Matt-Butler: x is not a real unit, what do you think about making this match the actual unit?

dkasper-was-taken: Still waiting on final values from ux, but I coded it that way because they are coefficient multipliers of the base 16/9 object. So they refer to 25x16/10x9, etc. I don't really know what these will end up being. But I know a grid style sizes are in the works. I'm all ears for alternatives.

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -0,0 +1,286 @@
...
+  /**
+   * The initial arrow position.
+   */
+  arrowPosition: PropTypes.oneOf(ARROW_POSITIONS),
+  /**
+   * The children to be presented as the popup's content.
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * CSS classnames that are append to the popup content body.
```
JakeLaCombe: *appended

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -0,0 +1,286 @@
...
+   */
+  classNameInner: PropTypes.string,
+  /**
+   * Whether or not the using the escape key should trigger the onRequestClose callback.
+   */
+  closeOnEsc: PropTypes.bool,
```
JakeLaCombe: Is it possible to use this convention for the bools in this file? https://github.cerner.com/terra/conventions/blob/master/react.md#for-boolean-props-define-defaultprop-value-as-false

Someline like isClosableOnEsc should work

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -0,0 +1,286 @@
...
+  /**
+   * A the px value of height to be applied to the content container.
+   */
+  contentHeight: PropTypes.number.isRequired,
+  /**
+   * The maximum height to set for popup content in px, also used with responsive behavior for actual height.
+   */
+  contentHeightMax: PropTypes.number,
+  /**
+   * A the px value of width to be applied to the content container.
```
JakeLaCombe: Reword sentence

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -0,0 +1,286 @@
...
+   */
+  contentWidthMax: PropTypes.number,
+  /**
+   * Should the default behavior, that inserts a header when constraints are breached, be disabled.
+   */
+  isHeaderDisabled: PropTypes.bool,
+  /**
+   * The function that should be triggered when a close is indicated.
+   */
+  onRequestClose: PropTypes.func.isRequired,
```
JakeLaCombe: required props should come first in the props hash.

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -0,0 +1,286 @@
...
+      return true;
+    }
+    return dimensionMax >= dimension + POPUP_MARGIN;
+  }
+
+  static shouldShowArrow(arrow, arrowPosition, contentHeight, contentHeightMax, contentWidth, contentWidthMax) {
+    if (arrow) {
+      if (['top', 'bottom'].indexOf(arrowPosition) >= 0) {
+        return PopupContent.isMarginValid(contentHeight, contentHeightMax);
+      }
```
JakeLaCombe: Should there also be a check for left and right?

packages/terra-popup/src/_TetherComponent.jsx
```diff
@@ -0,0 +1,251 @@
...
+  /**
+   * A hash of tether classes which should be changed or disabled.
+   */
+  classes: PropTypes.object,
+  /**
+   * The prefix placed at the beginning of the default classes.
+   */
+  classPrefix: PropTypes.string,
+  /**
+   * Rule set to pass to tether, contraining the content to by setting 'to', 'attachment', and/or 'pin'.
```
JakeLaCombe: Perhaps we should provide a link to the tether docs here? This is a very lengthy comment which corresponds to a variable doing multiple things.

packages/terra-site/src/examples/popup/PopupTopCenter.jsx
```diff
@@ -0,0 +1,55 @@
...
+    this.setState({ open: false });
+  }
+
+  handleOnChange() {
+    this.setState({ open: false });
+  }
+
+  render() {
+    return (
+      <div style={{ display: 'inline-block' }} ref={this.setButtonNode}>
```
JakeLaCombe: We need to be care when we have custom styles like this in our documentation. When you have code in the docs that don't use best practices, those easily manifest their way into their consuming applications. Perhaps there is a way we can remove this?

dkasper-was-taken: That's the topic of the findDomNode, refCallBack, and wrapping div that was brought up. Once we settle on a strategy I can update this pattern

packages/terra-popup/src/_TetherComponent.jsx
dkasper-was-taken: Thats a good idea.

packages/terra-popup/src/_PopupContent.jsx
dkasper-was-taken: Are we going by that or alphabetization?

JakeLaCombe: I believe we sort them by required alphabetized, then optional alphabetized

packages/terra-popup/src/Popup.jsx
```diff
@@ -0,0 +1,368 @@
...
+};
+
+const MIRROR_TB = {
+  middle: 'middle',
+  top: 'bottom',
+  bottom: 'top',
+};
+
+const BASE_WIDTH = 16;
+const BASE_HEIGHT = 9;
```
jmsv6d: Would we want to export these values for consumers who want to may want to calculate what popup size to use based on the content?

neilpfeiffer:
Added visual design updates for you here:  https://wiki.ucern.com/x/EeSZY
Mainly is the sizing, but updated some box-shadow updates.  Need to validate the scrolling behavior when only one dimension (height or width) is greater than window, vs. fullscreen with header when both are greater.

packages/terra-popup/src/_TetherComponent.jsx
dkasper-was-taken: Added a link in the readme to the tether docs.

dkasper-was-taken:
Closing, will reopen a new PR on a squashed branch.

PR #448, opened 2017-06-12
Terra form controls

JakeLaCombe:
### Summary
Control inputs are checkboxes and radio type inputs. For this component, they are wrapped inside a label so that the text can be clicked to activate the control element.

### Deployment Link
https://terra-core-deployed-pr-448.herokuapp.com/


packages/terra-form/src/Control.jsx
```diff
@@ -0,0 +1,103 @@
...
+   */
+  isInline: PropTypes.bool,
+  /**
+   * Text of the label
+   */
+  labelText: PropTypes.node,
+  /**
+   * Additional objects for the text object
+   */
+  labelTextAttrs: PropTypes.object,
```
bjankord: Should we add `labelAttrs` prop as well?

JakeLaCombe: That's actually covered by customProps, as the main container is a label. If you think that's confusing though, I can change customProps to be labelAttrs.

bjankord: Ok, I'm fine with the current implementation, no need to change it.

packages/terra-form/src/Control.jsx
```diff
@@ -0,0 +1,103 @@
...
+  const labelTextClasses = classNames(
+    'terra-Form-control-label-text',
+    labelTextAttrs.className,
+  );
+
+  const controlInputAttrs = Object.assign({}, inputAttrs);
+
+  if (checked !== undefined) {
+    controlInputAttrs.checked = checked;
+  } else {
```
emilyrohrbough: Is there ever an instance where both `checked` and `defaultChecked` are undefined due to the defaultProp values provided?

packages/terra-form/src/Control.scss
```diff
@@ -0,0 +1,23 @@
...
+  display: block;
+  margin-bottom: 0.714em;
+  width: auto;
+
+  .terra-Form-input {
+    vertical-align: middle;
+  }
+}
+
+@media screen and (min-width: 768px) {
```
emilyrohrbough: When doing the selection styles in table, Neil removed the logic for a responsive table in tiny mode. His reasoning was the base component should have a standard display in all viewports. Does this concept apply here? @neilpfeiffer

packages/terra-form/src/Control.jsx
JakeLaCombe: It would be an instance like this <Control name="name" />.

I think at a minimum that checked should default to undefined, as we need to determine if the user meant to set that value or not, but I think we can be different with defaultChecked, as setting it to false or not defining it sets the component to be uncontrolled. 

So the only other consideration is to determine if we want checked and defaultChecked to have a similar api, which includes having the same defaultChecked value.

Do others have an opinion?

emilyrohrbough: I agree check should default to undefined, especially if a HOC checkbox or radio button is to be implemented.

emilyrohrbough: If we don't give defaultChecked a value, I do wonder if the current implementation will cause an error if neither checked or defaultChecked are provided.

mmalaker:
Functional verification completed. The radio and checkbox control examples display and function correctly.

packages/terra-form/src/Control.jsx
JakeLaCombe: We talked in person. The default behavior for a react component when it receives undefined is to act as an uncontrolled value, and I'd like for consumers to be able to use the default react behavior for these inputs. As such, I believe it's best to keep it as undefined for the defaultChecked.

neilpfeiffer:
+1 with these comments:

**Fixes for this PR**
1. Remove/fix `height` on controls (radio and checkboxes)

2. For fields and controls set to inline, when they are below responsive-breakpoint and a no longer presented inline, set `margin-right:0;` 


**Fixes for next PR**
- [ ] Change the markup order of the help-text and error-text for Fieldsets (move above all fields, immediately after the Legend):
```
(FieldSet)
Legend
help-text
error
  (Field) label, field, help-text, error (/Field)
  (Field) label, field, help-text, error (/Field)
  (Field) label, field, help-text, error (/Field)
(/FieldSet)
```

- [ ] Change responsive breakpoint from `small (768px)` to `tiny (544px)` for all form elements that have responsive styling.

neilpfeiffer:
**Fixes for next PR**
- [ ] Change the markup order of the help-text and error-text for Fieldsets (move above all fields, immediately after the Legend):
```
(FieldSet)
Legend
help-text
error
  (Field) label, field, help-text, error (/Field)
  (Field) label, field, help-text, error (/Field)
  (Field) label, field, help-text, error (/Field)
(/FieldSet)
```

- [ ] Change responsive breakpoint from `small (768px)` to `tiny (544px)` for all form elements that have responsive styling.

PR #450, opened 2017-06-13
Remove SCSS theming capability

### LINKED ISSUE #445: Remove SCSS theming capability from terra components
# Issue Description
We currently support theming via SCSS variables in various components. As we plan to switch to a css-in-js solution for styling, the current implementation of theming variables leaks implementation details of how we are styling/theming components. 

With css-in-js, we'd plan to use JSON files for theming. Switching from SCSS variables to a JSON file for theming would be a non-passive change for our consumers.

We've discussed removing theming functionality via SCSS variables before the v1 release of the various components. This will allow use to roll out theming functionality via JSON files as a new feature rather than a non-passive change.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
All components use hard-coded values instead of SCSS theming variables.

## Current Behavior
Various terra components use SCSS theming variables.

bjankord:
### Summary
Resolves #445.

* Hard-coded values pulled from terra-legacy-theme.
* Removed `!default` from SCSS variables. This disables the ability to override the variables but they are still available for local file usage. Consumers won't be able to modify their definition initially set in each components `_variables.scss` file.
* Set `$terra-bidi` to true. This is to maintain bidi styling as we temporarily remove theming capability in the migration to css-in-js.
* Removed theming variables documentation.
* Removed SCSS config for importing `node_modules/terra-legacy-theme/lib/LegacyTheme.scss` in every SCSS file from webpack config.
* Removed terra-legacy-theme as a dependency from terra-site.

@cerner/terra


packages/terra-site/webpack.config.js
```diff
@@ -48,11 +48,9 @@ module.exports = {
               ];
             },
           },
-        }, {
+        },
+        {
           loader: 'sass-loader',
-          options: {
-            data: `@import "${path.resolve(path.join(__dirname, 'node_modules/terra-legacy-theme/lib/LegacyTheme.scss'))}"; $terra-bidi: true;`,
```
Matt-Butler: Do we still need to specify $terra-bidi: true?

bjankord: Its been hard coded in the the files that explicitly use the `$terra-bidi` variable. (mixins, datepicker, demographics banner, and slidepanel)

Matt-Butler: https://github.com/cerner/terra-core/pull/450/files#diff-a0157ec3f4609e3df477bbd4bf487b52L4

nm, I see it is hardcoded where needed

packages/terra-date-picker/src/_variables.scss
```diff
@@ -0,0 +1,4 @@
+// Variables
```
emilyrohrbough: Instead of adding a file to handle` $terra-bidi: true` should we update `DatePicker.scss` to use the `terra-margin-start` mixin [here](https://github.com/cerner/terra-core/blob/eccc6f0c81dd2e93419917e5f45fad861c430291/packages/terra-date-picker/src/DatePicker.scss#L20) and [here](https://github.com/cerner/terra-core/blob/eccc6f0c81dd2e93419917e5f45fad861c430291/packages/terra-date-picker/src/DatePicker.scss#L35)

packages/terra-demographics-banner/src/DemographicsBanner.scss
```diff
@@ -7,7 +7,7 @@
   margin-bottom: 0.714em;
   padding: 0.714em;
 
-  @media screen and (min-width: $terra-small-breakpoint) {
+  @media screen and (min-width: 768px) {
```
emilyrohrbough: Are we hard coding the breakpoints so consumers are unable to override them?

bjankord: Yes, all variables that consumers could theme are being hard-coded. This is to make the switch to a css-in-js solution more passive for consumers so they don't have to be concerned with migrating a sass file with variables to a JSON file for variables.
Once we've switched over to a css-in-js solution, we'll add the theming capabilities back as a enhancement which will be passive for consumers. That work is scheduled to start in July.

packages/terra-date-picker/src/_variables.scss
bjankord: Good call, updated: https://github.com/cerner/terra-core/pull/450/commits/1ca87545fb044227388ed097e1cb9b9e866bae80

## LTR after update
screen shot 2017-06-13 at 9 15 30 am

## RTL after update
screen shot 2017-06-13 at 9 15 39 am

packages/terra-demographics-banner/src/DemographicsBanner.scss
emilyrohrbough: Cool!

packages/terra-slide-panel/src/_variables.scss
```diff
@@ -1,4 +1,8 @@
-$terra-slide-panel-panel-width-small: 320px !default;
-$terra-slide-panel-panel-width-large: 60% !default;
-$terra-slide-panel-panel-border-color: $terra-grey-20 !default;
-$terra-slide-panel-shadow-color: $terra-grey-30 !default;
+// Variables
+// ==========================
+
+$terra-bidi: true;
```
emilyrohrbough: I just took a peak, I don't think we need to add `$terra-bidi` for the SlidePanel.scss file.

Issue #401, opened 2017-05-25
Feature to format the date and the value sent back in ISO8601 format


bjankord:
# Issue Description
Add feature to format the date and the value sent back in ISO8601 format

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior

When the value of the date picker is accessed / returned it is formatted in ISO8601 format

## Current Behavior

When the value of the date picker is accessed / returned, it is formatted as it was entered into the input.

<img width="794" alt="screen shot 2017-05-25 at 4 42 50 pm" src="https://cloud.githubusercontent.com/assets/633148/26471884/42e6328e-4169-11e7-844e-ac9a9e2fa78d.png">

dnguyenchevy:
The real concern with this issue is when you start introducing internationalization.

Looks like there's an internationalization issue already open, https://github.com/cerner/terra-core/issues/246.

PR #452, opened 2017-06-13
Issue-401 Create hidden input containing name and value attributes fo…

benbcai:
Create a hidden input to store the name and value attributes for for submission. The value will be the entered date or time in ISO 8601 format. When the entered date or time is invalid, the value will be whatever that entered value is.

For the hidden date input, I initially set the value to be the date followed by a time of 00:00:00 followed by the offset. Since the date component doesn't have the time part, I removed the time from ISO. So instead of 2017-06-14T00:00:00-05:00 initially, I changed it to just 2017-06-14. This is how blue-steel does it.

screen shot 2017-06-16 at 5 04 56 pm

For the hidden time input, the value would just be T followed by the hour and minute. For example: "T12:45"

screen shot 2017-06-16 at 5 06 10 pm

Resolves https://github.com/cerner/terra-core/issues/401

@cerner/terra


packages/terra-date-picker/README.md
```diff
@@ -3,7 +3,9 @@
...
+Terra-date-picker is a controlled input component that provides users a way to enter or select a date from the date picker. terra-date-picker is essentially a wrapper for [react-datepicker](https://github.com/Hacker0x01/react-datepicker) and leverages many of its props. One important difference between terra-date-picker and [react-datepicker](https://github.com/Hacker0x01/react-datepicker) is that all of the date props in [react-datepicker](https://github.com/Hacker0x01/react-datepicker) must be a [moment](http://momentjs.com/docs/) object whereas the date props in terra-date-picker are ISO 8601 representation of the date.
```
benbcai: Updated in https://github.com/cerner/terra-core/pull/452/commits/d5d1b6d6726fa521cca74547812186af5349ea0c

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -36,12 +37,30 @@ const propTypes = {
 // eslint-disable-next-line react/prefer-stateless-function
 class DatePickerInput extends React.Component {
   render() {
+    // If a name attribute is provided in inputAttributes, it will be used as the name attribute in the hidden input.
+    // If no name attribute is provided, a default name of 'terra-date' will be used.
+    let nameAttr = 'terra-date';
```
benbcai: As discussed in post scrum, a required name prop is added and we expect that the consumer will pass in a unique name for each instance of the input.

benbcai: As discussed in post scrum, a required name prop is added and we expect that the consumer will pass in a unique name for each instance of the input.

packages/terra-date-picker/src/DateUtil.js
```diff
@@ -28,6 +28,16 @@ class DateUtil {
 
     return momentDates.length > 0 ? momentDates : dates;
   }
+
+  // Converts date string to the ISO8601 format in UTC. If the date string is invalid and unable to convert, the originally provided string is returned.
+  static convertToISO8601(date, format) {
```
benbcai: I want it to return the original date that is passed in so that if the date is invalid, whatever date that was entered will be used as is to set as the value in the hidden input.

packages/terra-date-picker/tests/nightwatch/date-picker-spec.js
```diff
@@ -68,5 +68,28 @@ module.exports = {
     browser.expect.element('.terra-DatePicker-range > .terra-DatePicker').to.be.present;
   },
 
+  'Creates a hidden input with a name atribute of "test-name-attr" and an empty value attribute when no date is entered': (browser) => {
+    browser.url(`http://localhost:${browser.globals.webpackDevServerPort}/#/tests/date-picker-tests/default`);
+
+    browser.expect.element('.terra-hidden-date-input').to.have.attribute('name').which.equals('test-name-attr');
```
benbcai: I tried that but it fails to match. I basically followed the nightwatch API [documentation](http://nightwatchjs.org/api#expect-attribute) and this is how most of the other nigthwatch tests are written.

packages/terra-date-picker/src/DateUtil.js
JakeLaCombe: That sounds fair. I take it then that if the function is successful that it will never return the same value that was passed in. Since you documented that approach, it should be fine.

packages/terra-date-picker/tests/nightwatch/date-picker-spec.js
JakeLaCombe: Wow, I gave you bad syntax. I'm sorry. I think this is what I meant.

    browser.expect.element('.terra-hidden-date-input[name="test-name-attr"]')

However, this is just a suggestion. It makes it smaller, but your also spells it all out.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -11,7 +11,11 @@ const propTypes = {
    * Custom input attributes to apply to the time input.
    */
   inputAttributes: PropTypes.object,
-    /**
+  /**
+   * Name of the time input. The name should be unique and is required.
```
JakeLaCombe: Minor:

Wouldn't the prop tables list it as required? Perhaps potential duplication documentation?

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -70,20 +85,20 @@ class DatePicker extends React.Component {
...
   }
 
-  handleChange(date) {
+  handleChange(date, event) {
     this.setState({
       selectedDate: date,
     });
 
     if (this.props.onChange) {
       const dateString = date && date.isValid() ? date.format(this.state.dateFormat) : '';
```
emilyrohrbough: Should this be checked before setting the `selectedDate` state?

benbcai: I actually don't need to make this check because handleChange gets called from [here](https://github.com/Hacker0x01/react-datepicker/blob/master/src/datepicker.jsx#L234) only when an acceptable change has been made. So I've removed the check here.

https://github.com/cerner/terra-core/pull/452/commits/109a6590872e9de8cfee253af0a9fb3366ec60c6

packages/terra-time-input/src/TimeInput.jsx
benbcai: Updated: https://github.com/cerner/terra-core/pull/452/commits/b72d27942aa8839f8dede5b75d7e11ed20e5ee07

Issue #443, opened 2017-06-07
Add labelAttrs to the Field Component and legendAttrs to the Fieldset Component


JakeLaCombe:
# Issue Description
Fields have the ability to add custom attributes such as classes to the input node. There should also exist similar to field labels and fieldset legends.

## Issue Type
- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Passing labelAttrs to a Field should add those to the label, and passing legendAttrs to a FieldSet should add those to the labels

## Current Behavior
These props are not supported.

bjankord:
Resolved in #459

PR #459, opened 2017-06-14
Terra Form Label Attributes

JakeLaCombe:
### Summary
This pull request adds labelAttrs to fields and legendAttrs to fieldsets as requested in Issue #443.


packages/terra-form/docs/Field.md
```diff
@@ -10,6 +10,7 @@ import Field from 'terra-form/lib/Fieldset';
 
 <Field
   label="Do you have any Children?"
+  labelAttrs={{ className: 'healtheintent-application' }}
   error="This field is required"
   help="Families are eligible for family package plans"
   required
```
StephenEsser: Just a topic for a future discussion. Is there any known reason that consumers would also want to style the error and help fields?

In the worst case scenario would this scale to:
```

<Field
  label="Do you have any Children?"
  labelAttrs={{ className: 'healtheintent-application' }}
  error="This field is required"
  errorAttrs={{ className: 'healtheintent-error' }}
  help="Families are eligible for family package plans"
  helpAttrs={{ className: 'healtheintent-help' }}
  required
  isInline
 >
   {this.props.children}
</Field>
```
or
```
<Field required isInline>
   <Label className="healtheintent-application">Do you have any Children?</Label>
   <Error className="healtheintent-error">This field is required</Error>
   <HelpText className="healtheintent-help">Families are eligible for family package plans</HelpText>
   {this.props.children}
</Field>
```
or
```
<Field
  label={<Label className="healtheintent-application">Do you have any Children?</Label>}
  error={<Error className="healtheintent-error">This field is required</Error>}
  help={<HelpText className="healtheintent-help">Families are eligible for family package plans</HelpText>}
  required
  isInline
 >
   {this.props.children}
</Field>
```

mjhenkes:
@JakeLaCombe does this require Functional and/or UX review?

packages/terra-form/docs/Field.md
JakeLaCombe: If we go with styled components, I would see the 3rd option as the result api most likely. From looking at styled components, I don't think we would have the need potentially to attach individual classes to the elements to get the desired styles, as they could just style the Label they provided to Label prop, and use those styles as is.

As for option 2, the main thing I'm concerned for on that one is that Fields are designed to be an easy API where developers don't need to worry about having to determine where the label and error messages go, and the second option brings that concern back. We do have those components in place that allow for consumers to build their own Fields if they need the label, help, and error in different places, so I lean more towards having consumers build that if they need it.

JakeLaCombe:
I don't think we do @mjhenkes. This doesn't change any existing functionality, and just places new attributes on labels for fields.

packages/terra-form/docs/Field.md
bjankord: I prefer options 1 and 3 if we were to need to scale this out with more preference on option 1.

styled-components has a [`.attrs` method](https://www.styled-components.com/docs/api#attrs) which accepts an object of attributes to add to an element. We could pass the object from things like `labelAttrs` directly to that.

Adding attributes directly to components like `<Label className="healtheintent-application">` should be an issue either.

bjankord:
@JakeLaCombe  @mjhenkes Sounds good, lets get this merged.

Issue #460, opened 2017-06-14
Update terra-base to accept a placeholder component prop that is specifically there for when translations are loading


ryanthemanuel:
# Issue Description

Currently, terra-base returns null as its rendering while translations are loading.  Some consumers need to have a mix of React and non-React components on the screen at the same time.  If the translations are slow to load, this could cause potential flickering as the React components snap in.  To alleviate that, we will update terra-base to accept a placeholder component prop that is specifically there for when translations are loading.

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
While translations are loading for the first time, we will display the React component that is passed in as the optional prop translationsLoadingPlaceholder

## Current Behavior
terra-base displays nothing while translations are being loaded.

PR #461, opened 2017-06-15
Update base to support a “translations loading placeholder”

ryanthemanuel:
### Summary
See #460 for more details on this change

### DIFF
--- a/packages/terra-base/src/Base.jsx
+++ b/packages/terra-base/src/Base.jsx
@@ -24,6 +24,13 @@ const propTypes = {
       );
     }
   },
+  /**
+   * The component(s) that will be wrapped by `<Base />` ONLY
+   * in the event that translations have not been loaded yet.
+   * NOTE: Absolutely no locale-dependent logic should be
+   * utilized in this placeholder.
+   */
+  translationsLoadingPlaceholder: PropTypes.node,
 };
 
 const defaultProps = {
@@ -58,6 +65,7 @@ class Base extends React.Component {
       children,
       locale,
       customMessages,
+      translationsLoadingPlaceholder,
       ...customProps
     } = this.props;
 
@@ -70,7 +78,7 @@ class Base extends React.Component {
     const messages = Object.assign({}, this.state.messages, customMessages);
 
     if (locale === undefined) return childComponent;
-    if (!this.state.areTranslationsLoaded) return null;
+    if (!this.state.areTranslationsLoaded) return <div {...customProps}>{this.props.translationsLoadingPlaceholder}</div>;
     return (
       <I18nProvider locale={this.state.locale} messages={messages}>
         {childComponent}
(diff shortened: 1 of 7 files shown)


packages/terra-base/src/Base.jsx
```diff
@@ -70,7 +78,7 @@ class Base extends React.Component {
     const messages = Object.assign({}, this.state.messages, customMessages);
 
     if (locale === undefined) return childComponent;
-    if (!this.state.areTranslationsLoaded) return null;
+    if (!this.state.areTranslationsLoaded) return <div {...customProps}>{this.props.translationsLoadingPlaceholder}</div>;
```
windse7en: For this code, what's the reason of adding `customProps` to the placeholder's div? I think all style related props(class...) in customProps are specified for the loaded div.

I suggest we make the assumption, the loadingPlaceholder should be encapsulated with its own style config.

packages/terra-base/tests/nightwatch/NoTranslationsLoadedBase.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import Base from '../../lib/Base';
+
+const BaseNoTranslationsLoaded = () => {
+  const noTranslations = (
+    <Base locale="gabcdef" translationsLoadingPlaceholder={<div>No Translations</div>}>
```
windse7en: It should throw not supported locale error message in the console.
```
Uncaught Error: gabcdef is not supported, supported locales:en,en-US,en-GB,es,de,fi-FI,fr,pt
```
That may make the nightwatch test fragile. Could we capture this error message in the spec? I also don't know a better way to test this. Or maybe just remove the nightwatch test.

ryanthemanuel: The night watch test seems to pass.  I couldn't find a way to test this in night watch, other than to make translations completely not go through.  Which meant fail loading.  I'm cool with dropping night watch if we don't want that.

windse7en: If it doesn't throw error message in phantomjs, I'm fine with keeping it. If it does throw error message in phantomjs, I vote to remove this nightwatch test.

The concern is, the nightwatch test will depend on I18n component and Base component. That makes it quite fragile. If we move or update the supported locales validation, it will probably fail the nightwatch test in here.

ryanthemanuel: I'm going to keep it...but i might add a comment that this is ok to remove if it starts causing problems.

windse7en: OK. Sounds good to me. 👍

windse7en:
The heroku reviews app is weird. It can't find the react dependency: create-react-class@15.6.0. But I checked in https://semver.npmjs.com/, there is a 15.6.0 version. This issue also blocks 448 PR review deployment.

PR #468, opened 2017-06-15
Add Types to Buttons

JakeLaCombe:
### Summary
This adds the ability to add the type attribute to any terra buttons. This allows for more consistent and better browser performance when working with buttons.

### DIFF
--- a/packages/terra-button/src/Button.jsx
+++ b/packages/terra-button/src/Button.jsx
@@ -46,6 +46,10 @@ const propTypes = {
    */
   text: PropTypes.string,
   /**
+   * Sets the button type. One of button, submit, or reset
+   */
+  type: PropTypes.oneOf(['button', 'submit', 'reset']),
+  /**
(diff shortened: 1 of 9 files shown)


packages/terra-button/src/Button.jsx
```diff
@@ -46,6 +46,10 @@ const propTypes = {
    */
   text: PropTypes.string,
   /**
+   * Sets the button type. One of button, submit, or reset
+   */
+  type: PropTypes.oneOf(['button', 'submit', 'reset']),
```
emilyrohrbough: Did we come to a final consensus on using  enum objects for known prop values or strings?

bjankord: Let's stick with strings for Phase 1 work. As part of the Q3 work, we can look more into using enum objects. 

The enum objects would be an option for consumers to import that should be passive the way we've been doing current prop-types.

Issue #462, opened 2017-06-15
props-table cli


Matt-Butler:
# Issue Description
Add an executable to props-table to generate a markdown table for prop-types. The props-table currently does not work on terra-ui.com because generating the props-table requires the src component (es6). We want create a props table as documentation that will be distributed with the package.

How eslint created a cli tool:
https://github.com/eslint/eslint/blob/master/package.json#L6-L8
https://github.com/eslint/eslint/blob/master/bin/eslint.js

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

When this issue is complete, each component will distribute a props-table that will be consumable in terra-ui.com

Matt-Butler:
The initial arguments for the cli will be similar to [babel](https://babeljs.io/docs/usage/cli/)

```
props-table src/Arrange.jsx --out-dir docs
```

This would generate a props table for the arrange component and output a markdown file in the docs folder.

Matt-Butler:
| Prop Name | Type | Is Required | Default Value | Description | 
 |-|-|-|-|-|
| children | `node`  | optional |  | Child Nodes |
| href | `string`  | optional |  | Sets the href. When set will render the component as an anchor tag |
| icon | `element`  | optional |  | An optional icon. Nested inline with the text when provided |
| isBlock | `bool`  | optional | `false` | Whether or not the button should display as a block |
| isCompact | `bool`  | optional | `false` | Whether or not the button has reduced padding |
| isDisabled | `bool`  | optional | `false` | Whether or not the button should be disabled |
| isReversed | `bool`  | optional | `false` | Reverses the position of the icon and text |
| onClick | `func`  | optional |  | Callback function triggered when clicked |
| size | `enum('tiny','small','medium','large','huge')`  | optional |  | Sets the button size. One of tiny, small, medium, large, huge |
| text | `string`  | optional |  | Sets the button text |
| variant | `enum('default','link','primary','secondary')`  | optional | `'default'` | Sets the button variant. One of primary, secondary, or link |


This is what the cli is currently creating for the button component

Matt-Butler:
resolved in #473

PR #473, opened 2017-06-19
Props table cli

Matt-Butler:
### Summary
Props-table CLI. This tool generates a markdown table for es6 react components. 

@cerner/terra


packages/terra-arrange/docs/Arrange.md
```diff
@@ -0,0 +1,9 @@
+| Prop Name | Type | Is Required | Default Value | Description |
```
Matt-Butler: I think we only want to in the markdown table when we publish a package. I think we should add the each prop-table to our gitignore. What are your thoughts?

Matt-Butler:
https://github.com/cerner/terra-core/issues/462

packages/terra-arrange/docs/Arrange.md
emilyrohrbough: I am not opposed to providing the prop markdown in the documentation. I think it could be useful to consumers, though I would vote for a consistent naming scheme like `PROPS.md` that could be referenced from the `README.md` file.

packages/terra-form/docs/Field.md
```diff
@@ -1,21 +1,10 @@
-# Field
```
emilyrohrbough: It looks like this overrode the documentation for the form components.

emilyrohrbough:
It could be nice add the component name above the prop table since some packages have subcomponents.

packages/terra-arrange/docs/Arrange.md
Matt-Butler: My concern with keeping it in our repo is that we need to enforce the props-table to be updated when we change src code. We would need to add a prepush script to keep the src files in sync with the props-table.

I was going to make a consistent naming scheme except we have some components that have multiple props tables ([list](https://github.com/cerner/terra-core/tree/props-table-cli/packages/terra-list/docs) for example).

packages/terra-form/docs/Field.md
Matt-Butler: matt-butler--

I'll get this fixed.

packages/terra-arrange/docs/Arrange.md
JakeLaCombe: What all do we plan on using the cli for? Is the output designed to be used with normal documentation, or is this cli just for showing a page on github?

bjankord: I'd like to use it both as doc on GitHub as well as import it into the examples so it can on be used on the doc site.

The concern Matt noted about keeping the props table in sync with the src code when imported into the examples would be a good topic for Weds. meeting.

bjankord: Decided to just generate when we publish. And import them into terra-ui.com site.

Matt-Butler:
Spoke offline yesterday. We will not check the props-table into github. It will be generated when we publish a package. I added props-table to the .gitignore.


This is what a  few tables look like after my changes:



## Arrange Props Table
| Prop Name | Type | Is Required | Default Value | Description |
|-|-|-|-|-|
| align| `string`| optional| | The vertical orientation of all three containers. It will override the aligment of alignFitStart, alignFill and alignFitEnd if given. One of: `center`, `bottom`, `stretch`.|
| alignFill| `string`| optional| | The vertical orientation of fitEnd. One of: `center`, `bottom`, `stretch`.|
| alignFitEnd| `string`| optional| | The vertical orientation of fill. One of: `center`, `bottom`, `stretch`.|
| alignFitStart| `string`| optional| | The vertical orientation of fitStart. One of: `center`, `bottom`, `stretch`.|
| fill| `element`| `required`| | The content to display in the body of the fill.|
| fitEnd| `element`| optional| | The content to display in the body of the fitEnd.|
| fitStart| `element`| optional| | The content to display in the body of the fitStart.|





## Modal Props Table
| Prop Name | Type | Is Required | Default Value | Description |
|-|-|-|-|-|
| ariaLabel| `string`| `required`| `null`| String that labels the modal for screen readers|
| children| `node`| `required`| `null`| Content inside the modal dialog|
| classNameModal| `string`| optional| `null`| CSS classnames that are append to the modal|
| classNameOverlay| `string`| optional| `null`| CSS classnames that are append to the overlay|
| closeOnEsc| `bool`| optional| `true`| If set to true, the modal will close when the esc key is pressed|
| closeOnOutsideClick| `bool`| optional| `true`| If set to true, the modal will close when a mouseclick is triggered outside the modal|
| isFullscreen| `bool`| optional| `false`| If set to true, the modal will be fullscreen on all breakpoint sizes|
| isOpen| `bool`| `required`| `false`| If set to true, the modal will rendered as opened|
| isScrollable| `bool`| optional| `false`| If set to true, the modal dialog with have overflow-y set to scroll. It is recommended not to use this prop and instead create a HOC with the modal dialog body set to scroll.|
| onRequestClose| `func`| `required`| | Function to set isOpen={false} and close the modal.|
| role| `string`| optional| `'document'`| Role attribute on the modal dialog|





## SearchField Props Table
| Prop Name | Type | Is Required | Default Value | Description |
|-|-|-|-|-|
| minimumSearchTextLength| `number`| optional| `2`| The minimum number of characters to perform a search.|
| onSearch| `func`| optional| | A callback to perform search. Sends parameter {String} searchText.|
| placeholder| `string`| optional| `''`| Placeholder text to show while the search field is empty.|
| searchDelay| `number`| optional| `250`| How long the component should wait (in milliseconds) after input before performing an automatic search.|

Issue #466, opened 2017-06-15
[terra-image] Add onLoad and onError to determine if an image successfully loaded


mschile:
# Issue Description
In order to determine if an image successfully loaded or not, two new func properties, onLoad and onError, should be added to terra-image and set on the underlying img.

This functionality is needed for #429.

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
onLoad or onError is called when image loads or fails.

## Current Behavior
N/A

Matt-Butler:
resolved in https://github.com/cerner/terra-core/pull/474

Issue #429, opened 2017-06-02
Enhance or add a higher level component around the Image component to allow for a default image if the image fails to load


ryanthemanuel:
# Issue Description
Certain situations (especially patient photos) require that there is a back up image if the given image fails to load.  It would be ideal to have a mechanism to expose this first class to consumers.  This could either be a prop on the existing Image component or a higher level component if need be.

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

ryanthemanuel:
Let's go ahead and design this as if it is going to be a prop on the existing Image component.  Matt Schile, who is working on implementing the Demographics component, needs this functionality for the image in the demographics component.  We have requested that he work on this functionality.

mschile:
Discussed on June 7th during the Terra sync with @bjankord, @mjhenkes, and @ryanthemanuel and decided create a higher level component for the patient photo that will contain an avatar to be displayed while the actual patient photo is being loaded.  Once the patient photo is loaded it will replace the avatar image.  If the patient photo fails to load, the avatar will remain in place.

### Design

#### Component name:
terra-patient-photo
terra-patient-image
terra-profile-image

#### Props:
**image**: terra-image

#### State:
**isLoading**: boolean used to indicate if the patient image is still loading
**isLoaded**: boolean to indicate if the patient image was successfully loaded.  Will be false if the patient image failed to load.

#### Flow:
1. Copy the properties (other than src) set on image prop to the internal avatar image to maintain a consistent look for the avatar and patient images.  Not sure I like copying all of the properties since there could be other properties that we would not want to copy, such as the onLoad and onError properties I plan on adding to image.  Instead of copying the properties, I could just replicate the properties on this new component.  Thoughts?
2. Set state to isLoading: true and isLoaded: false.
3. During render return a div that contains both the avatar and patient images.  The patient image will be hidden until it is successful.
4. If the onLoad event fires, update the state to isLoading: false and isLoaded: true.  Render only the loaded patient image.
5. If the onError event fires, update the state to isLoading: false and isLoaded: false.  Render only the avatar image.

I'm going to log another issue for terra-image to add two new properties to determine when an image has loaded.

ryanthemanuel:
I like 'terra-profile-image'.  The question is would we want this to land in the terra-clinical space or the terra-core space.  I vote terra-core, as this profile could allow for various "people" concepts.

ryanthemanuel:
[...]

PR #474, opened 2017-06-19
[terra-Image] Adding onLoad and onError func props to Image.

mschile:
### Summary
#466 Adding onLoad and onError function properties to terra-Image.


emilyrohrbough:
It looks like there are no jest snapshots for the Image component. Is that something that should be updated?

Matt-Butler:
@mschile Thank you for your help on this.

I am a bit confused about what the onLoad and onError functions are supposed to do. Can you please provide more information?

mschile:
@Matt-Butler, the onLoad and onError functions are to allow the consumer to determine if an image was successfully loaded or not.  For my use-case, I want to load an avatar image while the patient photo is loading, if the patient photo successfully loads, I want to display that photo, otherwise, I want to continue to display the avatar image #429.

packages/terra-image/src/Image.jsx
```diff
@@ -59,12 +69,13 @@ const Image = ({
       alt={alt}
       height={height}
       width={width}
+      onLoad={onLoad}
+      onError={onError}
```
Matt-Butler: https://facebook.github.io/react/docs/events.html#image-events

mschile: Since I'm still new to React development I'm trying to decipher what you are implying by posting the link.  Looking at the code again, I think you are saying that customProps would handle the onLoad and onError events so they aren't needed?  If so, should src, alt, height, and width also be removed?  The description for terra-Image should also be updated to explicitly state it's a wrapper around image instead of stating "The terra-image component provides styling for visual imagery."

packages/terra-image/src/Image.jsx
```diff
@@ -29,6 +29,14 @@ const propTypes = {
    * Sets the width of the image.
    */
   width: PropTypes.string,
+  /**
+   * Function to be executed when the image load is successful.
+   */
+  onLoad: PropTypes.func,
```
JakeLaCombe: Do we need default props for theses functions? We have other func props that set the value to undefined, such onChange for terra-form.

packages/terra-image/src/Image.jsx
JakeLaCombe: I think the main concern is that if you pass onLoad and onError into the object without them being first class, then they would get added to the image anyway through customProps.

Matt-Butler: I didn't know about these image events so I was posting this so other people know they exist. I think what you have is fine.

packages/terra-image/src/Image.jsx
mschile: I don't think they need to be explicitly set to undefined as they will be undefined if they aren't passed.

packages/terra-image/src/Image.jsx
JakeLaCombe: I'm sorry. Got confused by the post. Thanks for the link!

packages/terra-image/tests/nightwatch/ImageLoading.jsx
```diff
@@ -0,0 +1,30 @@
...
+    return (
+      <div>
+        <h2>Loading Image Examples</h2>
+        <p>Image is successfully loaded.</p>
+        <div>
+          <Image id={this.state.loadedImageId} src={placeholder150x150} onLoad={() => { this.setState({ loadedImageId: 'loadedImage' }); }} />
+        </div>
+        <p>Image fails to load.</p>
+        <div>
+          <Image id={this.state.errorImageId} src="invalid.jpg" onError={() => { this.setState({ errorImageId: 'errorImage' }); }} />
```
Matt-Butler: Can you add onLoad to this image?

packages/terra-image/tests/nightwatch/ImageLoading.jsx
```diff
@@ -0,0 +1,30 @@
...
+    this.state = { errorImageId: undefined, loadedImageId: undefined };
+  }
+
+  render() {
+    return (
+      <div>
+        <h2>Loading Image Examples</h2>
+        <p>Image is successfully loaded.</p>
+        <div>
+          <Image id={this.state.loadedImageId} src={placeholder150x150} onLoad={() => { this.setState({ loadedImageId: 'loadedImage' }); }} />
```
Matt-Butler: Can you add onError to this image?

Matt-Butler:
Should we add an example 
- when an image fails to load then a fallback image is loaded?
- when an image loads successfully and a fallback image is not loaded?

mschile:
> Should we add an example
> when an image fails to load then a fallback image is loaded?
> when an image loads successfully and a fallback image is not loaded?

I'm creating terra-profile-image #429 that will demonstrate a similar concept except that the 'fallback' avatar image will be loaded by default and replaced when the image loads.

packages/terra-image/tests/nightwatch/ImageLoading.jsx
mschile: Added.

packages/terra-image/tests/nightwatch/ImageLoading.jsx
mschile: Added.

PR #477, opened 2017-06-19
Add i18n guide

### LINKED ISSUE #476: Internationalization guide upgrade
# Issue Description
Add Internationalization guide section to the root README.md and terra-i18n component READEME.md. Make consumer easier to set the config up.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

windse7en:
### Summary
Add Internationalization Guide section to root README.md. Make consumers easy to start with translations.

Fix #476 

In general, this Internationalization section should be the entry guide for any consumers who want to have translations in their application. Prefer to use existing doc resources to describe the each step. By this way, it should be easier to maintain.


windse7en:
I have synced with David Nguyen to see how helpful this guide will be. He's good with it.

README.md
```diff
@@ -66,6 +67,26 @@
 | Internet Explorer           | 10 & 11 |
 | Safari & Mobile Safari      | Current |
 
+## Internationalization(I18n)
+
+1. Please follow [Base Getting Started](packages/terra-base/README.md#getting-started) to install `Base`, and consume it with `locale` props.
+2. Install and config `terra-i18n-plugin`:
+    - Follow [terra-i18n-plugin Getting Started](packages/terra-i18n-plugin#getting-started) to install `terra-i18n-plugin` to aggregate translations in `node_modules`.
+    - Follow [terra-i18n-plugin Wiki Guide](https://github.com/cerner/terra-core/wiki/Terra-i18n-plugin-Guide) to config i18n plugin.
+3. Insatll and config `react-intl`
```
JakeLaCombe: *Install

dnguyenchevy: typo `3. Install`

dnguyenchevy:
I think it would be appropriate to update readme for [terra-base](https://github.com/cerner/terra-core/tree/master/packages/terra-base/docs) documentation at the same time.

packages/terra-i18n/README.md
```diff
@@ -12,6 +12,7 @@ The terra-i18n component provides the internationalization to the React componen
 
 ## Getting Started
 
+At most time, you don't need to install this package directly. Please follow [this internationalization section](../../README.md#internationalizationi18n) to set up. `Base` will install this package for you.
```
JakeLaCombe: I think this would be a better rephrasing of the sentence

    Note: If you are using `Base`, you do not need to install this component, as `Base` will include this module for you.

dnguyenchevy:
I think there's need to be two defining sections internationalization.

A section for, defines how to only internationalize terra components. Then a section for custom internationalization with consumers component using Base vs react-intl directly.

Current documentation makes it seem like we need all of these configurations with react-intl for terra components when in actuality consumers only need Base to encapsulate component with locale defined.

packages/terra-i18n/README.md
windse7en: We recommend to use `Base` component. Sometimes, consumers want to update add more translations at runtime or support different `locale` under the same `Base`. In those cases, they will want to use I18nProvider directly.

In here, I don't want to point out all those things. It may make consumers get confused. How do you think?

windse7en:
I try to keep it in one section for easy consuming. As you said, there are two cases. 
1. Consumers just want to use DatePicker or DemoBanner. They don't want to provide any translations at the application level.
2. Consumers want to provide their own translations.

My objective in here is recommending consumers follow the same way(assume if they have the locale set up now, they should use translations in future) and try to get familiar with translations strategy at first.

packages/terra-i18n/README.md
JakeLaCombe: Then we could rephrase it like so

    Note: It is recommend to use `Base` instead of using this module, as base will initialize your internationalization for you

If that doesn't work, then I think we are documenting this at the wrong level. We would probably need to document it as a higher Getting Started with Terra section as opposed to putting it inside this module's documentation.

JakeLaCombe:
From looking at Dave's comments, I don't think we should be documenting these concerns in the individual packages. Perhaps we can instead make use of our wiki space on this monorepo to talk about how to setup Internationalization?

dnguyenchevy:
> From looking at Dave's comments, I don't think we should be documenting these concerns in the individual packages.

Thinking about it now, @JakeLaCombe, internationalization for individual package should live within each component.

JakeLaCombe:
We found that approach to be limiting, as it would cause way too many files to be loaded at once. We wanted to ensure that all translations are loaded at once, and that the best work around was to create a webpack plugin that would look for all your translations files, then build out a single translation object that would be passed into the i18n wrapper.

README.md
```diff
@@ -66,6 +68,31 @@
...
+          extensions: ['.js', '.jsx'],
+          alias: {
+            'react-intl': path.resolve(__dirname, 'node_modules/react-intl'),
+          },
+        },
+        ```
+4. Based on your strategy to set locale by backend function and provide `customMessages` prop to `Base` by backend translations.
+5. Follow [react-intl wiki](https://github.com/yahoo/react-intl/wiki/API) to use `injectIntl`([pass translations to props](https://github.com/cerner/terra-core/wiki/terra-i18n-Guide#pass-translated-message-as-props)) or `FormattedMessage`([render translations](https://github.com/cerner/terra-core/wiki/terra-i18n-Guide#display-transalated-message-without-default-message-fallback)) to consume translations.
+
+### Packages Required I18n
```
JakeLaCombe: Packages Requiring*

README.md
```diff
@@ -66,6 +68,31 @@
...
+    - Add alias webpack config to avoid importing duplicate `react-intl`.
+        ```
+        resolve: {
+          extensions: ['.js', '.jsx'],
+          alias: {
+            'react-intl': path.resolve(__dirname, 'node_modules/react-intl'),
+          },
+        },
+        ```
+4. Based on your strategy to set locale by backend function and provide `customMessages` prop to `Base` by backend translations.
```
JakeLaCombe: I think the Based on your strategy part of the sentence is confusing. I would just say

    Provide values for locale and customMessages.

Perhaps we should mention that it's recommended to set the lang attribute on the html node as well?

dnguyenchevy:
>>Thinking about it now, @JakeLaCombe, should internationalization for individual package should live within each package?
>
> We found that approach to be limiting, as it would cause way too many files to be loaded at once. We wanted to ensure that all translations are loaded at once, and that the best work around was to create a webpack plugin that would look for all your translations files, then build out a single translation object that would be passed into the i18n wrapper.

Sorry, I should have given more context. I was talking about in regards to documentation. Each individual package documentation should contain, if applicable, how to internationalize the package.

README.md
windse7en: I'd like to see the `lang` attribute in some other doc specified for html tag config.

README.md
windse7en: Thanks, fix it.

README.md
```diff
@@ -66,6 +68,31 @@
 | Internet Explorer           | 10 & 11 |
 | Safari & Mobile Safari      | Current |
 
+## Internationalization(I18n)
```
Matt-Butler: Can we pull this information out into terra-i18n/docs so we can pull it into our doc site?

windse7en: As it's quite i18n setup guide, which includes installing packages and config `Base` component, I feel it should not be put into terra-i18n. It makes consumer like David, they need to consume terra-i18n directly. 

I'd like to see we will have some general section(Not specified for a particular component) in doc site to provide such information. Like in ant-design: https://ant.design/docs/react/i18n . Before that, I'd like to keep it in here.

ryanthemanuel: nit: Add a space between internationalization and (I18n)

README.md
windse7en: Fix it.

Issue #422, opened 2017-05-31
terra-date-picker calendar icon smaller than expected


viren140290:
# Issue Description

When trying to render terra-date-picker in our rails application, the calendar icon seems to be smaller than the expected size.
![terra-date-picker](https://cloud.githubusercontent.com/assets/1681774/26652694/6598c4b6-4616-11e7-9767-a45ea0cdaa56.png)

The only difference I could see was the calendar button seems to have height set to 24px in my app vs auto on the terra site example

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

## Expected Behavior

Button has the same height as the input box
## Current Behavior

Button doesn't have the same height as the input box

## Environment

* Component Version: 0.3.2
* Browser Name and Version: Chrome 58.0.3029.110 (64-bit)
* Operating System and version (desktop or mobile): OS X 10.11.6

Matt-Butler:
Hey @viren140290, is there an additional class adding a height of 24px to the button?

viren140290:
@Matt-Butler  the only classes I see on the button are terra-Button terra-Button--default terra-DatePicker-button . We have not added any custom classes or havent added any styles for the  Button element

viren140290:
The whole form is also wrapped with <Base> from terra-base

viren140290:
I dug a lil deeper and realized that in the box model the inner content for my app  is of size
14 * 14 vs on terra's site its 14 * 20

bjankord:
I'm gonna mark this as a bug, but we'll need to have a deeper look into this to figure out what the root cause of this is. Let's plan on syncing up tomorrow @viren140290.

benbcai:
@viren140290 The only [additional style](https://github.com/cerner/terra-core/blob/master/packages/terra-date-picker/src/DatePicker.scss#L31-L41) the date-picker adds to the button is to remove the start border radius and adjust the left margin a bit. Could you add just a [normal button](http://engineering.cerner.com/terra-core/#/site/button) with an icon to your app to see if you see the same thing? Just trying to see if this happens only with the calendar button or all buttons in your app.

Matt-Butler:
The difference between the two datepicker buttons is the padding:

![screen shot 2017-06-02 at 1 48 01 pm](https://cloud.githubusercontent.com/assets/8146159/26740176/2aa572ca-479a-11e7-9e09-74e47ae50905.png)
terra example is the padding on the right.

[...]

PR #481, opened 2017-06-20
Issue-422 - date input to consume form Input.

benbcai:
Updated date picker to use Form Input.
Fixed issue where the calendar button height is smaller than the input height when font-size is changed. 

font-size: 1em
screen shot 2017-06-20 at 9 18 21 am

font-size: 1.5em
screen shot 2017-06-20 at 9 21 28 am

Resolves https://github.com/cerner/terra-core/issues/422

Deployed site for verification: https://terra-core-deployed-pr-481.herokuapp.com/

@cerner/terra


emilyrohrbough:
Did the jest snapshots update with these changes?

benbcai:
No, the reason is because the date input is composed internal by react-datepicker (https://github.com/cerner/terra-core/blob/master/packages/terra-date-picker/src/DatePicker.jsx#L153) as the custom input so there is no reference to it in the jest snapshot.

mjhenkes:
@neilpfeiffer @mmalaker Does this issue need UX and Functional Review?

packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -38,7 +39,7 @@ class DatePickerInput extends React.Component {
   render() {
     return (
       (<div className="terra-DatePicker-customInput">
-        <input
+        <Input
           {...this.props.inputAttributes} // TODO: When forms is available, this.props.inputAttributes should be passed to the attrs props in the TextField component (attrs={this.props.inputAttributes}) instead of destructuring the inputAttributes prop here.
```
windse7en: May update to use props destructure in here to follow the convention.
```
const { inputAttributes, onClick, onKeyDonw ... } = this.props;
```

packages/terra-date-picker/tests/nightwatch/date-picker-spec.js
```diff
@@ -68,5 +67,18 @@ module.exports = {
     browser.expect.element('.terra-DatePicker-range > .terra-DatePicker').to.be.present;
   },
 
+  'Displays the calendar button with a height that matches the input ': (browser) => {
+    browser.url(`http://localhost:${browser.globals.webpackDevServerPort}/#/tests/date-picker-tests/default`);
+
+    let inputHeight;
+
+    browser.getElementSize('.terra-DatePicker-input', function getHeight(result) {
```
windse7en: Why don't use arrow callback in here and disable prefer-arrow-callback at top of the file?

packages/terra-date-picker/src/DateInput.jsx
benbcai: Updated. https://github.com/cerner/terra-core/pull/481/commits/f55037147df9d3f2b72601095c296500b22e71a6

packages/terra-date-picker/tests/nightwatch/date-picker-spec.js
benbcai: What I'm testing here is using [getElementSize](http://nightwatchjs.org/api#getElementSize) to get the height of the input and calendar button and expect that height values match. Once I have both height values, the way I found to compare two standalone values in nightwatch is using this.asset.equal() inside getElementSize. However, if I changed to use arrow-callback, the `this` keyword [no longer binds](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#No_binding_of_this) to the current scope and I can’t use assert.equal(). It is binded to the outer scope. This is the reason I couldn’t use arrow callback.

I couldn’t use getElementSize to get the height and then use assert.cssProperty or expect.css to compare the value because they return different height values for the same element. I think because of rounding, getElementSize returns 25 where as cssProperty returns 25.984375px.

After digging for more night watch APIs, there is actually getCssProperty that I can use to get the height, which returns a consistent value. In using getCssProperty, I no longer need to use this.assert.equal() to compare the values.
https://github.com/cerner/terra-core/pull/481/commits/f55037147df9d3f2b72601095c296500b22e71a6

windse7en: Cool. +1

mmalaker:
Functional verification completed. All date entry examples display and function correctly using form inputs.

neilpfeiffer:
Tested: PR [#481](https://github.com/cerner/terra-core/pull/481)

**Reviewed:**
- Follows pattern considerations documented in the current Cerner Design Standard: [Date-Pickers](https://design.cerner.com/styleguide/patterns/data-pickers/)
- Follows UX defined Terra design implementation for Date Picker.



**Pass** |
-----------|
UX Review|

Issue #486, opened 2017-06-20
Cleanup Control Docs and Props


JakeLaCombe:
# Issue Description
Control Props have a few minor concerns with them.

- The docs page has additional props that are not supported by the control (error, help)
- onChange should be a first class prop for the component

JakeLaCombe:
#487

bjankord:
Resolved in #487

PR #487, opened 2017-06-20
Fix Docs and First Class Props of Control

JakeLaCombe:
### Summary
Control Inputs have a few small issues with them

- The docs have incorrect props on them. They provide error and help examples, even though Control does not support those values.
- onChange should be a first class attribute that is passed into the input.


JakeLaCombe:
#486

PR #489, opened 2017-06-20
Add changelog for terra-grid

bjankord:
### Summary
Adding changelog for grid since its v1.0.0 release to now.

### Additional Details
We can use this review to decide what format we like for the CHANGELOG file. Based on this review, I'll add a change log to terra-mixins for v1.0.0+ releases and all the other components when we release them as v1.

@cerner/terra

### DIFF
--- a/packages/terra-grid/CHANGELOG.md
+++ b/packages/terra-grid/CHANGELOG.md
@@ -1,71 +1,71 @@
 Terra Grid Changelog
 ====================
 
-3.1.0 - 2017-06-20
+3.1.0 (June 20, 2017)
 ------------------
-### Enhancements
+### Dependency Updates
 * Bump dependency on terra-base to `^0.7.0`
 
-3.0.0 - 2017-06-13
+3.0.0 (June 13, 2017)
 ------------------
 ### Breaking Changes
 * Remove theming capability in grid component
 
-### Enhancements
+### Dependency Updates
 * Bump dependency on terra-base to `^0.6.0`
 * Bump dependency on terra-mixins to `^1.3.0`
 
-2.4.1 - 2017-06-07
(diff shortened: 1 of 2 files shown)


emilyrohrbough:
Was this auto generated or manual?

bjankord:
This was manually generated.

packages/terra-grid/CHANGELOG.md
```diff
@@ -0,0 +1,71 @@
+Terra Grid Changelog
+====================
+
+3.1.0 - 2017-06-20
```
emilyrohrbough: Total nit: I would vote towards formatting the date differently, maybe like React's changelog, which would use `(June 20, 2017)`

packages/terra-grid/CHANGELOG.md
```diff
@@ -0,0 +1,71 @@
...
+------------------
+### Enhancements
```
emilyrohrbough: Would it be helpful consumers to separate dependencies from enhancements by listing dependency changes under it's own section. Then a consumer could quickly see dependency changes verse enhancements in functionality?

windse7en: I'm OK with current CHANGLOG. For dependency, consumers can view the`package.json` directly to understand the required packages for the component. That's also easy to maintain.

bjankord: Yeah I like the idea of these sections:
* Bug Fix
* Breaking Changes
* Removals
* Enhancements

Sometimes the only change between a version are dependency updates. I feel like its good to note that in the changelog when we do that so consumers don't need to diff the code between packages to see what changed.
I'd propose calling this section "Dependency Updates"

Issue #485, opened 2017-06-20
Remove DateRange props


benbcai:
There is currently no desire to have a date range component in phase 1. The date range props currently are declared in the date picker component. There are questions on where these props should live. Since there is no need to have a date range component, the props should be removed for now so that they are not visible and accessible to consumers. We can continue to keep the date range component so that it is available when needed.

## Issue Type
- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [X] Other

PR #492, opened 2017-06-20
Issue-485 - Remove date range props.

benbcai:
Remove date range props and all references to it because there is no need for a date range in phase 1.

Resolves https://github.com/cerner/terra-core/issues/485

@cerner/terra


packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -52,15 +40,9 @@ const propTypes = {
    * An ISO 8601 string representation of the initial default date to show in the date input. This prop name is derived from react-datepicker but is analogous to defaultValue for a form input field.
    */
   selectedDate: PropTypes.string,
-  /**
-   * An ISO 8601 string representation of the start date for a date range.
-   */
-  startDate: PropTypes.string,
 };
 
 const defaultProps = {
```
benbcai: Defaults for other props are added in this PR (https://github.com/cerner/terra-core/pull/452/files#diff-06f07d3cca7012107347dc63be00c147R65)

emilyrohrbough: Looks like `endDate` and `startDate` defaults were added in the [PR](https://github.com/cerner/terra-core/pull/452/files#diff-06f07d3cca7012107347dc63be00c147R65) you listed.

emilyrohrbough:
Do we remove the `DateRange.jsx` file as well?

benbcai:
What the group agrees in the war room was to not expose any date range props/examples to consumers but still leave DateRange.jsx so that we have it we come back to use it again. It is there, just won't be used for now.

packages/terra-date-picker/src/DatePicker.jsx
benbcai: Yeah, there will be merge conflict when these are ready to be merged. I'll make sure that any reference to date range in that PR doesn't get merged down.

Issue #493, opened 2017-06-20
Button group onChange wrong index


cwalten:
# Issue Description
[This function](https://github.com/cerner/terra-core/blob/master/packages/terra-button-group/src/ButtonGroup.jsx#L82) assumes that setState is a synchronous function. However, this is not the case, and thus the onClick function passes the wrong selectedIndex to the onChange method.

## Issue Type
- [X] Bug

## Expected Behavior
The method should passes the new selectedIndex, not the old one.

## Current Behavior
The code as is passes the previously selected index. The visual updates fine as the state is updated, but because of the order of events, the onChange receives the wrong parameter.

## Environment

* Component Version: 0.1.0, but master is still wrong
* Browser Name and Version: All
* Operating System and version (desktop or mobile): desktop mac

Matt-Butler:
Resolved in #495

PR #495, opened 2017-06-21
Issue-493 Fix onChange not updating with correct index

madelineisabelle:
### Summary
#493 Fixing onChange receiving incorrect index because setState is not synchronous


Matt-Butler:
We should add a test to ensure this behavior is correct.

madelineisabelle:
@Matt-Butler that's fine, but is there an example somewhere of testing something similar?

PR #501, opened 2017-06-22
Adding terra-profile-image.

mschile:
### Summary
The terra-profile-image component displays an avatar image while the profile image is being loaded. If the profile image successfully loads, the avatar image is replaced with the profile image. Otherwise, the avatar image is left as is. All styling set on the profile image will be applied to the avatar image. #429


packages/terra-profile-image/package.json
```diff
@@ -0,0 +1,52 @@
+{
+  "name": "terra-profile-image",
+  "main": "lib/ProfileImage.js",
+  "version": "0.1.0",
```
Matt-Butler: When we do the initial release, the version will be bumped to 0.1.0. For now, please set this as 0.0.0

packages/terra-profile-image/docs/README.md
```diff
@@ -0,0 +1,21 @@
+# Terra Image
+
+The terra-profile-image component displays an avatar image while the profile image is being loaded. If the profile image successfully loads, the avatar image is replaced with the profile image. Otherwise, the avatar image is left as is. All styling set on the profile image will be applied to the avatar image.
+
+## Getting Started
+
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-image`
```
Matt-Butler: terra-profile-image

packages/terra-profile-image/docs/README.md
```diff
@@ -0,0 +1,21 @@
...
+  - `yarn add terra-image`
+
+## Usage
+
+```jsx
+import React from 'react';
+import ProfileImage from 'terra-profile-image';
+import Image from 'terra-image';
```
Matt-Butler: It seems weird that we need to import both image and profile image. Can profile image just be composed of image?

That way a consumer can just pass in the path for an image instead of creating an image component separately.

mschile: Yeah, we could definitely update this component to replicate the props from image (I assume we would want more than just src so styling could be applied).  I brought this question up here [1] and the consensus was to just pass in an image.

[1] https://github.com/cerner/terra-core/issues/429#issuecomment-308837639

packages/terra-profile-image/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "test:nightwatch-safari": "SPECTRE_TEST_SUITE=terra-image node ./node_modules/terra-toolkit/lib/scripts/nightwatch-non-parallel.js safari"
+  },
+  "devDependencies": {
+    "terra-toolkit": "^0.4.0"
+  },
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "terra-base": "^0.7.0",
+    "terra-mixins": "^1.3.0"
```
windse7en: Add `terra-image` version in here. If there is any version confliction, it will show in the console.

packages/terra-profile-image/src/ProfileImage.jsx
```diff
@@ -0,0 +1,69 @@
...
+  constructor(props) {
+    super(props);
+
+    this.state = { isLoading: true, isError: false };
+
+    this.updateImage(props.image.props);
+  }
+
+  componentWillReceiveProps(newProps) {
+    this.updateImage(newProps.image.props);
```
windse7en: Add props check to avoid extra render.
```
if (newProps == this.props)  {
  return ;
}
```

packages/terra-profile-image/src/ProfileImage.jsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  updateImage(imageProps) {
+    const avatarOverrideProps = { src: avatarImage, onLoad: undefined, onError: undefined };
```
windse7en: This `updateImage` has been called in `componentWillReceiveProps`, so it will be called every time if there is a props update to rerender the component. In that case, for each rerender, you will create a new object for hold the `avatorOverrideProps`. If we put such constant object outside of the Class, it will just create once. And pass the reference for each rerender to save the performance.

packages/terra-profile-image/src/ProfileImage.jsx
```diff
@@ -0,0 +1,69 @@
...
+    this.avatarImage = new TerraImage(Object.assign({}, imageProps, avatarOverrideProps));
+
+    const profileOverrideProps = {
+      onLoad: () => { this.handleOnLoad(imageProps.onLoad); },
```
windse7en: Is this safe to assume `imageProps` must provide the `onLoad` function?

packages/terra-profile-image/tests/nightwatch/ProfileImageDefault.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
+import Image from 'terra-image';
+
+import ProfileImage from '../../lib/ProfileImage';
+import placeholder150x150 from '../assets/150x150.jpg';
```
windse7en: Do we really want to have 150x150 in the variable name? As we have size modifiers like: tiny, small, medium, large, huge, could we rename the image file to `mediumPlaceHolder.jpg`?

mjhenkes:
@emilyrohrbough could you kick off a heroku build for this? It'd be nice to try the component out.

emilyrohrbough:
@mjhenkes Redeployed

packages/terra-profile-image/docs/README.md
mschile: @mjhenkes, @ryanthemanuel, @dkasper-was-taken, @bjankord, @windse7en, looking for additional feedback.  Thoughts?

mjhenkes: I think we should take the src, then the profile image can be opinionated about what styles it should expose from image.  For example maybe profile images should only be round etc.

packages/terra-profile-image/src/avatar.svg
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bjankord: Let's run this SVG through https://jakearchibald.github.io/svgomg/ 
It will trim some of the unneeded comments / code while maintaining the same icon rendering.

packages/terra-profile-image/src/ProfileImage.jsx
```diff
@@ -0,0 +1,74 @@
...
+  constructor(props) {
+    super(props);
+
+    this.state = { isLoading: true, isError: false };
+
+    this.updateImage(props.image.props);
+  }
+
+  componentWillReceiveProps(newProps) {
+    if (newProps === this.props) {
```
JakeLaCombe: I don't think this will work. newProps and this.props are different objects I believe, meaning this will always return false.

packages/terra-profile-image/src/avatar.svg
mschile: I ran the avatar.svg through svgomg using the default settings.

packages/terra-profile-image/docs/README.md
mschile: I went ahead and updated profile-image to accept all properties image currently accepts.  I am waiting to hear back from @neilpfeiffer to know whether or not profile-image should honor 'variant' and 'isFluid'.

packages/terra-profile-image/package.json
mschile: Reset to 0.0.0.

packages/terra-profile-image/docs/README.md
mschile: From @neilpfeiffer:

> Final answer - no. A profile image should always be a fixed size (usually square) and would never be used fill a space which is the intent of isFluid.  (checked with a few people just to validate and see if they though the same)

packages/terra-profile-image/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "terra-profile-image",
+  "main": "lib/ProfileImage.js",
+  "version": "0.0.0",
+  "description": "terra-profile-image",
```
emilyrohrbough: Add description

packages/terra-profile-image/package.json
```diff
@@ -0,0 +1,55 @@
...
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "scripts": {
+    "compile": "npm run compile:clean && npm run compile:build",
+    "compile:clean": "$(cd ..; npm bin)/rimraf lib",
+    "compile:build": "$(cd ..; npm bin)/babel src --out-dir lib --copy-files",
+    "lint": "npm run lint:js && npm run lint:scss",
+    "lint:js": "$(cd ..; npm bin)/eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
+    "lint:scss": "$(cd ..; npm bin)/stylelint src/**/*.scss",
+    "test": "npm run test:spec && npm run test:nightwatch-default",
```
emilyrohrbough: add ` "props-table": "props-table ./src/ProfileImage.jsx --out-dir ./docs/props-table",`

packages/terra-profile-image/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "test:spec": "$(cd ..; npm bin)/jest --config ../../jestconfig.json",
+    "test:all": "npm run test:nightwatch-default && npm run test:nightwatch-chrome && npm run test:nightwatch-firefox && npm run test:nightwatch-safari",
+    "test:nightwatch-default": "SPECTRE_TEST_SUITE=terra-image node ./node_modules/terra-toolkit/lib/scripts/nightwatch.js",
+    "test:nightwatch-chrome": "SPECTRE_TEST_SUITE=terra-image node ./node_modules/terra-toolkit/lib/scripts/nightwatch.js chrome",
+    "test:nightwatch-firefox": "SPECTRE_TEST_SUITE=terra-image node ./node_modules/terra-toolkit/lib/scripts/nightwatch.js firefox",
+    "test:nightwatch-safari": "SPECTRE_TEST_SUITE=terra-image node ./node_modules/terra-toolkit/lib/scripts/nightwatch-non-parallel.js safari"
+  },
+  "devDependencies": {
+    "terra-toolkit": "^0.4.0"
```
emilyrohrbough: add `"terra-props-table": "^0.8.0",`

packages/terra-profile-image/src/ProfileImage.jsx
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * Function to be executed when the profile image load errors.
+   */
+  onError: PropTypes.func,
+};
+
+const defaultProps = {
+  variant: 'default',
+  isFluid: false,
+  alt: ' ',
```
emilyrohrbough: Do we want to provide a default value when it is a required prop?

packages/terra-profile-image/package.json
mschile: What's the props-table command used for?

packages/terra-profile-image/src/ProfileImage.jsx
mschile: In the general sense, I think it's okay to mark it as required and provide a default value.  Basically it's just preventing someone from trying to set the value to null.  However, I'm not sure there is much value in defaulting the 'alt' prop to a blank space.  I copied this over from terra-image.  Does anyone know why 'alt' is required but defaulted to blank space?  Seems like we could just not require the field.  @aganapuram @bjankord @Matt-Butler

packages/terra-profile-image/package.json
mschile: Updated the description.

packages/terra-profile-image/tests/nightwatch/ProfileImageDefault.jsx
mschile: Renamed to exampleProfileImage.

mhemesath:
Can we update this branch to use CSS Modules before its merged? Its a relatively small and simple change.

mjhenkes:
@mhemesath We'll make the changes for adding CSS Modules. @mschile has had this PR out for 14 days and I'd rather not add additional burden on him as a contributor.

packages/terra-profile-image/src/ProfileImage.jsx
mschile: I went ahead and removed the 'alt' default value and made the prop optional.

mmalaker:
Functional verification completed. The profile image examples and test pages display and function as expected.

bjankord:
@neilpfeiffer This is ready for UX verification.

neilpfeiffer:
Profile Image:
 Tested: PR [#501](https://github.com/cerner/terra-core/pull/501)

**Reviewed:**
- Follows existing patterns and similar implementations for a Profile Image


**Pass** |
-----------|
UX Review|

ryanthemanuel:
Resolves #429

Issue #504, opened 2017-06-22
Form `for` attribute


Matt-Butler:
# Issue Description

We should update our form labels to include the `for` attribute. 

Also, our generic field component includes the `htmlFor` prop but the rest of our field components do not include that prop. Should this prop be added/required for our field components?
https://github.com/cerner/terra-core/blob/master/packages/terra-form/src/Field.jsx#L24

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

JakeLaCombe:
I think it would be better to create an id prop, and pass that through. Any other thoughts?

Matt-Butler:
resolved in #511

PR #511, opened 2017-06-26
Add inputId to various field components

JakeLaCombe:
### Summary
The base form component contains an htmlFor attribute for setting the label of a Field. However, there is not an easy way currently to interface field components (Text, Number, Textarea) with the htmlFor attribute.

To make this easier, I have attached an inputId prop to the field components that will set both the id of the input, and the htmlFor attribute of the label.

### DIFF
--- a/packages/terra-form/src/NumberField.jsx
+++ b/packages/terra-form/src/NumberField.jsx
@@ -113,7 +113,7 @@ const NumberField = (
     label={label}
     labelAttrs={labelAttrs}
     help={help}
-    htmlFor
+    htmlFor={inputId}
     isInline={isInline}
     required={required}
     {...customProps}
--- a/packages/terra-form/src/NumberField.jsx
+++ b/packages/terra-form/src/NumberField.jsx
@@ -22,6 +22,10 @@ const propTypes = {
    */
   inputAttrs: PropTypes.object,
   /**
+   * Id of the input. Also populates the 'htmlFor' prop of the field.
+   */
+  inputId: PropTypes.string,
+  /**
    * Whether the field is inline
    */
   isInline: PropTypes.bool,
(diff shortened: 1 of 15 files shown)


packages/terra-form/src/NumberField.jsx
```diff
@@ -107,6 +113,7 @@ const NumberField = (
     label={label}
     labelAttrs={labelAttrs}
     help={help}
+    htmlFor
```
benbcai: Should this be htmlFor={inputId}?

Missed setting id={inputId} for the Input component.

JakeLaCombe: You are right. I missed it in this pull request, thus the failing build. Added it here.

https://github.com/cerner/terra-core/pull/511/commits/41218bd352a23895f9d8866b73d23100126d0faa

packages/terra-form/src/TextField.jsx
```diff
@@ -101,13 +107,15 @@ const TextField = ({
...
     isInline={isInline}
+    htmlFor={inputId}
     required={required}
     {...customProps}
   >
     <Input
       maxLength={maxLength}
       minLength={minLength}
       name={name}
+      id={inputId}
```
benbcai: Looks like this input id is not being set in NumberField.

JakeLaCombe: Done. https://github.com/cerner/terra-core/pull/511/commits/41218bd352a23895f9d8866b73d23100126d0faa

JakeLaCombe: Looks like I was wrong still, and mixed myself up with another issue that occurred. Try this commit out. https://github.com/cerner/terra-core/pull/511/commits/6b284e7359b926fbaac5bf1a52b05a495b1c5aad

Matt-Butler:
https://github.com/cerner/terra-core/issues/504

PR #517, opened 2017-06-27
Testing heroku deployment

bjankord:
Testing heroku branch deployments.


bjankord:
Branching of master directly, this seems to deploy fine.

PR #519, opened 2017-06-27
Migrated badges to be CSS Modules

mhemesath:
### Summary
This change converts terra badges over to use CSS localized modules. Changes include:
* Localized CSS for badges
* Disabled SUITS CSS linting for localized Badge CSS
* Created stylelint rule for custom properties
* Renamed CSS class names to be simpler now they are localized
* Updated JSX
* Updated tests

### Deployment Link
https://terra-core-deployed-pr-519.herokuapp.com/static/#/site/badge


packages/terra-badge/src/Badge.scss
```diff
@@ -1,78 +1,80 @@
...
-  @include terra-badge-color($terra-badge-negative-color-scheme);
-}
-
-.terra-Badge--warning {
-  @include terra-badge-color($terra-badge-warning-color-scheme);
-}
-
-.terra-Badge--info {
-  @include terra-badge-color($terra-badge-info-color-scheme);
+:local {
```
mhemesath: It was brought up that if someone is rolling their own theme (and not compiling custom props server side) that we may want an option to not include the default theme. Is it worth creating a Sass variable that default includes the theme but can be configured to leave it out for those that want to roll their own and not have duplicate variable definitions

packages/terra-badge/src/_variables.scss
```diff
@@ -1,57 +1,80 @@
-$terra-badge-font-sizes: (
-  tiny: 0.5rem,
-  small: 0.8rem,
-  medium: 1rem,
-  large: 1.4rem,
-  huge: 1.7rem
-);
+:root {
```
mhemesath: It was brought up that if someone is rolling their own theme (and not compiling custom props server side) that we may want an option to not include the default theme. Is it worth creating a Sass variable that default includes the theme but can be configured to leave it out for those that want to roll their own and not have duplicate variable definitions

packages/terra-badge/tests/jest/__snapshots__/Badge.test.jsx.snap
```diff
@@ -1,20 +1,20 @@
 exports[`test should render a badge component in the order, icon and text with medium size 1`] = `
 <span
-  className="terra-Badge terra-Badge--medium terra-Badge--default">
+  className="badge medium default">
   <img
     alt="Test icon" />
   <span
-    className="terra-Badge-text">
+    className="text">
```
emilyrohrbough: Do we want to keep the same class naming structure as before? I could see class names like `text` become confusing as more components are used together.

mhemesath: We can change it, but its worth noting that is localized to the component so that although the test shows `text`, the actual HTML output will either be a hash or will be a source path location of the selctor (in dev).

mhemesath: You can view source of the badge here to see what I'm referring to https://terra-core-deployed-pr-519.herokuapp.com/static/#/site/badge

emilyrohrbough: Oh I see what you mean. I feel like what you have is totally fine then. Out of curiosity, will `-terra-badge-lib- alway be prepended? Is this something the CSS Modules does? 

> _-terra-badge-lib-___Badge__badge___8Kz6n _-terra-badge-lib-___Badge__small___9T8V4 _-terra-badge-lib-___Badge__positive___16N_U

packages/terra-badge/src/_variables.scss
bjankord: Yeah I think this is worth exploring.

.stylelintrc
```diff
@@ -3,6 +3,12 @@
...
     "max-nesting-depth": 3,
     "scss/at-mixin-pattern": "^(terra-)[a-z]+([a-z0-9-]+[a-z0-9]+)?$",
+    "custom-property-pattern": [
+      "^terra-(?:[A-Z][a-zA-Z0-9]+)*(?:-[a-z0-9][a-zA-Z0-9]+)*?(?:--[a-z0-9][a-zA-Z0-9-]+)*?$",
+      {
+        "message": "Custom property names should match the SUIT CSS naming convention"
+      }
+    ],
     "selector-class-pattern": [
       "^terra-(?:[A-Z][a-zA-Z0-9]+)*(?:-[a-z0-9][a-zA-Z0-9]+)*?(?:--[a-z0-9][a-zA-Z0-9-]+)*?$|^(?:terra-)?(?:u|is|has|js)(?:-[a-z0-9][a-zA-Z0-9-]*)+$",
```
Matt-Butler: If we no longer enforce suitcss, should this linting rule be removed?

packages/terra-site/webpack.config.js
```diff
@@ -45,6 +51,7 @@ module.exports = {
                     'iOS >= 8',
                   ],
                 }),
+                CustomProperties(),
               ];
```
Matt-Butler: Do we want to add the rtl plugin?

packages/terra-badge/tests/jest/__snapshots__/Badge.test.jsx.snap
bjankord: > Is this something the CSS Modules does?

I believe so, [this webpack config](https://github.com/cerner/terra-core/blob/4edb57182de142580595cd60f61d159cd140b751/packages/terra-site/webpack.config.js#L38) is used to format the generated class names.

This means the nightwatch tests won't be able to rely on specific classNames. We talked about using IDs in nightwatch test files to identify components and then switch to inspecting the actually CSS using http://nightwatchjs.org/api#expect-css or http://nightwatchjs.org/api#assert-cssProperty

The jest tests have access to the CSS classes before they run through css modules / css loader so we can reference class names like "text" or "primary" in the jest tests.

.stylelintrc
bjankord: Yeah this can be removed. New components should be written with CSS modules.

mhemesath: I don't think we remove it until all components are converted over IMO

packages/terra-site/webpack.config.js
bjankord: Lets add it for our config. In our documentation on webpack config, I wouldn't include it, but add a section on how to add it. This way we can continue to test LTR/RTL styles but consumers have the choice to add the RTL plugin in their own webpack config if they need it.

packages/terra-badge/src/Badge.jsx
```diff
@@ -1,8 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
-import 'terra-base/lib/baseStyles';
```
mhemesath: was this OK to remove?

bjankord: We'll want to keep this, this file acts as a CSS reset.

packages/terra-badge/src/_variables.scss
mhemesath: Looks like this is very easy to do with a PostCSS plugin https://github.com/cbracco/postcss-remove-root

Even if we don't want to use that plugin, the impl of that plugin is very short.

packages/terra-site/webpack.config.js
mhemesath: I recommend we add it, i dig into the code and the plugin returns if the `[dir]` attribute is detected https://github.com/vkalinichev/postcss-rtl/blob/master/src/index.js#L32

packages/terra-badge/tests/jest/__snapshots__/Badge.test.jsx.snap
mhemesath: > The jest tests have access to the CSS classes before they run through css modules

This is partially true, Jest actually mocks out the style module so that the class name is the property be accessed off the style object. For CSS modules, this coincidentally is the original class name.

packages/terra-badge/src/Badge.jsx
```diff
@@ -46,20 +46,16 @@ const Badge = ({
...
-  ...additionalProps
+  ...customProps
 }) => {
-  const classes = classNames([
-    'terra-Badge',
-    { [`terra-Badge--${size}`]: size },
-    { [`terra-Badge--${intent}`]: intent },
-    additionalProps.className,
-  ]);
+  const badgeClassNames = classNames(styles.badge,
```
JakeLaCombe: Is there a new linting rule that changed this? I kinda like how each property was aligned and on it's own separate line.

PR #532, opened 2017-06-28
Update status img examples

### LINKED ISSUE #526: Status example update to relative path
# Issue Description

Update the status examples to use a relative path for images.

```
Mixed Content: The page at 'https://terra-ui.herokuapp.com/components/core/status' was loaded over HTTPS, but requested an insecure image 'http://placehold.it/350x150'. This content should also be served over HTTPS.
```

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

Update this:
https://github.com/cerner/terra-core/blob/master/packages/terra-site/src/examples/status/StatusArrange.jsx#L6

To something like this:
https://github.com/cerner/terra-core/blob/master/packages/terra-image/tests/nightwatch/ImageDefault.jsx#L4

bjankord:
### Summary
Resolves #526

### DIFF
--- a/packages/terra-site/src/examples/status/StatusArrange.jsx
+++ b/packages/terra-site/src/examples/status/StatusArrange.jsx
@@ -2,8 +2,9 @@
 import React from 'react';
 import Arrange from 'terra-arrange';
 import Status from 'terra-status';
+import placeholderPic150x150 from '../../assets/150x150.jpg';
 
-const image = <img style={{ display: 'block' }} height="150" width="300" src="http://placehold.it/350x150" alt="placeholder" />;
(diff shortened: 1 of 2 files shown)


packages/terra-site/src/examples/status/StatusArrange.jsx
```diff
@@ -2,8 +2,9 @@
 import React from 'react';
 import Arrange from 'terra-arrange';
 import Status from 'terra-status';
+import placeholderPic150x150 from '../../assets/150x150.jpg';
```
emilyrohrbough: Should we update from using `` to our `Image` component while we are updating the examples?

Issue #525, opened 2017-06-28
Break examples out into their own files


Matt-Butler:
# Issue Description

The datepicker and time input examples are all in the same file. Because of this, we need to update terra-ui.com whenever we update these examples. We should break these examples into their own files so the updates occur automatically. 

- https://github.com/cerner/terra-core/blob/master/packages/terra-site/src/examples/date-picker/Index.jsx
- https://github.com/cerner/terra-core/blob/master/packages/terra-site/src/examples/time-input/Index.jsx

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

bjankord:
Resolved in #537

PR #537, opened 2017-06-29
Breakout DatePickers and TimeInput Examples

emilyrohrbough:
### Summary
Address Issue #525


packages/terra-site/src/examples/date-picker/DatePickerOnChange.jsx
```diff
@@ -1,15 +1,30 @@
...
+  }
+
+  handleDateChange(event, date) {
+    this.setState({ date });
+  }
+
+  render() {
+    return (
+      <div>
+        <div id="selected-index">
```
JakeLaCombe: Is there a need for the extra div wrappers? I maybe missing it, but I don't see the selected-index id used in other places.

emilyrohrbough: No there isn't really isn't a need. I used this format in the single select list nightwatch example to test the onChange occurred correctly. Must have missed removing them :)

PR #547, opened 2017-07-04
Make base variables into custom properties

### LINKED ISSUE #574: Convert Base component to CSS modules
# Issue Description

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior

## Current Behavior

mhemesath:
### Summary
Made font size, font family and base color into custom properties.

Resolves #574

### DIFF
--- a/packages/terra-base/src/Base.scss
+++ b/packages/terra-base/src/Base.scss
@@ -4,16 +4,16 @@
 // More info: https://css-tricks.com/box-sizing/#article-header-id-6
 html {
   box-sizing: border-box;
-  font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
-  font-size: $terra-base-font-size;
+  font-family: var(--terra-Base-fontFamily);
+  font-size: var(--terra-Base-fontSize);
   height: 100%;
   margin: 0;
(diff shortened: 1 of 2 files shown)


packages/terra-base/src/Base.scss
```diff
@@ -4,16 +4,16 @@
 // More info: https://css-tricks.com/box-sizing/#article-header-id-6
```
Matt-Butler: When we migrate to css modules, do we want to convert these from scss to css?

mhemesath: We can could flip this file to CSS, there is no reason it needs to be scss at this point.

mhemesath:
One thing I'm not sure of is how this component will integrate into an MPages environment (once not in iframes). I think we should  try to identify strategies for mitigating the global html/body properties and push all others to a wrapper div we can scope to Terra components. Thoughts?

packages/terra-base/src/Base.scss
Matt-Butler: +1 I added this to our discussion on Friday. I don't want this to block your review.

PR #594, opened 2017-07-10
Update stylelint dependencies

bjankord:
### Summary
This _should_ help reduce the deprecation warnings we see in travis CI when running stylelint.

@cerner/terra


bjankord:
CI build looks like it might be failing on a timeout issue in the status component. Taking a look now.

Issue #538, opened 2017-06-29
Allow for assignment of SelectField option values


madelineisabelle:
# Issue Description
SelectFormField only lets you set the display of a select field, which will be inconvenient with translations. Adding the ability to set both the display and value of each option in a SelectField would fix this.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The SelectField should let the consumer set both the value and display for an option.

## Current Behavior
The SelectField only takes in a display for each option and uses that as the option's value.

emilyrohrbough:
Released in terra-forms v1.2.0

PR #595, opened 2017-07-11
Allow setting of value for option in Select, SelectField

doogthemushroom:
### Summary
This change allows the user to pass an object with a key/value pair to set the value and display of options in a Select or Select Field. Addresses Issue #538


packages/terra-form/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ Unreleased
 ### Added
 * Add missing props-table
 
+### Breaking Changes
+
+* Select and SelectField no longer accepts a list of strings, but a list of objects to set their value/display
```
bjankord: Is it worth keeping this passive by leaving the `choices` prop as-is and adding a `options` prop for this functionality. cc/ @mjhenkes @mhemesath

packages/terra-form/src/Select.jsx
```diff
@@ -62,7 +62,7 @@ const Select = ({
       defaultValue={defaultValue}
       className={classNames('terra-Form-select', additionalSelectFieldProps.className)}
     >
-      {choices.map(choice => <option key={choice.toString()} value={choice}>{choice}</option>)}
+      {choices.map(choice => <option key={choice.value} value={choice.value}>{choice.display}</option>)}
```
emilyrohrbough: Out of curiosity, how does this change affect the onChange callback? Would we need to add a handler function to return `choice.value` onChange?

packages/terra-form/CHANGELOG.md
mjhenkes: Given that we plan to enhance this API yet again to offer more functionality that may be non passive. I'd vote for the time being keep the API passive and cause the choices prop to warn that it will be removed in a future release.

packages/terra-form/src/Select.jsx
JakeLaCombe: I'm curious too. From looking here, it looks like you need both a value and onChange for a controlled select box. https://facebook.github.io/react/docs/forms.html.

packages/terra-form/src/Select.jsx
```diff
@@ -7,9 +7,18 @@ import './Select.scss';
...
    */
-  choices: PropTypes.array.isRequired,
+  choices: PropTypes.arrayOf(PropTypes.shape({
+    value: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.number,
+    ]).isRequired,
+    display: PropTypes.oneOfType([
+      PropTypes.string,
+      PropTypes.number,
```
tbiethman: Can we lock down the display type to just strings? Or do numbers make things easier in some cases?

madelineisabelle: @tbiethman We do have a case where a select box will display number options and the user will pick a number. Leaving this will save us having to convert to a string everytime, but it is not a big deal if we have to do that conversion.

mjhenkes: I'd vote for string only.

bjankord: I'm fine with oneOfType of string or number but I'll go with whatever the team consensus is.

packages/terra-form/CHANGELOG.md
```diff
@@ -20,6 +20,10 @@ Unreleased
 ### Removed
 * Removed nightwatch.config file
 
+### Breaking Changes
+
+* Select and SelectField no longer accepts a list of strings, but a list of objects to set their value/display
```
windse7en: `value/display` a little misunderstanding. It seems value or display, but it's the value and display value pair. Could we update it and make it clear?

doogthemushroom:
Apologies for leaving this out to dry everyone, I'll be getting around to this PR today.

packages/terra-form/src/Select.jsx
```diff
@@ -62,7 +74,7 @@ const Select = ({
       defaultValue={defaultValue}
       className={classNames('terra-Form-select', additionalSelectFieldProps.className)}
     >
-      {choices.map(choice => <option key={choice.toString()} value={choice}>{choice}</option>)}
```
windse7en: 1. To guarantee the unique key for each option, how about using `choice.value + choice.display` instead of just value. Is there probably some special case that same value for different displays? I get an example, as:
```
<option value="----">----
<option value="----">All countries
```
2. Follow the facebook example: https://facebook.github.io/react/docs/lists-and-keys.html#keys, the value of key must be string. If value is number, we'd better to use string interpolation to convert it to string.

packages/terra-form/CHANGELOG.md
doogthemushroom: Updated to have an `options` prop, with `choices` listed as a deprecated prop that logs a warning message if used.

packages/terra-form/src/Select.jsx
doogthemushroom: I've changed them to just be strings for simplicity.

packages/terra-form/CHANGELOG.md
doogthemushroom: Clarified to value and display.

packages/terra-form/src/Select.jsx
doogthemushroom: Updated to be a combination of value and display parameters, converted to a string.

packages/terra-form/src/Select.jsx
doogthemushroom: Looking into this at the moment, my initial understanding of controlled forms was not up to snuff when I first wrote this.

doogthemushroom: Ok yep I should definitely have a value prop for controlled forms. They should work like all the other inputs now, pass an onchange and value prop.

doogthemushroom:
I'm going to add some new controlled select tests but this should be close to final changes for this PR.

packages/terra-form/CHANGELOG.md
windse7en: +1

packages/terra-form/src/Select.jsx
windse7en: +1, Nice work.

packages/terra-form/tests/nightwatch/select/PopulatedSelectControlled.jsx
```diff
@@ -0,0 +1,40 @@
...
+        <Select
+          name="foo"
+          options={[{ value: 'puppies', display: 'Puppies' },
+                    { value: 'kittens', display: 'Kittens' },
+                    { value: 'micros', display: 'Microprocessors' },
+                    { value: 'snaps', display: 'Snappers' }]}
+          onChange={this.updateName}
+          value={this.state.animal}
+        />
+        <button type="submit" />
```
emilyrohrbough: Give button a label. Also the follow error occurs when the button is clicked:

> warning.js:35 Warning: Failed form propType: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`. Check the render method of `Select`.

packages/terra-site/src/examples/form/examples/Select.jsx
```diff
@@ -7,9 +7,9 @@ import Select from 'terra-form/lib/Select';
 const SelectExamples = () => (
   <form>
     <Select
-      choices={['Puppies', 'Kittens', 'Snappers', 'Bumblers', 'Joeys', 'Microprocessors']}
+      choices={['wow', 'this', 'that']}
```
emilyrohrbough: options? This could be what is causing the forms page to fails to load in the heroku instance.

packages/terra-form/tests/nightwatch/select/PopulatedSelectControlled.jsx
emilyrohrbough: Actually, looks like you need to update your onChange function to be this.updateAnimal

packages/terra-form/tests/nightwatch/select/PopulatedSelectControlled.jsx
```diff
@@ -0,0 +1,40 @@
...
+
+  updateAnimal(e) {
+    this.setState({
+      animal: e.target.value,
+    });
+  }
+
+  render() {
+    return (
+      <form>
```
emilyrohrbough: Add a div to display the value of the submitted option.

packages/terra-form/docs/Select.md
```diff
@@ -9,7 +9,9 @@ import React from 'react';
 import Select from 'terra-form/lib/Select';
 
 <Select
-  choices={['moo', 'boo', 'bar']}
+  choices={[ { value: 'm', display: 'moo' },
```
dnguyenchevy: This should be updated to `options={[ ...`, as well documentation below in SelectField.md.

packages/terra-form/tests/nightwatch/select/PopulatedSelectControlled.jsx
doogthemushroom: Yep I've resolved this now

packages/terra-site/src/examples/form/examples/Select.jsx
doogthemushroom: Ah I changed the initialized value to be null instead of undefined so it was skipping my choices/options check. Both props should work to populate the Select so that is backwards compatible. I've fixed it now but I'll re-add tests with the choices to prop to ensure it's working.

packages/terra-form/tests/nightwatch/select/PopulatedSelectControlled.jsx
doogthemushroom: Added.

packages/terra-form/docs/Select.md
doogthemushroom: Ah thanks, have it fixed now.

packages/terra-form/tests/jest/Select.test.jsx
```diff
@@ -26,13 +26,13 @@ it('should render a Select when all the possible props are passed into it', () =
 });
 
 it('should render as uncontrolled when just the required fields and a default value is passed into the Select', () => {
-  const select = <Select choices={['moo', 'boo']} defaultValue="boo" />;
+  const select = <Select options={[{ value: 'm', display: 'moo' }, { value: 'b', display: 'boo' }]} defaultValue="boo" />;
```
emilyrohrbough: Update defaultValue to 'b'

packages/terra-form/tests/jest/SelectField.test.jsx
```diff
@@ -26,13 +27,13 @@ it('should render a SelectField when all the possible props are passed into it',
 });
 
 it('should render as uncontrolled when just the required fields and a default value is passed into the SelectField', () => {
-  const selectField = <SelectField choices={['moo', 'boo']} defaultValue="boo" />;
+  const selectField = <SelectField options={[{ value: 'm', display: 'moo' }, { value: 'b', display: 'boo' }]} defaultValue="boo" />;
```
emilyrohrbough: Update defaultValue to 'b'

packages/terra-form/tests/jest/Select.test.jsx
doogthemushroom: Good catch, updated.

packages/terra-form/tests/jest/SelectField.test.jsx
doogthemushroom: Good catch, updated.

packages/terra-form/src/Select.jsx
```diff
@@ -56,16 +78,28 @@ const Select = ({
     additionalSelectProps['aria-required'] = 'true';
   }
 
+  let finalOptions;
+  if (choices && options === null) {
+    // Build out key & value array for deprecated choices prop
+    finalOptions = choices.map(choice => ({ value: choice, display: choice }));
+    /* eslint-disable no-console */
+    console.warn('The choices prop for the Terra Form Select component is deprecated and will be removed in a later release. Please use the options prop instead https://terra-ui.herokuapp.com/components/core/form#select.');
```
bjankord: +1

PR #606, opened 2017-07-13
terra-popup

dkasper-was-taken:
### Summary
Terra Popup, new feature, a tether enabled implementation.

Addressed Issue: #374 

A clean pr from the previous #444 

### Visual Design
 https://wiki.ucern.com/x/EeSZY

### Deployment Link
https://terra-core-deployed-pr-606.herokuapp.com/static/#/site/popup


packages/terra-popup/docs/README.md
```diff
@@ -0,0 +1,109 @@
...
+    );
+  }
+}
+
+export default PopupExampleComponent;
+```
+
+### Implementation Note:
+
+As stateless react elements do not return a ref, but may be desired targets for a popup to atttch to, ,additional code is required to retreive the elements ref. If a ref callback has not been plumbed through in the elements implementation there are a few options:
```
ryanthemanuel: attach not atttch.  Do something with ", ,"

packages/terra-popup/docs/README.md
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  render() {
+    <div>
+      return (
+        <Button id="my-popup-button" />
+        <Popup
+          targetRef={this.getButtonNode}
+```
+#### Wrap the element in span or div. The weakness of this option is that you need to ensure the wrapping div fits the content, while still styled correct on the page.
```
ryanthemanuel: nit: styled "correctly"

packages/terra-popup/src/Popup.jsx
```diff
@@ -0,0 +1,274 @@
...
+
+const HEIGHT_KEYS = ['40', '80', '120', '160', '240', '320', '400', '480', '560', '640', '720', '800', '880'];
+const HEIGHT_VALUES = [40, 80, 120, 160, 240, 320, 400, 480, 560, 640, 720, 800, 880];
+
+const WIDTH_KEYS = ['160', '240', '320', '640', '960', '1280', '1760'];
+const WIDTH_VALUES = [160, 240, 320, 640, 960, 1280, 1760];
+
+const CREATE_MAPPING = ((keys, values) => {
+  const combinations = {};
+  for (let i = 0; i < keys.length; i += 1) {
```
ryanthemanuel: You could make this a map:

```
keys.map((element, index) => (
  combinations[element] = values[index]
));
```

I don't care too much one way or another

packages/terra-popup/docs/README.md
emilyrohrbough: nit: extra space & comma before additional code is required

packages/terra-popup/src/Popup.jsx
```diff
@@ -0,0 +1,274 @@
...
+  /**
+   * Callback function indicating a close condition was met, should be combined with isOpen for state management.
+   */
+  onRequestClose: PropTypes.func.isRequired,
+  /**
+   * Target element for the popup to anchor to.
+   */
+  targetRef: PropTypes.func.isRequired,
+  /**
+   * Bounding container for the popup, will use window if no value provided.
```
emilyrohrbough: Matt once mentioned we want to avoid binding to window. @mjhenkes is the desired default?

mjhenkes: In this case window is appropriate. Popup, as a construct that floats above other content, should try to take up as much space possible. It's very similar to modal (which is also bound to window). Content should be bound to it's frame.

packages/terra-popup/src/Popup.jsx
```diff
@@ -0,0 +1,274 @@
...
+          targetRef={targetRef}
+          targetAttachment={mirrorAttachment(bidiContentAttachment)}
+        />
+      );
+
+      portalContent = this.createPortalContent(tetherCotent, !allowScrolling);
+    }
+
+    return (
+      <Portal {...customProps} isOpened={isOpen}>
```
emilyrohrbough: Does adding customProps to the Portal potential cause the same issue as https://github.com/cerner/terra-core/issues/625?

packages/terra-popup/src/Popup.scss
```diff
@@ -0,0 +1,5 @@
+@import './variables';
+
```
emilyrohrbough: convert to css modules?

dkasper-was-taken: This one can't be, unfortunately as it's the class prefix passed to the tether component, and we have no control over tethers implementation.

packages/terra-popup/src/_variables.scss
```diff
@@ -0,0 +1,33 @@
+:root {
```
emilyrohrbough: Variables naming format needs updated to `--terra-popup-arrow-size` format.

packages/terra-popup/src/Popup.jsx
dkasper-was-taken: It would, i was using the same pattern. Though the change in implementation isn't quite as simple.

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -0,0 +1,289 @@
...
+
+class PopupContent extends React.Component {
+  static getContentStyle(arrow, arrowPosition, height, maxHeight, width, maxWidth) {
+    const validHeight = maxHeight <= 0 || height <= maxHeight ? height : maxHeight;
+    const validWidth = maxWidth <= 0 || width <= maxWidth ? width : maxWidth;
+    return { height: `${validHeight.toString()}px`, width: `${validWidth.toString()}px` };
+  }
+
+  static addPopupHeader(children, onRequestClose) {
+    const icon = <IconClose height="20" width="20" />;
```
emilyrohrbough: Would there ever be a reason this height and width should scale or resize?

packages/terra-popup/src/_PopupUtils.js
```diff
@@ -0,0 +1,121 @@
...
+  let offset = (targetBounds.top - contentBounds.top) + arrowOffset + (targetBounds.height / 2);
+  if (offset < (2 * arrowOffset) + cornerOffset) {
+    offset = (2 * arrowOffset) + cornerOffset;
+  } else if (offset > contentBounds.height - cornerOffset) {
+    offset = contentBounds.height - cornerOffset;
+  }
+  return (`${offset}px`);
+};
+
+export { parseStringPair, isVerticalAttachment, primaryArrowPosition, primaryMarginStyle, switchAttachmentToRTL, mirrorAttachment, getContentOffset, arrowPositionFromBounds, leftOffset, topOffset };
```
emilyrohrbough: Would it be cleaner to add these functions to a utils object and just export the object?

packages/terra-popup/src/_TetherComponent.jsx
```diff
@@ -0,0 +1,251 @@
...
+   * Rule set to pass to tether, contraining the content to by setting 'to', 'attachment', and/or 'pin'.
+   * Also allows for class aliasing or 'outOfBoundsClass' and 'pinnedClass'.
+   */
+  constraints: PropTypes.array,
+  /**
+   * The content to be tethered.
+   */
+  content: PropTypes.element,
+  /**
+   * String pair of top, middle, bottom, and left, center, right.
```
emilyrohrbough: nit: specify the pair must be separated by a space

packages/terra-popup/src/_TetherComponent.jsx
```diff
@@ -0,0 +1,251 @@
...
+   */
+  disableOnPosition: PropTypes.bool,
+  /**
+   * Should element be tethered to the page.
+   */
+  isEnabled: PropTypes.bool,
+  /**
+   * The ability to disable "gpu" and "moveElement" optimizations.
+   */
+  optimizations: PropTypes.object,
```
emilyrohrbough: It might be helpful to include a reference somewhere on how these these tether optimizations might look and where to learn about them.

packages/terra-popup/src/_TetherComponent.jsx
```diff
@@ -0,0 +1,251 @@
...
+
+  handleOnUpdate(event) {
+    if (this.props.onUpdate) {
+      const bounds = this.getNodeBounds();
+      this.props.onUpdate(event, bounds.targetBounds, bounds.contentBounds);
+    }
+  }
+
+  handleOnRepositioned(event) {
+    if (this.props.disableOnPosition) {
```
emilyrohrbough: does this need to be an if-else?

packages/terra-popup/src/_TetherComponent.jsx
```diff
@@ -0,0 +1,251 @@
...
+      target: targetRef(),
+      ...customProps,
+    };
+
+    // Aliased parameters
+    if (contentOffset) {
+      tetherOptions.offset = contentOffset;
+    }
+    if (isEnabled !== undefined) {
+      tetherOptions.enabled = true;
```
emilyrohrbough: Do we want `tetherOptions.enabled = isEnabled`?

packages/terra-popup/src/_TetherComponent.jsx
dkasper-was-taken: I have a link in the readme to the tether doc site, but I can add one here too.

packages/terra-popup/src/_TetherComponent.jsx
dkasper-was-taken: There aren't mutually exclusive, if both are present, we want to call both.

packages/terra-popup/tests/nightwatch/components/SizeLargePopup.jsx
```diff
@@ -0,0 +1,65 @@
...
+            boundingRef={this.getParentNode}
+            classNameArrow="test-arrow"
+            classNameContent="test-content"
+            contentHeight="240"
+            contentWidth="960"
+            isOpen={this.state.open}
+            onRequestClose={this.handleRequestClose}
+            targetRef={this.getButtonNode}
+          >
+            <p>This popup is 100x by 100x.</p>
```
emilyrohrbough: Nit: popup is 240 x 960

packages/terra-popup/src/_variables.scss
dkasper-was-taken: Updated.

packages/terra-popup/src/_PopupContent.jsx
dkasper-was-taken: Likely, though this is still waiting on ux feedback.

packages/terra-popup/tests/nightwatch/popup-spec.js
```diff
@@ -0,0 +1,233 @@
...
+  },
+
+  afterEach: (browser, done) => {
+    screenshot(browser, 'terra-Popup', done);
+  },
+
+  'isOpen prop - RESIZE': (browser) => {
+    browser
+      .url(`http://localhost:${browser.globals.webpackDevServerPort}/#/tests/popup-tests/default`)
+      .waitForElementPresent('#default-button', 1000)
```
emilyrohrbough: While going through the css module conversions, we have been removing waitForElementPresent because its causing our nightwatch tests to take significantly longer waiting a second for each assertion.

dkasper-was-taken: What have we been replacing them with?

packages/terra-popup/src/Popup.scss
emilyrohrbough: Might be good to add a comment about that so people know in the future?

packages/terra-popup/tests/nightwatch/popup-spec.js
emilyrohrbough: We have been removing them. @bjankord mentioned a screenshot could be included if you wanted to capture the browser changes on events like click button.

packages/terra-popup/src/Popup.scss
dkasper-was-taken: Updated.

packages/terra-popup/src/_TetherComponent.jsx
dkasper-was-taken: Updated.

packages/terra-popup/src/Popup.jsx
dkasper-was-taken: I'll probably leave it for now. While a little cleaner code, not sure I like using the map without using it's yielded value.

packages/terra-popup/tests/nightwatch/popup-spec.js
dkasper-was-taken: So there'd no longer be tests, just taking a screenshot?

emilyrohrbough: There would still be tests to ensure the element is present on the page like [here](https://github.com/cerner/terra-core/pull/623/files#diff-54e3ca03d43643bd79a2bae13b4f9cb3R17) if no events are occurring. If we do have events occurring, like this your test here with the button click, we have something like [this](https://github.com/cerner/terra-clinical/pull/151/files#diff-a71d5e5423e814bff432415bc9f88fabR107). So each `.waitForElementPresent` in your tests could be changed to `.assert.elementPresent`.

dkasper-was-taken: Unfortunately the layout occurs after mounting, so element present always fails. Even if you launch the popup by default it fails.

packages/terra-popup/src/Popup.jsx
```diff
@@ -0,0 +1,274 @@
...
+import Portal from 'react-portal';
+import PopupContent from './_PopupContent';
+import PopupArrow from './_PopupArrow';
+import PopupOverlay from './_PopupOverlay';
+import TetherComponent from './_TetherComponent';
+import PopupUtils from './_PopupUtils';
+import './Popup.scss';
+
+const HEIGHT_KEYS = ['40', '80', '120', '160', '240', '320', '400', '480', '560', '640', '720', '800', '880'];
+const HEIGHT_VALUES = [40, 80, 120, 160, 240, 320, 400, 480, 560, 640, 720, 800, 880];
```
Matt-Butler: If the key and the value match up, you should only need to define it once. 

You could even do something like this:
```
const HEIGHT_VALUES = [40, 80, 120, 160, 240, 320, 400, 480, 560, 640, 720, 800, 880];

const HEIGHT_KEYS = HEIGHT_VALUES.map((value) => value.toString());
```

Matt-Butler: However, I would probably just make height & width their own files as
```
const height = {
'40': 40, 
'80': 80,
}
export default height;
```

In this file, import height

packages/terra-popup/docs/README.md
```diff
@@ -0,0 +1,121 @@
...
+        </button>
+      </div>
+    );
+  }
+}
+
+export default PopupExampleComponent;
+```
+
+### Implementation Notes:
```
bjankord: I like this level of detail, this will be really helpful for consumers

packages/terra-popup/src/_PopupArrow.jsx
```diff
@@ -0,0 +1,50 @@
...
+  refCallback: PropTypes.func,
+};
+
+const PopupArrow = ({
+    refCallback,
+    ...customProps
+  }) => <div {...customProps} className={cx(['popupArrow', customProps.className])} ref={refCallback} />;
+
+PopupArrow.propTypes = propTypes;
+PopupArrow.positionAttrs = ARROW_ATTRS;
```
bjankord: nit: Add `positionAttrs`, `mirroredPositionAttrs`, and `arrowSize` to a PopupArrow.Opts object. This lines up with what we are doing in [overlay](https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/Overlay.jsx#L174-L178) and [alert](https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/Overlay.jsx#L174-L178).

packages/terra-popup/src/_PopupContent.jsx
dkasper-was-taken: Talking with Neil, for now we are going to go with the px value instead of scaling, but that may change down the road as the design for the modal and detail view dialogs are firmed up.

mmalaker:
Functional verification completed. All popup examples and test pages display and function correctly.

bjankord:
I'd really like to find a fix for the shifting that occurs when clicking on popups as soon as possible.
popup-shift

bjankord:
@neilpfeiffer This is ready for UX review

packages/terra-popup/src/PopupContent.scss
```diff
@@ -0,0 +1,71 @@
...
+      justify-content: flex-end;
+      width: 100%;
+    }
+
+    .close {
+      background: var(--terra-popup-content-close-background);
+      border: var(--terra-popup-content-close-border);
+      color: var(--terra-popup-content-close-color);
+      cursor: pointer;
+      font-size: var(--terra-popup-content-close-fontSize);
```
jmsv6d: * font-size

packages/terra-popup/src/PopupContent.scss
```diff
@@ -0,0 +1,71 @@
...
+      font-weight: var(--terra-popup-content-close-font-weight);
+      line-height: var(--terra-popup-content-close-line-height);
+      margin-left: var(--terra-popup-content-close-margin-left);
+      margin-right: var(--terra-popup-content-close-margin-right);
+      padding: var(--terra-popup-content-close-padding);
+
+      &:hover {
+        color: var(--terra-popup-content-close-hoverColor);
```
jmsv6d: * hover-color

neilpfeiffer:
Tested: PR [#606](https://github.com/cerner/terra-core/pull/606)

**Reviewed:**
- Follows pattern considerations documented in the current Cerner Design Standard: [Popup](https://design.cerner.com/styleguide/healthe-intent/tooltips/)
- Follows UX defined design implementation for [Terra Popups](https://wiki.ucern.com/x/EeSZY) 



**Pass** |
-----------|
UX Review|

Issue #608, opened 2017-07-13
terra-i18n-plugin runs mkdirp asynchronously


brhoades:
# Issue Description
When aggregating translations, terra-i18n-plugin runs mkdirp asynchronously and then follows it with a synchronous write call. This doesn't cause an issue locally, but on our slow Jenkins node it arbitrarily fails when writing:

```
11:33:51 > react-clipboards@1.6.2 build:webpack /opt/jenkins_virtualenvs/06c35fc2/virtualenvs/d41d8cd9/src/trunk/iqh/clipboard/node_modules/react-clipboards
11:33:51 > NODE_ENV=production webpack --config config/prod.webpack.config.js --progress --colors --display-error-details
11:33:51 
11:33:51 fs.js:584
11:33:51   return binding.open(pathModule._makeLong(path), stringToFlags(flags), mode);
11:33:51                  ^
11:33:51 
11:33:51 Error: ENOENT: no such file or directory, open '/opt/jenkins_virtualenvs/06c35fc2/virtualenvs/d41d8cd9/src/trunk/iqh/clipboard/aggregated-translations/en-US.js'
11:33:51     at Object.fs.openSync (fs.js:584:18)
11:33:51     at Object.fs.writeFileSync (fs.js:1316:33)
11:33:51     at /opt/jenkins_virtualenvs/06c35fc2/virtualenvs/d41d8cd9/src/trunk/iqh/clipboard/node_modules/terra-i18n-plugin/lib/I18nAggregatorPlugin.js:51:20
11:33:51     at Array.forEach (native)
11:33:51     at aggregateTranslations (/opt/jenkins_virtualenvs/06c35fc2/virtualenvs/d41d8cd9/src/trunk/iqh/clipboard/node_modules/terra-i18n-plugin/lib/I18nAggregatorPlugin.js:31:20)
11:33:51     at Compiler.apply (/opt/jenkins_virtualenvs/06c35fc2/virtualenvs/d41d8cd9/src/trunk/iqh/clipboard/node_modules/tapable/lib/Tapable.js:164:16)
11:33:51     at WebpackOptionsApply.process (/opt/jenkins_virtualenvs/06c35fc2/virtualenvs/d41d8cd9/src/trunk/iqh/clipboard/node_modules/webpack/lib/WebpackOptionsApply.js:62:18)
[...]

PR #609, opened 2017-07-13
#608 - terra-i18n-plugin now uses synchronous mkdirp

brhoades:
See #608 

terra-i18n-plugin runs mkdirp asynchronously before doing a synchronous write. This switches mkdirp to the synchronous variant.


Matt-Butler:
Thanks for finding this @brhoades 

Can you please add this to the [changelog](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n-plugin/CHANGELOG.md)?

```
Unreleased
----------
* Updated to synchronous mkdirp
```

brhoades:
Didn't catch that. Updated the changelog and rebased.

Issue #625, opened 2017-07-17
Modal custom props do not get passed to modal correctly


bjankord:
# Issue Description
Setting a custom prop like `id="foo"` on a `<Modal>` component fails to render an `id` in the DOM.

This looks like it can be fixed by assigning custom props to the `<Modal>` component rather than the `<Portal>` component.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Setting a custom prop like `id="foo"` on a `<Modal>` component should render an `id` in the modal container element.

## Current Behavior
Setting a custom prop like `id="foo"` on a `<Modal>` component fails to render an `id` in the DOM.

## Steps to Reproduce

1. Add `id="foo"` to `<Modal>` component

## Environment

* Component Version: Modal v1.0.0

Matt-Butler:
https://github.com/cerner/terra-core/blob/master/packages/terra-modal/src/Modal.jsx#L117

Custom props are currently being added to the portal component. To resolve this issue, I'll move custom props to modal component.

PR #627, opened 2017-07-17
Modal custom props

Matt-Butler:
### Summary
Custom props are now added to the modal, not the portal component.

Resolves: https://github.com/cerner/terra-core/issues/625

@cerner/terra


Matt-Butler:
@benbcai @dkasper-was-taken You two have been looking at getting the datepicker working inside of the modal. Do you need the ability to add custom props on the portal component?

benbcai:
@Matt-Butler No, I don't have any need to add custom props to the portal component.

Issue #383, opened 2017-05-22
React Form Validation


bjankord:
# Issue Description
Terra-Legacy uses [jQuery-validation| https://jqueryvalidation.org/] for form validation. We are migrating to React and are no longer using jQuery. We want to investigate alternative libraries for form validation. There have been suggestions on not including form validation as a part of terra and instead leave that up to the application.  This issue is for discussion and to document the approach we will take. 

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

## Expected Behavior
Spike on form validation solutions for use with [terra-form](https://github.com/cerner/terra-core/tree/master/packages/terra-form).

## Current Behavior
Terra does not provide any form validation.

bjankord:
This project was one that has been suggested to check out: https://github.com/skaterdav85/validatorjs

JakeLaCombe:
How about for uncontrolled inputs? I take it we don't need to worry about that.

I really like how that validator takes in a hash of data as opposed to being attached to dom nodes. I feel that makes overriding the onSubmit function much easier than if we tried to attach validation to individual dom elements.

JakeLaCombe:
Here is another approach that we could leverage. https://github.com/Lesha-spr/react-validation

While this approach has components that are already defined for inputs and other form components,  we are able to extend our own components to leverage this validation api. One thing I like about this api is that we can write our own field components to have a hint area that is automatically populated when a validation error occurs.

JakeLaCombe:
React Bootstrap Validation contains validation methods that are similar to how Rails handles their validations. https://www.npmjs.com/package/react-bootstrap-validation

I'll write up something for the next war room meeting on how we envision consumers interfacing with any validations we provide.

bjankord:
Today's war room discussion notes on form validations:

* KAs recommend validations should not be done at the component level, it should be done at store/redux level
* Revisit prior validation spike to see what options there are listed on that
* KAs commented that it is going to be hard for terra to recommend a form validation solution that fits all apps. App teams will be better suited to determine validation strategies.
* We can recommend a solution once there is a consensus around a validation implementation.
* KAs recommended that we should find a recommendation otherwise there will be inconsistency with validation implementations. Everyone will do their own thing.

JakeLaCombe:
[...]

PR #646, opened 2017-07-20
Terra Form CSS Modules

### LINKED ISSUE #553: Convert Form components to CSS modules
# Issue Description

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior

## Current Behavior

JakeLaCombe:
### Summary
Terra Form has been converted to CSS Modules.

Closes #553 

### Deployment URL
https://terra-core-deployed-pr-646.herokuapp.com/


packages/terra-form/src/Control.scss
```diff
@@ -1,22 +1,26 @@
 @import '~terra-mixins';
```
emilyrohrbough: As these components have been updated to use css modules, terra-mixins have been replaced with the defaulted `dir=ltr` behavior since the `postcss-rtl` webpack plugin can handle the directionality.

JakeLaCombe: Good catch! This isn't needed at all for the project, so I removed it all together. https://github.com/cerner/terra-core/pull/646/commits/cc1792fce39e86d5056e2c948ae93b3d26db371a

packages/terra-form/src/Control.jsx
```diff
@@ -82,16 +84,16 @@ const Control = ({
...
 
-  const labelTextClasses = classNames(
-    'terra-Form-control-label-text',
+  const labelTextClasses = cx([
+    'control-label-text',
     labelTextAttrs.className,
-  );
+  ]);
 
   const controlInputAttrs = Object.assign({}, inputAttrs);
```
emilyrohrbough: In the control.scss file, you changed `.terra-Form-input` to `.input` that is locally scoped to the control component. Do we need to add `className={cx('input')}` to the `controlInputAttrs`?

emilyrohrbough: I just checked the input styling on the heuoku compared to the deployed site and it does appear the Control.scss input styling has been lost.

packages/terra-form/tests/nightwatch/field/field-spec.js
```diff
@@ -14,17 +14,15 @@ module.exports = {
...
     browser
       .url(`http://localhost:${browser.globals.webpackDevServerPort}/#/tests/form-tests/field/populated`)
-      .assert.elementPresent('.terra-Form-field')
-      .assert.elementPresent('.terra-Form-field--inline')
-      .assert.elementPresent('.terra-Form-label.healtheintent-application')
+      .assert.elementPresent('#populated-field')
       .assert.containsText('label', 'Do you have any Children?')
-      .assert.containsText('.terra-Form-helpText', 'Families are eligible for family package plans')
-      .assert.containsText('.terra-Form-error', 'This field is required');
+      .assert.containsText('small', 'Families are eligible for family package plans')
```
emilyrohrbough: Where is small coming from?

packages/terra-form/src/Control.jsx
emilyrohrbough: I just checked the input styling on the heuoku compared to the deployed site and it does appear the Control.scss input styling has been lost.

emilyrohrbough:
Why did we remove the Input Controlled example?

JakeLaCombe:
@emilyrohrbough That was left over from previous code when we decided to remove the controlled tests. Essentially, the controlled tests did nothing more than test React code itself, and our largest priority with the tests was to make sure it received the appropriate arguments for value and onChange, which our jest tests covered.

packages/terra-form/src/Control.jsx
JakeLaCombe: Good Eye! I completely missed that! It's fixed here. https://github.com/cerner/terra-core/pull/646/commits/8ab614c9491b5a2a96b418202f0aab54a2f5de3a

packages/terra-form/tests/nightwatch/field/field-spec.js
JakeLaCombe: Right there. https://github.com/cerner/terra-core/blob/master/packages/terra-form/src/Field.jsx#L72

mjhenkes:
@JakeLaCombe could you resolve the conflicts on this branch?

PaulSolDev:
In the Default Number Field test I'm able to enter letters but the field clears the content when select outside of the entry field.  I was under the impression this type of field would not even show letters in this case.  Is this the intended behavior?

PaulSolDev:
This looks busted.  Selected the default selected field when my browser was not full screen.  

default selected field ie 10

PaulSolDev:
Another odd example.  This is the populated select field when the browser is less than full screen and I moved the browser when the drop down was expanded.  I would expect this workflow to dismiss the menu as I am selecting outside of the intended menu area.

populated select field with browser movement ie10

JakeLaCombe:
@PaulSolDev The behavior inputs are defaulting to the browser behavior. We don't have any validation inputs in place at this moment that do keyboard masing and ensuring only numbers are inputed into a field.

JakeLaCombe:
Also, on those defect you found, I see them when I resize the virtual box machine, but not when I reset the browser inside the virtual box. I think that may just be a defect with the virtual box as opposed to a browser issue.

PaulSolDev:
@JakeLaCombe Great, thank you for that clarification.

PaulSolDev:
I will comment back when the we have finalized testing in the remaining browsers.

JakeLaCombe:
@PaulSolDev can you also check the deployed branch as well. It looks like you are using a different url than what this branch was deployed too. Try https://terra-core-deployed-pr-646.herokuapp.com/

PaulSolDev:
@JakeLaCombe I'm not even sure how I jumped to the wrong pull request in that browser but I'll go back and review again via the links in this thread.

PaulSolDev:
Functional review complete.  No issues identified outside of the need to limit character type input for specific fields.  Has this issue been logged yet?

JakeLaCombe:
I think this is what you are looking for @PaulSolDev. https://github.com/cerner/terra-core/issues/666

PaulSolDev:
@JakeLaCombe Yes and no.  I think that is important too, but what I wanted to see what a number field that only allowed input/display of numbers and vice versa for text only fields.

JakeLaCombe:
I've logged https://github.com/cerner/terra-core/issues/675. We also have https://github.com/cerner/terra-core/issues/383 that does validations as well.

PR #650, opened 2017-07-23
Update stylelint

bjankord:
### Summary
Update stylelint and our stylelint config to latest major versions.

@cerner/terra

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,8 +90,8 @@
     "react-test-renderer": "^15.5.4",
     "rimraf": "^2.6.1",
     "shelljs": "^0.7.7",
-    "stylelint": "^7.12.0",
-    "stylelint-config-sass-guidelines": "^2.2.0",
+    "stylelint": "~8.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -90,8 +90,8 @@
     "react-test-renderer": "^15.5.4",
     "rimraf": "^2.6.1",
     "shelljs": "^0.7.7",
-    "stylelint": "^7.12.0",
-    "stylelint-config-sass-guidelines": "^2.2.0",
+    "stylelint": "~8.0.0",
```
Matt-Butler: Just curious, why `~8.0.0` instead of `^8.0.0`

bjankord: stylelint documented their new [semantic version policy](README.md#semantic-versioning-policy). They now recommend the use of the tilde (`~`) in `package.json`, e.g. `"stylelint": "~8.0.0"`, to guarantee the results of your builds ([#1865](https://github.com/stylelint/stylelint/issues/1865)).

PR #652, opened 2017-07-24
Badge: use classnames/bind

Matt-Butler:
### Summary
Updating Badge to use classnames/bind

@cerner/terra


packages/terra-badge/src/Badge.jsx
```diff
@@ -48,7 +50,7 @@ const Badge = ({
   icon,
   ...customProps
 }) => {
-  const badgeClassNames = classNames(
+  const badgeClassNames = cx(
       styles.badge,
```
emilyrohrbough: this should be updated to 'badge' and 'size', etc

PR #656, opened 2017-07-24
Migrate Table to CSS Modules.

### LINKED ISSUE #567: Convert Table components to CSS modules
# Issue Description

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior

## Current Behavior

emilyrohrbough:
### Summary
Resolves #567

@cerner/terra

### DIFF
--- a/packages/terra-table/src/Table.scss
+++ b/packages/terra-table/src/Table.scss
@@ -1,176 +1,178 @@
-@import '~terra-mixins';
 @import './variables';
+/* stylelint-disable selector-class-pattern, selector-max-compound-selectors, max-nesting-depth */
+:local {
+  .table {
+    background-color: var(--terra-table-background-color);
+    border: var(--terra-table-border);
+    border-collapse: collapse;
+    border-spacing: 0;
+    max-width: 100%;
+    width: 100%;
 
-.terra-Table {
-  background-color: $terra-table-background-color;
-  border: $terra-table-border;
-  border-collapse: collapse;
-  border-spacing: 0;
-  max-width: 100%;
-  width: 100%;
-
-  caption {
-    padding-bottom: 0.5em;
-  }
-
-  tr {
-    display: table-row;
-    margin-bottom: 0;
-  }
-
-  th,
-  td {
-    @include terra-text-align-start;
-    border: 0;
-    @include terra-border-start($terra-table-cell-border);
-    display: table-cell;
-    vertical-align: top;
-
-    &:first-of-type {
-      @include terra-border-start($terra-table-row-border-vertical);
+    caption {
+      padding-bottom: 0.5em;
     }
-
-    &:last-of-type {
-      @include terra-border-end($terra-table-cell-border);
-      @include terra-border-end($terra-table-row-border-vertical);
-    }
-  }
-
-  thead {
-    background-color: $terra-table-thead-background-color;
-    display: table-header-group;
 
     tr {
-      border-bottom: $terra-table-thead-tr-border-bottom;
-      border-top: $terra-table-thead-tr-border-top;
+      display: table-row;
+      margin-bottom: 0;
     }
 
-    th {
-      color: $terra-table-thead-th-color;
-      font-size: $terra-table-thead-th-font-size;
-      font-weight: $terra-table-thead-th-font-weight;
-      vertical-align: bottom;
-    }
-  }
+    th,
+    td {
+      border: 0;
+      border-left: var(--terra-table-cell-border);
+      display: table-cell;
+      text-align: left;
+      vertical-align: top;
 
-  tbody {
-    tr {
-      border-top: $terra-table-row-border-horizontal;
+      &:first-of-type {
+        border-left: var(--terra-table-row-border-vertical);
+      }
 
       &:last-of-type {
-        border-bottom: $terra-table-row-border-horizontal;
+        border-right: var(--terra-table-cell-border);
+        border-right: var(--terra-table-row-border-vertical);
+      }
+    }
+
+    thead {
+      background-color: var(--terra-table-thead-background-color);
+      display: table-header-group;
+
+      tr {
+        border-bottom: var(--terra-table-thead-tr-border-bottom);
+        border-top: var(--terra-table-thead-tr-border-top);
+      }
+
+      th {
+        color: var(--terra-table-thead-th-color);
+        font-size: var(--terra-table-thead-th-font-size);
+        font-weight: var(--terra-table-thead-th-font-weight);
+        vertical-align: bottom;
+      }
+    }
+
+    tbody {
+      tr {
+        border-top: var(--terra-table-row-border-horizontal);
+
+        &:last-of-type {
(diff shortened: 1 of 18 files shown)


packages/terra-table/src/TableSubheader.jsx
```diff
@@ -19,21 +22,14 @@ const TableSubheader = ({
...
-  return (
-    <tr className="terra-Table-subheaderRow">
-      <td {...customProps} className={contentClassName} colSpan={colSpan}>
-        {content}
-      </td>
-    </tr>
-  );
-};
+  <tr data-class="subheader-row">
+    <td {...customProps} className={cx('subheader', customProps.className)} colSpan={colSpan}>
```
windse7en: Why include customProps.className and provide the feature that consumers can use all Table classes in here?

emilyrohrbough: On [line 23](https://github.com/cerner/terra-core/pull/656/files/8934dc09cbc70d876252a6bbc8ca26d7a0e09e67#diff-b21a6067f2e7eca6ed91fa3729b329f3L23), `customProps.className` was included. I moved the className rendering to be inline, as it has been a suggestion on other PR by @mhemesath.

emilyrohrbough: > why ... provide the feature that consumers can use all Table classes in here?
I am not quite sure what you mean by this comment. Could you explain further?

packages/terra-table/src/_variables.scss
```diff
@@ -1,55 +1,56 @@
...
+  --terra-table-subheader-background-color: #c8cacb;
+  --terra-table-thead-background-color: #dedfe0;
+
+  // Base Border Styles
+  --terra-table-base-border-color: #c8cacb;
+  --terra-table-base-border-width: 1px;
+
+  // Borders
+  --terra-table-border: 0;
+  --terra-table-thead-tr-border-top: var(--terra-table-base-border-width) solid var(--terra-table-base-border-color);
```
bjankord: With the work we are planning to do in the [theme provider component](https://github.com/cerner/terra-core/tree/theme-adapter/packages/terra-theme-provider), CSS custom property values need to be static, rather than dynamic. With run-time theming, the plan is to find/replace CSS custom property with their value. If those values are dynamic e.g. (other CSS custom properties) that becomes a lot more challenging to polyfill in the run time environment. It will work with postcss-custom-properties plugin at build time though as the plugin can parse the AST and find the dynamic values.

tl:dr We should avoid using CSS custom properties as values to other CSS custom properties.

I'd recommend removing CSS custom properties like `--terra-table-thead-tr-border-top` and putting their value inline where they were used. 

e.g 
```css
 tr {
  border-bottom: var(--terra-table-base-border-width) solid var(--terra-table-base-border-color);
  border-top: var(--terra-table-base-border-width) solid var(--terra-table-base-border-color);
}
```

or make a variable for the entire short-hand value

```css
:root {
  --terra-table-thead-tr-border-bottom: 1px solid #c8cacb;
  --terra-table-thead-tr-border-top: 1px solid #c8cacb;
}
 tr {
  border-bottom: var(--terra-table-thead-tr-border-bottom);
  border-top: var(--terra-table-thead-tr-border-top);
}
```

I'll look into working on a stylelint rule for this case.

emilyrohrbough: I removed the dynamic custom properties going with the short hand approach. I assume as the themes are created they will handle the border styles in full so by using the shorthand method, there are less css custom properties to maintain.

packages/terra-table/src/TableSubheader.jsx
JakeLaCombe: Should customProps.className be inside the cx method? I feel that if the class is a custom class name, then we wouldn't expect table to be in front of it.

emilyrohrbough: customProps.className have been placed within the cx method for serveral components. If the custom class name is not locally scoped, it will not be name spaced/compiled with the css modules naming. 

I added a custom class name to a table nightwatch example and the result was this:

screen shot 2017-07-26 at 8 31 40 am

JakeLaCombe: You are right. Had a brain moment yesterday. This looks fine.

windse7en: Sorry for unclear question. I wonder what if consumer uses some class name inside `Table.scss`. In that case, our local css style code will be applied or not. For example:
```javascript
// customProps.className includes 'striped' class
<td {...customProps} className={cx('subheader', customProps.className)} .../>
```
In this case, striped style code in [here](https://github.com/cerner/terra-core/blob/8934dc09cbc70d876252a6bbc8ca26d7a0e09e67/packages/terra-table/src/Table.scss#L118) will be applied to `td` or not?

emilyrohrbough: Oh. I see what you mean. I would assume it would use our locally scoped styling, but not 100% certain. @mhemesath @bjankord do you know the answer to Tao's question?

windse7en: I just will use our locally scoped styling. I checked with code base. It appends fill class twice, one if from `fill={true}` and one is from `class="fill"`.
screen shot 2017-07-26 at 10 23 23 am

windse7en: In such case, I wonder if we want to add customized classes after the cx() values.

emilyrohrbough: I wonder as well. If we do there would be several components that would need this change that have already been converted.

mjhenkes:
@emilyrohrbough could you resolve the conflict on this branch?

mjhenkes:
@bjankord could you take another look at this review?

packages/terra-table/src/TableSubheader.jsx
bjankord: The concern is still unclear to me. Let's talk more in our meeting later today.

windse7en: After the discussion, this issue won't be avoided if consumers follow suit naming convention, we decided to keep it.
+1

packages/terra-table/src/Table.scss
```diff
@@ -1,176 +1,178 @@
...
-    }
+      &:first-of-type {
+        border-left: var(--terra-table-row-border-vertical);
+      }
 
-    &:last-of-type {
-      @include terra-border-end($terra-table-cell-border);
-      @include terra-border-end($terra-table-row-border-vertical);
+      &:last-of-type {
+        border-right: var(--terra-table-cell-border);
```
emilyrohrbough: Hmm not sure how that got in there?

emilyrohrbough:
@bjankord Thank you for updating the inline fallback!

mmalaker:
Functional verification completed. Table examples display and function correctly after conversion to CSS modules.

PR #665, opened 2017-07-26
Issue 246 internationalize date picker

### LINKED ISSUE #246: Internationalize terra-date-picker and terra-time-input
# Issue Description
Internationalize terra-date-picker and terra-time-input

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Strings, locale, and date and time format in terra-date-picker and terra-time-input need to be internationalized to support formats in other locales. This work needs to wait until we have a strategy in place on how to get the translated strings and locale from i18nProvider. 

## Current Behavior
Strings, locale, and date and time format in terra-date-picker and terra-time-input are currently hard-coded in English and using the US date/time formats.

benbcai:
Internationalize DatePicker.
 - Translated strings
 - Support date format in other locales

Resolves #246 

@cerner/terra


packages/terra-date-picker/package.json
```diff
@@ -21,6 +21,7 @@
   },
   "homepage": "https://github.com/cerner/terra-core#readme",
   "devDependencies": {
+    "react-intl": "^2.3.0",
```
Matt-Butler: This should probably be a peerDependency.  The component depends on react-intl but we want to allow the consumer to choose the version of react-intl.

benbcai: react-intl is added as a devDependency because it is used only for jest testing just like [Overlay](https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/package.json#L24) and [Demographics Banner](https://github.com/cerner/terra-core/blob/master/packages/terra-demographics-banner/package.json#L24)

emilyrohrbough:
The `Packages Requiring I18n` in the project ReadMe needs updated as well.

benbcai:
The  [Packages Requiring I18n](https://github.com/cerner/terra-core/blob/master/README.md#packages-requiring-i18n) section in the ReadMe already included DatePicker. But your commented reminded me that I need to add `Internationalization Support` in the `Component Features` section in the component ReadMe.

packages/terra-date-picker/tests/jest/__snapshots__/DatePicker.test.jsx.snap
```diff
@@ -1,639 +1,11 @@
-exports[`test should render a date picker with disabled dates 1`] = `
```
windse7en: Due to the shallow rendering, this snapshot of jest isn't rendered correctly. It just contains `<datePicker />`. We may want to use deep render to render all children inside it.

benbcai: Ahh, I didn't realize that.

bjankord: Yeah, I'd recommend using `render` to render snapshots so we have more of the structure captured in the snapshot.

bjankord:
+1 provided the merge conflicts are resolved.

packages/terra-date-picker/tests/nightwatch/components/DatePickerExcludeDates.jsx
```diff
@@ -1,12 +1,17 @@
...
 const DatePickerExcludeDates = () => (
-  <DatePicker
-    name="date-input"
-    excludeDates={['2017-04-01', '2017-04-03']}
-    selectedDate={'2017-04-02'}
-  />
+  <Base locale={locale}>
+    <DatePicker
+      name="date-input"
+      excludeDates={['2017-04-01', '2017-04-03']}
```
emilyrohrbough: On PR #671, we had this discussion regarding the consistency of the date format: 

screen shot 2017-08-01 at 10 24 57 am

I think it would be important to include/consider how this affects `excludeDates`, as well as `selectedDate`.

mmalaker:
Functional verification completed. All date picker examples display and function correctly in each supported locale.

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -68,13 +68,20 @@ const defaultProps = {
   selectedDate: undefined,
 };
 
+const contextTypes = {
+  /* eslint-disable consistent-return */
+  intl: (context) => {
```
ryanthemanuel: Do we like this pattern vs. using injectIntl to add an intl prop?  I don't think this needs to hold us up on this PR, but I'm more curious if we prefer one vs. the other outright or have situations where one makes sense over the other.

benbcai: I think using a required intl prop would be less code but the reason for using this pattern to check for it is explained by this comment: https://github.com/cerner/terra-core/pull/431/files#r120640131

PR #689, opened 2017-08-01
Exposing fullscreen presentation style to Modal Manager

tbiethman:
### Summary
The ModalManager today supports 5 standard display sizes (tiny through huge). We want to provide an additional fullscreen size that will utilize the Modal's already defined isFullscreen prop.

### Additional Details
A couple of functionality tweeks to consider:
* When fullscreen, maximize and minimize functions should not be provided to the modal contents (as they're already maximized and cannot minimize).
* Animations are disabled when fullscreen (because the transition is way too fast for the translating of such a large piece of UI). We could consider adding a different animation type when fullscreen, but that'd increase the scope of the PR a bit.


neilpfeiffer:
Tested: PR #689 

**Reviewed:**
- ModalManager goes fullscreen or maximizes/minimizes as expected
- ModalManager tests: nesting disclosed Modals functions as expected, with no animations when fullscreen/maximized (probably ideal to not use animations for this one).

**Pass** |
-----------|
UX Review|

 

**Potentially solve with Future Issue/PR:**
- Relook at ModalManager sizing, see if more/specific are needed.
- Possibly change isFullscreen styling to remove outer margins, rounded borders, and shown overlay so that modal, popover, sidepanel, etc. all look similar when full screen (mainly when used for smallscreens/mobile). Will need to validate use cases.

mmalaker:
Functional verification completed. All modal manager examples display and function correctly. +1 UX comments.

PR #690, opened 2017-08-02
dynamic grid

### LINKED ISSUE #726: Create Dynamic Grid component
# Issue Description
Create Dynamic Grid component.

## Issue Type

- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Create a dynamic grid based on CSS Grid spec. This grid should only be used for dynamic configurations (determined at runtime, not build time). This component will be used to power our configurable dashboards/content for embedded components.

mhemesath:
### Summary
Pull request to create a dynamic grid based on CSS Grid spec.  Documentation needs updating yet, but this grid should only be used for dynamic configurations (determined at runtime, not build time).  This component will be used to power our configurable dashboards/content for embedded components.

Resolves #726 

### Additional Details
The code is complete and ready for review, however I still need to add a lot of tests and documentation before this pull request acan be merged.

Deployment URL: https://terra-core-deployed-pr-690.herokuapp.com/static/#/site/dynamic-grid


packages/terra-dynamic-grid/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "terra-toolkit": "^1.0.0"
+  },
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "terra-base": "^1.0.0",
+    "terra-mixins": "^1.0.0"
+  },
+  "dependencies": {
+    "aphrodite": "^1.2.3",
```
mhemesath: Each CSS-In-JS approach is quite unique and aphrodite nicely fits the requirements.  Here is the reasoning:

* styled-components - Styles are required to be bound to react element. This makes it difficult/impossible to create multiple response layouts to power a single grid
* styles-jsx - Doesn't allow for dynamic styles
* Aphrodite - Styles are decoupled from component. They are almost the same as CSS modules in that you generate the stylesheet and pass reference by key to class names. This makes it easy to generate all the responsive styles and properly pass them down to the regions.

mhemesath:
URL to doc https://terra-core-deployed-pr-690.herokuapp.com/static/#/site/dynamic-grid

packages/terra-dynamic-grid/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Dynamic Grid
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-dynamic-grid.svg)](https://www.npmjs.org/package/terra-dynamic-grid)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+{insert description}
```
emilyrohrbough: Add Description

packages/terra-dynamic-grid/docs/README.md
```diff
@@ -0,0 +1,28 @@
+# Terra Dynamic Grid
+
+{insert description}
```
emilyrohrbough: Add description

packages/terra-site/src/examples/dynamic-grid/Dashboard.jsx
```diff
@@ -0,0 +1,80 @@
...
+      style: {
+        padding: '5px 5px 0',
+      },
+    },
+  ],
+},
+];
+
+const DashboardLayout = () => (
+  (<DynamicGrid layouts={layouts}>
```
mhemesath: Generally I'd prefer to not pass in a monolithic config info the top component, but rather build out each child with the config it needs. However, this was needed to support responsive layouts that change based on media query. I don't have a doc example of this yet, but it would be challenging to pass in the responsive config for each region separate (as its dependent on the grid config) so I determined it made the most sense to just pass in one large object.

packages/terra-dynamic-grid/docs/README.md
```diff
@@ -0,0 +1,28 @@
...
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-dynamic-grid`
+
+## Usage
+
+```jsx
+import React from 'react';
+import DynamicGrid from 'terra-dynamic-grid';
+
+<DynamicGrid {props...} />
```
emilyrohrbough: Update Example

packages/terra-dynamic-grid/docs/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+## Component Features
+
+<!-- Uncomment supported features.
```
emilyrohrbough: Uncomment & remove Internationalization & Localization support

packages/terra-dynamic-grid/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "terra-dynamic-grid",
+  "main": "lib/DynamicGrid.js",
+  "version": "0.0.0",
+  "description": "{insert description}",
```
emilyrohrbough: Add description

packages/terra-dynamic-grid/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "DynamicGrid",
+    "UI"
+  ],
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
```
emilyrohrbough: Remove toolkit dependency, as well as terra-mixins

packages/terra-dynamic-grid/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "lint:scss": "$(cd ..; npm bin)/stylelint src/**/*.scss",
+    "props-table": "props-table ./src/DynamicGrid.jsx --out-dir ./docs/props-table",
+    "test": "npm run test:spec && npm run test:nightwatch-default",
+    "test:spec": "$(cd ..; npm bin)/jest --config ../../jestconfig.json",
+    "test:all": "npm run test:nightwatch-default && npm run test:nightwatch-chrome && npm run test:nightwatch-firefox && npm run test:nightwatch-safari",
+    "test:nightwatch-default": "SPECTRE_TEST_SUITE=terra-dynamic-grid node ./node_modules/terra-toolkit/lib/scripts/nightwatch.js",
+    "test:nightwatch-chrome": "SPECTRE_TEST_SUITE=terra-dynamic-grid node ./node_modules/terra-toolkit/lib/scripts/nightwatch.js chrome",
+    "test:nightwatch-firefox": "SPECTRE_TEST_SUITE=terra-dynamic-grid node ./node_modules/terra-toolkit/lib/scripts/nightwatch.js firefox",
+    "test:nightwatch-safari": "SPECTRE_TEST_SUITE=terra-dynamic-grid node ./node_modules/terra-toolkit/lib/scripts/nightwatch-non-parallel.js safari",
+    "test:remote": "SPECTRE_TEST_SUITE=terra-dynamic-grid REMOTE=true node ./node_modules/terra-toolkit/lib/scripts/nightwatch-process.js --config tests/nightwatch.conf.js",
```
emilyrohrbough: Remove remote test scripts.

packages/terra-dynamic-grid/src/DynamicGrid.jsx
```diff
@@ -0,0 +1,39 @@
...
+  return (<div className={css(layoutStyles('grid'))}>
+    {React.Children.map(props.children,
+     region => React.cloneElement(region, {
+       styles: layoutStyles(region.props.name),
+     }),
+    )}
+  </div>);
+};
+
+DynamicGrid.propTypes = {
```
emilyrohrbough: Add prop definitions

packages/terra-dynamic-grid/tests/nightwatch.conf.js
```diff
@@ -0,0 +1,9 @@
+/* eslint import/no-extraneous-dependencies: ["error", {"devDependencies": true}] */
```
emilyrohrbough: This file can be removed.

packages/terra-dynamic-grid/README.md
mhemesath: Yeah, need to work on Doc and tests for the PR.

packages/terra-site/src/Index.jsx
```diff
@@ -97,6 +98,7 @@ ReactDOM.render((
       <Route path="content-container" component={ContentContainerExamples} />
       <Route path="datepicker" component={DatePickerExamples} />
       <Route path="demographics-banner" component={DemographicsBannerExamples} />
+      <Route path="dynamic-grid" component={DynamicGridExamples} />
       <Route path="form" component={FormExamples} />
       <Route path="grid" component={GridExamples} />
       <Route path="heading" component={HeadingExamples} />
```
emilyrohrbough: Need to add dynamic Grid TestRoutes

packages/terra-site/src/examples/dynamic-grid/Index.jsx
```diff
@@ -0,0 +1,36 @@
...
+import DynamicGridSrc from '!raw-loader!terra-dynamic-grid/src/DynamicGrid';
+
+// Example Files
+
+const DynamicGridExamples = () => (
+  <div>
+    <div id="version">Version: {version}</div>
+    <Markdown id="readme" src={ReadMe} />
+    <PropsTable id="props" src={DynamicGridSrc} />
+    <h2>One Column</h2>
```
emilyrohrbough: Nit: add <br /> between examples

packages/terra-dynamic-grid/tests/jest/DynamicGrid.test.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import DynamicGrid from '../../src/DynamicGrid';
+
+describe('DynamicGrid', () => {
+  const defaultRender = <DynamicGrid />;
```
emilyrohrbough: Children is a required prop of DynamicGrid. Does this render correctly? I am not seeing the snapshots.

mhemesath: I mentioned this in PR description, but I haven't updated doc or tests yet.

packages/terra-dynamic-grid/src/DynamicGrid.jsx
```diff
@@ -0,0 +1,39 @@
...
+const DynamicGrid = (props) => {
+  const stylesheets = props.layouts
+    .map(generateStyles)
+    .map(StyleSheet.create);
+
+  const layoutStyles = name => stylesheets.map(x => x[name]);
+  return (<div className={css(layoutStyles('grid'))}>
+    {React.Children.map(props.children,
+     region => React.cloneElement(region, {
+       styles: layoutStyles(region.props.name),
```
mhemesath: this passes the generated region classnames down to the regions

packages/terra-dynamic-grid/src/Region.jsx
```diff
@@ -0,0 +1,18 @@
+
+import React from 'react';
+import { css } from 'aphrodite';
+import PropTypes from 'prop-types';
+
+const Region = props => (
+  (<div data-region-name={props.name} className={css(props.styles)}>
```
kafkahw: nit: parentheses are not needed.

packages/terra-dynamic-grid/src/styles.js
```diff
@@ -0,0 +1,110 @@
...
+
+  return { grid: layout.media
+    ? { [layout.media]: styles }
+    : styles };
+};
+
+const generateRegionStyles = layout => (
+  (region) => {
+    const styles = {
+      ...gridColumnStart(layout, region),
```
mhemesath: I think i can remove layout from being passed in.. I though I needed to have access to it to determine if I needed to calculate grid gaps, but to keep things simple, I just always include grid gaps and set them to 0.

packages/terra-dynamic-grid/src/DynamicGrid.jsx
bjankord: Could you add a code comment about this, took me a while to grok what all this was doing here.

packages/terra-dynamic-grid/src/styles.js
```diff
@@ -0,0 +1,110 @@
+// IE < 11 doesn't honor window.CSS.Supports. Edge has supports, but will fail on grid.
+const cssGridsEnabled = window.CSS && window.CSS.supports && window.CSS.supports('display', 'grid');
+
+const gridTemplateColumns = (layout) => {
+  const columns = layout['grid-template-columns'];
+  const msColumns = columns.split(/\s+/).join(` ${layout['grid-gap'] || 0} `);
```
kafkahw: do we allow users to specify [grid-column-gap](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-column-gap)?

packages/terra-dynamic-grid/src/styles.js
```diff
@@ -0,0 +1,110 @@
...
+
+  return {
+    'grid-template-columns': columns,
+    '-ms-grid-columns': msColumns,
+  };
+};
+
+const gridTemplateRows = (layout) => {
+  const rows = layout['grid-template-rows'];
+  const msRows = rows.split(/\s+/).join(` ${layout['grid-gap'] || 0} `);
```
kafkahw: similarly, do we allow users to specify [grid-row-gap](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-row-gap)?

packages/terra-dynamic-grid/src/DynamicGrid.jsx
dkasper-was-taken: Definite prop types at the top of the file.

packages/terra-dynamic-grid/src/styles.js
mhemesath: I  was lazy and left it off.. I could easily add support for it though. I'll be sure to add it

packages/terra-dynamic-grid/src/styles.js
mhemesath: same comment as above :)

packages/terra-dynamic-grid/src/styles.js
```diff
@@ -0,0 +1,110 @@
...
+
+  return {
+    'grid-template-rows': rows,
+    '-ms-grid-rows': msRows,
+  };
+};
+
+
+const gridColumnStart = (layout, region) => {
```
kafkahw: didn't see param 'layout' being used anywhere in this function. probably should take it out. same applies to a couple following functions.

kafkahw: Oops, didn't see your comment below. Ignore this comment then.

packages/terra-dynamic-grid/src/styles.js
```diff
@@ -0,0 +1,110 @@
...
+      : styles };
+  }
+);
+
+
+const generateStyles = (layout) => {
+  const grid = generateGridStyles(layout);
+  return layout.regions
+    .map(generateRegionStyles(layout))
+    .reduce((regions, region) => Object.assign(regions, region), grid);
```
kafkahw: I feel this line would break region's style if there are multiple media query styles for the same region.
For instance, if we got two different media query styles for region 'r1', 
`{r1: {'@media @smartphone': {background: 'red'}}}`
 `{r1: {'@media @tablet': {background: 'green'}}}`. 
In the final result, the later style would override the former one because they're sharing the same key 'r1'. So it would become:  
`{r1: {'@media @tablet': {background: 'green'}}}`.

mhemesath: I have to test media queries yet.. so this may need modification so thanks for pointing out

packages/terra-dynamic-grid/src/styles.js
mhemesath: So I dont think we should support `grid-column-gap` and `grid-row-gap` but rather support shorthand syntax of `grid-gap: <grid-column-gap> <grid-row-gap>` This is easier to implement and doesnt' take away any functionality.

packages/terra-dynamic-grid/src/styles.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+);
+
+
+const generateStyles = (layout) => {
+  const grid = generateGridStyles(layout);
+  const regionStyles = generateRegionStyles(layout);
+  return layout.regions
+    .map(regionStyles)
+    .reduce((regions, region) => ({ ...regions, ...region }), grid);
```
kafkahw: Leaving the same comment again because the old comment's gone with the new code changes. If a region has multiple media queries as follows:
`{r1: {'@media @smartphone': {background: 'red'}}}`
`{r1: {'@media @tablet': {background: 'green'}}}`
`{r1: {background: 'yellow'}}`
The last one will override the former ones because they're sharing the same key.

packages/terra-dynamic-grid/src/DynamicGrid.jsx
```diff
@@ -0,0 +1,99 @@
...
+  const stylesheets = layouts
+    .map(generateStyles)
+    .map(StyleSheet.create);
+
+  const layoutStyles = name => stylesheets.map(x => x[name]);
+  const gridClasses = classNames(customProps.className, css(layoutStyles('grid')));
+
+  // Clone the child regions and inject their generated styles
+  return (<div className={gridClasses} {...customProps}>
+    {(React.Children || []).map(children,
```
tbiethman: Is React.Children ever undefined?

mhemesath: yeah... it seems to be undefined when running snapshot tests with jest

packages/terra-dynamic-grid/tests/jest/DynamicGrid.test.jsx
mhemesath: update: didn't render correctly, this should be fixed now

packages/terra-dynamic-grid/src/DynamicGrid.jsx
JakeLaCombe: Minor:

If you wanted, you maybe able to unmock the React object here with [jest.unmock](https://facebook.github.io/jest/docs/jest-object.html#jestunmockmodulename), but I don't think what you have is bad. I'll leave it up to you.

packages/terra-dynamic-grid/src/styles.js
```diff
@@ -0,0 +1,156 @@
...
+
+  return {
+    'grid-row-start': `${start}`,
+    '-ms-grid-row': `${msStart}`,
+  };
+};
+
+const gridRowEnd = (region) => {
+  if (region['grid-row-end'] === undefined
+    && region['grid-row-end'] === undefined) {
```
JakeLaCombe: I think you meant start for one of these checks?

packages/terra-dynamic-grid/src/styles.js
```diff
@@ -0,0 +1,156 @@
...
+  };
+};
+
+const gridColumnEnd = (region) => {
+  if (region['grid-column-end'] === undefined
+    && region['grid-column-end'] === undefined) {
+    return {};
+  }
+
+  const start = region['grid-column-start'] || 0;
```
JakeLaCombe: I find it a little odd that we require this parameter to be set, yet default it to 0 in this function. Personally, I think it's better not to default it to 0 here, but I understand the concern of getting potential undefined errors here.

packages/terra-site/src/examples/dynamic-grid/Dashboard.jsx
JakeLaCombe: Personally, I think this approach is better. I think it shows the consumer exactly what the component needs to be initialized correctly.

packages/terra-dynamic-grid/src/styles.js
mhemesath: hahah yeah found that while I was writing tests. Will push that update

packages/terra-dynamic-grid/src/styles.js
mhemesath: you bring up a point.... I only need it defaulted for calculating the `-ms-grid-column-span`.  For grid-column-end i can just use whats defined, or not include it if missing.

mhemesath: actually, thinking about this... it may be possilbe to just not render this attribute if `grid-coloumn-end` is not specified.

packages/terra-dynamic-grid/src/styles.js
JakeLaCombe: Awesome! Just a heads up I saw the issue as well on line 51, gridColumnEnd

mhemesath:
Hey @kafkahw getting around to answering this:

>  If a region has multiple media queries as follows:
{r1: {'@media @smartphone': {background: 'red'}}}
{r1: {'@media @tablet': {background: 'green'}}}
{r1: {background: 'yellow'}}
The last one will override the former ones because they're sharing the same key.

So I don't think that can happen. The media queries are specified at the layout level, and a layout can only have 1 media query. It is possible that multiple regions could be created with the same name within a layout, but I'd expect in that scenario that the last one would override the first because there really can't be two regions (they'd overlay each other if there were two html elements for that region). 

What will happen if there are multiple layouts is it will generate multiple stylesheets. Each stylesheet will have a unique class name and all those class names will be appended to the DOM to take effect for their respective responsive ranges.

kafkahw:
@mhemesath I see. yeah, if the media queries are specified only at the layout level, I think we should be fine then.

mhemesath:
Ok, this is almost ready for final review. The code is unit tested, examples written. I just need to write a few nightwatch tests and this should be good to go.

packages/terra-dynamic-grid/src/DynamicGrid.jsx
```diff
@@ -0,0 +1,99 @@
...
+  children: PropTypes.node,
+  /**
+  * The layout configuration that defines the grid. Multiple layouts can be passed per mediaquery to make the grid responsive.
+  */
+  layouts: PropTypes.arrayOf(PropTypes.shape({
+    /**
+    * The column definitions of the grid. See https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns.
+    * For IE support, verify syntax is supported here https://msdn.microsoft.com/en-us/library/hh772246(v=vs.85).aspx.
+    */
+    'grid-template-columns': PropTypes.string,
```
zhongyn: any consideration of adding singe-quote for prop name here?

zhongyn: nvm, it's due to the dash in the name

kafkahw: because the prop name contains hypens.

mhemesath:
I have no further changes to make to this branch. Tests, doc and code is all completed.

packages/terra-dynamic-grid/src/DynamicGrid.jsx
```diff
@@ -0,0 +1,101 @@
...
+
+const propTypes = {
+  /**
+  * The child Region components that make up the grid.
+  */
+  children: PropTypes.node,
+  /**
+  * The layout configuration that defines the grid. Multiple layouts can be passed per mediaquery to make the grid responsive.
+  */
+  layouts: PropTypes.arrayOf(PropTypes.shape({
```
bjankord: The prop-types doc seems to be blowing up on this.

screen shot 2017-08-14 at 10 20 47 am

I thought we accounted for `arrayOf`, but maybe that logic only made it into the markdown scripts. We may need to port that code over to the code that generates the react table.

mhemesath: Yeah... I wanted to discuss changing our doc structure to more match [grommet](https://grommet.github.io/docs/notification). This would allow us to better represent complex properties better than a table can

packages/terra-dynamic-grid/README.md
bjankord: For the description, you can add `The terra-dynamic-grid component provides users a way to build dynamic layouts using CSS Grids.`

packages/terra-dynamic-grid/docs/README.md
bjankord: For the description, you can add `The terra-dynamic-grid component provides users a way to build dynamic layouts using CSS Grids.`

packages/terra-dynamic-grid/README.md
bjankord: For the description, you can add `The terra-dynamic-grid component provides users a way to build dynamic layouts using CSS Grids.`

packages/terra-site/src/examples/dynamic-grid/Dashboard.jsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import DynamicGrid from 'terra-dynamic-grid';
+
+import Card from './Card';
+
+const layouts = [{
+  'grid-template-columns': '300px 1fr 300px',
```
bjankord: nit: Switching to use percentages here would make these fluid and avoid a horizontal scrollbar on smaller widths on the doc site. 
`300px 1fr 300px` => `30% 1fr 30%`

bjankord:
@mmalaker This is ready for functional review.

emilyrohrbough:
https://github.com/cerner/terra-core/issues/726

packages/terra-dynamic-grid/src/DynamicGrid.jsx
```diff
@@ -0,0 +1,101 @@
...
+  const stylesheets = layouts
+    .map(generateStyles)
+    .map(StyleSheet.create);
+
+  const layoutStyles = name => stylesheets.map(x => css(x[name]));
+  const gridClasses = classNames(customProps.className, layoutStyles('grid'));
+
+  // Clone the child regions and inject their generated styles
+  return (<div className={gridClasses} {...customProps}>
+    {(React.Children || []).map(children,
```
alex-bezek: Should children just be required to avoid having to do this similar to the grid component?
https://github.com/cerner/terra-core/blob/master/packages/terra-grid/src/Grid.jsx#L11

mhemesath: its not that.. its that Jest mocks them to be null which breaks this during test runs.

packages/terra-dynamic-grid/src/DynamicGrid.jsx
```diff
@@ -0,0 +1,101 @@
...
+}) => {
+  const stylesheets = layouts
+    .map(generateStyles)
+    .map(StyleSheet.create);
+
+  const layoutStyles = name => stylesheets.map(x => css(x[name]));
+  const gridClasses = classNames(customProps.className, layoutStyles('grid'));
+
+  // Clone the child regions and inject their generated styles
+  return (<div className={gridClasses} {...customProps}>
```
alex-bezek: I think you want to put `{...customProps}` before `className={gridClasses}` otherwise a passed in `className`, since its not being de-constructed out of `customProps`,  will override the gridClasses `className`
https://github.com/cerner/terra-core/blob/master/packages/terra-badge/src/Badge.jsx#L61

mhemesath: Good call, i pushed a fix

mmalaker:
Seeing some overlap on the dashboard grid example in IE. See attachment. 
gridoverlap

mhemesath:
@mmalaker That example isn't responsive (300px left/right columns). I updated it to be based off fractional units instead. Once it deploys it should be good

packages/terra-site/src/examples/dynamic-grid/ResponsiveGrid.jsx
```diff
@@ -0,0 +1,172 @@
...
+      'grid-row-start': 3,
+    },
+    {
+      name: 'cell',
+      'grid-column-start': 1,
+      'grid-row-start': 1,
+    },
+  ],
+}, {
+  media: '@media (max-width: 1500px)',
```
mjhenkes: Are these media layouts overlaid on top of the base configuration specified above? I'm wondering If theres a way to reorganize the API to make that more clear. Maybe the base config has an array of Media to regions pairs? Maybe at that point we could break up the monolithic API too.

mhemesath: All this does is basically generate CSS that reflects the config. The downside to localizing the config to the regions is you are making an assumption/constraint that the base grid configuration (template/columns) doesn't change across media queries.  This reduces the flexibility of the grid which may or may not be OK.

mhemesath: In this specific example, the base configuration does not change across media queries.. so your suggestion is understandable in this context.

mhemesath: The second problem is that once we drop IE support we can start to use [implicit grids](https://css-tricks.com/difference-explicit-implicit-grids/).. this means that we don't even need to have named regions.... the grid implicitly places them.

mjhenkes: Makes sense. Just trying to think through options for breaking up that big config block.

mmalaker:
Functional verification completed. The dynamic grid examples display and function correctly in all supported browsers.

mjhenkes:
@mhemesath is making a couple more changes. Lets hold off merging this down until they are completed.

mhemesath:
Pushed a bunch of changes to this repo. The primary change is that the API is cleaned up a bunch. It now resembles the API of ResponsiveElement. While this eliminates flexibility in media queries, it does simplify usage of the component to our pre-defined breakpoints. Regarding the breakpoints, one difference between this component and ResponsiveElement as that I import the breakpoints from a CSS file, rather than getting them from the DOM.

packages/terra-site/src/examples/dynamic-grid/ResponsiveGrid.jsx
```diff
@@ -0,0 +1,115 @@
...
+import Card from './Card';
+
+const template = {
+  'grid-template-columns': '1fr 1fr 1fr 1fr 1fr',
+  'grid-template-rows': '100px',
+  'grid-gap': '10px',
+};
+
+const region1 = {
+  defaultPosition: {
```
mhemesath: I'm actually happy with how this reads.. I think most scenarios will have a single default template.. with regions having positions across multiple breakpoints configured onto that template.

packages/terra-site/src/examples/dynamic-grid/ULayout.jsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import DynamicGrid from 'terra-dynamic-grid';
+
+import Card from './Card';
+
+const template = {
+  'grid-template-columns': '1fr 1fr 1fr',
+  'grid-template-rows': 'auto',
```
bjankord: IE/Edge are rendering the ULayout without a gap between rows but the other browsers render with a gap between rows. The grid-gap on line 9 should be applied to both columns and rows right?

## IE Screenshot
screen shot 2017-08-29 at 10 57 06 am

## Chrome Screenshot
screen shot 2017-08-29 at 11 16 58 am

Switching this to `auto 10px auto` or `auto 10px` seems to fix this in IE/Edge

## IE after switching grid-template-rows to `auto 10px auto`
screen shot 2017-08-29 at 11 14 18 am

## IE after switching grid-template-rows to `auto 10px`
screen shot 2017-08-29 at 11 23 23 am


Not sure if the 10px grid-gap on line 9 is being applied to rows via -ms properties to simulate gap.

emilyrohrbough:
@mhemesath You said the only difference between this component and the responsive element component is how the breakpoints are imported--is this a change we could make to the responsive element such that this component would only need to be wrapped by it if responsiveness was desired?

bjankord:
@mmalaker This is ready to functionally verify again.

mhemesath:
@emilyrohrbough technically yes you can achieve responsive behavior with responsive element. The only issue is that window resizes re render the DOM and with iframes and embedded content that means a full reload. DynamicGrid rearranged based on CSS which doesn't affect react DOM and is likely much faster and passive to transition across breakpoints with.

emilyrohrbough:
@mhemesath I figured there was a reason, thank you for the explanation!

mmalaker:
Functional verification completed. the reworked dynamic grid examples display and function correctly.

PR #694, opened 2017-08-02
Merge data-class customProps

emilyrohrbough:
### Summary
In the conversion to CSS Modules, some components were given the `data-class` attribute for classes that had no styling applied. This PR updates to merge custom data-classes with the existing component's data-attribute value.

@cerner/terra

### DIFF
--- a/packages/terra-table/src/TableCell.jsx
+++ b/packages/terra-table/src/TableCell.jsx
@@ -12,11 +12,14 @@ const propTypes = {
 const TableCell = ({
   content,
   ...customProps
-}) => (
-  <td {...customProps} data-class="cell">
-    {content}
-  </td>
-);
+}) => {
+  const dataClassNames = customProps['data-class'] ? `cell ${customProps['data-class']}` : 'cell';
+  return (
+    <td {...customProps} data-class={dataClassNames}>
+      {content}
+    </td>
+  );
+};
 
 
 TableCell.propTypes = propTypes;
--- a/packages/terra-table/CHANGELOG.md
+++ b/packages/terra-table/CHANGELOG.md
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Merge data-class attributes to allow for custom data-classes
 
 1.4.0 - (August 1, 2017)
 ------------------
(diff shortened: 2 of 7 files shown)


packages/terra-table/src/TableCell.jsx
```diff
@@ -12,11 +12,14 @@ const propTypes = {
 const TableCell = ({
   content,
   ...customProps
-}) => (
-  <td {...customProps} data-class="cell">
-    {content}
-  </td>
-);
+}) => {
+  const dataClassNames = customProps['data-class'] ? `cell ${customProps['data-class']}` : 'cell';
```
StephenEsser: Is there a discussion thread on the decision to introduce these data attributes?

emilyrohrbough: There is not a thread that discusses the decision to include data attributes, this was something @bjankord and I discussed offline during the CSS Module conversion. These are intended to be dom descriptions.

Before the CSS Module conversion, some elements had terra class names added to represent the structure of the component but did not apply any styling. In the conversion, the class names of un-styled elements were not locally scoped to the component because their names were not bound to the style sheet. We decided it was appropriate to keep these names, but change them to data attributes, to maintain the visual structure of our components.

StephenEsser: This seems like it adds an awkward level of maintenance and potential inconsistency throughout all of our modules we maintain. Is there enough value in keeping these data attributes solely for examining the dom structure? 

The consumer controls the css module classes through the css loader options in the webpack config. This is something some consumers might alter to significantly reduce the size of their assets. This will transform the css module classes into a unique hash, but leave data classes untransformed making the visual structure less valuable in the opposite direction.

screen shot 2017-08-03 at 11 59 32 am


If there's enough value in keeping it it might be easier to own a data attribute so we no longer need to merge potential attributes from custom props. Something like 'data-terra-class-identifier'

emilyrohrbough: @bjankord @mjhenkes  What are your thoughts on this?

mjhenkes: I agree with @StephenEsser, if we want to keep the class names to represent the structure, we should recommend using name-spaced data attributes.

emilyrohrbough:
I am closing this PR and will open a new one for name-spaced data attributes. This will remove the need to merge data-class attributes.

Issue #420, opened 2017-05-31
Menu Component


jmsv6d:
# Issue Description
We want a menu component that will be able to have items with drill-in submenus

## Issue Type

- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Menu would be a popup that displays menu items.
Menu items could have actions occur on click, be toggle-able, or display a submenu

## Current Behavior
N/A

jmsv6d:
Tech-design PR: https://github.cerner.com/terra/tech-designs/pull/32

jmsv6d:
Requirements PR: https://github.cerner.com/terra/design-input/pull/2

neilpfeiffer:
Added UX Visuals for initial Design Input: https://wiki.ucern.com/x/aMjMY

We can validate and determine which items should be included with version 1.0 requirements and which items we will leave for future enhancements.

tm4027:
we are dividing the elements up, but don't have requirements on the TRM side.

jmsv6d:
@mmalaker has done work on writing the requirements for this.

PaulSolDev:
Orion JIRA created

Issue #647, opened 2017-07-20
Slide Group Keyboard Focus


Matt-Butler:
# Issue Description

When tabbing through the slide group examples, the isAnimate examples allow the user to focus elements in the background. 

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior

Keyboard navigation is only able to focus the first component in the slide group.

## Current Behavior

Keyboard navigation is able to focus components behind the first component in the slide group.
![slide_group](https://user-images.githubusercontent.com/8146159/28439012-d0fbe5aa-6d65-11e7-9af9-865f111ca944.gif)

1. Navigate to http://engineering.cerner.com/terra-core/#/site/slide-group
2. Hit increment a couple times on the isAnimated example.
3. Users are able to focus buttons behind the first component.

Matt-Butler:
The toggle component has to handle this same type of behavior. @bjankord added `visibility: hidden` to containers that should not be focused.

https://github.com/cerner/terra-core/blob/master/packages/terra-toggle/src/Toggle.scss#L12

bjankord:
This has been released in terra-slide-group@1.9.0

PR #702, opened 2017-08-07
Terra Menu

jmsv6d:
### Summary
Terra-menu component completes #420 

### Deployment Links
* https://terra-core-deployed-pr-702.herokuapp.com/static/#/site/menu
* https://terra-core-deployed-pr-702.herokuapp.com/static/#/tests/menu-tests
* https://terra-core-deployed-pr-702.herokuapp.com/static/#/tests/menu-item-tests
* https://terra-core-deployed-pr-702.herokuapp.com/static/#/tests/menu-item-group-tests

@cerner/terra


packages/terra-menu/src/MenuItem.scss
```diff
@@ -0,0 +1,40 @@
+/* stylelint-disable selector-class-pattern  */
+
+:local {
+  .item {
+    cursor: pointer;
+    padding: 0.286rem 0;
+
+    &:hover,
+    &:focus {
+      background-color: #ebf6fd;
```
JakeLaCombe: Should it be themed?

packages/terra-menu/src/Menu.scss
```diff
@@ -0,0 +1,104 @@
...
+  .submenu:not(.fullscreen) .close-button {
+    display: none;
+  }
+
+  .list {
+    padding: 6px 0;
+  }
+
+  .main-menu .content {
+    background-color: var(--terra-menu-background-color, #fff);
```
JakeLaCombe: Typically, it's best practice to set the color if you set the background color. We should allow theming for the foreground of the element as well.

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -0,0 +1,139 @@
...
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+   * Title the should be displayed in header
+   */
+  title: PropTypes.string,
+  /**
+   * Callback function for when close button is clicked
```
JakeLaCombe: *back button

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -0,0 +1,139 @@
...
+    this.buildHeader = this.buildHeader.bind(this);
+    this.isSelectable = this.isSelectable.bind(this);
+  }
+
+  getChildContext() {
+    return { isSelectableMenu: this.isSelectable() };
+  }
+
+  isSelectable() {
+    for (let i = 0; i < this.props.children.length; i += 1) {
```
JakeLaCombe: Minor:

It'd be a great coding win if we can use the some function here. I think it's applicable. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -0,0 +1,139 @@
...
+    ) : null;
+
+    const backIcon = <IconLeft tabIndex="0" />;
+    const backButton = this.props.index > 0 ? (
+      <button className={cx(['header-button'])} onClick={this.props.onRequestBack}>
+        {backIcon}
+      </button>
+    ) : null;
+
+    const titleElement = <h1 className={cx(['header-title'])}>{this.props.title}</h1>;
```
JakeLaCombe: Minor:

Heading module here? I know there's been a bit of a push to start using that.

packages/terra-menu/src/_MenuUtils.js
```diff
@@ -0,0 +1,57 @@
+const getPopupHeight = (contentHeight) => {
+  if (contentHeight <= 40) {
```
JakeLaCombe: We can get rid of this if statement with the Math.ceil and Math.max methods. Perhaps we can switch over to using those?

packages/terra-menu/src/_MenuWidths.js
```diff
@@ -0,0 +1,10 @@
+const MenuWidths = {
```
JakeLaCombe: From how this is used, perhaps this should be an array instead of a hash? I don't see a lot of value of having as a hash atm.

packages/terra-menu/src/Menu.jsx
```diff
@@ -0,0 +1,210 @@
...
+   */
+  targetRef: PropTypes.func.isRequired,
+  /**
+   * Bounding container for the menu, will use window if no value provided.
+   */
+  boundingRef: PropTypes.func,
+  /**
+   * CSS classnames that are append to the arrow.
+   */
+  classNameArrow: PropTypes.string,
```
jmsv6d: This is copied from the popup api. Since it'll be passed through, I don't think it would make sense for the api's to differ

packages/terra-menu/src/Menu.jsx
```diff
@@ -0,0 +1,210 @@
...
+   * Otherwise, the menu will display without an arrow and right aligned.
+   */
+  isArrowDisplayed: PropTypes.bool,
+};
+
+const defaultProps = {
+  isArrowDisplayed: false,
+  children: [],
+  isOpen: false,
+  contentWidth: '240',
```
jmsv6d: No. The popup expects a string with just the number of pixels.

packages/terra-menu/src/Menu.scss
```diff
@@ -0,0 +1,104 @@
...
+    width: 100%;
+
+    .header-title {
+      color: var(--terra-menu-submenu-header-title-color, #1c1f21);
+      font-size: var(--terra-menu-submenu-header-title-font-size, 12px);
+      font-weight: normal;
+      line-height: var(--terra-menu-submenu-header-title-line-height, 1.1667);
+      margin: var(--terra-menu-submenu-header-title-margin, 7px 0 7px 0);
+      overflow: hidden;
+      text-overflow: ellipsis;
```
jmsv6d: I'll have to leave that up to @neilpfeiffer. During initial design discussions we decided that the title could be truncated since the user would see the full string on the previous menu page, where the item was clicked. I don't have a problem with making it opt-in if UX agrees.

packages/terra-menu/src/_MenuWidths.js
jmsv6d: I'm using a hash for this so that I don't have to convert between string and number, since the popup api requires a string and I need the number for fullscreen calculations.

packages/terra-menu/docs/README.md
```diff
@@ -0,0 +1,125 @@
...
+import Menu from 'terra-menu';
+
+class BasicMenu extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleButtonClick = this.handleButtonClick.bind(this);
+    this.handleRequestClose = this.handleRequestClose.bind(this);
+    this.setButtonNode = this.setButtonNode.bind(this);
+    this.getButtonNode = this.getButtonNode.bind(this);
+    this.handleToggle1OnClick = this.handleToggle1OnClick.bind(this);
```
jmsv6d: Sure.

packages/terra-menu/docs/README.md
```diff
@@ -0,0 +1,125 @@
...
+    this.setState(prevState => ({ toggle2Selected: !prevState.toggle2Selected }));
+  }
+
+  handleOnChange(event, index) {
+    this.setState({ groupSelectedIndex: index });
+  }
+
+  render() {
+    return (
+      <div style={{ display: 'inline-block' }} ref={this.setButtonNode}>
```
jmsv6d: Removing it

packages/terra-menu/src/_MenuContent.jsx
jmsv6d: I'll have to look at it more. I'm not sure there's an easy way to use the heading component here unless I actually override the styles it uses (which really defeats the purpose). The header has a different styles when the menu is full screen (slightly larger, and size is rem instead of px). There would be a fair amount of refactoring that would have to happen to make it so that the _MenuContent would know if the menu is fullScreen or not.

packages/terra-menu/.npmignore
```diff
@@ -0,0 +1,3 @@
+*.log
```
emilyrohrbough: Add
`target
reports
tests`

packages/terra-menu/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "Menu",
+    "UI"
+  ],
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
```
emilyrohrbough: `terra-toolkit`  and `terra-mixins` can be removed.

packages/terra-menu/package.json
```diff
@@ -0,0 +1,66 @@
...
+    "terra-slide-group": "^1.1.0"
+  },
+  "scripts": {
+    "compile": "npm run compile:clean && npm run compile:build",
+    "compile:clean": "$(cd ..; npm bin)/rimraf lib",
+    "compile:build": "$(cd ..; npm bin)/babel src --out-dir lib --copy-files",
+    "lint": "npm run lint:js && npm run lint:scss",
+    "lint:js": "$(cd ..; npm bin)/eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
+    "lint:scss": "$(cd ..; npm bin)/stylelint src/**/*.scss",
+    "props-table": "props-table ./src/List.jsx ./src/SingleSelectList.jsx ./src/MultiSelectList.jsx --out-dir ./docs/props-table",
```
emilyrohrbough: oops, these need updated to the menu components

packages/terra-menu/src/MenuDivider.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import classNames from 'classnames/bind';
+import 'terra-base/lib/baseStyles';
+import styles from './Menu.scss';
+
+const cx = classNames.bind(styles);
+
+
```
emilyrohrbough: Nit: extra line. Also cx('divider') will result in the same style-binded classname generation.

packages/terra-menu/src/MenuDivider.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
```
emilyrohrbough: Is this component ever expected to be used outside of Menu? I am curious if it should be an internal component.

packages/terra-menu/src/MenuItemGroup.jsx
```diff
@@ -0,0 +1,53 @@
...
+};
+
+class MenuItemGroup extends React.Component {
+  getChildContext() {
+    return { isGroupItem: true };
+  }
+  render() {
+    const { children, onChange, ...customProps } = this.props;
+
+    const attributes = Object.assign({}, customProps);
```
emilyrohrbough: nit: could we just spread customProps directly on SingleSelectList.Item?

packages/terra-menu/src/_MenuContent.jsx
emilyrohrbough: One thing to consider is <Heading level={1}> does not apply the same text size as defaulted by the h1 (before any custom styling). UX feedback has been requested for this difference for implementing Heading in the demographic banner.

packages/terra-site/src/examples/menu/BasicMenu.jsx
```diff
@@ -0,0 +1,123 @@
...
+            text="Nested Menu 2"
+            key="Nested2"
+            subMenuItems={[
+              <Menu.Item text="Default 2.1" key="2.1" />,
+              <Menu.Item text="Default 2.2" key="2.2" />,
+              <Menu.Item text="Default 2.3" key="2.3" />,
+            ]}
+          />
+          <Menu.Divider key="Divider2" />
+          <Menu.Item text="Alert Action" key="Action1" onClick={() => alert('Action 1')} />
```
emilyrohrbough: It looks like these triggered alerts triggered by onChange is causing Travis to fail.

packages/terra-menu/tests/nightwatch/menu-item-group/menu-item-group-spec.js
```diff
@@ -0,0 +1,30 @@
+/* eslint import/no-extraneous-dependencies: ["error", {"devDependencies": true}] */
+
```
emilyrohrbough: adding `// eslint-disable-next-line import/no-extraneous-dependencies` to you spec files will remove the following error seen in Travis log;

>  3:20  error  'terra-toolkit' should be listed in the project's dependencies. Run 'npm i -S terra-toolkit' to add it  import/no-extraneous-dependencies

packages/terra-menu/src/MenuDivider.jsx
jmsv6d: This is component would be used by consumers when the menu items need to be separated into sections. I had considered making a Menu.Section component for this, but then there would be css logic needed to hide the top divider if the section is at the beginning of the list, hide the bottom divider if the section is at the end, and hide one if there are two sections next to each other. And that also complicates the height logic.

packages/terra-menu/src/_MenuContent.jsx
jmsv6d: It turns out that the some function doesn't work when there's only one child element.

packages/terra-menu/README.md
```diff
@@ -0,0 +1,26 @@
+# Terra Menu
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-menu.svg)](https://www.npmjs.org/package/terra-menu)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+The menu is a popup component that displays a list of items, item groups, and dividers. Menu Items can be actionable, have toggle-style selection, or have nested submenu items. Menu Item groups are a sigle-select grouping of menu items.
```
JakeLaCombe: typo here: sigle

packages/terra-menu/src/Menu.jsx
```diff
@@ -0,0 +1,215 @@
...
+import MenuUtils from './_MenuUtils';
+import styles from './Menu.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+   * List of Menu.Item(s)/Menu.ItemGroup(s)/Menu.Divider(s) to be displayed as content within the Menu.
+   */
+  children: PropTypes.node.isRequired,
```
emilyrohrbough: Should this be `PropTypes.array.isRequired`?

packages/terra-site/src/examples/menu/Index.jsx
```diff
@@ -0,0 +1,51 @@
...
+import BasicMenu from './BasicMenu';
+import MenuBounded from './MenuBounded';
+import MenuWidths from './MenuWidths';
+
+const MenuExamples = () => (
+  <div>
+    <div id="version">Version: {version}</div>
+    <Markdown id="readme" src={ReadMe} />
+    <h2>Menu</h2>
+    <PropsTable id="menu-props" src={MenuSrc} />
```
emilyrohrbough: PropsTable has been updated recently. You should be able to directly pass in the component name with the `componentName="Menu"` props.

packages/terra-menu/src/Menu.jsx
```diff
@@ -0,0 +1,215 @@
...
+  classNameOverlay: PropTypes.string,
+  /**
+   * Should the menu be presented as open.
+   */
+  isOpen: PropTypes.bool,
+  /**
+   * A string representation of the width in px, limited to:
+   * 160, 240, 320, 640, 960, 1280, 1760
+   */
+  contentWidth: PropTypes.oneOf(Object.keys(MenuWidths)),
```
emilyrohrbough: In the propsTable markdown, this proptype is undefined. I wonder if this is something we log as an issue with propsTable or if we just want to write out `PropTypes.oneOf(160, 240, 320, 640, 960, 1280, 1760)`

packages/terra-menu/src/Menu.jsx
```diff
@@ -0,0 +1,215 @@
...
+  /**
+   * Indicates if the menu should have an center aligned arrow displayed on dropdown.
+   * Otherwise, the menu will display without an arrow and right aligned.
```
emilyrohrbough: How come with or without and arrow is not either right or center aligned?

packages/terra-menu/src/_MenuContent.jsx
JakeLaCombe: That's strange. At any rate, thanks for attempting it!

packages/terra-menu/src/Menu.scss
```diff
@@ -0,0 +1,104 @@
...
+      white-space: nowrap;
+      word-break: normal;
+      word-wrap: normal;
+    }
+
+    .header-button {
+      background: 0 0;
```
jmsv6d: The default background color is gray for buttons

packages/terra-menu/src/Menu.jsx
jmsv6d: I don't care much either way, I just thought that are standard was to use prop type node for children.

packages/terra-menu/src/Menu.jsx
```diff
@@ -0,0 +1,215 @@
...
+    const content = (
+      <MenuContent
+        key={`MenuPage-${index}`}
+        title={item.props.text}
+        index={index}
+        onRequestBack={this.pop}
+        onRequestClose={this.props.onRequestClose}
+        onRequestNext={this.push}
+      >
+        {item.props.subMenuItems}
```
emilyrohrbough: I am curious, what is the difference between item.props.subMenuItems here vs this.props.children in the forEach loop of the getContentHeight? I'm missing it.

packages/terra-menu/src/Menu.jsx
jmsv6d: There are two menu styles defined by UX: a center aligned arrow or right aligned without an arrow. I'm completely open to other ways to define the two styles, I just couldn't think of a succinct way to do it.

packages/terra-menu/src/Menu.jsx
```diff
@@ -0,0 +1,215 @@
...
+
+    const arrowClass = cx([
+      'arrow',
+      { submenu: isSubMenu },
+      classNameArrow,
+    ]);
+
+    return (
+      <Popup
+        {...attributes}
```
emilyrohrbough: this could be spread directly with {...customProps} with reassignment to attributes

packages/terra-menu/src/Menu.scss
```diff
@@ -0,0 +1,104 @@
+:local {
+  /* stylelint-disable selector-class-pattern */
+  .submenu.arrow[data-align-bottom='true']::after {
```
emilyrohrbough: I am curious, how does this outputted by css-loader?

packages/terra-menu/src/Menu.jsx
jmsv6d: The content height is calculated based on the items in the menu, it is not recalculated for submenus. This function accepts a Menu.Item, creates a new menu content with that item's subMenuItems and adds that new menu content to the stack so that it is displayed instead of the previous menu content.

packages/terra-menu/src/MenuItem.jsx
```diff
@@ -0,0 +1,145 @@
...
+class MenuItem extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+    this.wrapOnClick = this.wrapOnClick.bind(this);
+    this.wrapOnKeyDown = this.wrapOnKeyDown.bind(this);
+    this.handleSelection = this.handleSelection.bind(this);
+    this.state = { isSelected: props.isSelected && props.isSelectable && !context.isGroupItem };
+  }
+
+  handleSelection(event) {
```
emilyrohrbough: Should handleSelection check `!context.isGroupItem`?

jmsv6d: Doesn't need to. That check is being done before wrapping onClick and onKeyDown on this line https://github.com/cerner/terra-core/pull/702/files/3e26dbfcb281f12b23fb3bf98ed2bf60a5e1dcac#diff-29f754ea95dc722a29df6d9782f5cca6R107

emilyrohrbough:
I have noticed a few things.
1- Tabbing through list items, at the end of the list, the next focused item is the menu container then the first item
2 - On space or enter, items are selected but for items that indicated submenus, these submenus are not shown
3 - When menu is open and the page is resized it moves around a bit then is dropped from the screen. Not sure if it should remain open on resize or not

packages/terra-menu/src/MenuItem.jsx
```diff
@@ -0,0 +1,145 @@
...
+
+    const { isGroupItem, isSelectableMenu } = this.context;
+
+    const attributes = Object.assign({}, customProps);
+    attributes.tabIndex = '0';
+    // This is passed down by the single select list in group item and not needed
+    delete attributes.hasChevron;
+
+    if (isSelectable && !isGroupItem) {
+      attributes.onClick = this.wrapOnClick;
```
emilyrohrbough: I am curious, would we want to change this to onChange to be consistent with table and list? I would expected we would want the same callback for both actions

packages/terra-menu/src/MenuItem.jsx
emilyrohrbough: Actually would we want to check for `isSelectableMenu`?

packages/terra-menu/src/_MenuContent.jsx
emilyrohrbough: Did you try https://facebook.github.io/react/docs/react-api.html#react.children.map? This is what we used in Selectable Table and List.

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -0,0 +1,139 @@
...
+    return (event) => {
+      this.props.onRequestNext(item);
+
+      if (onClick) {
+        onClick(event);
+      }
+    };
+  }
+
+  buildHeader() {
```
emilyrohrbough: Is this the same function that builds subheader menus? I'm not sure the tabIndex='0' is applying correctly for icons

packages/terra-menu/src/MenuItemGroup.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+class MenuItemGroup extends React.Component {
+  getChildContext() {
+    return { isGroupItem: true };
+  }
+  render() {
+    const { children, onChange, ...customProps } = this.props;
+
+    return (
+      <SingleSelectList.Item
```
emilyrohrbough: Would we want to default the isSelectable behavior to `isSelectable`? It seems the would be the expected behavior of items in menus. @bjankord @mjhenkes

mjhenkes:
Should just clicking on the header of the sub menu take you back to the parent menu? the little arrow seems like a small click target.
Could you add examples of long strings in a menu, to display text wrapping or truncation.

packages/terra-menu/src/_MenuUtils.js
jmsv6d: I've looked at this some. And maybe I'm missing something obvious, but I'm not seeing how those could be used here to get rid of the if statement.

packages/terra-menu/.npmignore
jmsv6d: Added here: https://github.com/cerner/terra-core/pull/702/commits/08dd03e6617951d5c381426e92d20ea41019af5d

packages/terra-menu/package.json
jmsv6d: Removed here: https://github.com/cerner/terra-core/pull/702/commits/08dd03e6617951d5c381426e92d20ea41019af5d

packages/terra-menu/src/MenuItem.jsx
jmsv6d: Add an onChange callback here: https://github.com/cerner/terra-core/pull/702/commits/cace25bb7b886c5df972f65a359cfedfa5a1cb3f

packages/terra-menu/src/_MenuUtils.js
JakeLaCombe: I was thinking this
```
Math.max(Math.ceil(contentHeight / 40) * 40, 880)
```

jmsv6d:
@emilyrohrbough 
1. That tabbing behavior happens because the Popup sets `tabIndex="0"` on the PopupContent. So if we don't want that behavior it needs to be changed in Popup
2.  Fixed here: https://github.com/cerner/terra-core/pull/702/commits/cace25bb7b886c5df972f65a359cfedfa5a1cb3f
3. Resize functionality should determined by the Popup component as well.

packages/terra-menu/src/_MenuUtils.js
jmsv6d: That would always return 880, or else an invalid popup height greater than that.

JakeLaCombe: Sorry! I was thinking Math.min here. It looks like you want to return a height that is a multiple of 40 up to 880, correct? If so, I think the min function would work here.

jmsv6d: Okay. But it's only multiples of 40 up to 160, then it's multiples of 80. I could do an if statement for those two cases which would still greatly reduce the lines of code.

jmsv6d: I mean it's still multiples of 40 at that point, but not all multiples of 40 are valid beyond 160.

JakeLaCombe: > Okay. But it's only multiples of 40 up to 160, then it's multiples of 80. I could do an if statement for those two cases which would still greatly reduce the lines of code.

Let's go with this approach. If it reduces it by a lot, I still think it's worth it.

packages/terra-menu/src/Menu.scss
jmsv6d: `[dir] ._-terra-menu-lib-___Menu__submenu___3nHR-._-terra-menu-lib-___Menu__arrow___xOZjo[data-align-bottom='true']::after`

packages/terra-menu/src/_MenuUtils.js
JakeLaCombe: Thanks! That looks good!

jmsv6d:
@mjhenkes 
I agree that the back arrow is small. If I made the change so that clicking anywhere on the header would go back, how would you expect it to behave when the menu is fullScreen and the header also has a close button? The icons and header are larger for the fullScreen menu so it does have larger click targets.
Also, I do have a Menu.Item nightwatch test that shows the text wrapping. I'll add another one that shows text truncating in the submenu header. I didn't put these in the site examples because after a discussion with @neilpfeiffer and @dkasper-was-taken the ideal situation would be for an appropriate width to be selected when designing the component so that the text wouldn't wrap, since I can only accurately calculate height if the text doesn't wrap, without doing multiple renders. I figured that I'd just have the edge case scenarios in nightwatch test and just have the happy path scenarios in the site examples. I can add some documentation on the limitations of this component as well.

packages/terra-menu/src/Menu.jsx
emilyrohrbough: I initially saw this while reading down through the PR. I actually thing node is good.

mjhenkes:
@jmsv6d I would expect that clicking anywhere in the header except on the x button would take you back. But before we make any changes I would like @neilpfeiffer's input. This could be solvable in other ways.

As a consumer of the doc, I think an example showing how text wrapping works in the menu would be appropriate. Nothing extravagant but something to show how it would work.

neilpfeiffer:
Thanks @mjhenkes and @jmsv6d -
I'll take a look today to see about the wrapping and what the hit/target areas are or what we might need to adjust, but I'm initially thinking the same as Matt with the full arrow+title being one target separated from the X.

mhemesath:
It's a little weird that I can't scroll when a menu is open. Is that a side-effect of the popover?

packages/terra-menu/src/Menu.jsx
```diff
@@ -0,0 +1,214 @@
...
+      if (child.props.children && child.props.children.length > 0) {
+        itemCount += child.props.children.length;
+      } else if (child.type === <MenuDivider />.type) {
+        dividerCount += 1;
+      } else {
+        itemCount += 1;
+      }
+    });
+
+    return (itemCount * MENU_ITEM_HEIGHT) + (dividerCount * MENU_DIVIDER_HEIGHT) + MENU_PADDING_TOP + MENU_PADDING_BOTTOM;
```
mhemesath: this make this component not themeable in any way.  Beyond that, what happens if we increase accessibility font-size? We really need to find a way for popups to fit to their content. I'd be willing to work with someone on this if they can explain to me current complexities.

jmsv6d: It doesn't necessarily prevent theming. This is really just a best guess for the proper height, since if the consumer doesn't select a width that prevents wrapping the height wouldn't be calculated accurately either. In either of the cases the menu content within the popup will scroll, while not ideal, it is still accessible. If you want to discuss the limitations of popup @dkasper-was-taken is your guy.

jmsv6d:
@mhemesath yes, the popup was designed with a clear overlay behind it that prevents scrolling while it is open, which was our intention. This is how we have our legacy popup component designed as well.

mhemesath:
@jmsv6d That makes sense for popups.. but not for menus and dropdowns and things that use popup but aren't functionally a popup. 

From an interface perspective, it just feels buggy to have scrolling blocked when a menu is open.

jmsv6d:
I can see your point. Particularly in a menu bar situation where the menu is being displayed from a fixed location, I think allowing the other content to scroll wouldn't be a unreasonable expectation. Perhaps we should discuss in the war room.

packages/terra-menu/src/_MenuContent.jsx
jmsv6d: For future reference. Any of the array functions won't work properly when there is only one child element, because when only one element is passed as a child it is actually of type element and not array. This is why it is best to use React.Children functions for children : https://facebook.github.io/react/docs/react-api.html#react.children

packages/terra-menu/src/Menu.jsx
dkasper-was-taken: @mhemesath The fixed sizes for popups prevent the need for multiple renders and multiple re-positions when dealing with popups bounded by screen/bounding rects.
There are also technical limitations when used with headers/flex layouts. If the consumer wants a header, and we are changing the size of its parent container when constrained, it will break the content’s scrolling. 

The general thought from UX was that the content in the popup should be responsive, rather than the popup respond to the content. It also gives UX a common toolset to leverage for consistent popup sizing.

To summarize, if the consumer wants the popup to by dynamically sized:
 - It will be forced to render multiple times
 - Popup content cannot use flex positioning or sizing using percentages
 - If consumers provide their own set content size to allow for flex/percentage sizing, we cannot realistically manage bounding conflicts.

mhemesath: > Popup content cannot use flex positioning or sizing using percentages
I think this is the assumption introducing the complexity. Not all popups need flex positioning, so forcing it is causing a lot of downstream complexities where they don't need to exist. 

It's fine having popups with fixed sizes for the situations that need them, but I see no reason for not allowing popups to dynamically fit their content (perhaps to a max-height/max-width) for sutations that don't need them

neilpfeiffer:
@jmsv6d - I just reviewed the functionality of the menu header and back actions, and would like to make an update so the header-title has the same action as the arrange-fitStart (arrow icon), so it either _is_ one hit target - or perceived as one large target (with the text included with the icon).

So end result will be the entire menu header will be actionable, and there should end up being two perceivable hit targets, the back action ("arrow" and title) and the close "X" 

I'll be updating the visuals to match:
menu-backplustitle-target

packages/terra-menu/src/Menu.jsx
bjankord: +1 for node.

packages/terra-menu/src/Menu.jsx
bjankord: Lets write it out as you mentioned @emilyrohrbough. That matches what we did in the `<Text />` component as well: https://github.com/cerner/terra-core/blob/master/packages/terra-text/src/Text.jsx#L29

packages/terra-menu/src/Menu.scss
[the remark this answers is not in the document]
bjankord: Are you wanting to adjust the background position or the background color?

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -0,0 +1,171 @@
...
+      if (onKeyDown) {
+        onKeyDown(event);
+      }
+    });
+  }
+
+  buildHeader() {
+    const closeIcon = <IconClose />;
+    const closeButton = this.props.onRequestClose ? (
+      <button className={cx(['header-button', 'close-button'])} onClick={this.props.onRequestClose}>
```
emilyrohrbough: Does this allow for keyboard navigation? I am wondering if we want onChange

emilyrohrbough:
IMO things look pretty good on this PR. Two things:
1. total nit: add periods after prop descriptions.
2. I am wondering if we want to add a few more onChange, onClick and/or onKeyDown tests in nightwatch to ensure the desired behavior.

packages/terra-menu/src/_MenuContent.jsx
jmsv6d: Yes keyboard navigation is built into button components. What kind of use case are you thinking that onChange would be needed?

emilyrohrbough: I was wondering if the close button allowed for keyboard changes. I just tested it and all seems to work, however, I am noticing when I navigate to a submenu from the keyboard, the sub menu shows, but does not gain focus until the remaining main menu has been tabbed through.

packages/terra-menu/src/Menu.scss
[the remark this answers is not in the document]
jmsv6d: Honestly, since the callouts for this component said same as popup I just copied the popup styles for the header and buttons. It really just needs to have a transparent background color.

packages/terra-menu/src/_MenuContent.jsx
jmsv6d: Yes. That's an issue with the slide-group. It's logged here: https://github.com/cerner/terra-core/issues/647

mmalaker:
Functional verification completed. Menu examples display and function correctly with the exception of keyboard navigation. Keyboard navigation will be addressed in issue #784.

bjankord:
@neilpfeiffer This is ready for UX review.

neilpfeiffer:
Tested: PR #702
Resolves: Issue #420

**Reviewed:**
- Follows pattern considerations documented in the current Cerner Design Standards: [Menus](https://wiki.ucern.com/display/UserExperience/Menus)
- Follows UX defined design implementation for [Terra Menus](https://wiki.ucern.com/x/aMjMY) 
- Follows all defined theme designs: terra-legacy-theme

**Fix with Upcoming Issues:**
- Added an additional enhancement to issue #784 : the Action and Close Actions respond for mouse press but not keyboard “enter/return” in menu examples
- Styling of Header area has some UI variance between the bounded and unbounded version (text, back icon and close icon sizes and positions). Probably need to look at underlying integration with Popup to fix, will log issue to investigate.


**Pass** |
-----------|
UX Review|

Issue #337, opened 2017-05-10
Add document on 3rd party libraries


Matt-Butler:
# Issue Description
Add a document with the 3rd party libraries we use in Terra.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

## Expected Behavior
Documentation is available with the 3rd party libraries we use in Terra.

## Current Behavior
N/A

Matt-Butler:
react-portal
focus-trap-react
react-datepicker
react-intl
moment
react
react-dom
prop-types
classNames
resize-observer-polyfill

mask
react-docgen

Matt-Butler:
| dependency               | version | react version      | description                                                                                                                                                              |
|--------------------------|---------|--------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| classNames               | ^2.2.5  | --                 | A simple utility for conditionally joining classNames together                                                                                                           |
| focus-trap-react         | ^3.0.2  | 0.14.x || ^15.0.0  | A React component that traps focus                                                                                                                                       |
| intl                     | ^1.2.5  | --                 | Polyfill the ECMA-402 Intl API (except collation)                                                                                                                        |
| marked                   | ^0.3.6  | --                 | A markdown parser built for speed                                                                                                                                        |
| moment                   | ^2.17.1 | --                 | Parse, validate, manipulate, and display dates                                                                                                                           |
| prop-types               | ^15.5.8 | --                 | PropType validation extracted from React                                                                                                                                 |
| react                    | ^15.4.2 | --                 | React is a JavaScript library for building user interfaces.                                                                                                              |
| react-datepicker         | ^0.46.0 | ^0.14.0 || ^15.0.0 | A simple and reusable datepicker component for React                                                                                                                     |
| react-docgen             | 2.13.0  | --                 | A CLI and toolkit to extract information from React components for documentation generation.                                                                             |
| react-dom                | ^15.4.2 | --                 | React package for working with the DOM.                                                                                                                                  |
| react-intl               | ^2.3.0  | ^0.14.9 || ^15.0.0 | Internationalize React apps. This library provides React components and an API to format dates, numbers, and strings, including pluralization and handling translations. |
| react-portal             | ^3.0.0  | --                 | React component for transportation of modals, lightboxes, loading bars... to document.body                                                                               |
| resize-observer-polyfill | ^1.4.1  | --                 | A polyfill for the Resize Observer API                                                                                                                                   |

[...]

Issue #20, opened 2017-02-24
Look into drying up the npm-scripts used in each package


bjankord:
# Issue Description
There are duplicate npm-scripts used across all terra-core packages. @ryanthemanuel had some ideas on ways we could reduce this duplication. Some options to look into are having all the scripts point to a .js file in the root directory for each script and using shelljs to run tasks. Another option is to investigate https://www.npmjs.com/package/npm-run to run tasks.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Duplicate npm-scripts are reduced

## Current Behavior
Duplicate npm-scripts are in every package.json

bjankord:
link-parent-bin could help with some cleanup: https://www.npmjs.com/package/link-parent-bin

bjankord:
We could fork react-scripts and explore adapting it to our use-cases

emilyrohrbough:
Creating a 'terra-scripts' package or something similar with

```js
"bin": {
    "terra-scripts": "./bin/teraa-scripts.js"
  },
```

in the package.json will create a symlink to `terra-scripts.js` in the `node_modules/.bin directory` on npm install. 

This project, [kyt-coe/cli](https://github.com/NYTimes/kyt/tree/master/packages/kyt-core/cli), is a simple example of how we could dry up our scripts will still providing arguments for the scripts that need it.

bjankord:
Closing this out, I no longer think this is something we need to prioritize.

PR #708, opened 2017-08-08
Dependency Markdown Script

emilyrohrbough:
### Summary
This PR adds the `DependencyMarkdownGenerator.js` script to the repository to generate dependency documentation for terra-core and each of its packages.

This script aggregates the list of dependencies, devDepenencies and peerDependencies listed in each `package.json` and then generates markdown with the following information:
1. dependency name
2. dependency version specified in the package.JSON
3. supported react version (if applicable) 
4. dependency description

This information satisfies the requests made in #337 to generated third party documentation.

#### Note
Currently this PR is for the script itself. Once there are enough reviews, I will run the script and add the generated `DEPENDENCIES.md` files.

### Additional Discussion
@bjankord and I discussed online that this script would be a candidate to be a standalone `terra-script` which would be closely tied to issue #20.


emilyrohrbough:
Here is an example of the generated markdown:

screen shot 2017-08-08 at 4 28 26 pm

mjhenkes:
Should we use this as the first script to add to Terra Toolkit?

bjankord:
I lean towards completing this quickly in core and then move it to toolkit later.

Issue #447, opened 2017-06-12
Break up large example pages into subpages


mmalaker:
# Issue Description
Some of the Terra example pages (like forms) are getting too large to be usable. Recommend breaking the separate examples into subpages.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
For components with many examples, each example has its own subpage.

## Current Behavior
Consumers have to scroll through explanatory text, code, and examples on a single page.

emilyrohrbough:
The components that should be broken out:
- [ ] Form 
- [ ] Icons - static vs themeable
- [ ] List - list item, list, single select list, multi select list
- [ ] Overlay - overlay container, overlay, loading overlay
- [ ] Table - Table,  Header vs Subheaders, Header Cell vs Cell, Rows vs Row, Selectable Table

mmalaker:
In addition to the above, I'd also like to break up these core component examples:
* Alert - basic, basic with actions/dismissible, and custom/build your own.
* Popup - default and arrow, attachments, size variants, and special popups: bounded, class styles, modal
* Slide Panel - overlay, squish, start, end, sizes, and fill

bjankord:
We'll address the remaining pages in individual issues.

Issue #697, opened 2017-08-03
Terra Site to use squished Slide Panel


benbcai:
# Issue Description
The RTL/LTR and locale switchers are located at the upper left corner of the [engineering doc site](http://engineering.cerner.com/terra-core/#/site). When a selection is made and to see the change, you would then have to scroll to the component examples at the bottom of the page. To make the doc site easier to use, a squished slide panel can be used in terra-site so that the main content and panel content can scroll independently.

## Issue Type
- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The list of components on the left side and the selected component content on the right side should scroll independent of each other.

## Current Behavior
Both the list of components and component content scroll together.

PR #709, opened 2017-08-08
Create Separate Pages for Form Components

emilyrohrbough:
### Summary
The terra-form site page was very long and difficult to use. This PR breaks out each form component's documentation and examples into their own sub-pages which are linked from the terra-form site page. Now the terra-form page gives navigations to the subcomponents and displays a full example.

Addresses issue #447.

@cerner/terra


packages/terra-site/src/examples/form/Index.jsx
```diff
@@ -1,122 +1,36 @@
...
-      <Button text="Select Field" variant="link" onClick={() => { scrollToComponent('#select-field'); }} />
-    </div>
-    <div>
-      <Button text="Controlled Input Example" variant="link" onClick={() => { scrollToComponent('#controlled-input-example'); }} />
+    <div className={styles['form-markdown']}>
+      <h1>Component Information</h1>
+      <p> Below are links to further documentation and implementaiton of each form component.</p>
+      <List className={styles['site-nav']}>
+        <List.Item content={<Link to="site/form/field">Field</Link>} />
+        <List.Item content={<Link to="site/form/fieldset">Feildset</Link>} />
```
JakeLaCombe: *Fieldset

packages/terra-site/src/site.scss
```diff
@@ -9,6 +9,24 @@
...
+
+  // to match the markdown styling
+  .form-markdown {
+    h1 {
+      border-bottom: 1px solid #eaecef;
+      padding-bottom: 0.3rem;
+    }
+
+    p {
+      font-size: 16px;
```
JakeLaCombe: Can we switch to em's here?

JakeLaCombe:
From looking at the PR, it would be great if we could always have the sidebar navigation present. I don't like how I have to click the upper left click both times to get to the homepage.

If we can't get the secondary navigation always present, perhaps we can use modals to display all of the form examples?

packages/terra-site/src/Index.jsx
```diff
@@ -84,83 +86,86 @@ import ToggleButtonTestRoutes from 'terra-toggle-button/tests/nightwatch/ToggleB
...
+        <Route path="app-delegate" component={AppDelegateExamples} />
+        <Route path="arrange" component={ArrangeExamples} />
+        <Route path="badge" component={BadgeExamples} />
+        <Route path="base" component={BaseExamples} />
+        <Route path="button" component={ButtonExamples} />
+        <Route path="button-group" component={ButtonGroupExamples} />
+        <Route path="content-container" component={ContentContainerExamples} />
+        <Route path="datepicker" component={DatePickerExamples} />
+        <Route path="demographics-banner" component={DemographicsBannerExamples} />
+        <Route path="form" component={FormExamples} />
```
StephenEsser: Adding onto Jake's comment below it may be easier to flatten form elements here. Rather than one form group that contains a chain of drilling into, it would be easier to find specific components if we provided first class links here to Input, Control, Field, etc...

I think there would be value into providing categories and grouping system into this sidebar. Several other UI toolkits group items. We're working on a grouping system for Kaiju and I think the hierarchy in Kaiju should closely match anything we come up with here, I'll post what we've started on. Any feedback would be welcome.

Atoms
- Heading
- Badge
- Image
- Progress Bar
- Text
- Button
- Icons
- Forms
  - Input
  - Control
  - Field
  - Fieldset
  - Select
  - etc

Molecules 
- Search Field
- Arrange
- Button Group
- Content Container
- Toggle
- Toggle Button
- Status

Organisms
- List
- Multi-Select List
- Single-Select List
- Demographics Banner
- Responsive Element
- Slide Panel
- Grid
- Table

JakeLaCombe: I like that idea!

emilyrohrbough:
@StephenEsser @JakeLaCombe I updated the side navigation to display the form components under the form link with a bit of padding to indicate the sub-navigation. Let me know if this is what you might have expected.

JakeLaCombe:
That looks great!

One thing we should do is move the gigantic controlled input form into it's own page IMO. I think the small controlled and uncontrolled sections on the main page is enough to detail the difference between the two.

emilyrohrbough:
@JakeLaCombe I put the full code example in a toggle button so people can open or close if they want to view the source code of the example.

JakeLaCombe:
Looking great! 

My only remaining section is to add a header right after the uncontrolled doc section that says 'Full Functional Example' or something like that.

emilyrohrbough:
I created a new branch and opened PR #718 to track down the Travis failure.

It looks like the issue with the PR was the styling fix I had applied to the site page. Before the css modules conversion, each page of the application had a padding style applied, with the conversion, this styling was lost on the component test links pages and example pages. To fix the styling, I had wrapped the site Router in a div that applied the padding. 

This was the same change I made to terra-clinical's site page, so I am not sure why this caused Travis to fail in terra-core. Regardless, this was an extra nicety that should not hold this PR back, so I remove the code change.

packages/terra-site/src/Index.jsx
```diff
@@ -121,6 +122,7 @@ ReactDOM.render((
       <Route path="time-input" component={TimeInputExamples} />
       <Route path="toggle" component={ToggleExamples} />
       <Route path="toggle-button" component={ToggleButtonExamples} />
+      {FormIndexRoutes}
     </Route>
```
StephenEsser: At some point in the future it might be nice to have the side navigation scroll independently so you don't have to scroll back up after making a selection.

emilyrohrbough: [This issue](https://github.com/cerner/terra-core/issues/697) was logged to handle just that.

PR #718, opened 2017-08-11
Reattempting PR for 'Create Separate Pages for Form Components'

emilyrohrbough:
### Summary
PR #709 keeps failing. This PR is to slowly commit changes to see what is triggering the failure. 

Once a commit passes, the next change will be committed until the issue is pegged.

### DIFF
--- a/packages/terra-site/CHANGELOG.md
+++ b/packages/terra-site/CHANGELOG.md
@@ -5,7 +5,6 @@ Unreleased
 ----------
 ### Changed
 * Break out form component examples into their own packages
-* Update styling lost in CSS Module conversion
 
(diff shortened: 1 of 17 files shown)


emilyrohrbough:
It looks like the issue with the PR was the styling fix I had applied to the site page. Before the css modules conversion, each page of the application had a padding style applied, with the conversion, this styling was lost on the component test links pages and example pages. To fix the styling, I wrapped the site in a div that applied the padding. 

This was the same change I made to terra-clinical's site page, so I am not sure why this is causing Travis to fail in terra-core. Regardless, this was an extra nicety that should not hold this PR back, so I remove the code change.

Issue #313, opened 2017-05-04
Card Component


bjankord:
# Issue Description
Create card component.

## Issue Type

- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Needs expected behavior defined

## Current Behavior
N/A

WhiteLight1107:
# Tech Design for Card
## Summary:
The Card Component is used as container for other divs. 

PR:
TBD

## React Props:
### Card:
| Prop     | Type | Default | Description  |
|----------|------|---------|--------------|
| padding | array or bool | true | option to pass in a single boolean to turn on/off, or an array to specify each side using css short hand https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties |
| children | node | none | Child Nodes |

## Example:
```
import React from 'react';
import Card from 'terra-card';

const CardDefault = () => (
  <div>
    <Card children={"Hello World!!"}/>
</div>
);

export default CardDefault;

```

## CSS Classes:
| Selector | Description |
|----------|------|
| .well-card | Outer Container of the Card |
| .well-card-container | Inner part of the card used to contain children prop |
| .u-pa--sm | 10px padding |
| .u-pv--sm | 10px padding vertical |
| .u-ph--sm | 10px padding horizontal |
| .u-pt--sm | 10px padding top |
| .u-pb--sm | 10px padding bottom |
| .u-pl--sm | 10px padding left |
| .u-pr--sm | 10px padding right |

WhiteLight1107:
UX Designs: https://jira2.cerner.com/browse/PORTALDEV-35541

mhemesath:
CSS classes aren't really needed anymore in tech designs as we localize those in modules. I will say though to remove the well prefix from them.

dkasper-was-taken:
As far as the selectors, they needs to be update to remove the wellness name spacing.
We are no longer using the '--' selector nomenclature.
5px margins are not enforceable as a stand alone component relative, so HOC or usage of list would be needed.
I'm a little uncertain about using a bool & array type, generally multiple typed variables are discouraged.
If we are using array, we want to set a type or shape of the array contents.
The padding itself it currently under discussion if we need a component to provide such padding styles, or if a cerner standard padding should be applied.

jmsv6d:
It looks like the UX designs also have at concept of a selectable card and a selectable list of cards that is missing from this tech design. I also propose renaming this component.
[...]

PR #721, opened 2017-08-14
cards component

WhiteLight1107:
### Summary
Terra-card component completes #313 
### Deployment Links 
TBD


emilyrohrbough:
We won't want to check in the `package-lock.json` file

packages/terra-card/docs/README.md
```diff
@@ -0,0 +1,29 @@
+# Terra Card
+
+{insert description}
```
emilyrohrbough: Add description

packages/terra-card/docs/README.md
```diff
@@ -0,0 +1,29 @@
...
+  - `npm install terra-card`
+  - `yarn add terra-card`
+
+## Usage
+
+```jsx
+import React from 'react';
+import Card from 'terra-card';
+
+<Card {props...} />
```
emilyrohrbough: Could you give an example or two of the expected use?

packages/terra-card/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "terra-card",
+  "main": "lib/Card.js",
+  "version": "0.0.0",
+  "description": "{insert description}",
```
emilyrohrbough: Add description

packages/terra-card/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "terra-base": "^1.0.0"
```
emilyrohrbough: Update to `terra-base: ^1.3.0`

packages/terra-card/docs/README.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+## Component Features
+
+<!-- Uncomment supported features.
```
emilyrohrbough: Uncomment the supported features

packages/terra-card/src/Card.jsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import 'terra-base/lib/baseStyles';
+import './Card.scss';
+
```
emilyrohrbough: To create name-spaced class names for css modules, please update
`import './Card.scss';` to `import styles from './Card.scss';`
 and then add
`const cs = classNames.bind(styles);`

To incorporate this style binding, you might need to update how you are applying classNames for vertical and horizontal padding. [Here](https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx) is an example that would be good to look at for guidance.

packages/terra-card/src/Card.scss
```diff
@@ -0,0 +1,27 @@
+
```
emilyrohrbough: For the css module styling, this file needs updated to be locally scoped. [Here](https://github.com/cerner/terra-core/blob/master/packages/terra-button-group/src/ButtonGroup.scss) is a simple example of what your styling might look like. 

Our team has decided to use css styling syntax for class names, so `.cardContainer` should be changed to `.card-container`.

packages/terra-card/src/Card.jsx
```diff
@@ -0,0 +1,60 @@
...
+  padHorizontal: PropTypes.bool,
+  /**
+   * Child Nodes
+   */
+  children: PropTypes.node,
+};
+
+const defaultProps = {
+  padVertical: true,
+  padHorizontal: true,
```
alex-bezek: the defaults should be no padding per this
https://github.com/cerner/terra-core/issues/313#issuecomment-322214889

packages/terra-card/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Card
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-card.svg)](https://www.npmjs.org/package/terra-card)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+{insert description}
```
jmsv6d: Add description here

packages/terra-card/src/Card.jsx
```diff
@@ -0,0 +1,60 @@
...
+};
+
+const Card = ({
+  padVertical,
+  padHorizontal,
+  children,
+  ...customProps
+}) => {
+  // Follows these conventions https://developer.mozilla.org/en-US/docs/Web/CSS/Shorthand_properties
```
alex-bezek: This comment is no longer applicable as we are just doing the 2 properties instead of the css shorthand style array of values

packages/terra-card/src/Card.scss
```diff
@@ -0,0 +1,27 @@
...
+/* stylelint-disable selector-class-pattern  */
+.card {
```
jmsv6d: You should probably discuss with @neilpfeiffer to determine if any of these values should be themeable.

packages/terra-card/src/Card.jsx
```diff
@@ -0,0 +1,60 @@
...
+    'card',
+    customProps.className,
+  ]);
+
+  if (padVertical) {
+    containerClass += ' veritcalPadding';
+  }
+  if (padHorizontal) {
+    containerClass += ' horizontalPadding';
+  }
```
StephenEsser: ClassNames allows conditional class names. This if statement can be replaced with the classNames API.
Example:
```
  const cardClass = cx([
    'card',
    { 'veritcal-padding': padVertical },
    { 'horizontal-padding': padHorizontal },
    customProps.className,
  ]);
```
( Assuming the containers below are flattened into one contianer )

packages/terra-card/src/Card.jsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import 'terra-base/lib/baseStyles';
+import './Card.scss';
+
+const propTypes = {
+  /**
+   * Provides 10px on top and bottom
```
jmsv6d: Is this padding going to be themeable? If so, you should probably reword this to "Provide themeable padding to top and bottom" same with vertical padding.

packages/terra-card/tests/nightwatch/card-spec.js
```diff
@@ -0,0 +1,19 @@
...
+
+  afterEach: (browser, done) => {
+    screenshot(browser, 'terra-card', done);
+  },
+
+  'Displays a default card': (browser) => {
+    browser
+      .url(`http://localhost:${browser.globals.webpackDevServerPort}/#/tests/card-tests/default`)
+      .assert.elementPresent('.card');
+  },
```
jmsv6d: You aren't testing all of your nightwatch test files.

packages/terra-card/src/Card.jsx
```diff
@@ -0,0 +1,60 @@
...
+  if (padHorizontal) {
+    containerClass += ' horizontalPadding';
+  }
+
+
+  const content = <div className={containerClass}> {children}</div>;
+
+  return (
+    <div {...customProps} className={cardClass}>
+      {content}
```
StephenEsser: I recommend using `children` here directly, styling the existing outer div, and removing the `content` div wrapper unless the double nested div is needed independent of the wrapper the card class is being applied to.

packages/terra-card/src/Card.scss
StephenEsser: +1 I think it is likely the background color will need to be theme-able, which would probably require the box shadow to theme

packages/terra-card/src/Card.scss
```diff
@@ -0,0 +1,27 @@
...
+  background-color: #fff;
+  border-radius: 3px;
+  box-shadow: 0 1px 1px 0 #1c1f21;
+  margin: 10px;
```
StephenEsser: Adding on to Jaime's comment above, I would reach out to @neilpfeiffer to see if we want a constant margin around each Card or if the Component the Card is within should control the spacing

packages/terra-card/src/Card.jsx
jmsv6d: +1 to remove nested content div, if possible.

alex-bezek: The content wrapper div originally was because we were handling clickable cards and it was needed for something. Agreed, let's remove it and can always bring it back later if required for something

packages/terra-card/src/Card.scss
```diff
@@ -0,0 +1,27 @@
...
+}
+
+.cardContainer {
```
alex-bezek: With the removal of the extra div, do we need this as a separate class anymore?

packages/terra-card/src/Card.scss
```diff
@@ -0,0 +1,27 @@
...
+  color: inherit;
+  display: block;
+  height: 100%;
+  text-decoration: none;
```
alex-bezek: I think this was specifically for our clickable card. Is this still required?

packages/terra-card/src/Card.scss
neilpfeiffer: I remember from Friday's conversation that there wouldn't be any margin on the card but would be on the responsibility of card-collection or similar grid based parent. Is that a correction assumption of always the containing component handling layout/placement?

It still could be a themeable option, I suppose in addition, but would need to investigate/validate further.

packages/terra-card/src/Card.scss
neilpfeiffer: I would say all used style props are probably themeable, my list would include:
border, background, box-shadow, padding (based on relative "sizing set" of css-variables for entire theme)

packages/terra-card/src/Card.scss
neilpfeiffer: Ditto - I would be careful of `text-decoration: none` without good reason.

packages/terra-card/src/Card.scss
jmsv6d: I would say we should remove the margin for now and a themeable option could be added later if a need for it occurs.

packages/terra-card/src/Card.scss
alex-bezek: Just curious, are we going to start making most style's themeable right out of the gate, or add in theme-ability as the need comes? We can always add more customizability, but it's harder to take it away

mhemesath: I'd say we only make theme-able what needs to be. Its fairly trivial to expose additional properties.

packages/terra-card/src/Card.jsx
mhemesath: I'm actually in favor of vertica/horizontal padding being a sub-component (or 2 sub-components). This enables us to reuse those padding styles  in sub ocmponents of the card. Example, we need a side-to-side horizontal rule and want to have 2 padded containment regions above and below it.

WhiteLight1107: So should I remove the padding properties all together from card? @mhemesath

PR #722, opened 2017-08-14
Update node and npm to latest major versions

bjankord:
### Summary
Testing updating to node v8 and npm v5.

@cerner/terra


emilyrohrbough:
I noticed on this PR https://github.com/cerner/terra-core/pull/721/files/71af1ebcf529ed24ce0f50b3388ac98f3758e255..1c46f4ca1bc68c795b9297238048c2f4597b4d82 that each packaged also created a few package-lock.json files for different components. Do we need to add `package-lock.json` to each package's gitignore as well?

emilyrohrbough:
Also, do we need to update the node version in the package.json?

alex-bezek:
@emilyrohrbough I don't think each individual package containers a .gitignore as they are all part of the same git repo. Adding the filename `package-lock.json` in the root level .gitignore without a path has the desired affect.

Concerning the engine in package.json I'm assuming this should be updated as well. Looks like consumers that don't use NVM use this to determine the version to use
https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version

.nvmrc
```diff
@@ -1 +1 @@
-6.11.1
+8.3.0
```
JakeLaCombe: It looks like 8.4.0 is released. https://github.com/nodejs/node/tree/v8.4.0

bjankord: Bumping to 8.4.0. https://github.com/cerner/terra-core/pull/722/commits/8072a4cfec1e27b832b6f4a630a82a77f064c938 and https://github.com/cerner/terra-core/pull/722/commits/846f1d8fb93e5f8d9ea1a833f75f97d21ef4affb

PR #728, opened 2017-08-14
fixing generate-example script to create IconStatic.jsx component in terra-site

amf9t2:
### Summary
Running `npm run migrate-cerner-one-icons` is not generating the IconStatic.jsx component in terra-site.

### DIFF
--- a/packages/terra-icon/scripts/src/generate-example/generateStaticIcons.js
+++ b/packages/terra-icon/scripts/src/generate-example/generateStaticIcons.js
@@ -1,12 +1,12 @@
 /* eslint-disable */
 import fs from 'fs';
 
-const outputfile = fs.createWriteStream('../../packages/terra-site/src/examples/icon/IconThemeable.jsx', { flags: 'w' });
+const outputfile = fs.createWriteStream('../../packages/terra-site/src/examples/icon/IconStatic.jsx', { flags: 'w' });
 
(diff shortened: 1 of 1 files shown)


packages/terra-icon/scripts/src/generate-example/generateStaticIcons.js
```diff
@@ -1,12 +1,12 @@
 /* eslint-disable */
 import fs from 'fs';
 
-const outputfile = fs.createWriteStream('../../packages/terra-site/src/examples/icon/IconThemeable.jsx', { flags: 'w' });
+const outputfile = fs.createWriteStream('../../packages/terra-site/src/examples/icon/IconStatic.jsx', { flags: 'w' });
 
 const generateIconAll = iconObjs => new Promise((resolve) => {
 
   const staticIcons = iconObjs.filter(function(iconObj){
```
mhemesath: being [filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) just expects a boolean if the result should be filtered, this can be simplified to 
```js
const staticIcons = iconObjs.filter(icon => !icon.themeable);
```

amf9t2: [updated](https://github.com/cerner/terra-core/pull/728/commits/6d48f214d3368499f958190a381c15355f22af29) Thanks.

PR #732, opened 2017-08-16
ThemeProvider component

### LINKED ISSUE #723: [spike] Investigate our Theming Implementation
# Issue Description
We currently have two approaches that we are looking into with doing our theming approach. There exist a [theme provider](https://github.com/cerner/terra-core/blob/theme-adapter/packages/terra-theme-provider/src/CSSVariableProvider.jsx#L29) branch that takes in a json file of attributes, and uses JavaScript to manually set the styles inside the current CSS. Another approach being investigated is using classes to set the variable names and swapping classes on the theming component to update the values. This approach can be viewed [here](https://www.google.com/search?q=theming+with+css+variables&rlz=1C5CHFA_enUS719US722&oq=theming+with+css+variables&aqs=chrome..69i57j69i60.4583j0j7&sourceid=chrome&ie=UTF-8)

The goal of this spike is to determine what we would like to do in order to implement theming. While we have talked about the two approaches logged up above, our team is open to other approaches that meets our needs as well.

We also should document what browser support we will have to dynamic theming as well. Older browsers don't support css variables, which can make some of our suggested implementations invalid.

bjankord:
### Summary
Resolves #723.

### Deployment Link
https://terra-core-deployed-pr-732.herokuapp.com/static/#/site/theme-provider
https://terra-core-deployed-pr-732.herokuapp.com/static/#/tests/theme-provider-tests

@cerner/terra


packages/terra-badge/src/_mixins.scss
```diff
@@ -2,3 +2,34 @@
   background-color: map-get($color-map, background-color);
   color: map-get($color-map, color);
 }
+
+$default-theme: 'consumer' !default;
+$bundled-themes: 'consumer' 'ocs' !default;
+
+@mixin terra-consumer-theme {
+  @if index($bundled-themes, 'consumer') {
```
mhemesath: should we wrap this in `:global` in the event someone accidentally calls this within `:local`

packages/terra-badge/src/_mixins.scss
```diff
@@ -2,3 +2,34 @@
   background-color: map-get($color-map, background-color);
   color: map-get($color-map, color);
 }
+
+$default-theme: 'consumer' !default;
```
mhemesath: in reality this file wouldn't live in`terra-badge` right?

bjankord: Correct, we'd put this in [terra-mixins](https://github.com/cerner/terra-core/tree/master/packages/terra-mixins) and import it into each component `scss` file similar to how we did bidi mixins. 

I put it here just as a p.o.c. for how these mixins could work.

packages/terra-site/src/ThemeProvider.jsx
```diff
@@ -0,0 +1,21 @@
...
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * Name of class for specified theme. e.g `terra-OSC-theme`
+   */
+  themeName: PropTypes.string.isRequired,
+};
+
+const ThemeProvider = ({ themeName, children, ...customProps }) => {
+  return <div className={themeName}>{children}</div>;
```
JakeLaCombe: If this is all the theme provider has, I would rather us have a prop on the Terra Base that takes in a theme, and places that on the class. This feels too small to be it's own component.

JakeLaCombe: Or document the theme module in a way that tells the consumer just to attach the theme as a class to a component they are using.

mhemesath: >  I would rather us have a prop on the Terra Base that takes in a theme, and places that on the class

I'm fine having this logic in terra base.. but I do think solutions will need the ability for nested sub-themes.

> Or document the theme module in a way that tells the consumer just to attach the theme as a class to a component they are using.
The goal is to not have consumers aware of class names as they are an implementation detail.

alex-bezek: Was discussing with @mhemesath and @zhongyn , updating the theme name would effectively re-render the children which could be the whole site, where really I think it just needs to add the new className, declaring the new variable, and forcing a CSS update.
Could we wrap the children in another component that implements ShouldComponentUpdate, or extends React.PureComponent to avoid re-rendering the whole page on theme change?

StephenEsser: Does this solution work with nested themes?

mhemesath: technically yes.. but nested themes are basically freetexted as nested selectors under the primary theme.

JakeLaCombe: Agreed with not having consumers aware of class names, but I don't know if this approach resolves that though, as themeName corresponds to a class name. I'll see if I can think of other approaches.

bjankord: Good call, I'll take a look at adding that.

bjankord: I created a `<ThemeProvider />` and a `<ThemeProviderChildren />` component. Inside of the `<ThemeProviderChildren />` I used ImmutableJS's is function to compare `props.children`, e.g. 
```js
shouldComponentUpdate(nextProps) {
    // performs a deep equality check if necessary
    return is(this.props.children, nextProps.children);
 }
```
This works well, but I found changeing routes via react-router / react-router's link component doesn't trigger shouldComponentUpdate to return true. There's a way to check routes with previous route [documented here](https://stackoverflow.com/questions/30078260/react-js-shouldcomponentupdate-and-react-router-link) but I'm thinking for now, leaving out the `shouldComponentUpdate` check to keep it simple and compatible with react-router.

packages/terra-badge/src/_mixins.scss
bjankord: Wrapped the `class` and `:root` selector generated by the mixin in `:global` in this commit: https://github.com/cerner/terra-core/pull/732/commits/38794f6d48344762ac35788da9d04a3b16150798

packages/terra-badge/src/_mixins.scss
bjankord: Moved this into terra-mixins in this commit: https://github.com/cerner/terra-core/pull/732/commits/38794f6d48344762ac35788da9d04a3b16150798

packages/terra-site/src/ThemeProvider.jsx
bjankord: After this PR lands, I plan to have `<Base />` wrap the `<ThemeProvider />` component so people can interact with it through Base via `<Base themeName="my-theme">`.

`<ThemeProvider />` could still be nested within `<Base themeName="my-theme">` if people wanted nested theming e.g. 
```js
<Base themeName="my-theme">
  <ThemeProvider themeName="dark">
    <Sidebar content={<SidebarContent/>} />
  </ThemeProvider>
  <Body content={<BodyContent/>} />
</Base>

Comment (bjankord) [inline packages/terra-site/src/ThemeProvider.jsx]: I'm gonna look into this some more today and see if we can use `document.referrer` to check when react-router updates routes.

File: packages/terra-theme-provider/src/ThemeProvider.jsx [hunk 4]
```diff
@@ -0,0 +1,28 @@
...
+  themeName: PropTypes.string,
+};
+
+const ThemeProvider = ({ themeName, children, ...customProps }) => (
+  <div {...customProps} className={themeName}>{children}</div>
+);
+
+ThemeProvider.propTypes = propTypes;
+ThemeProvider.Opts = {};
+ThemeProvider.Opts.Themes = ThemeProviderThemes;
```
zhongyn: can user pass a theme name which is not included in the opts? Is it just a predefined mapping so that user can use it like: 
`<ThemeProvider themeName={ThemeProvider.Opts.Themes.MOCK}>`
 instead of 
`<ThemeProvider themeName="terra-mock-theme"}>`?

bjankord: Yeah, the user can pass in whatever string they want and that will be applied as the class name. Using the `ThemeProvider.Opts.Themes` object, users have access to the list of names of all the first class themes.

packages/terra-badge/src/Badge.scss
```diff
@@ -1,6 +1,24 @@
...
+  --terra-badge-color-negative: #fff;
+  --terra-badge-background-color-warning: #e67e22;
+  --terra-badge-color-warning: #fff;
+  --terra-badge-background-color-info: #bdc3c7;
+  --terra-badge-color-info: #333;
+}
+
 :local {
   /* stylelint-disable selector-class-pattern  */
   .badge {
```
mhemesath: if we add a new style property to badge

```css
filter: var(--terra-badge-filter, none);
```

we can very easily create a grayscale theme by doing


```css
@include cerner-mock-theme {
  --terra-badge-filter: grayscale(1);
}
```

mhemesath: Actually, thinking about this... we'd only need that property on base.css. So the theme would only need to be defined in that one spot (and would turn the whole doc site gray).

packages/terra-theme-provider/src/ThemeProvider.jsx
zhongyn: if there are sub-themes like consumer button: 
```
@include terra-consumer-theme {
    &.terra-consumer-theme-concept1 {
        --terra-consumer-color: #123456;
    }
     --terra-button-active-background-color-primary: darken(var(--terra-consumer-color, #1c1f21), 50%);
    ....
}
```
Can we use it like:
const theme = `${ThemeProvider.Opts.Themes.CONSUMER} terra-consumer-theme-concept1`
<ThemeProvider themeName={theme}>

zhongyn: I am thinking about two designs of sub-theme handling:
1. Multi-class selector
```
@include terra-consumer-theme {
    &.terra-consumer-theme-concept1 {
        --terra-consumer-color: #123456;
    }
     --terra-button-active-background-color-primary: darken(var(--terra-consumer-color, #1c1f21), 50%);
    ....
}

const theme = `${ThemeProvider.Opts.Themes.CONSUMER} terra-consumer-theme-concept1`
<ThemeProvider themeName={theme}>
```

2. Nested class selector
```
@include terra-consumer-theme {
    .terra-consumer-theme-concept1 {
        --terra-consumer-color: #123456;
    }
     --terra-button-active-background-color-primary: darken(var(--terra-consumer-color, #1c1f21), 50%);
    ....
}

<ThemeProvider themeName={ThemeProvider.Opts.Themes.CONSUMER}>
    <ThemeProvider themeName="terra-consumer-theme-concept1" >
        ...
        <Button />
    <ThemeProvider/>
<ThemeProvider/>
```
@bjankord @mhemesath any thought?

mhemesath: I think we'd already require scenario 2.

bjankord: For sub-theming, I'd use a nested class selector approach as you have in example 2

mmalaker:
Functional verification completed. The button theming examples display and function correctly. Note that the dynamic theming capability is not supported in IE or Edge.

PR #737, opened 2017-08-17
Terra Embedded content consumer

rasthana:
### Summary
An Embedded content consumer component to interface with embedded iframe content. This pull request includes the initial component API.

### Additional Details
This component was initially created for terra-clinical and is being moved to terra-core.
Terra-Clinical PR: https://github.com/cerner/terra-clinical/pull/167
Terra-Clinical Design: https://github.com/cerner/terra-clinical/issues/70


packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,106 @@
...
+  }),
+  /**
+  * A set of event handlers keyed by the event name.
+  */
+  eventHandlers: PropTypes.arrayOf(PropTypes.shape({
+    key: PropTypes.string,
+    handler: PropTypes.func,
+  })),
+  /**
+   * Whether or not the container should expanded to fill its parent.
```
ryanthemanuel: Reword this to "should be expanded"

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,106 @@
...
+  fill: PropTypes.bool,
+};
+
+class EmbeddedContentConsumer extends React.Component {
+  componentDidMount() {
+    // Mount the provided source as the application into the content wrapper.
+    this.xfcFrame = Consumer.mount(this.embeddedContentWrapper, this.props.src, this.props.options);
+
+    // Attach the event handlers to the xfc frame.
+    this.addEventListener('xfc.launched', this.props.onLaunch);
```
tbiethman: What is the result of binding an undefined function?

tbiethman: D'oh, you check for their presence below

slor:
Merging blocked?

packages/terra-embedded-content-consumer/docs/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+```jsx
+import React from 'react';
+import EmbeddedContentConsumer from 'terra-embedded-content-consumer';
+
+<EmbeddedContentConsumer
+  src="https://google.com"
+  onLaunch={() => { console.log('launched'); }}
+  onAuthorize={() => { console.log('Authorized'); }}
+  onFullScreen={() => { window.open('https://yahoo.com'); }}
```
slor: I think we should remove full-screen hook or at least deprecate it or hide it from doc. Since full-screen was first coded, we have shifted our strategy away from discrete API calls and towards using a generic _trigger_ api call to pass in the event desired. 

Removing this hook will streamline the API so that there are no first-lass event handlers outside of the life-cycle events.

slor:
General question: there are aspects to this code I'm not familiar with like _fill_ and what a "data status update consumer/provider" represent. Are those larger Terra concepts that I just need to go learn or are those things specific to this component?

packages/terra-embedded-content-consumer/docs/README.md
rasthana: My preference would be to remove this functionality from this PR if there is consensus that this should be handled via the custom event handlers.

rasthana:
> General question: there are aspects to this code I'm not familiar with like fill and what a "data status update consumer/provider" represent. Are those larger Terra concepts that I just need to go learn or are those things specific to this component?

Some of these terms are similar to other components while quite a few are specific to this component.

**fill** - Component expands to its parent, used in [terra-content-container](https://github.com/cerner/terra-core/blob/master/packages/terra-content-container/src/ContentContainer.jsx#L21)
**data-status** - Attribute set on an xfc frame that notifies the life cycle status. I can change the example names to LifeCycleStatus if that is clearer.
**consumer/provider** - This is xfc terminology for the container and the application.

packages/terra-embedded-content-consumer/docs/README.md
bjankord: +1 for removing it

slor:
👍

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,97 @@
...
+   */
+  options: PropTypes.shape({
+    secret: PropTypes.string,
+  }),
+  /**
+  * A set of event handlers keyed by the event name.
+  */
+  eventHandlers: PropTypes.arrayOf(PropTypes.shape({
+    key: PropTypes.string,
+    handler: PropTypes.func,
```
kafkahw: maybe some docs/comments needed for this custom event handler: One thing I'd like to call out is that `this` will refer to `this.xfcFrame` of EmbeddedContentConsumer by default unless you bind it with something else.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * Whether or not the container should be expanded to fill its parent.
+   */
+  fill: PropTypes.bool,
+};
+
+class EmbeddedContentConsumer extends React.Component {
+  componentDidMount() {
+    // Mount the provided source as the application into the content wrapper.
+    this.xfcFrame = Consumer.mount(this.embeddedContentWrapper, this.props.src, this.props.options);
```
kafkahw: This will sound more like a comment for xfc. I noticed that here we mount frame first and then add listeners to the frame. It should be fine in most cases since mounting will take some time to complete, however, events can potentially be triggered before event listeners get added. I wish we can pass event handlers as part of options to Consumer's `mount` method. Thoughts? @mhemesath @bjankord

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+  addEventListener(eventName, eventHandler) {
+    if (eventName && eventHandler) {
+      this.xfcFrame.on(eventName, eventHandler);
+    }
+  }
+
+  addEventListeners(customEvents) {
+    if (customEvents) {
+      customEvents.forEach((event) => {
```
kafkahw: nit: parens around a single param can be omitted.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+  render() {
+    const {
+      src,
+      onLaunch,
+      onAuthorize,
+      options,
+      eventHandlers,
```
kafkahw: didn't see any consts from line 71 to line 75 being used in render method

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * Notifies the component that the container has been authorized.
+   */
+  onAuthorize: PropTypes.func,
+  /**
+   * The component can be configured with an authorization secret.
+   * secret - The authorization secret to be used if the embedded application does not know which domain to trust.
+   */
+  options: PropTypes.shape({
+    secret: PropTypes.string,
```
kafkahw: [`resizeConfig`](https://github.com/cerner/xfc#iframe-resizing-config) needs to be added here as well.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.scss
```diff
@@ -0,0 +1,19 @@
...
+:local {
+  .fill {
+    height: 100%;
+
+    :global {
+      .xfc {
+        height: 100%;
+     
+        iframe {
+          height: 100%;
```
kafkahw: `height: 100%` would possibly break the iframe's auto resizing feature, because the height of iframe should be changed dynamically based on the height of its embedded content. That's the default config. 
If user wants to change the default resizing logic, they can choose to pass in an additional param called `resizeConfig` to prop `options` (I made a comment above about the missing of resizeConfig in option's shape).

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.scss
```diff
@@ -0,0 +1,19 @@
+/* stylelint-disable max-nesting-depth */
+/* stylelint-disable selector-class-pattern */
+/* stylelint-disable selector-max-compound-selectors */
+:local {
+  .fill {
+    height: 100%;
+
+    :global {
+      .xfc {
+        height: 100%;
```
kafkahw: `height: 100%` is not needed.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
rasthana: What is the intended workflow to initialize an xfc frame if not for mounting it? Initializing the consumer just defaults the globalHandlers and mount returns the frame after emitting the xfc.mounted event. I agree that it would have been helpful to mount a frame with event handlers.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
rasthana: I added all the props based on the feedback I received and after looking at some of the terra components. Personally, I would prefer to omit the unused props but I'd prefer to opt for current practice. @tbiethman thoughts?

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
rasthana: Due to the scope of the changes needed for the resizeConfig options we opted to introduce that on another pull request.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
rasthana: Removing the parens around event causes a arrow-parens lint error. Let me know if I misunderstood your suggestion.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
rasthana: Wouldn't it be the responsibility of the component providing the handler to bind this to its scope? @tbiethman thoughts?

packages/terra-embedded-content-consumer/docs/README.md
```diff
@@ -0,0 +1,30 @@
+# Terra Embedded Content Consumer
+
+The Embedded Content Consumer is the application component which is embedding content within it.
+
```
emilyrohrbough: Your PR description is: 
> An Embedded content consumer component to interface with embedded iframe content

I feel it would be good to note this is meant specifically for iframe content.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
emilyrohrbough: What sort of changes would be required to include the resizeConfig?

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,98 @@
...
+  }),
+  /**
+  * A set of event handlers keyed by the event name.
+  * Note: Binding the event handler is necessary to make `this` work in the callback.
```
emilyrohrbough: Could we update the readme example to show this concept?

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
emilyrohrbough: You want to include these unused props here. Not doing so would result in them being spread via customProps on line 88.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
rasthana: Here is the resizeConfig that needs to be supported based on the [xfc documentation](https://github.com/cerner/xfc#iframe-resizing-config). The resize config can be added to the options prop without any api changes. In addition, testing all possible resize config options would increase the size of this PR significantly.

packages/terra-embedded-content-consumer/docs/README.md
rasthana: How about "embedding web content" instead of calling out the internal iframe details?

emilyrohrbough: Can you make this change in all description places?

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
rasthana: If I were to add the resizeConfig to the options would you prefer an object prop or a prop with the shape defined (see below). If I define a shape then I would have to restrict xfc to a specific version. @kafkahw @tbiethman thoughts?

Defining a shape will force a specific version of xfc.
```
options: PropTypes.shape({
  secret: PropTypes.string,
  resizeConfig: PropTypes.shape({
    scrolling: PropTypes.boolean,
    autoResizeWidth: PropTypes.boolean,
    fixedHeight: PropTypes.string,
    fixedWidth: PropTypes.string,
    heightCalculationMethod: PropTypes.oneOf([
      'bodyOffset',             // uses document.body.offsetHeight
      'bodyScroll',             // uses document.body.scrollHeight
      'documentElementOffset',  // uses document.documentElement.offsetHeight
      'documentElementScroll',  // uses document.documentElement.scrollHeight
      'max',                    // max of all of above options.
      'min',                    // min of all of above options.
	]),
    widthCalculationMethod: PropTypes.oneOf([
      'bodyOffset',             // uses document.body.offsetHeight
      'bodyScroll',             // uses document.body.scrollHeight
      'documentElementOffset',  // uses document.documentElement.offsetHeight
      'documentElementScroll',  // uses document.documentElement.scrollHeight
      'scroll',                 // max of bodyScroll and documentElementScroll.
      'max',                    // max of all of above options.
      'min',                    // min of all of above options.
	]),
	customCalculationMethod: PropTypes.func,
  }),
}),
```

Or, define options to be a generic object. This leaks that the component is a xfc wrapper.
```
options: PropTypes.object,
```

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.scss
rasthana: By default the frame is dynamically resized to fit the content however in our use case we need to expand the frame to fit its parent. In order to achieve this I added an opt-in fill property that allowed the frame to expand to fit its parent. However, we could create a custom wrapper component that provides this functionality.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -0,0 +1,114 @@
...
+   */
+  onLaunch: PropTypes.func,
+  /**
+   * Notifies the component that the container has been authorized.
+   */
+  onAuthorize: PropTypes.func,
+  /**
+   * The component can be configured with an authorization secret and the resize configuration.
+   */
+  options: PropTypes.shape({
```
mhemesath: I'd just leave this to be a generic config object for XFC.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
mhemesath: I agree... event handlers should be able to be passed ahead of mount

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
mhemesath:
```js
(customEvents || []).forEach(e => this.addEventListener(e.key, e.handler))
```

PR #741, opened 2017-08-22
Updated terra-i18n example to avoid race conditions

brhoades:
The previous terra-i18n example was very light. We used it for our implementation then had to make changes twice due to race conditions. I've added a more verbose example to hopefully prevent future developers from repeating our mistakes.


packages/terra-i18n/docs/README.md
```diff
@@ -28,6 +53,24 @@ Note that the state of the object needs to contain keys as follows for the i18nL
 }
 ```
 
+Additionally, do not render translated components until areTranslationsLoaded is true.
```
emilyrohrbough: Nit: I feel its a little confusing. Could we move this explanation to be an inline comment of the provided example?

ryanthemanuel: Agreed.  I like the idea of explaining what we need to do in the context of an example.

brhoades: Updated and rebased. Good recommendation.

PR #765, opened 2017-08-30
Fix focus management issues in Firefox

### LINKED ISSUE #706: In Firefox, when the hour is entered in the TimeInput, the focus does not automatically move to the minute input
# Issue Description
This issue only occurs in Firefox. The focus does not automatically move to the minute input after entering an hour. This occurs only when the onChange prop is set.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior
When an hour is entered, the focus should automatically move to the minute input.

## Current Behavior
In Firefox, the focus does not to the minute input after entering an hour.

## Steps to Reproduce
1. Open Firefox and go to the TimeInput component (http://engineering.cerner.com/terra-core/#/site/time-input)
2. Go to the "With Default Time" example and clear both the defaulted hour and minute.
3. Enter an hour (e.g. 10)
4. The focus should automatically move to the minute input but it does not.

## Environment
* Component Version: 1.2.0
* Browser Name and Version: Firefox

bjankord:
### Summary
Resolves #706 

@cerner/terra

### DIFF
--- a/packages/terra-time-input/CHANGELOG.md
+++ b/packages/terra-time-input/CHANGELOG.md
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Fixed focus management issue between hour and time input in Firefox.
 
 1.5.0 - (August 15, 2017)
 ------------------
--- a/packages/terra-time-input/src/TimeInput.jsx
+++ b/packages/terra-time-input/src/TimeInput.jsx
@@ -129,12 +129,12 @@ class TimeInput extends React.Component {
       }
     }
 
-    this.handleValueChange(type, inputValue);
-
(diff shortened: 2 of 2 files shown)


emilyrohrbough:
With this logged bug, @benbcai  commented:

> Anywhere that calls handleValueChange should pass in event as it is being used in handleValueChange.

In debugging we noticed `event` is not being passed to the handleValueChange function, which in turn is not being returned in the onChange function. Not sure if we want to handle that issue here as well.

bjankord:
@emilyrohrbough I'll take a look at the event stuff as well.

mmalaker:
Functional verification completed. Time input focus now functions correctly in Firefox and other supported browsers.

Issue #755, opened 2017-08-28
Create consumer theme for form input


bjankord:
# Issue Description
To match consumer UX, we want to enable the theming of terra input.
Here is a screenshot of the current component:
<img width="692" alt="screen shot 2017-08-28 at 9 27 53 am" src="https://user-images.githubusercontent.com/633148/29777908-3df3c214-8bd3-11e7-9580-cf9128e45c01.png">

Evaluate if we will solve this issue as well
https://github.com/cerner/terra-core/issues/755

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
CSS custom properties are added to enable the theming of the `<Input />` component to match consumer UX visuals.

WhiteLight1107:
# Tech Design
4 Consumer Theme states need to be added to a themes.scss file for input 
-Default
-Focus
-Active
-Disabled 

In order to achieve this variables in input.scss will be needed to change into variables so they can be modified in consumer theme for example with height currently it is

```
//Input.scss
:local {
  ...
  height: 1.857em;
  ...
   &[type='checkbox'],
   &[type='radio'] {
      height: auto;
      ...
    }
 ...
}
```
This will need to change to 

```
//Input.scss
:local {
  ...
  height: var(--terra-input-height, 1.857em_;
  ...
   &[type='checkbox'],
   &[type='radio'] {
      height: auto;
      ...
    }
 ...
}

//_themes.scss
@include cerner-consumer-theme {
   --terra-input-height: 32px;
}

```

This way the input text box can be modified for the consumer theme while keeping the defaults for legacy terra-core input, this "pattern" will need to be followed for all of the css props that would change in the consumer theme opacity,border etc.

alex-bezek:
With the currently recommended tech design, we would have to override every value we set back to the original values specifically for radio/check boxes. This could get messy and easy to forget or get things wrong. Additionally, when we start theming the control.scss file more, and creating our custom inputs, there would be css affecting radio/checkbox in multiple css files, and i'm not actually even sure how the specificity wins out here.

What if we instead used a not selector, or maybe radio/check shouldn't even consume the same base input as its going to look very different. @JakeLaCombe  thoughts?

WhiteLight1107:
@alex-bezek  perhaps that would be the right move would be to move scss for checkbox and radio from Input.scss to the Control.scss file

WhiteLight1107:
So for example it would work like this 

```
[...]

PR #767, opened 2017-08-30
Add OCS Consumer input theme

WhiteLight1107:
### Summary
1. Added Consumer Theme for input component
2. Moved css for radio and checkmark that was inside Input.scss and moved to Control.scss,
added logic in Input.jsx and Control.jsx to reflex this change 

**NOTE:** Still waiting for concept colors for the focus, currently just using a purple color, so this is not the final product just looking for some feedback.

Callouts: [forms-formFields callouts.pdf](https://github.com/cerner/terra-core/files/1264535/forms-formFields_PORTALDEV-35680.Design.and.Callouts2.pdf)

Deployment Link:
https://terra-core-deployed-pr-767.herokuapp.com/static/#/site/form/input


packages/terra-form/src/Control.jsx
```diff
@@ -97,6 +97,7 @@ const Control = ({
 
   const controlInputClasses = cx([
     'input',
+    (customProps.required ? 'required' : ''),
```
StephenEsser: ClassNames allows conditional checks first class for classes. 

Example: https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L88

packages/terra-form/src/Input.jsx
```diff
@@ -55,7 +55,8 @@ class Input extends React.Component {
       ...customProps
     } = this.props;
     const additionalInputProps = Object.assign({}, customProps);
-    const inputClasses = cx(['input', additionalInputProps.className]);
+    const inputClasses = cx([(customProps.type === 'radio' || customProps.type === 'checkbox') ?
+      '' : 'input', additionalInputProps.className]);
```
jmsv6d: Put each class on a new line for readability and use first class conditional checks provided by classNames as Stephen mentioned in the Control.

emilyrohrbough:
Resolves https://github.com/cerner/terra-core/issues/755

packages/terra-form/src/Input.jsx
```diff
@@ -55,7 +55,9 @@ class Input extends React.Component {
       ...customProps
     } = this.props;
     const additionalInputProps = Object.assign({}, customProps);
-    const inputClasses = cx(['input', additionalInputProps.className]);
+    const inputClasses = cx([
+      { input: (customProps.type !== 'radio' && customProps.type !== 'checkbox') },
+      additionalInputProps.className]);
```
StephenEsser: I would get a few more opinions on this before making any changes. This feels really awkward to me that we use this Input for Control but ignore all styles related to the Input component. Essentially just trying to tie into the couple extra lines of default / value and get the `<input>` tag. It feels like a really roundabout way of inheritance which isn't recommended in React. https://facebook.github.io/react/docs/composition-vs-inheritance.html

My opinion is that Input is an independent functioning component from Control and Control should use its' own input tag as Alex suggest on the tech design.

## Input
image

## Control
image

jmsv6d: I agree that this seems awkward. And it will probably become more so as we add OCS styles and special functionality for different types. I think that the control implementing it's own input tag and value logic would be fine, but I think another valid option would be for the Input to not have any styling at all and having HOCs that wrap it and add any styling/logic that's type specific.

bjankord:
It sounds like UX is reworking concept colors and the forms are in review for the OCS stuff. @WhiteLight1107 I'm leaning towards closing this PR and reopening as we get the updated design input for consumer form themes.

bjankord:
Closing this PR for the time being. We can reopen as we get the updated design input for consumer form themes

Issue #303, opened 2017-05-03
Datepicker enforce formatting


Matt-Butler:
# Issue Description
When typing data into the datepicker, the component should enforce the format provided by the user. 

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The datepicker should enforce the format provided by the user. 

## Current Behavior
The datepicker does not enforce format for user entered data. In this example, we have placeholder text for the format of the date input field. 
![dateformatting](https://cloud.githubusercontent.com/assets/8146159/25665258/ad1ff7ea-2fe2-11e7-8840-befa413a1228.gif)

Nick-Kroeker:
Plan needs this as well.

bjankord:
Related: https://github.com/Hacker0x01/react-datepicker/issues/1503

Jmacek:
I would also like some form of input enforcement. It would act similarly to the `pattern` attribute of the `TimePicker` component that only allows a correct date format (probably determined by the locale) as text entry.

bjankord:
Example added of how to add validation to date picker in https://github.com/cerner/terra-framework/pull/442

PR #768, opened 2017-08-30
Date time picker

### LINKED ISSUE #500: Date Time Picker Component
# Issue Description
Create Date Time Picker component. 

## Issue Type

- [x] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The DateTimePicker component has a date picker for selecting date and a time input for entering time. DateTimePicker supports the Spring and Fall daylight saving time changes in the following behaviors:

Spring: If you enter a date and time that falls within the missing hour during the Spring time change, the time will automatically be adjusted to the next relevant hour when the component loses focus. In the central time zone, the missing hour occurs between 2:00 and 2:59. For example, if the entered time is 2:15, the component will automatically adjust the time to the next hour at 3:15.

Fall: If you enter a date and time that falls within the ambiguous hour during the Fall time change, the component will display a dialog to require you to select daylight or standard time. After a selection is made a button will be displayed with a label showing the selected DST option.

### Considerations:
- [moment-timezone](http://momentjs.com/timezone/) will be a required dependency as it is needed to determine when daylight saving time occurs.
- Need UX guidance on whether or not time zone selection is needed.

### Responsiveness:
Inherits from terra-date-picker and terra-time-input

### Bidirectionality
Inherits from terra-date-picker and terra-time-input

### Accessibility
Inherits from terra-date-picker and terra-time-input

### React Props:
| Prop | Type | Default | Description |
|-------|-------|---------|---------------|
| inputAttributes | object | undefined | Custom input attributes to apply to the date and time inputs. |
| excludeDates | array | undefined | An array of ISO 8601 string representation of the dates to disable in the picker. |
| filterDate | func  | undefined | A function that gets called for each date in the picker to evaluate which date should be disabled. A return value of true will be enabled and false will be disabled. |
| includeDates        | array  | undefined      | An array of ISO 8601 string representation of the dates to enable in the picker. All Other dates will be disabled. |
| maxDateTime         | string | undefined      | An ISO 8601 string representation of the maximum date/time that can be entered. |
| minDateTime         | string | undefined      | An ISO 8601 string representation of the minimum date that can be entered. |
| name                | string | undefined      | Name of the date and time input. The name should be unique. |
| onChange            | func   | undefined      | A callback function to execute when a valid date or time is selected or entered. |
| releaseFocus        | func   | undefined      | A callback function to let the containing component (e.g. modal) to regain focus. |
| requestFocus        | func   | undefined      | A callback function to request focus from the containing component (e.g. modal). |
| value               | string | undefined | An ISO 8601 string representation of the initial default date/time to show in the date and time inputs. |

### Example:
```
import React from 'react';
import DateTimePicker from 'terra-date-time-picker';

//Renders a date time picker
<DateTimePicker
  name="admission-date"
  maxDateTime={'2017-08-16T00:00:00'}
  minDateTime={'2017-07-16T00:00:00'}
  value={'2017-07-25T10:30:00'}
/>
```
### CSS Classes:
| Selector                    | Description |
|-----------------------------|-------------|
| .date-time-picker           | The root class for the CSS component |
| .offset-button              | Class for the offset button |
| .time-container             | The container class for the time input and offset button |
| .time-clarification         | Class for the time clarification modal |

benbcai:
### Summary
terra-date-time-picker component
Fixes #500 

### Deployment Link
https://terra-core-deployed-pr-768.herokuapp.com/static/#/site/date-time-picker

@cerner/terra


packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -86,19 +101,28 @@ class DatePicker extends React.Component {
...
     this.handleOnClickOutside = this.handleOnClickOutside.bind(this);
   }
 
-  handleOnSelect() {
-    this.requestFocus();
+  handleOnSelect(selectedDate, event) {
+    this.releaseFocus();
+
+    if (this.props.onSelect) {
+      this.props.onSelect(event, selectedDate ? selectedDate.format() : '');
```
emilyrohrbough: `selectedDate ? selectedDate.format() : ''` 

In what instance would onSelect be triggered without a selectedDate? Should this check be done above in the if statement?

packages/terra-date-time-picker/docs/README.md
```diff
@@ -0,0 +1,49 @@
...
+//Renders a date picker
+<Base locale='en'>
+  <DateTimePicker
+    name="admission-date"
+    value={'2017-08-15T10:30'}
+  />
+</Base>
+```
+
+## Implementation Notes:
```
emilyrohrbough: This is a good detailed instruction on how to implement. Would it be useful to also provide a simple example of a date-picker inside a modal?

packages/terra-date-time-picker/docs/README.md
```diff
@@ -0,0 +1,49 @@
+# Terra Date Time Picker
+
+The DateTimePicker component has a date picker for selecting date and a time input for entering time. DateTimePicker supports the Spring and Fall daylight saving time changes in the following behaviors:
+
+Spring: If you enter a date and time that falls within the missing hour during the Spring time change, the time will automatically be adjusted to the next relevant hour when the component loses focus. In the central time zone, the missing hour occurs between 2:00 and 2:59. For example, if the entered time is 2:15, the component will automatically adjust the time to the next hour at 3:15.
+
+Fall: If you enter a date and time that falls within the ambiguous hour during the Fall time change, the component will display a dialog to require you to select daylight or standard time. After a selection is made a button will be displayed with a label showing the selected option.
+
+
```
emilyrohrbough: Do we want to indicate this component is composed of the date-picker and time-input components?

packages/terra-date-time-picker/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "DateTimePicker",
+    "UI"
+  ],
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
```
emilyrohrbough: `terra-app-delegate`, `terra-date-picker` and `terra-modal-manger` can be dropped from devDependencies

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,359 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import 'terra-base/lib/baseStyles';
```
emilyrohrbough: @bjankord With the css module conversion and the newest updates to Base, do we still need `import 'terra-base/lib/baseStyles'`

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,359 @@
...
+  /**
+   * An ISO 8601 string representation of the minimum date time.
+   */
+  minDateTime: PropTypes.string,
+  /**
+   * Name of the date input. The name should be unique.
+   */
+  name: PropTypes.string.isRequired,
+  /**
+   * A callback function to execute when a valid date is selected or entered. The first parameter is the event. The second parameter is the changed date time value.
```
emilyrohrbough: Is the onChange only triggered with a valid date selection or a valid date and time combination?

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,359 @@
...
+   */
+  onChange: PropTypes.func,
+    /**
+   * A callback function to execute when a change is made in the date or time input. The first parameter is the event. The second parameter is the changed date value.
```
emilyrohrbough: nit: second parameter is the changed **input** value

emilyrohrbough:
It might be good to add some tests for the added props in date-picker.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,359 @@
...
+      updatedDateTime = DateTimeUtil.updateTime(this.state.dateTime, this.timeValue);
+    }
+
+    this.checkAmbiguousTime(updatedDateTime);
+  }
+
+  checkAmbiguousTime(dateTime) {
+    // To prevent multiple time clarification dialogs from rendered, ensure that it is not open before checking for the ambiguous hour.
+    // ONe situation is when using the right arrow key to move focus from the hour input to the minute input will invoke onBlur and check for ambiguous hour.
+    // If the hour is ambiguous, the dialog would display and steals focus from the minute input, which again will invoke onBlur and check for ambiguous hour.
```
jmsv6d: Typos:
* rendering
* One
* steal

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,359 @@
...
+    const previousDateTime = this.state.dateTime ? this.state.dateTime.clone() : null;
+
+    // If both date and time are valid, check if the time is the missing hour and invoke onChange.
+    // If the date is valid but time is invalid, the time in the dateTime state needs to be cleared and render.
+    if (validDate && validTime) {
+      const updatedDateTime = DateTimeUtil.updateTime(previousDateTime, time);
+
+      if (event.keyCode === keyCodes.ARROWDOWN &&
+        previousDateTime && updatedDateTime && previousDateTime.format() === updatedDateTime.format()) {
+        updatedDateTime.add(-1, 'hours');
```
jmsv6d: Is there a reason you're adding -1 hours instead of subtracting 1?

packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -86,19 +106,36 @@ class DatePicker extends React.Component {
...
     this.handleChange = this.handleChange.bind(this);
+    this.handleChangeRaw = this.handleChangeRaw.bind(this);
     this.handleOnSelect = this.handleOnSelect.bind(this);
     this.handleOnClickOutside = this.handleOnClickOutside.bind(this);
   }
 
-  handleOnSelect() {
-    this.requestFocus();
+  handleOnSelect(selectedDate, event) {
+    // onSelect should only be invoked when selecting a datr from the picker.
```
StephenEsser: Nit: `datr` I think was meant to be `date`

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,358 @@
...
+    this.handleOnSelect = this.handleOnSelect.bind(this);
+    this.handleOnDateBlur = this.handleOnDateBlur.bind(this);
+    this.handleOnTimeBlur = this.handleOnTimeBlur.bind(this);
+
+    this.handleDaylightSavingButtonClick = this.handleDaylightSavingButtonClick.bind(this);
+    this.handleStandardTimeButtonClick = this.handleStandardTimeButtonClick.bind(this);
+  }
+
+  componentWillMount() {
+    const dateTime = DateTimeUtil.createSafeDate(this.props.value);
```
StephenEsser: Is this value different than `this,state.dateTime` declared on line 103?

packages/terra-date-picker/tests/nightwatch/date-picker-spec.js
```diff
@@ -93,6 +93,39 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
     browser.expect.element('h3').text.to.not.contain('2017-07-12');
   },
 
+  'Triggers onChangeRaw when a date input value is changed': (browser) => {
+    browser.url(`http://localhost:${browser.globals.webpackDevServerPort}/#/tests/date-picker-tests/on-change-raw`);
```
emilyrohrbough: With the newest nightwatch changes, brower.url is specified as:

`browser.url('${browser.launchUrl}/#/tests/date-picker-tests/on-change-raw')`

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,358 @@
...
+
+  handleOnSelect(event, selectedDate) {
+    const dateFormat = DateUtil.getFormatByLocale(this.context.intl.locale);
+    this.dateValue = DateTimeUtil.formatISODateTime(selectedDate, dateFormat);
+    const previousDateTime = this.state.dateTime ? this.state.dateTime.clone() : null;
+    const updatedDateTime = DateTimeUtil.syncDateTime(previousDateTime, selectedDate, this.timeValue);
+
+    if (!previousDateTime || previousDateTime.format() !== updatedDateTime.format()) {
+      this.checkAmbiguousTime(updatedDateTime);
+    }
```
emilyrohrbough: Would we want to call this.props.OnChange for selections?

packages/terra-date-time-picker/src/_TimeClarification.jsx
```diff
@@ -0,0 +1,171 @@
...
+          classNameModal="time-clarification"
+          ariaLabel="Time Clarification"
+          isOpen={this.state.isOpen}
+          onRequestClose={this.handleCloseModal}
+          closeOnEsc={false}
+          closeOnOutsideClick={false}
+        >
+          <div>
+            <header className={cx('header')}>
+              {<h1 className={cx(['title'])}>{title}</h1>}
```
emilyrohrbough: nit: don't need the extra brackets around classname

packages/terra-date-time-picker/tests/nightwatch/components/DateTimePickerExcludeDates.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import DateTimePickerExampleTemplate from './DateTimePickerExampleTemplate';
+
+const DateTimePickerExample = () => (
+  <DateTimePickerExampleTemplate
+    excludeDates={['2017-08-14']}
```
emilyrohrbough: Does this component exclude or include specific time selection in addition to dates?

packages/terra-date-time-picker/tests/nightwatch/date-time-picker-spec.js
```diff
@@ -0,0 +1,271 @@
+/* eslint import/no-extraneous-dependencies: ["error", {"devDependencies": true}] */
+/* eslint-disable no-unused-expressions */
+// eslint-disable-next-line import/no-extraneous-dependencies
+const screenshot = require('terra-toolkit').screenshot;
```
emilyrohrbough: One thing to note about the switched to ChromeDriver. When using click, it will click the center of the element specified. Not sure if/how that will affect selecting these imputs.

packages/terra-site/src/examples/date-time-picker/Index.jsx
```diff
@@ -0,0 +1,66 @@
...
+    <DateTimePickerExcludeDates />
+    <h2 id="filter-dates">Filter Dates</h2>
+    <DateTimePickerFilterDates />
+    <h2 id="include-dates">Include Dates</h2>
+    <DateTimePickerIncludeDates />
+    <h2 id="min-max">Min Max Selection Options</h2>
+    <DateTimePickerMinMax />
+    <h2 id="dst">Daylight Saving</h2>
+    <p>
+      The missing hour occurs when DST begins at these recent past and near future dates between 2:00 and 3:00.
```
emilyrohrbough: Are these specific to the US or world wide?

packages/terra-date-picker/tests/nightwatch/date-picker-spec.js
benbcai: Updated.
https://github.com/cerner/terra-core/pull/768/commits/32ba54d38fa0e035681655c69a78110d176ddfca

packages/terra-date-picker/src/DatePicker.jsx
benbcai: Ideally onSelect should only be invoked when selecting a date from the picker. However, react-datepicker has an issue (https://github.com/Hacker0x01/react-datepicker/issues/990) where onSelect is invoked both when selecting a date from the picker as well as manually entering a valid date or clearing the date. If the date is manaully cleared, onSelect would get triggered and selectedDate would be null. I originally had a check in handleOnSelect in DateTimePicker.jsx to ignore the event if it was a manual change. But I've moved that check to here so that consumers of DatePicker doesn't need to handle it.

https://github.com/cerner/terra-core/pull/768/commits/7f704a5dd8211fc1627f14d09ab2fa2b9f92f5ec

packages/terra-date-time-picker/docs/README.md
benbcai: I added links to the examples.
https://github.com/cerner/terra-core/commit/7f704a5dd8211fc1627f14d09ab2fa2b9f92f5ec

packages/terra-date-time-picker/docs/README.md
benbcai: Updated: https://github.com/cerner/terra-core/commit/7f704a5dd8211fc1627f14d09ab2fa2b9f92f5ec

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: Updated: https://github.com/cerner/terra-core/commit/7f704a5dd8211fc1627f14d09ab2fa2b9f92f5ec

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,359 @@
...
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+   * An array of ISO 8601 string representation of the dates to disable in the picker.
+   */
+  excludeDates: PropTypes.arrayOf(PropTypes.string),
+  /**
+   * A function that gets called for each date in the picker to evaluate which date should be disabled. A return value of true will be enabled and false will be disabled.
```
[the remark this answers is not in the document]
benbcai: Updated: https://github.com/cerner/terra-core/commit/7f704a5dd8211fc1627f14d09ab2fa2b9f92f5ec

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: Updated: https://github.com/cerner/terra-core/commit/7f704a5dd8211fc1627f14d09ab2fa2b9f92f5ec

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: I agree that subtracting 1 hour is easier to understand than adding -1 hour. https://github.com/cerner/terra-core/commit/7f704a5dd8211fc1627f14d09ab2fa2b9f92f5ec

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: When selecting a date from the picker, both onChange and onSelect will be invoked, so handleDateChange will take care of calling this.props.onChange.

packages/terra-date-picker/src/DatePicker.jsx
benbcai: Updated: https://github.com/cerner/terra-core/pull/768/commits/3763caa7628bc29b961a755d2d760b627a28bbfa

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: They both should have the same value when mounting. Updated to use `this.state.dateTime`.  https://github.com/cerner/terra-core/pull/768/commits/1823173fe270eecfe65c432c23a279e77cdeacde

Also added documentation to clarify the difference between `this.state.dateTime` and `this.dateValue`/`this.timeValue`. https://github.com/cerner/terra-core/pull/768/commits/3763caa7628bc29b961a755d2d760b627a28bbfa

packages/terra-date-time-picker/src/_TimeClarification.jsx
benbcai: Updated: https://github.com/cerner/terra-core/pull/768/commits/3763caa7628bc29b961a755d2d760b627a28bbfa

packages/terra-site/src/examples/date-time-picker/Index.jsx
benbcai: These daylight saving dates are specific to the US. I updated the message to call out that these are specific to the US and added an additional link for daylight saving dates for other time zones. https://github.com/cerner/terra-core/pull/768/commits/3763caa7628bc29b961a755d2d760b627a28bbfa

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: onChange is triggered when there is a valid date without a time or a valid date with a valid time. If no time is entered, the default time passed to onChange is 00:00. It does not get triggered with only a valid time and no date.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,359 @@
...
+   */
+  filterDate: PropTypes.func,
+  /**
+   * An array of ISO 8601 string representation of the dates to enable in the picker. All Other dates will be disabled.
+   */
+  includeDates: PropTypes.arrayOf(PropTypes.string),
+  /**
+   * Custom input attributes to apply to the date and time inputs. Use the name prop to set the name for the input. Do not set the name in inputAttribute as it will be ignored.
+   */
+  inputAttributes: PropTypes.object,
```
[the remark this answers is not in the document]
benbcai: I'm using the same InputAttributes prop for both the date input and time input because I didn't think that there would be a use case where we would want to apply one set of attributes for the date input and a different set for the time input. If there is such use case, I can certainly separate this prop.

packages/terra-date-picker/src/DatePicker.scss
```diff
@@ -1,20 +1,23 @@
 @import '~react-datepicker/dist/react-datepicker.min';
-
+/* stylelint-disable selector-class-pattern, selector-max-compound-selectors, max-nesting-depth */
```
emilyrohrbough: The selector-class-pattern stylelint rule has been removed in [this PR](https://github.com/cerner/terra-core/pull/841) we can remove the disable now

packages/terra-date-time-picker/src/DateTimeUtil.js
```diff
@@ -0,0 +1,105 @@
+import moment from 'moment-timezone';
+
+class DateTimeUtil {
```
emilyrohrbough: nit: DateTimeUtils

packages/terra-date-time-picker/src/_TimeClarification.jsx
```diff
@@ -0,0 +1,171 @@
...
+        >
+          <div>
+            <header className={cx('header')}>
+              {<h1 className={cx('title')}>{title}</h1>}
+            </header>
+            <br />
+            <div className={cx('body')}>
+              <p>{message}</p>
+            </div>
+            <br />
```
emilyrohrbough: Do we want to `<br />`s or should we included styling on the 'body' and 'buttons' divs?

packages/terra-date-time-picker/tests/nightwatch/DateTimePickerTests.jsx
```diff
@@ -0,0 +1,24 @@
...
+  <div>
+    <ul>
+      <li><Link to="/tests/date-time-picker-tests/default">DateTimePicker - Default</Link></li>
+      <li><Link to="/tests/date-time-picker-tests/date-only">DateTimePicker - Date Only</Link></li>
+      <li><Link to="/tests/date-time-picker-tests/date-time">DateTimePicker - Date Time</Link></li>
+      <li><Link to="/tests/date-time-picker-tests/dst">DateTimePicker - DST</Link></li>
+      <li><Link to="/tests/date-time-picker-tests/exclude-dates">DateTimePicker - Exclude Dates</Link></li>
+      <li><Link to="/tests/date-time-picker-tests/filter-dates">DateTimePicker - Filter Dates</Link></li>
+      <li><Link to="/tests/date-time-picker-tests/include-dates">DateTimePicker - Include Dates</Link></li>
+      <li><Link to="/tests/date-time-picker-tests/min-max">DateTimePicker - Min Max</Link></li>
```
emilyrohrbough: By Min Max do you mean date range?

packages/terra-date-picker/src/DatePicker.scss
benbcai: Removed selector-class-pattern. https://github.com/cerner/terra-core/pull/768/commits/079249a481ca7ff307578b55d0d3dfadeae48124

packages/terra-date-time-picker/src/DateTimeUtil.js
benbcai: Updated. https://github.com/cerner/terra-core/commit/079249a481ca7ff307578b55d0d3dfadeae48124

packages/terra-date-time-picker/src/_TimeClarification.jsx
benbcai: Removed the line breaks and added styling. https://github.com/cerner/terra-core/commit/079249a481ca7ff307578b55d0d3dfadeae48124

packages/terra-date-time-picker/tests/nightwatch/DateTimePickerTests.jsx
benbcai: This is for setting a minimum and maximum date that are allowed to be selected and no so much a date range. I look at date range as selecting two separate dates to indicate a from date and a to date.

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: I still see `terra-base/lib/baseStyles` being imported in all the components. 

@bjankord, Is `import 'terra-base/lib/baseStyles'` still needed in all the components?

bjankord: Yeah, it should still be in all the components. This ensures if terra-date-time-picker is the only component a consumer uses, it will also include base styles we assume to be in use in the application which all components are built on top of.

packages/terra-date-time-picker/package.json
bjankord: I'm guessing these are included to facilitate with tests scenarios around including date-picker in a modal. I'm fine keeping them as devDependecies for this case.

benbcai: Yes, these are included to facilitate testing.

packages/terra-date-time-picker/tests/nightwatch/components/DateTimePickerExcludeDates.jsx
benbcai: These are used only in the date picker.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,361 @@
...
+  minDateTime: PropTypes.string,
+  /**
+   * Name of the date input. The name should be unique.
+   */
+  name: PropTypes.string.isRequired,
+  /**
+   * A callback function to execute when a valid date is selected or entered.
+   * The first parameter is the event. The second parameter is the changed input value.
+   */
+  onChange: PropTypes.func,
```
nickpith: Seems like the parameters should be flipped. From most if not all consumers perspective, I could care less what event triggered the change. I just want to know that the value changed so I can store the value off. If the parameters are flipped then really only those that care about the event have to add them to their handler.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,361 @@
...
+    }
+  }
+
+  handleChange(event, newDateTime) {
+    this.setState({
+      dateTime: newDateTime,
+    });
+
+    if (this.props.onChange) {
+      this.props.onChange(event, newDateTime && newDateTime.isValid() ? newDateTime.format() : '');
```
nickpith: How will consumers of this component have to handle invalid input such as MM/DD/YY so that they can provide some feedback to a user that a value entered was invalid? Or is the expected behavior users should see is the date/time picker will clear because the input is invalid and an empty value is returned by onChange() which then tells the app to render the date/time picker with a value of ''?

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: I agree with your suggestion and don't mind making the flip. However, callback props in our other React components have the event as the first parameter and additional data as the second parameter. Flipping them here would make it inconsistent with other React components.

nickpith: I would argue those components are following a bad pattern

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,361 @@
...
+    let updatedDateTime;
+
+    if (isDateTimeValid) {
+      updatedDateTime = DateTimeUtils.updateTime(this.state.dateTime, this.timeValue);
+    }
+
+    this.checkAmbiguousTime(updatedDateTime);
+  }
+
+  checkAmbiguousTime(dateTime) {
```
nickpith: I'm not quite convinced that this component should be handling the ambiguous time frame. Doing so means that you are making assumptions of time zone or basing it off the browser's time zone and I'm not really sure every consumer will want that behavior. For instance, HealtheIntent wants to handle all times in UTC.

At a minimum, I think this behavior should be opt-in for consumers or broken off to another wrapper component and consumers that want to deal with it can.

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: This component currently does not enforce formatting. There is an [enhancement](https://github.com/cerner/terra-core/issues/303) to enforce formatting or some sort of validation. onChange will get triggered when entry is valid. There is another prop called onChangeRaw that gets triggered on every change event.

nickpith: Thanks. That's good to know for me because that will be something needed before I think broad usage of the component will happen.

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: I think that is a valid point. I'll bring this up in our war room discussion and see what others think about making this an opt-in behavior.

benbcai: The decision from the war room was this component will honor the ambiguous hour functionality by default and when there is a need to exclude this we can then add an opt-out option passively.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,361 @@
...
+  /**
+   * An array of ISO 8601 string representation of the dates to enable in the picker.
+   * All Other dates will be disabled.
+   */
+  includeDates: PropTypes.arrayOf(PropTypes.string),
+  /**
+   * Custom input attributes to apply to the date and time inputs. Use the name prop to set the name for the input.
+   * Do not set the name in inputAttribute as it will be ignored.
+   */
+  inputAttributes: PropTypes.object,
```
StephenEsser: It seems odd to spread the same attributes onto two different inputs. Is there a case where the same name would be useful being spread onto two different functioning inputs?

benbcai: I agree that there might be times where a different id or some other attributes that need to be set differently on these inputs.  https://github.com/cerner/terra-core/pull/768/commits/25cebdfa34847f2b2087c425778e17a310af4d94

mmalaker:
Functional verification completed. Date/time examples display and function correctly, including the daylight saving workflow.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,368 @@
...
+};
+
+const defaultProps = {
+  dateInputAttributes: undefined,
+  excludeDates: undefined,
+  filterDate: undefined,
+  includeDates: undefined,
+  maxDateTime: undefined,
+  minDateTime: undefined,
+  name: undefined,
```
kschuste: do you really need to default these to undefined?  wouldn't they be undefined automatically if you didn't default them to undefined

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,368 @@
...
+      isTimeClarificationOpen: false,
+    };
+
+    // The dateValue and timeValue variables represent the actual value in the date input and time input respectively.
+    // They are used to keep track of the currently entered value to determine whether or not the entry is valid.
+    // Unlike dateValue and timeValue, this.state.dateTime is the internal moment object representing both the date and time as one entity
+    // It is used for date/time minuipulation and used to calculate the missing/ambiguous hour.
+    // The dateValue and timeValue are tracked outside of the react state to limit the number of renderings that occur.
+    this.dateValue = '';
+    this.timeValue = '';
```
kschuste: maybe put these in default props?

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -0,0 +1,368 @@
...
+
+    this.setState({
+      isAmbiguousTime: isDateTimeAmbiguous,
+      isTimeClarificationOpen: isDateTimeAmbiguous && !isOldTimeAmbiguous,
+    });
+  }
+
+  handleDateChange(event, date) {
+    if (event.type === 'change') {
+      this.dateValue = event.target.value;
```
kschuste: should we continue on if the event is not 'change'?  why not just return right away?

benbcai: When a date is selected from the picker, as opposed to manually entering a date, the event.type is 'click' and this function would still need to get executed to handle the date change. However, on selecting a date from the picker, event.target.value is null so it should 
 not overwrite this.dateValue that is already set in handleOnSelect.

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: These are just properties on the component and not props so putting them in defaultProps would actually throw an error from PropsTable.

packages/terra-date-time-picker/src/DateTimePicker.jsx
benbcai: At one point the value and defaultValue were defaulted to null, which causes a warning, and I had to default to undefined to make it work as desired.

neilpfeiffer:
Tested: PR #768 
Resolves: Issue #500 
 
**Reviewed:**
- Follows UX defined design implementation for Terra Date-Time component with support for Daylight Savings time and the Ambiguous Hour.
 
 
**Pass** |
-----------|
UX Review|

Issue #692, opened 2017-08-02
Signature Component


vikonava:
# Issue Description
Create a Signature Component.

## Issue Type

- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Should generate a canvas element that captures the user input when dragging mouse/finger on the device's screen. Line segments of the signature would be returned on a callback method along with the most recently drawn line.  Line Color and Line Width can be defined by the consumer.

e3ujwh:
# Tech Design for terra-signature
## Summary:
The terra-signature component provides users a way to define a signature with a series of line segments.  The component can be define the line width, line color and optionally display a signature if the line segments have already been recorded.

## React Props:
| Prop        | Type | Default | Description  |
|----------|------|---------|--------------|
| lineWidth | number | 1    | Sets the line width |
| lineColor | string | #000000    | Sets the line color |
| lineSegments | array | undefined    | Line segments that define signature |
| onChange | func | undefined    | Callback function that returns all line segments and last drawn line segment |

## Example:
```
import React from 'react';

class SignatureExample extends React.Component {

  handleChange(event, lineSegments, lastSegment){
    console.log(JSON.stringify(lastSegment));
  }

  render() {
    return (      
      <div style={{ height: '100px', width: '100%', border: '1px solid black'}} >
        <Signature ref={instance => { this.signature = instance; }} 
               lineWidth={4} 
               lineColor={'#00008B'} 
               lineSegments={[{"x1":116,"y1":53,"x2":116,"y2":53}] 
               onChange={this.handleChange.bind(this)}/>
        <button onClick={() => {this.signature.clearSignature()}}>Clear</button>
      </div>
    );
  }
}
```

## CSS Classes:
| Selector         | Description |
|------------------|-------------|
| .terra-Signature | Used in canvas of signature component |

dkasper-was-taken:
@neilpfeiffer  Do we want lineWidth and lineColor to be themeable via css-modules, or do we want to keep the params? And if we keep the params, do we need UX predefined values for available options?

neilpfeiffer:
Lets reduce the props to just:
- lineWidth (enum of 4 sizes: I can provide list)
- lineSegments (array)
- onChange (func)

And we'll switch these to themeable variables:
- color
- background-color

PaulSolDev:
Orion JIRA created

bjankord:
+1 on tech design with addition of feedback from @neilpfeiffer.

[...]

PR #769, opened 2017-08-31
Create terra-signature component for eSignature

e3ujwh:
### Summary
Resolves https://github.com/cerner/terra-core/issues/692

### Additional Details
Will add additional nightwatch tests during and after component review.

@cerner/terra

### Deployment
https://terra-core-deployed-pr-769.herokuapp.com/static/#/site/signature


packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
...
+import 'terra-base/lib/baseStyles';
+import './Signature.scss';
+
+const LineWidths = {
+  EXTRAFINE: 1,
+  FINE: 2,
+  MEDIUM: 4,
+  HEAVY: 6,
+};
+
```
JakeLaCombe: Extra line.

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
...
+  componentWillReceiveProps(nextProps) {
+    if (this.props.lineSegments !== nextProps.lineSegments) {
+      const newState = Object.assign({}, this.state, { lineSegments: nextProps.lineSegments });
+      this.drawSignature(nextProps.lineSegments);
+      this.setState(newState);
+    }
+  }
+
+  componentWillUnmount() {
```
JakeLaCombe: What about the mouseup and touchcancel events? Should those be removed as well?

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
...
+
+    this.node.width = canvas.getBoundingClientRect().width;
+    this.node.height = canvas.getBoundingClientRect().height;
+
+    const context = this.node.getContext('2d');
+
+    context.lineWidth = this.props.lineWidth;
+
+    const style = window.getComputedStyle(this.node);
+    const color = style.getPropertyValue('color');
```
JakeLaCombe: Given on how this is used here and also that lineWidth is a prop, it almost feels like we should have the color as a first class prop for this component. I could see the case where the ink color is different than that of the surrounding text.

packages/terra-signature/src/Signature.scss
```diff
@@ -0,0 +1,8 @@
+.terra-Signature {
```
JakeLaCombe: with our use of css modules, I don't think we need to prefix the classname with terra nymore.

packages/terra-signature/tests/jest/Signature.test.jsx
```diff
@@ -0,0 +1,18 @@
...
+  // Snapshot Tests
+  it('should render a default component', () => {
+    const signature = shallow(defaultRender);
+    expect(signature).toMatchSnapshot();
+  });
+
+  // Structure Tests
+  it('should have the class terra-Signature', () => {
+    const signature = shallow(defaultRender);
+    expect(signature.prop('className')).toContain('terra-Signature');
```
JakeLaCombe: Is this test needed? I know we brought it up before, but it feels like our snapshot tests check this behavior for us.

packages/terra-signature/tests/nightwatch/signature-spec.js
```diff
@@ -0,0 +1,20 @@
+/* eslint import/no-extraneous-dependencies: ["error", {"devDependencies": true}] */
```
JakeLaCombe: Should we introduce tests to ensure that the drawing functionality is behaving as expected? I'm not sure how feasible that is though, but it'd be great to have tests to ensure the drawing behavior is working as expected.

packages/terra-signature/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+Changelog
+=========
+
+Unreleased
+----------
+
+1.0.0 - (August 31, 2017)
+------------------
+Initial stable release
```
jmsv6d: Please put this under the Unreleased section. A 1.0.0 section will be created with the release date when the component actually gets released.

packages/terra-signature/docs/README.md
```diff
@@ -0,0 +1,30 @@
...
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-signature`
+
+## Usage
+
+```jsx
+import React from 'react';
+import Signature from 'terra-signature';
+
+<Signature {props...} />
```
jmsv6d: Put an actual usage example here

packages/terra-signature/docs/README.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+## Component Features
+
+<!-- Uncomment supported features.
+ * [Cross-Browser Support](https://github.com/cerner/terra-core/wiki/Component-Features#cross-browser-support)
```
jmsv6d: Uncomment the features that this component supports

packages/terra-signature/package.json
```diff
@@ -0,0 +1,58 @@
+{
+  "name": "terra-signature",
+  "main": "lib/Signature.js",
+  "private": true,
+  "version": "0.0.0",
+  "description": "{insert description}",
```
jmsv6d: Add a description here

packages/terra-signature/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "UI"
+  ],
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
+    "terra-toolkit": "^0.x"
```
jmsv6d: Update dependencies to use most recent versions.

packages/terra-signature/src/Signature.jsx
jmsv6d: In the text design it was determined that the line color should be theme-able. https://github.com/cerner/terra-core/issues/692

packages/terra-signature/package.json
```diff
@@ -0,0 +1,58 @@
...
+  },
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "terra-base": "^0.x",
+    "terra-mixins": "^1.0.0"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "jquery": "^3.2.1",
```
StephenEsser: Introducing jQuery will probably require some discussion. Can you give a short explanation on where and why this is needed? 

The versioning and introduction of this could potentially cause thrashing with legacy applications in the process of uplifting.

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import 'terra-base/lib/baseStyles';
+import './Signature.scss';
+
+const LineWidths = {
+  EXTRAFINE: 1,
```
StephenEsser: Is EXTRAFINE a constant in a previous API? What are your thoughts on EXTRA_FINE ?

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
...
+  * Line segments that define signature.
+  */
+  lineSegments: PropTypes.array,
+
+  /**
+   * A callback function to execute when a line segment is drawn. The first parameter is the event, the
+   * second parameter is all the line segments, and the last parameter is the most recent line segment drawn.
+   */
+  onChange: PropTypes.func,
+
```
StephenEsser: Nit: It might not be worth updating. I don't think there is an official standard for the propTypes comment spacing, but looking at a few other examples it seems like there is normally no spaces between comments.

https://github.com/cerner/terra-core/blob/master/packages/terra-alert/src/Alert.jsx#L26-L65
https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L9-L58

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
...
+    const context = this.node.getContext('2d');
+
+    context.lineWidth = this.props.lineWidth;
+
+    const style = window.getComputedStyle(this.node);
+    const color = style.getPropertyValue('color');
+    context.strokeStyle = color;
+
+    if (this.props.lineSegments) {
+      this.state.lineSegments = this.props.lineSegments;
```
StephenEsser: State should never be assigned / manipulated directly. It should only ever be changed with setState. In this case you can initialize lineSegments in the constructor with something similar to the following: 

```
  constructor(props) {
     super(props);
 
     this.state = { lineSegments: props.lineSegments. };
   }
```

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
...
+  render() {
+    return (
+      <canvas className="terra-Signature" ref={(node) => { this.node = node; }} />
+    );
+  }
+}
+
+Signature.propTypes = propTypes;
+Signature.defaultProps = defaultProps;
+Signature.Width = LineWidths;
```
StephenEsser: There was a meeting a while back to export variables like this as opts. Here are a couple examples:

https://github.com/cerner/terra-core/blob/master/packages/terra-theme-provider/src/ThemeProvider.jsx#L25-L26
https://github.com/cerner/terra-core/blob/master/packages/terra-alert/src/Alert.jsx#L205-L206

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
...
+    this.node.addEventListener('mousemove', e => this.mouseXY(e));
+    document.body.addEventListener('mouseup', e => this.mouseUp(e));
+
+    // For mobile
+    this.node.addEventListener('touchstart', e => this.mouseDown(e), false);
+    this.node.addEventListener('touchmove', e => this.mouseXY(e), true);
+    this.node.addEventListener('touchend', e => this.mouseUp(e), false);
+    document.body.addEventListener('touchcancel', e => this.mouseUp(e).bind(this), false);
+
+    this.node.width = canvas.getBoundingClientRect().width;
```
StephenEsser: Can you explain what this is doing? Canvas is set to this.node, so this appears to be setting 
```
this.node.width = this.node.width
```

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
...
+  componentWillReceiveProps(nextProps) {
+    if (this.props.lineSegments !== nextProps.lineSegments) {
+      const newState = Object.assign({}, this.state, { lineSegments: nextProps.lineSegments });
+      this.drawSignature(nextProps.lineSegments);
+      this.setState(newState);
+    }
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', () => this.updateDimensions());
```
StephenEsser: Anonymous functions can't be removed

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
...
+
+    this.state = { lineSegments: [] };
+  }
+
+  componentDidMount() {
+    const canvas = this.node;
+
+    this.node.addEventListener('mousedown', e => this.mouseDown(e));
+    this.node.addEventListener('mousemove', e => this.mouseXY(e));
+    document.body.addEventListener('mouseup', e => this.mouseUp(e));
```
StephenEsser: Is there a reason these are bound using arrow functions?

I think these should be bound as:
```
document.body.addEventListener('mouseup', this.mouseUp);
```

Anonymous functions cannot be unbound

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
...
+  onChange: undefined,
+};
+
+class Signature extends React.Component {
+
+  constructor() {
+    super();
+
+    this.state = { lineSegments: [] };
+  }
```
StephenEsser: All class functions should be bound here. Example: 
https://github.com/cerner/terra-core/blob/db6c8ea1b2d9807dcf243da15bae9463c96e2d79/packages/terra-site/src/examples/modal-manager/ContentContainer.jsx#L10-L17

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
...
+      this.state.lineSegments = this.props.lineSegments;
+      this.drawSignature(this.state.lineSegments);
+    }
+
+    window.addEventListener('resize', () => this.updateDimensions());
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (this.props.lineSegments !== nextProps.lineSegments) {
+      const newState = Object.assign({}, this.state, { lineSegments: nextProps.lineSegments });
```
StephenEsser: I don't think this line is needed. With only one state object you should be able to call:
```
this.setState({ lineSegments: nextProps.lineSegments })
```

Is the if statement needed? You could also simplify this a little more by deconstructing lineSegments

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
...
+      <canvas className="terra-Signature" ref={(node) => { this.node = node; }} />
+    );
+  }
+}
+
+Signature.propTypes = propTypes;
+Signature.defaultProps = defaultProps;
+Signature.Width = LineWidths;
+
+export default Signature;
```
StephenEsser: I don't know how common the use-case is but I would test at least two Signature Components on the screen at the same time and see if the events are bound appropriately and allow the signatures to function independently.

packages/terra-signature/tests/nightwatch/SignatureExample.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+import Signature from '../../lib/Signature';
+
+class SignatureExample extends React.Component {
+  constructor() {
+    super();
+
+    this.state = { show: false };
```
StephenEsser: Is this state used?

packages/terra-signature/src/Signature.jsx
dkasper-was-taken: +1 for EXTRA_FINE

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,204 @@
...
+    if (this.props.lineSegments) {
+      this.state.lineSegments = this.props.lineSegments;
+      this.drawSignature(this.state.lineSegments);
+    }
+
+    window.addEventListener('resize', () => this.updateDimensions());
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (this.props.lineSegments !== nextProps.lineSegments) {
```
dkasper-was-taken: You may need to compare the values and length, rather than compare references.

packages/terra-signature/src/Signature.jsx
dkasper-was-taken: You probably don't need to use Object.assign here.

this.setState({ lineSegments: nextProps.lineSegments });

Should carry over the existing values.

packages/terra-signature/package.json
e3ujwh: The initial version of the code used jQuery but I replaced it based on feedback from Dave Kasper.  Will remove the dependency, it is no longer needed.

packages/terra-signature/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
+    "terra-toolkit": "^0.x"
+  },
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "terra-base": "^0.x",
```
emilyrohrbough: terra-base needs updated

packages/terra-signature/package.json
```diff
@@ -0,0 +1,58 @@
...
+  },
+  "scripts": {
+    "compile": "npm run compile:clean && npm run compile:build",
+    "compile:clean": "$(cd ..; npm bin)/rimraf lib",
+    "compile:build": "$(cd ..; npm bin)/babel src --out-dir lib --copy-files",
+    "lint": "npm run lint:js && npm run lint:scss",
+    "lint:js": "$(cd ..; npm bin)/eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
+    "lint:scss": "$(cd ..; npm bin)/stylelint src/**/*.scss",
+    "props-table": "props-table ./src/Signature.jsx --out-dir ./docs/props-table",
+    "test": "npm run test:spec && npm run test:nightwatch-default",
```
emilyrohrbough: These test scripts need updated based on the newest changes in terra-toolkit v2.0.0. [Here](https://github.com/cerner/terra-core/blob/master/packages/terra-alert/package.json#L51) is an example of the newest changes.

packages/terra-signature/tests/nightwatch/signature-spec.js
```diff
@@ -0,0 +1,20 @@
...
+
```
emilyrohrbough: This needs updated with the newest release to terra-toolkit v2.0.0. [Here](https://github.com/cerner/terra-core/blob/master/packages/terra-alert/tests/nightwatch/alert-spec.js#L1) is an example of what needs to change.

We will also want to update this to use the expect syntax. With this, the wait for 1000 can be dropped because the newest nightwatch configuration allows for this by deault.

emilyrohrbough:
Tests need to be added for this component.

packages/terra-signature/src/Signature.jsx
e3ujwh: I cannot really explain why but the canvas always assumes a default width of 300 and default height of 150.  I must set the width and height explicitly based on its container dimensions.  There are several ways to determine those dimensions but I thought using api getBoundingClientRect was the cleanest.

packages/terra-signature/src/Signature.jsx
e3ujwh: Recording a signature with line segments takes a lot of line segments!  We were a bit worried about the size of the array but with a variety of real world examples, it typically results in a 50kb or less array.  Still a concern that during the course of signing, there might be several arrays hanging around in the browser memory but we plan to test this aggressively when we add the terra-signature component to esignature.  

But more to your original point, whether we need to compare the actual element values in the array.  Do you know how expensive array comparison is in javascript or should we look for some simple comparisons like length to decide if we are dealing with the same signature?  Thoughts?

dkasper-was-taken: I'd say maybe start with at least comparing the length, but if we are worried about array performance we may need to consider switching to an updated key/forcing a remount on clear. As this comparison will occur with every state change and render.

e3ujwh: I will add a length comparison in addition to reference.  Given the vast number of line segments to describe a signature, it turns out to be a fairly good evaluator for sameness. 

I will revisit when the signature component is added to the eSignature app.

packages/terra-signature/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "terra-mixins": "^1.0.0"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "jquery": "^3.2.1",
+    "prop-types": "^15.5.8",
+    "terra-base": "^1.4.0",
+    "terra-mixins": "^1.0.0"
+  },
+  "scripts": {
```
bjankord: There is some clean up needed in scripts. Take a look at some of the other packages to see what's been updated. Most should be able to be copy/pasted back to here and adjusted for this package.

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,211 @@
...
+
+  componentDidMount() {
+    const canvas = this.node;
+
+    this.node.addEventListener('mousedown', e => this.mouseDown(e));
+    this.node.addEventListener('mousemove', e => this.mouseXY(e));
+    document.body.addEventListener('mouseup', e => this.mouseUp(e));
+
+    // For mobile
+    this.node.addEventListener('touchstart', e => this.mouseDown(e), false);
```
bjankord: You might want to test for support of touch events in the document before using them, I forget if older browsers throw errors on this or not.

packages/terra-signature/tests/nightwatch/signature-spec.js
dkasper-was-taken: I'm not sure that there's currently a feasible way to test it, but I do agree it would be good to ensure that it is functioning correctly. I believe the ultimate solution for this is screenshot comparison, so maybe in the mean time we add a test that executes drawing functionality, but doesn't tests anything, so when screenshot comparison is added, we already have a test ready. Plus it may catch the odd javascript error in the mean time.

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,211 @@
...
+
+    this.node.width = canvas.getBoundingClientRect().width;
+    this.node.height = canvas.getBoundingClientRect().height;
+
+    this.drawSignature(this.state.lineSegments);
+  }
+
+  render() {
+    return (
+      <canvas className="terra-Signature" ref={(node) => { this.node = node; }} />
```
dkasper-was-taken: Pass through customProps.

emilyrohrbough:
This components needs to use CSS Modules. Here are a few references to what changes that need to be made:
- [Example](https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-header/src/Header.jsx#L7) updates in source
- [Example](https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-header/src/Header.scss) updates in styling
- [Example](https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-header/tests/nightwatch/header-spec.js#L44) updates in nightwatch tests

packages/terra-badge/CHANGELOG.md
```diff
@@ -6,10 +6,6 @@ Unreleased
 ### Changed
 * Update nightwatch tests and test scripts.
```
emilyrohrbough: readd

packages/terra-signature/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "UI"
+  ],
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
+    "terra-toolkit": "^2.0.0"
```
emilyrohrbough: Remove terra-toolkit dependency and add `"terra-props-table": "^1.7.0"`

packages/terra-signature/package.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "terra-base": "^2.0.0",
+    "terra-mixins": "^1.0.0"
```
emilyrohrbough: Remove terra-mixins

packages/terra-signature/package.json
```diff
@@ -0,0 +1,51 @@
...
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "prop-types": "^15.5.8",
+    "terra-base": "^2.0.0",
+    "terra-mixins": "^1.0.0"
```
emilyrohrbough: Remove terra-mixins

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,215 @@
...
+    context.strokeStyle = color;
+
+    this.drawSignature(this.props.lineSegments, this.props.lineWidth);
+
+    window.addEventListener('resize', () => this.updateDimensions());
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if ((this.props.lineSegments !== nextProps.lineSegments) ||
+      (this.props.lineWidth !== nextProps.lineWidth)) {
```
emilyrohrbough: nit: keep if on same line

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,215 @@
...
+      this.setState({ lineSegments: nextProps.lineSegments, lineWidth: nextProps.lineWidth });
+      this.drawSignature(nextProps.lineSegments, nextProps.lineWidth);
+    }
+  }
+
+  componentWillUnmount() {
```
emilyrohrbough: Will we want to unmount 'onTouchStart' event listeners?

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,215 @@
...
+    window.removeEventListener('touchcancel', this.mouseUp);
+    window.removeEventListener('mouseup', this.mouseUp);
+  }
+
+  mouseDown(event) {
+    const canvas = this.node;
+
+    this.setState({ painting: true });
+
+    const rect = canvas.getBoundingClientRect();
```
emilyrohrbough: In componentDidMount you set this.node.height & this.node.width, should we also set rect.left and rect.top here as well?

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,215 @@
...
+    } else {
+      this.node.addEventListener('mousedown', this.mouseDown);
+      this.node.addEventListener('mousemove', this.mouseXY);
+      document.body.addEventListener('mouseup', this.mouseUp);
+    }
+
+    this.node.width = canvas.getBoundingClientRect().width;
+    this.node.height = canvas.getBoundingClientRect().height;
+
+    const context = this.node.getContext('2d');
```
emilyrohrbough: set this to this.context? I noticed this is called in other functions. Will this value be changing?

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,215 @@
...
+    }
+  }
+
+  clearSignature() {
+    this.setState({ lineSegments: [] });
+
+    const canvas = this.node;
+    const context = canvas.getContext('2d');
+
+    context.clearRect(0, 0, canvas.getBoundingClientRect().width, canvas.getBoundingClientRect().height);
```
emilyrohrbough: Should we use `this.node.width` & `this.node.height` instead of `canvas.getBoundingClientRect().width` & `canvas.getBoundingClientRect().height`

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,215 @@
...
+    this.drawSignature(this.state.lineSegments);
+  }
+
+  render() {
+    const { ...custProps } = this.props;
+
+    delete custProps.lineSegments;
+    delete custProps.lineWidth;
+    delete custProps.onChange;
+
```
emilyrohrbough: Update to `const {lineSegments, lineWidth, onhange, ...custProps } = this.props;`. This is the approach our team as been using for handling unused props and it removes the need to delete.

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,215 @@
...
+    this.draw = this.draw.bind(this);
+    this.drawSignature = this.drawSignature.bind(this);
+    this.clearSignature = this.clearSignature.bind(this);
+    this.updateDimensions = this.updateDimensions.bind(this);
+  }
+
+  componentDidMount() {
+    const canvas = this.node;
+
+    if ('ontouchstart' in document.documentElement) {
```
dkasper-was-taken: Might want to verify this is working on a touch screen laptop, with both touch and mouse events.

packages/terra-signature/src/Signature.scss
```diff
@@ -0,0 +1,11 @@
+/* stylelint-disable selector-class-pattern */
+/* stylelint-disable property-no-unknown */
+:local {
+	.signature {
```
dkasper-was-taken: Looks like this is using 4 spaces.

packages/terra-signature/docs/DEPENDENCIES.md
```diff
@@ -0,0 +1,23 @@
+# Dependency Information
+
+## dependencies
```
jmsv6d: run npm install so this dependency doc gets regenerated with the updated dependencies.

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,198 @@
...
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.updateDimensions);
+    window.removeEventListener('touchcancel', this.mouseUp);
+    window.removeEventListener('mouseup', this.mouseUp);
+  }
+
+  mouseDown(event) {
+    this.setState({ painting: true });
+
+    this.canvasRect = this.canvas.getBoundingClientRect();
```
emilyrohrbough: This was already set in updateDimensions

packages/terra-signature/src/Signature.jsx
```diff
@@ -0,0 +1,198 @@
...
+      this.canvas.addEventListener('touchmove', this.mouseXY, true);
+      this.canvas.addEventListener('touchend', this.mouseUp, false);
+      document.body.addEventListener('touchcancel', this.mouseUp, false);
+    } else {
+      this.canvas.addEventListener('mousedown', this.mouseDown);
+      this.canvas.addEventListener('mousemove', this.mouseXY);
+      document.body.addEventListener('mouseup', this.mouseUp);
+    }
+
+    const context = this.canvas.getContext('2d');
```
emilyrohrbough: Update to `this.context = this.canvase.getContext('2d')` and remove `const context = this.canvas.getContext('2d');` from draw, drawSignature and clearSignature

packages/terra-signature/src/Signature.jsx
e3ujwh: When I fixed the problem with the Usage section in the README.md (to show examples of how to use the signature component), it had the side effect of pushing the signature canvas off screen.  And if the component mounts but the canvas is not visible, getBoundingClientRect() returns bogus values.  I need to re-evaluate getBoundingClientRect when the canvas becomes visible so I can correctly draw signature lines in the canvas.  It also needs to stay in updateDimensions() so that if the window resizes the canvas takes on its new width and height.

packages/terra-badge/CHANGELOG.md
e3ujwh: fixed: https://github.com/cerner/terra-core/pull/769/commits/3a8c2f78444b155b7b6ec315126022c03c9bb5d9

packages/terra-signature/docs/DEPENDENCIES.md
e3ujwh: fixed: https://github.com/cerner/terra-core/pull/769/commits/3a8c2f78444b155b7b6ec315126022c03c9bb5d9

packages/terra-signature/src/Signature.jsx
e3ujwh: fixed: https://github.com/cerner/terra-core/pull/769/commits/3a8c2f78444b155b7b6ec315126022c03c9bb5d9

bjankord:
@e3ujwh Need to revert all the CHANGELOG additions that added the following:

```
### Changed
* Update nightwatch tests and test scripts.
```
Not sure where thats coming from.

vikonava:
We made all changes, all we need is final +1's from @ryanthemanuel and @bjankord

dkasper-was-taken:
After some discussion, we would like to see this component with both a managed and unmanaged version.  Matching a similar pattern of implementation of List, ModalManager, Table, etc.

ven-sub:
On Dave's comment, we can talk about this for version 0.2 but I think we need to move with this round 0.1 ASAP.

ryanthemanuel:
Let's leave this at 0.1 for now with what you have.  For any additional changes, we'll need to assess what the next version is.  If we're overall solid with it, we can make it 1.0, otherwise 0.2, etc.

packages/terra-signature/src/Signature.jsx
ryanthemanuel: Please address this comment.

packages/terra-signature/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "terra-signature",
+  "main": "lib/Signature.js",
+  "private": true,
+  "version": "0.1.0",
```
ryanthemanuel: This needs to be 0.0.0.  When we release the version gets bumped from the "current version" to whatever we decide (in this case it'll be 0.1.0).

packages/terra-signature/src/Signature.jsx
vikonava: Checked with Emily regarding this. We cannot have an instance variable for the getContext method due to the fact that on every redraw that happens the canvas reference will change and so the getContext reference needs to be updated as well. We tried another approach by pointing the canvas ref to a function that sets both values, but it didn't worked either. In the end it was decided to leave those const generated when needed.

BEPPSY:
Functional requirements and testing complete.  Approved

bjankord:
UX has approved changes and provided review here: https://github.com/cerner/terra-core/issues/692#issuecomment-330562687

neilpfeiffer:
Tested: PR #769
Resolves: Issue #692

**Reviewed:**
- New component, follows UX defined design implementation for initial release of Terra Signature component


**Pass** |
-----------|
UX Review|

PR #783, opened 2017-09-06
Update var name to avoid conflicts with legacy code

bjankord:
### Summary
We've found in some cases the `$terra-breakpoints` variable in the responsive element can conflict with the variable named the same in terra-legacy-theme. This causes the CSS modules export to be reported as string values instead of numeric values. This update ensures this variable is unique in our code and avoids conflicts with existing terra variable names.

@cerner/terra


tbiethman:
Very Nice

PR #786, opened 2017-09-08
Slide Group - Fix Focus & Add Nightwatch

### LINKED ISSUE #647: Slide Group Keyboard Focus

### LINKED ISSUE #774: Slide Group - Add Nightwatch Tests
# Issue Description
The slide group component needs nightwatch tests written to functionally verify the component behaves as expected. The current tests only tests slide entry. 

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

emilyrohrbough:
### Summary
This PR 
- Closes #647 - Removes keyboard focus of hidden slides
- Closes #774 - Adds nightwatch tests 

## Notes
I am open to suggestions on improving the test step which tests interaction with the hidden elements on a slide.


packages/terra-slide-group/src/Slide.scss
```diff
@@ -17,6 +17,10 @@
     }
   }
 
+  .slide[aria-hidden] {
+    display: none;
```
tbiethman: This style will hide the element immediately, causing the animations to not look quite right.  We're going to have to uplift to the react-transition-group ^2.0, which exposes handlers we can use to hide elements at the appropriate time.

emilyrohrbough:
Spoke with @tbiethman. My initial fix hid the 'old' slide immediately, which affected how animations looked. We decided it would be good to bump the react-transitions-group dependency from v.1.1.3 to v2.2.0. This bump allows lifecycle callbacks that are used to manipulate when aria-hidden is applied on an 'old' slide while the 'new' slide is transitioned in.

packages/terra-slide-group/src/Slide.scss
emilyrohrbough: Changed made in [this commit](https://github.com/cerner/terra-core/pull/786/commits/cc81fa06ee69492644309d10f470f8686107c96b)

bjankord: I think we might be able to use `visibility: hidden;` instead of `display: none;` to prevent focus. This would play nicer with the animations.

mmalaker:
Keyboard navigation functions correctly except for IE and Edge. In those browsers, the slide panels do not maintain tab focus. This is most evident when selecting the Decrement button. You can trigger it only once from the keyboard. Thereafter, the focus shifts out of the slide panel and on to the page navigation.

emilyrohrbough:
@mmalaker, @tbiethman and I discussed this behavior. The content of the slide group can by anything and we do not know the context of intended use or placement, so we determined the component utilizing the slide group should handle the focus shifts.

emilyrohrbough:
@mmalaker Are these README updates clear/what you would expect?

mmalaker:
Still have a typo: "accessibly" should be accessibility. While we're at it, let's specify that we mean keyboard accessibility. Once that text is corrected we are good to go.

mmalaker:
Functional verification completed. The slide group examples display and function correctly and the documentation is correct regarding keyboard accessibility.

Issue #637, opened 2017-07-18
Datepicker popup arrow rtl


Matt-Butler:
# Issue Description
When dir=rtl, the datepicker popup arrow does not point at the text-field. 

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Current Behavior

![datepicker-arrow](https://user-images.githubusercontent.com/8146159/28339698-9105b37a-6bd2-11e7-9d44-6df0396bb715.gif)

bjankord:
This has been released in terra-date-picker@1.11.0

PR #792, opened 2017-09-11
Fix Datepicker's Bidirectional Rendering

emilyrohrbough:
### Summary
This PR sets react-datepicker's popup attachment to fix bidi display. Closes https://github.com/cerner/terra-core/issues/637.


mmalaker:
Functional verification completed. The calendar popup's positioning is now correct in both LTR and RTL.

Issue #794, opened 2017-09-11
isSelected prop of Table.SingleSelectableRow child not honored without isSelectable


dv297:
# Issue Description
For terra-table, there are examples of the SingleSelectableRows in the project in test files that show a row as "selected".

```jsx
<Table.SingleSelectableRows>
  <Table.Row isSelected key={'PERSON_0'}>
    <Table.Cell content="John Smith" key="NAME" />
    <Table.Cell content="123 Adams Drive" key="ADDRESS" />
    <Table.Cell content="111-222-3333" key="PHONE_NUMBER" />
  </Table.Row>
</Table.SingleSelectableRows>
```

However, this does not cause any visible change in the UI. It's only when the `isSelectable` property is `true` that `isSelected` is honored. I'm not sure if this is deliberate, but the fact that it shows up in so many tests and in the documentation as having isSelected as being true but having `isSelectable` as `undefined` makes it seem like a bug.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [X] Other

## Expected Behavior
The highlighting of the row should be independent of `isSelectable`. If a row defines `isSelectable` as false, the row could still take `isSelected` as `true`, which would allow the row to be permanently selected. If `isSelectable` is `undefined` on the row, the row should be selectable from the SingleSelectableRows component

## Current Behavior
The row is not highlighted initially when the `isSelected` prop is `true`.

## Side Note
I am also working on contributing a MultipleSelectableRows component. With this component, multiple rows can be selected, but this table should also allow the user to deselect rows. This means that an `isSelectable` prop might not be as appropriate. Something like `isSelectionChangeable` might be more accurate, and could work for both multiple and single.

PR #795, opened 2017-09-12
Added MultipleSelectableRows to terra-table

dv297:
### Summary
This pull request adds a new component called `MultipleSelectableRows` to terra-table. This allows the user to select multiple rows from a table.

### Additional Details
I logged issue #794 , where the `isSelected` prop is not honored without the `isSelectable` prop set as true in `SingleSelectableRows`. I changed the behavior of this in the `MultipleSelectableRows` component so that if you defined `isSelected`, it would be selected, regardless if you defined `isSelectable`. As mentioned in the issue, `isSelectable` is kind of a confusing name at this point, as `isSelected` could be `true` while `isSelectable` is false (in cases where you would want the row to be permanently hard-coded as selected). The desired behavior and interactions of these two props should probably be discussed.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -29,6 +29,7 @@ Cerner Corporation
 - Michael Hemesath [@mhemesath]
 - Umesh Shimpi [@us044466]
 - Rishi Asthana [@rasthana]
+- Daniel Vu [@dv297]
 
 
 [@ryanthemanuel]: https://github.com/ryanthemanuel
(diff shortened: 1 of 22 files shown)


emilyrohrbough:
@dv297 I checked out #794 and it appears to be a bug with the rendering of SingleSelect Table. I am currently in the process of addressing this issue. That being said, could you log an enhancement issue for the MultiSelect Table Rows and include a tech design for this PR?

jmsv6d:
Should this PR be closed until it's been updated and ready for review?

dv297:
We can do that. With the new changes to SelectableTable, I have to rebase anyways. Closing.

PR #800, opened 2017-09-12
add new locales

zhongyn:
### Summary
Added new locales: ar, es-US, es-ES, pt-BR, fr-FR
Added placeholder for translation in alert/overlay/i18n/demographic-banner/date-picker

fix issue: https://github.com/cerner/terra-core/issues/805

@cerner/terra


dkasper-was-taken:
Is there an issue logged for this work? Are these additional locale aliases, or are we including new languages?

packages/terra-i18n/src/translationLoaders.js
```diff
@@ -1,3 +1,11 @@
+const loadArTranslation = (callback, scope) => {
```
alex-bezek: these all seem to be the same method. Would it be simpler to just make a single function that can create these based off a variable passed in?

Something along the lines of
```
const loadTranslationGenerator = (locale => {
  return (callback, scope) => {
   require.ensure([], (require) => {
     // eslint-disable-next-line
    const i18n = require(locale.toString());
    callback.call(scope, i18n);
  }, `${locale}-translations`);
})

...

const translationLoaders = {
  en: loadTranslationGenerator('en'),
  'en-GB': loadTranslationGenerator('en-GB'),
  ....
};
```

not sure on the syntax

emilyrohrbough:
I agree with @dkasper-was-taken. I think we should log an issue for this PR and log issues for the subsequent changes that will result from this change (i.e. updating translation files in components in core & clinical).

zhongyn:
@dkasper-was-taken @emilyrohrbough agree, will log related issues

packages/terra-alert/translations/es-ES.json
```diff
@@ -0,0 +1,9 @@
+{
```
emilyrohrbough: These aren't spanish?

zhongyn: they are placeholders for translation team to update. Need to log a jira in trans team so they can update the json and send a PR

packages/terra-i18n/src/translationLoaders.js
zhongyn: i tried that but webpack was not happy:
screen shot 2017-09-14 at 5 02 04 pm

bjankord: Yes, I believe we need to leave as is, webpack is picky about being able to statically analyze modules.

packages/terra-alert/translations/es-ES.json
bjankord: I'm ok with landing the PR with english translations as placeholders for the new locales. Its not ideal, but it avoids teams getting JS error is they use these new locales. HL team is on a time crunch to add this to one of their projects. I'll work with the translation team to prioritize getting these translations asap so we don't go too long with placeholder translations.

zhongyn:
@emilyrohrbough @bjankord can you approve this PR if no more change needed

PR #811, opened 2017-09-14
Update theming-plugin to always generate json file in same location

bjankord:
### Summary
This update resolves an issue where users would change directories into one of the component directories and run nightwatch tests. This would trigger webpack to run, hitting the theming-plugin and generating a `themeable-variables.json` file inside of the component directory.

After this update, when users change directories into one of the component directories and run nightwatch tests or run `npm start` from the root or `terra-site` directory, the `themeable-variables.json` file will always be output to `packages/terra-site`.

@cerner/terra


packages/terra-site/tests/jest/theming-plugin-spec.test.jsx
```diff
@@ -54,9 +54,8 @@ describe('theming-plugin', () => {
         },
       };
       rootNode.source = { input: { file: 'TestingComponent.scss' } };
-
       expect(testingFunction(rootNode)).toEqual(JSON.stringify(expectedHash));
-      expect(outputtedFileName).toEqual('themeable-variables.json');
+      expect(outputtedFileName.slice(outputtedFileName.length - 55)).toEqual('terra-core/packages/terra-site/themeable-variables.json');
```
emilyrohrbough: What does -55 do?

packages/terra-site/theming-plugin.js
```diff
@@ -33,6 +34,6 @@ module.exports = postcss.plugin('theming-plugin', () => {
       }
     });
 
-    return fs.writeFileSync('themeable-variables.json', JSON.stringify(sortHash(variables)), 'utf8');
+    return fs.writeFileSync(path.resolve(__dirname, 'themeable-variables.json'), JSON.stringify(sortHash(variables)), 'utf8');
```
mjhenkes: Could we pretty print the json?
`JSON.stringify(sortHash(variables)), 'utf8', 2)`

packages/terra-site/tests/jest/theming-plugin-spec.test.jsx
bjankord: Wanted to grab the end of the file / path, or the last 55 characters. Using this to check that its within the `terra-site` folder. This slice returns `terra-core/packages/terra-site/themeable-variables.json`. Depending on where this code is run, your machine, my machine, travis, the outputtedFileName's file path is different. Slicing the string to just the path thats consistent gives us the info we need when run in various environments. Also, its a nod to not driving slow.

packages/terra-site/theming-plugin.js
bjankord: Yeah, good idea. That will make diffs of this file easier to reason about.

bjankord: Updated: https://github.com/cerner/terra-core/pull/811/commits/205672d5bd4d59163e3899587fee1cfded292e07

packages/terra-site/themeable-variables.json
```diff
@@ -1 +1,278 @@
...
+    "--terra-popup-content-close-font-size": "1.45rem",
+    "--terra-popup-content-close-font-weight": "bold",
+    "--terra-popup-content-close-line-height": "0",
+    "--terra-popup-content-close-margin-left": "0.3571rem",
+    "--terra-popup-content-close-margin-right": "0",
+    "--terra-popup-content-close-padding": "4px",
+    "--terra-popup-content-close-hover-color": "#999",
+    "--terra-popup-content-inner-background-color": "#fff",
+    "--terra-popup-content-inner-border": "1px solid #b2b5b6",
+    "--terra-popup-content-inner-box-shadow, 0 4px 8px 0 rgba(0, 0, 0, 0.35), 0 4px 3px -3px rgba(0, 0, 0": "0.5)",
```
bjankord: Logged an issue about this malformatting: https://github.com/cerner/terra-core/issues/818
It should not impact users much as I don't think anyone is using this file other than us to generate the list of themable components here on https://terra-ui.herokuapp.com/getting-started#theming. For that we only use the top level keys.

packages/terra-site/themeable-variables.json
```diff
@@ -1 +1,278 @@
...
+    "--terra-badge-color-positive": "#fff",
+    "--terra-badge-background-color-negative": "#bc0203",
+    "--terra-badge-color-negative": "#fff",
+    "--terra-badge-background-color-warning": "#da3b03",
+    "--terra-badge-color-warning": "#fff",
+    "--terra-badge-background-color-info": "#004c76",
+    "--terra-badge-color-info": "#fff"
+  },
+  "Base": {
+    "--terra-base-font-family, \"Helvetica Neue\", Helvetica, Arial": "sans-serif",
```
bjankord: Logged an issue about this malformatting: #818

bjankord:
I need to talk with @JakeLaCombe tomorrow about the utf-8 setting. I'm seeing the JSON files saved as `us-ascii` prior to this change and after the change, and this is what the jest tests are failing on.

packages/terra-site/theming-plugin.js
```diff
@@ -33,6 +33,6 @@ module.exports = postcss.plugin('theming-plugin', () => {
         variables[component][matches[1]] = matches[2];
       }
     });
-    return fs.writeFileSync(path.resolve(__dirname, 'themeable-variables.json'), JSON.stringify(sortHash(variables), 'utf8', 2));
+    return fs.writeFileSync(path.resolve(__dirname, 'themeable-variables.json'), JSON.stringify(sortHash(variables), null, 2), 'utf8');
```
mjhenkes: writeFileSync should default to utf8, but it doesn't hurt to specify.

PR #821, opened 2017-09-15
Use uniqueIds for ids on terra-arrange examples

### LINKED ISSUE #389: Use uniqueIds for ids on terra-arrange examples.
# Issue Description
This implementation of an id for the terra-arrange components doesn't feel like we can be guaranteed a unique id (https://github.com/cerner/terra-core/blob/f9846b168e06d5e13a346b6176234d3ebf73b38a/packages/terra-site/src/examples/arrange/ArrangeAlignment.jsx#L35). We should look into using uniqueId to generate one, and then setting it in componentWillMount as opposed to the constructor.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Look into using uniqueId to generate one, and then setting it in componentWillMount as opposed to the constructor.

## Current Behavior
Id for example is set with the following:

```jsx
class ArrangeAlignment extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      id: Math.floor(Math.random() * 0xFFFF),
    };
    this.getId = this.getId.bind(this);
    this.handleSelectChange = this.handleSelectChange.bind(this);
  }

  getId(name) {
    return name + this.state.id;
  }
```

saedar:
### Summary
This PR
- Closes #389 - Use uniqueIds for ids on terra-arrange examples.

### DIFF
--- a/packages/terra-site/src/examples/arrange/ArrangeAlignment.jsx
+++ b/packages/terra-site/src/examples/arrange/ArrangeAlignment.jsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Arrange from 'terra-arrange';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import _ from 'lodash';
 import { ArrangeWrapper, alignLabels, alignOptions } from './examplesetup';
 
 const alignmentTypes = ['all', 'individual'];
@@ -32,10 +34,14 @@ class ArrangeAlignment extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      id: Math.floor(Math.random() * 0xFFFF),
+      id: undefined,
     };
-    this.getId = this.getId.bind(this);
+
     this.handleSelectChange = this.handleSelectChange.bind(this);
+  }
+
+  componentWillMount() {
+    this.state.id = _.uniqueId();
   }
 
   getId(name) {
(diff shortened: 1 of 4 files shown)


packages/terra-site/src/examples/arrange/ArrangeAlignment.jsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Arrange from 'terra-arrange';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import _ from 'lodash';
```
mjhenkes: Lodash isn't declared in the package.json. I assume it's getting pulled in transitively, could you add it to the package for site?

saedar: https://github.com/cerner/terra-core/pull/821/commits/7d270f071a4ca1d17ce4f6445ecc82aafcc709db

packages/terra-site/src/examples/arrange/ArrangeAlignment.jsx
```diff
@@ -32,12 +33,16 @@ class ArrangeAlignment extends React.Component {
...
     this.handleSelectChange = this.handleSelectChange.bind(this);
   }
 
+  componentWillMount() {
+    this.state.id = _.uniqueId();
+  }
+
   getId(name) {
     return name + this.state.id;
   }
```
emilyrohrbough: Should we update line 58 to set id={this.state.id}?

saedar: Do you mean: `id={this.getId('align')}`?

If so, I don't think so. It seems to be using `getId()` to generate a related but unique identifier to be use in the form for the labels/select boxes.

PR #822, opened 2017-09-15
Clean up package.json via link-parent-bin module

bjankord:
### Summary
There was an issue after the lerna v2 update where `lerna exec` commands in the nested packages would not run for users who don't have lerna installed globally. Installing lerna globally is not required to contribute on this project so this was a bug. 

Luckily theres a package exactly for this use case which links bin files in the root directory to the nested packages to use in their npm scripts.

@cerner/terra


bjankord:
@emilyrohrbough [Added nightwatch to the root package.json](https://github.com/cerner/terra-core/pull/822/commits/f5c7b6a9ad8c9794c2c1a99caa357ef681f4096e) and then tested running the `test:nightwatch` npm script from a nested package and it worked.

PR #834, opened 2017-09-19
Table - Fix Preselected Row & Test Structure Refactor

### LINKED ISSUE #794: isSelected prop of Table.SingleSelectableRow child not honored without isSelectable

emilyrohrbough:
### Summary
This PR:
1. Fixed bug to allow for preselected single select table row. Closes #794.
2. Adds nightwatch test for disabled single select table row.
3. Refactors nightwatch tests into `table` and `single-select-table` folders to separate examples and tests. This change mimics the list nightwatch file structure, to make the addition of MultipleSelectRows ( #795 ) tests easier to manage.

### Deployments
1. Preselected Test: https://terra-core-deployed-pr-834.herokuapp.com/static/#/tests/single-select-table-tests/preselected
2. Non-Selectable Test: https://terra-core-deployed-pr-834.herokuapp.com/static/#/tests/single-select-table-tests/non-selectable
3. Refactored Test Links: 
`table` https://terra-core-deployed-pr-834.herokuapp.com/static/#/tests/table-tests & `single-select-table` https://terra-core-deployed-pr-834.herokuapp.com/static/#/tests/single-select-table-tests


packages/terra-table/src/SingleSelectableRows.jsx
```diff
@@ -31,7 +31,7 @@ class SingleSelectableRows extends React.Component {
     }
     // Find the first row which is selected and is selectable
```
dkasper-was-taken: Update this comment to remove selectable.

mmalaker:
Functional verification complete. Table examples and test pages display and function as expected.

Issue #812, opened 2017-09-14
Remove "selector-class-pattern" Rule


emilyrohrbough:
# Issue Description
Remove "selector-class-pattern" rule from .stylelintrc file. This rule is no longer needed with the conversion to css-modules complete. With removing this rule, all scss files will need cleaned up to remove `stylelint-disable selector-class-pattern` comments.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

emilyrohrbough:
Resolved by #841

PR #841, opened 2017-09-20
Removed "selector-class-pattern" Rule in stylelint.rc 

msaikaushik1992:
### Summary
Removed selector-class-pattern rule in stylelint.rc and also removed related css comments in the packages wherever possible. After the migration to css modules complete, the selector-class-pattern-rule becomes obsolete, hence removed the rule and also its related comments in the packages.

Related issue : https://github.com/cerner/terra-clinical/issues/193

@cerner/terra


packages/terra-popup/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ Changelog
 Unreleased
 ----------
 
+### Removed
```
jmsv6d: You also changed classnames to match selector class pattern

packages/terra-text/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ Changelog
 Unreleased
 ----------
 
+### Removed
```
jmsv6d: Changed classnames here too

msaikaushik1992: I talked to Brett about this and he said I don't need to log class name changes because it is trackable through this https://github.com/cerner/terra-core/issues/812

bjankord: Im fine with no bullet point for the class name changes, I'd consider it an implementation detail.

PR #848, opened 2017-09-21
Menu keyboard Navigation

### LINKED ISSUE #784: Menu - keyboard sub-menu navigation and arrow key functions
# Issue Description
The menu component needs enhancements for keyboard navigation:
- Tab key should set focus within a sub-menu to the first item in that sub-menu. Currently you have to tab several times to get to the sub-menu items.
- Enable arrow keys (in addition to Tab) to move focus between menu items.
- Action and Close Actions available in the menu and submenu examples respond to mousepress but not "enter" on keyboard

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

Here is a link to keyboard navigation recommendations:
http://webaim.org/techniques/keyboard/

jmsv6d:
### Summary
Resolves #784 

- Added support to use arrow keys to navigate menu items
- Added keyboard functionality to the actionable menu item in the site example

Note: Testing this has proven to be very challenging. If anyone has any ideas on how to more thoroughly test this I'm open ears.

### Deployment Link
https://terra-core-deployed-pr-848.herokuapp.com/static/#/site/menu

@cerner/terra


packages/terra-button/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Added
```
emilyrohrbough: I think this should be in Menu's changelog? 😄

packages/terra-menu/tests/nightwatch/menu/menu-spec.js
```diff
@@ -61,4 +62,14 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
       .assert.hidden('div[class*="_header"] > div:last-child > button[class*="_header-button"]')
       .keys([browser.Keys.ESCAPE]);
```
emilyrohrbough: Add tests for up and down navigation?

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -90,11 +97,25 @@ class MenuContent extends React.Component {
...
       if (event.nativeEvent.keyCode === MenuUtils.KEYCODES.ENTER || event.nativeEvent.keyCode === MenuUtils.KEYCODES.SPACE) {
-        this.props.onRequestNext(item);
+        if (item.props.subMenuItems && item.props.subMenuItems.length > 0) {
+          this.props.onRequestNext(item);
+        }
+      } else if (event.nativeEvent.keyCode === MenuUtils.KEYCODES.RIGHT_ARROW) {
+        if (item.props.subMenuItems && item.props.subMenuItems.length > 0) {
+          this.props.onRequestNext(item);
+        } else {
+          this.props.onRequestClose();
```
emilyrohrbough: Is using the right arrow key to close a menu standard?

packages/terra-button/CHANGELOG.md
jmsv6d: Geeze, my brain is just gone today.

packages/terra-menu/tests/nightwatch/menu/menu-spec.js
jmsv6d: I haven't been able to find a way to verify that the focus changes on up/down... thoughts?

packages/terra-menu/src/_MenuContent.jsx
jmsv6d: That was for the ability to move between menus in a menu bar... that functionality would probably be better placed in the collapsible menu view and not applicable to a stand alone menu. I'll remove it from here.

packages/terra-menu/tests/nightwatch/menu/menu-spec.js
emilyrohrbough: Ah. I have yet to find a good way to determine what element on the page has focus either.

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -143,12 +162,34 @@ class MenuContent extends React.Component {
...
+
+        newItem = React.cloneElement(item, {
+          onClick,
+          onKeyDown,
+          isActive,
+        });
+      } else if (item.props.children) {
+        const children = React.Children.map(item.props.children, (child) => {
+          index += 1;
+          return React.cloneElement(child, { onKeyDown: this.wrapOnKeyDown(child, index), isActive: index === this.state.focusIndex });
```
emilyrohrbough: nit: on line 179 you place each cloneElement prop on a new line and here is same line

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -143,12 +162,34 @@ class MenuContent extends React.Component {
...
+        const isActive = this.state.focusIndex === index;
+
+        newItem = React.cloneElement(item, {
+          onClick,
+          onKeyDown,
+          isActive,
+        });
+      } else if (item.props.children) {
+        const children = React.Children.map(item.props.children, (child) => {
+          index += 1;
```
emilyrohrbough: Could we move index+=1 and isActive outside of if logic?

jmsv6d: No, that logic shouldn't be applied to Divider elements, and in the case of ItemGroups the isActive should be applied to it's children, not the ItemGroup.

emilyrohrbough: oh right. At first glance I didn't realize that. Maybe we should add a comment?

packages/terra-menu/tests/nightwatch/menu/menu-spec.js
bjankord: Here is a [custom nightwatch assertion for testing focus](https://gist.github.com/Rykus0/fa62d641dfeb75316d04). We don't need to add in it to this review, but it would be nice to add this in the future as we get into more focus management on various components.

bjankord:
@jmsv6d When tabbing to the `Close Action` menu item and pressing the enter key in Chrome, it looks like it fails to close the popup menu. However, pressing spacebar, when focused on that menu item, closes the popup menu.

packages/terra-menu/tests/nightwatch/menu/menu-spec.js
emilyrohrbough: That's pretty neat! I do think there is value to adding it. There are a few components were testing focus would be important (this one, time input, slide group). I think adding it to the toolkit would be a nice to have.

mjhenkes:
In the doc site after you've selected a drill in menu item via keyboard(enter or space), you have to tab again to select the menu again. Is this expected?

jmsv6d:
@mjhenkes That's because the item that did have focus is no longer visible so nothing really has focus at that point. Since the first item with a tabIndex is the popup container you have to tab to that first. I could try to force focus into the menu when the keyboard is used to navigate into the submenu. I just not sure if it's worth doing that work, since when we get the popup updates I believe it should behave more intuitively with the first tab bringing focus to the first element in the menu instead of the popup container.

mjhenkes:
@jmsv6d I'm ok with how it is.

packages/terra-menu/tests/nightwatch/menu/menu-spec.js
bjankord: I halfway implemented the hasFocus assertion in a local branch, need to get it out for review.

PR #846, opened 2017-09-21
Create Separate Pages for Icon Examples

emilyrohrbough:
### Summary
The terra-icon site page was very long and difficult to use. This PR breaks out the static and themeable icon examples into their own sub-pages.

Addresses #447.

PR #853, opened 2017-09-22
Create Separate Pages for List Examples

saedar:
### Summary
This PR moves the List examples into their own subpages.

Addresses #447.

### Additional Details
"List Item" was called out in the original comment on the issue, but I decided to leave it out since there isn't much semantic meaning for the component outside of the List itself. If there is disagreement, I can go ahead and create an example.

I opted to not move the propTables to the example pages, so that the main List page is the full "API" for the component. Again, if people think otherwise, I can move or replicate them in the example pages.

Finally, we've now used the `site-form-nav` style in Form, Icon, and List. Are we fine continuing to use that style name for our page sublinks or should it be renamed to something more generic?


packages/terra-site/src/examples/list/ListIndexRoutes.jsx
```diff
@@ -0,0 +1,20 @@
...
+// Form Component Indexes
+import List from './components/List';
+import MultiSelect from './components/MultiSelect';
+import MultiSelectMaxSelection from './components/MultiSelectMaxSelection';
+import SingleSelect from './components/SingleSelect';
+
+const routes = (
+  <div>
+    <Route path="/site/list/list" component={List} />
+    <Route path="/site/list/multi-select" component={MultiSelect} />
```
emilyrohrbough: My vote would be to combine the MultiSelect examples on one example page.

saedar: https://github.com/cerner/terra-core/pull/853/commits/9c3f80105944f72a9da5a84cb7b941399f8f31a8

emilyrohrbough:
After looking at the deployment, here are my thoughts:
1. The (original) List index should include the props table for List and List Items, the usage example for only List and then the list example.
2. The List sublink should be removed
3. For the Single & Multi lists, add the list version, create and include a README doc that includes a description (indication the use of List Items) and the usage examples. This would make these pages appear consistent with the Form component links.

saedar:
> After looking at the deployment, here are my thoughts:
> 
> The (original) List index should include the props table for List and List Items, the usage example for only List and then the list example.
> The List sublink should be removed
> For the Single & Multi lists, add the list version, create and include a README doc that includes a description (indication the use of List Items) and the usage examples. This would make these pages appear consistent with the Form component links.

@emilyrohrbough I think I got all of those points. Thanks for the feedback! https://github.com/cerner/terra-core/pull/853/commits/03074bb227ba54e9ac3b0194f29440e8cad1c18e

packages/terra-list/docs/MultiSelect.md
```diff
@@ -0,0 +1,16 @@
+# Multi Select List
+
+The Terra List is a structural component to arrange content within list/listitems with a configurable number of selectable items.
```
jmsv6d: Update to say Terra Multi Select List

saedar: https://github.com/cerner/terra-core/pull/853/commits/6c35ba1ab5c8608c7f9c3fdc74081d6512d96d8c

emilyrohrbough:
Based on discussion with #846 with @mmalaker & @neilpfeiffer, we will want to make sure and include links to the single and multi select list READMEs from the list README to be consistent with the form's README. Moving forward we want this linkage to subcomponents.

saedar:
> Based on discussion with #846 with @mmalaker & @neilpfeiffer, we will want to make sure and include links to the single and multi select list READMEs from the list README to be consistent with the form's README. Moving forward we want this linkage to subcomponents.

Done. https://github.com/cerner/terra-core/pull/853/commits/94a25ea78cdc745ba9c116fac654eaace848fce4

mmalaker:
Nit, but can we we switch the nav order of the pages? The simpler, single select list should display first, followed by the multi select list.

saedar:
> Nit, but can we we switch the nav order of the pages? The simpler, single select list should display first, followed by the multi select list.

@mmalaker https://github.com/cerner/terra-core/pull/853/commits/0f115e9c5cdc474006a53754d868b0e629aece6b

mmalaker:
Functional verification completed. List examples divided into single and multiselect list pages, as expected.

Issue #850, opened 2017-09-21
Nightwatch Checking Browser Specific CSS Causing Local Failures


emilyrohrbough:
# Issue Description
Arrange component tests are failing on this test step when running `npm run clean:install && npm run test`.

<img width="1149" alt="screen shot 2017-09-21 at 5 44 42 pm" src="https://user-images.githubusercontent.com/14099737/30722073-b1b41f26-9ef4-11e7-8cdc-713d68cfe44b.png">

This error is occurring due to [this test step](https://github.com/cerner/terra-core/blob/master/packages/terra-arrange/tests/nightwatch/arrange-spec.js#L14) checking a css value that we are not defining, but that the browser is actually defining. Overall we should avoid checking css values in nightwatch, especially if we do not define them given each browser is different.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

## Steps to Reproduce

1. `npm run clean:install && npm run test`

## Environment

*  Failure occurs on ChromeDrive v2.32.3 with Chrome v61.0.3163.91 installed

PR #858, opened 2017-09-22
updating test to not check against browser defined values

kschuste:
### Summary
This PR updates the nightwatch tests for the terra-arrange component to have them pass again.

Addresses #850 

### Additional Details
The nightwatch tests for the arrage component fail as they are checking a css value that we are not defining, but that the browser is actually defining.  We should not be checking against a css value that we do not define as browsers can be different.  Therefore, I have changed the tests to test against the css value not getting set to one of our defined values.


packages/terra-arrange/tests/nightwatch/arrange-spec.js
```diff
@@ -101,7 +104,10 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
       .url(`${browser.launchUrl}/#/tests/arrange-tests/aligned-individually-fit-start`);
 
     browser.expect.element('#center div:nth-child(1)').to.have.css('align-self').which.equals('center');
-    browser.expect.element('#center div:nth-child(2)').to.have.css('align-self').which.equals('auto');
+    browser.expect.element('#default div:nth-child(2)').to.have.css('align-self').to.not.equal('flex-start');
```
emilyrohrbough: Overall we should really avoid checking styling in nightwatch. Since we are not applying a style here via a prop, I'd say we should not be testing this at all.

jmsv6d: I would rather we tested that the classNames are placed on the element rather than the css itself.

PR #860, opened 2017-09-23
Break out SelectableList subcomponent to allow for managed selectable lists

### LINKED ISSUE #856: Create selectable sub-component for Single and Multi Lists
# Issue Description
Create sub-component as a common underlying implementation for the single and multi select list components.  Allowing consumers to directly consume this component if they wish to managed there own state.  Also break out common functions to a utilities file, for consumers to use in their custom implementations.

## Issue Type
- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

dkasper-was-taken:
### Summary
Break out SelectableList subcomponent to allow for managed selectable lists.

Fixes #856 

@cerner/terra


packages/terra-list/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ Unreleased
 ### Removed
 * Remove stylelint-disable selector-class-pattern
 
+### Changed
```
emilyrohrbough: Should this be :
###Added 
*SelectableList component

###Changed
*Update SingleSelectList and MultiSelectList components to use SelectableList

packages/terra-list/src/SelectableList.jsx
```diff
@@ -0,0 +1,158 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import 'terra-base/lib/baseStyles';
+import List from './List';
+import SelectableUtils from './SelectableUtils';
+
+const KEYCODES = {
```
emilyrohrbough: Move KEYCODES to Utils

packages/terra-list/src/SelectableList.jsx
```diff
@@ -0,0 +1,158 @@
...
+  /**
+   * Whether or not the child list items has a disclosure indicator presented.
+   */
+  hasChevrons: PropTypes.bool,
+  /**
+   * A callback event that will be triggered when selection state changes.
+   */
+  onChange: PropTypes.func,
+  /**
+   * An array of the currectly selected indexes
```
emilyrohrbough: nit: missing period

packages/terra-list/tests/nightwatch/selectable-list/DisableUnselectedSelectableList.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+
+import SelectableList from '../../../src/SelectableList';
```
emilyrohrbough: update to lib

packages/terra-list/tests/nightwatch/selectable-list/NonSelectableItemSelectableList.jsx
```diff
@@ -0,0 +1,31 @@
...
+    this.state = { selectedIndex: 'none' };
+    this.handleSelection = this.handleSelection.bind(this);
+  }
+
+  handleSelection(event, selectedIndex) {
+    this.setState({ selectedIndex });
+  }
+
+  render() {
+    return (
```
emilyrohrbough: Add context to indicate which item is non-selectable

packages/terra-list/tests/nightwatch/selectable-list/NonSelectableItemSelectableList.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+
+import SelectableList from '../../../src/SelectableList';
```
emilyrohrbough: src -> lib

packages/terra-list/tests/nightwatch/selectable-list/OnChangeSelectableList.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+
+import SelectableList from '../../../src/SelectableList';
```
emilyrohrbough: src -> lib

packages/terra-list/src/SelectableList.jsx
```diff
@@ -0,0 +1,158 @@
...
+    // If selectable, add tabIndex on items to navigate through keyboard tab key for selectable lists and add
+    // onClick and onKeyDown functions.
+    if (newProps.isSelectable) {
+      newProps.tabIndex = '0';
+      newProps.onClick = onClick;
+      newProps.onKeyDown = onKeyDown;
+    }
+
+    // Uses the props.hasChevron value, unless the consumer specifies the item hasChevron attribute as false.
+    if (this.props.hasChevrons) {
```
emilyrohrbough: I know multi select lists are not intended to render chevrons. Should we add some sort of check to handle that?

packages/terra-site/src/examples/list/Selectable.jsx
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable no-alert */
+import React from 'react';
+
+import SelectableList from 'terra-list/lib/SelectableList';
+
+const list = () => (
+  <SelectableList hasChevrons isDivided onChange={(event, index) => alert(`Index: ${index} Clicked!`)}>
```
emilyrohrbough: Its been mentioned before we don't want to trigger alerts as onChange examples.

packages/terra-list/src/SelectableList.jsx
dkasper-was-taken: True, though I'm not sure this agnostic component can distinguish between multi vs single select. Be open to opinions on how this should function if we intend to adopt that behavior.

emilyrohrbough: Hmm.. yeah that is a very good point. The cleanest way to enforce this might be to add a boolean prop called `isMultiSelectable` which we could add to the hasChevrons check? That sort-of removes the flexibility of this component, but would kept the distinction between the two? 

Regardless, we should probably update the hasChevrons description to indicated it should only be used when implementing SingleSelect lists.

packages/terra-list/tests/nightwatch/selectable-list/NonSelectableItemSelectableList.jsx
dkasper-was-taken: Updated Item name.

packages/terra-list/src/SelectableList.jsx
```diff
@@ -0,0 +1,154 @@
...
+   */
+  hasChevrons: PropTypes.bool,
+  /**
+   * A callback event that will be triggered when selection state changes.
+   */
+  onChange: PropTypes.func,
+  /**
+   * An array of the currectly selected indexes.
+   */
+  selectedIndexes: PropTypes.array,
```
StephenEsser: To match the existing API I think this makes sense.

In the future it might be worth discussing if selectedKeys might be a more appropriate identifier of selected items. Each List item should be provided a unique key ( ideally not index ). The key would always map to the item, whereas an index has no constraints or requirements to always uniquely identify an item but a key by definition **should**. It would help remove the awkwardness of remapping indexes to items after any additions / removals from the list.

dkasper-was-taken: We could change this under the hood. And just have the current multi and single selects return indexes. Since this is a new subcomponent, we could start the pattern here.

dkasper-was-taken: After some looking, react doesn't allow you to use the key. It's considered protected and not stored on the child. We'd need special item key props or meta data to accomplish this.

StephenEsser: The key is not accessible within props, but it should still be accessible on the child. If this is causing too many changes I wouldn't make the changes on this pull request. 

Here is a small code sample I use that accesses the key in a similar way: 
```
    const children = React.Children.map(this.props.children, child => (
      <SelectableItem onClick={() => this.handleSelection(child.key)} isSelected={child.key === this.state.selectedKey}>
        {child}
      </SelectableItem>
    ));
```

https://github.com/facebook/react/issues/2429#issuecomment-277810636

dkasper-was-taken: I got this working with the keys, but after some discussion, maintaining both the index and key for passivity is too complicated for now. So we'll just handled indexes in this PR.

packages/terra-list/src/SelectableList.jsx
```diff
@@ -0,0 +1,154 @@
...
+   */
+  children: PropTypes.node,
+  /**
+   * Whether or not the child list items should have a border color applied.
+   */
+  isDivided: PropTypes.bool,
+  /**
+   * Whether or not unselected items should be disabled.
+   */
+  disableUnselectedItems: PropTypes.bool,
```
bjankord: Can we add a comment here "Helpful for enabling max row selection."

dkasper-was-taken:
@mmalaker Regression verification of single and multi select is needed to ensure that replacing the underlying implementation hasn't changed anything.
Also there are new tests for that that implementation:
https://terra-core-deployed-pr-860.herokuapp.com/static/#/tests/selectable-list-tests

Selectable List has been given it's own site page as well under the List directory.

PaulSolDev:
Functional review complete.

PR #862, opened 2017-09-25
Update Base.scss

### LINKED ISSUE #861: Update terra-base to default body to 100% rather than auto
# Issue Description
Most of our consumers have been using legacy terra which has assumed something along the lines of a 100% body.  terra-base defaults it to auto.  We should change the default to be more in line with what was done previously.  Consumers can still override this if need be.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [X] Other

ryanthemanuel:
### Summary
This will fix #861   It will set the default of body height to be 100% rather than auto.

### DIFF
--- a/package.json
+++ b/package.json
@@ -98,5 +98,11 @@
     "stylelint-config-sass-guidelines": "^3.0.0",
     "stylelint-suitcss": "^1.0.0",
     "terra-toolkit": "^2.0.0"
+  },
+  "dependencies": {
+    "terra-mixins": "^1.10.0"
+  },
+  "peerDependencies": {
+    "terra-mixins": "^1.10.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -98,11 +98,5 @@
     "stylelint-config-sass-guidelines": "^3.0.0",
     "stylelint-suitcss": "^1.0.0",
     "terra-toolkit": "^2.0.0"
-  },
-  "dependencies": {
-    "terra-mixins": "^1.10.0"
-  },
-  "peerDependencies": {
-    "terra-mixins": "^1.10.0"
   }
 }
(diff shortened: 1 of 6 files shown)


packages/terra-base/src/Base.scss
```diff
@@ -13,7 +13,7 @@ body {
   background-color: var(--terra-base-background-color, #fff);
   color: var(--terra-base-color, #1c1f21);
   font-size: 1rem;
-  height: var(--terra-base-body-height, auto);
+  height: var(--terra-base-body-height, 100%);
```
mhemesath: Please update consumer theme to have a body height of auto. This change is very non-passive to embedded content.

bjankord: With this change, can you add `terra-mixins` to dependencies / peerDependencies.

e.g.
```json
"peerDependencies": {
    "terra-mixins": "^1.10.0"
  },
  "dependencies": {
    "terra-mixins": "^1.10.0"
  },
```

And then we also need to import the mixins file in Base.scss and _theme.scss. Add this to the top of those files
```scss
@import '~terra-mixins';
```

package.json
```diff
@@ -98,5 +98,11 @@
     "stylelint-config-sass-guidelines": "^3.0.0",
     "stylelint-suitcss": "^1.0.0",
     "terra-toolkit": "^2.0.0"
+  },
```
bjankord: These should be added to the package.json in `packages/terra-base`

ryanthemanuel: Ha....oops.  That's what I get for trying to update everything in git.  Sorry about that.

ryanthemanuel:
After talking with @bjankord, we decided this didn't warrant functional verification.

PR #879, opened 2017-10-02
Terra Hookshot

### LINKED ISSUE #739: Terra Hookshot
# Issue Description
Create Tether alternative capable of handling arrow offsets, target positions, mobile zoom, and shrink package size.

## Issue Type

- [x] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Current Behavior
Currently the tether implementation used for popups is unable to handle mirrored offsets and correct positioning when zoomed on mobile.  It also has a considerable code footprint to account for a myriad of positioning options and support for browsers that are no longer needed. Also as a closed system, it does not allow for seamless arrow positioning, as positional margins are not an option.

jmsv6d:
### Summary
Resolves #739 Creates Tether alternative capable of handling arrow offsets, target positions, mobile zoom, and shrink package size.

### Deployment
https://terra-core-deployed-pr-879.herokuapp.com/


packages/terra-hookshot/LICENSE
```diff
@@ -0,0 +1,203 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
JakeLaCombe: Does this copyright need a year and owner name?

packages/terra-hookshot/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Hookshot
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-hookshot.svg)](https://www.npmjs.org/package/terra-hookshot)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+The Terra Hookshot componenet positions content according to a targeted attachment, ensuring they stay connected.
```
JakeLaCombe: Component*

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,336 @@
...
+
+const ATTACHMENT_BEHAVIORS = [
+  'auto',
+  'flip',
+  'none',
+];
+
+const propTypes = {
+  /**
+   * If the primary attachment in not available, how should the content be positioned.
```
JakeLaCombe: I think this would be worded better if we have the if statement at the end.

    How the content should be positioned when the primary attachment is not available.

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,336 @@
...
+   * Valid values: 'auto', 'flip', and 'none'.
+   */
+  attachmentBehavior: PropTypes.oneOf(ATTACHMENT_BEHAVIORS),
+  /**
+   * Value in px of the margin to place between the target and the content.
+   */
+  attachmentMargin: PropTypes.number,
+  /**
+   * Reference to the bounding container, wil use window if nothing is provided.
```
JakeLaCombe: will*

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,336 @@
...
+   */
+  boundingRef: PropTypes.func,
+  /**
+   * The HookshotContent to be attached.
+   */
+  content: PropTypes.element.isRequired,
+  /**
+   * String pair seperated by a space using values of top, middle, bottom, and left, center, right.
```
JakeLaCombe: *separated.

Also, I think providing an example in the documentation comment can help make this easier to understand. (ex: 'top middle'

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,336 @@
...
+   */
+  contentAttachment: PropTypes.oneOf(ATTACHMENT_POSITIONS).isRequired,
+  /**
+   * String pair of top and left offset, ie "10px -4px".
+   */
+  contentOffset: PropTypes.string,
+  /**
+   * Should content be actively positioned via hookshot.
```
JakeLaCombe: Minor:

I think phrasing it as more a statement than a question helps with the docs

    Determines whether the content should be actively positioned via hookshot

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,336 @@
...
+   */
+  isEnabled: PropTypes.bool,
+  /**
+   * Should the content be presented.
+   */
+  isOpen: PropTypes.bool,
+  /**
+   * Callback function when the content have been positioned.
```
JakeLaCombe: *content has been*

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,336 @@
...
+  getNodeRects() {
+    const targetRect = HookshotUtils.getBounds(this.props.targetRef());
+    const contentRect = HookshotUtils.getBounds(this.contentNode);
+    const boundingRect = HookshotUtils.getBoundingRect(this.props.boundingRef ? this.props.boundingRef() : 'window');
+    return { targetRect, contentRect, boundingRect };
+  }
+
+  tick() {
+    if (this.lastDuration && this.lastDuration > 16) {
+      // Throttle to 60fps, in order to handle sarafi and mobile performance
```
JakeLaCombe: safari*

packages/terra-hookshot/src/Hookshot.jsx
emilyrohrbough: We need to list out the combinations

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,336 @@
...
+   */
+  attachmentMargin: PropTypes.number,
+  /**
+   * Reference to the bounding container, wil use window if nothing is provided.
+   */
+  boundingRef: PropTypes.func,
+  /**
+   * The HookshotContent to be attached.
+   */
+  content: PropTypes.element.isRequired,
```
nickpith: Is there a reason to not use the standard `children` prop? `content` props like this always feel wrong to me and only as a last resort if `children` cannot be used.

packages/terra-hookshot/tests/jest/HookshotDefaultExample.jsx
```diff
@@ -0,0 +1,42 @@
...
+
+    return (
+      <div>
+        <Hookshot
+          content={hookshotContent}
+          contentAttachment="top center"
+          isEnabled
+          isOpen={this.state.isOpen}
+          targetRef={() => document.getElementById('hookshot-standard-button')}
+        />
```
nickpith: Using `children` as opposed to `content` make the usage and examples much more natural in my opinion.

```
<Hookshot
  contentAttachment="top center"
  isEnabled
  isOpen={this.state.isOpen}
  targetRef={() => document.getElementById('hookshot-standard-button')}
>
  {hookshotContent}
</Hookshot>
```

packages/terra-hookshot/CHANGELOG.md
```diff
@@ -0,0 +1,9 @@
+Changelog
+=========
+
+Unreleased
+----------
+
+1.0.0 - (July 26, 2017)
```
ryanthemanuel: This hasn't been released yet :)

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,336 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Portal from 'react-portal';
+import HookshotContent from './_HookshotContent';
+import HookshotUtils from './_HookshotUtils';
+
+const ATTACHMENT_POSITIONS = [
```
ryanthemanuel: This is really nit picky, but is there a reason we're structuring this array of constants like this instead of 'TOP_LEFT', etc.

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,336 @@
...
+      targetAttachment,
+      targetOffset,
+      onPosition,
+    } = this.props;
+    /* eslint-enable no-unused-vars */
+    if (!isOpen) {
+      return null;
+    }
+
+    this.cOffset = HookshotUtils.parseOffset(contentOffset);
```
ryanthemanuel: Can we not abbreviate these?  It's tough when looking elsewhere in the code that this is content and target offset/attachment/etc.

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,336 @@
...
+    const boundingRect = HookshotUtils.getBoundingRect(this.props.boundingRef ? this.props.boundingRef() : 'window');
+    return { targetRect, contentRect, boundingRect };
+  }
+
+  tick() {
+    if (this.lastDuration && this.lastDuration > 16) {
+      // Throttle to 60fps, in order to handle sarafi and mobile performance
+      this.lastDuration = Math.min(this.lastDuration - 16, 100);
+
+      // // Just in case this is the last event, remember to position just once more
```
ryanthemanuel: Don't need both '//'s

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,336 @@
...
+      this.pendingTimeout = setTimeout(this.tick, 100);
+      return;
+    }
+
+    if (this.lastCall && (Hookshot.now() - this.lastCall) < 10) {
```
ryanthemanuel: How did we decide on 10 out of curiosity?

dkasper-was-taken: It's was just trial and error to find a happy medium as i was testing it. 10ms throws out enough without visible jittering.

packages/terra-hookshot/tests/jest/HookshotDefaultExample.jsx
dkasper-was-taken: I went with content in that we only take one element, and it's only ever expect it to be of one type. The goal was to draw attention to the fact that this element isn't being wrapped, so it wouldn't handle an array, and didn't want to introduce logic to ensure it's only one child. It also aligns with the other prop names, so it's clear that content itself is matched with the other content named props.

nickpith: Can't you enforce that with PropTypes though and say that it only supports one. I really think the API is important. This is really a pet peeve of mine as a consumer.

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,336 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Portal from 'react-portal';
+import HookshotContent from './_HookshotContent';
+import HookshotUtils from './_HookshotUtils';
```
emilyrohrbough: We are exposing HookshotContent so we can drop the '_'

packages/terra-hookshot/src/_HookshotUtils.js
```diff
@@ -0,0 +1,420 @@
...
+  right: 'left',
+};
+
+const MIRROR_TB = {
+  middle: 'middle',
+  top: 'bottom',
+  bottom: 'top',
+};
+
+// Same as native getBoundingClientRect, except it takes into account parent <frame> offsets
```
emilyrohrbough: we should document each of these functions to explain what they do and what parameters are passed in

packages/terra-hookshot/docs/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Getting Started
+
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-hookshot`
+  - `yarn add terra-hookshot`
+
+## Usage
+
+```jsx
```
emilyrohrbough: We might want to "beef" up this example a bit to include an example of `<Hookshot.Content >` to pass through and then include or remove the `this.getTargetRef`, `this.handleOnPositions`, etc references that might be confusing here.

packages/terra-hookshot/src/_HookshotUtils.js
```diff
@@ -223,6 +279,13 @@ const rotateContentAttachment = (attachment, angle) => {
   return rAttachment;
 };
 
+/**
+ * This function returns whether or not there is available space for the provided bounds and positions.
+ *
+ * @ param {Object} positions - The positional data to evaluate.
+ * @ param {Object} cRect - The content rectance.
```
emilyrohrbough: *rectangle

packages/terra-hookshot/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "terra-hookshot",
+  "main": "lib/Hookshot.js",
+  "version": "0.0.0",
+  "description": "The Terra Hookshot component positions content according to a targeted attachment, ensuring they stay connected.",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/cerner/terra-core.git"
+  },
+  "keywords": [
```
mjhenkes: add some more to the key words, maybe `position, attached`? As awesome a name as hookshot is, it's not really descriptive

dkasper-was-taken: Updated.

packages/terra-hookshot/src/_HookshotUtils.js
```diff
@@ -0,0 +1,537 @@
...
+ * @ param {Object} tRect - The target rectangle.
+ * @ param {number} margin - The px value of the attachmentMargin.
+ * @ param {string} behavior - The fallback pattern for available rotations.
+ */
+const getRotatedPositions = (positions, cRect, bRect, tRect, margin, behavior) => {
+  if (isValidPositions(positions, cRect, bRect)) {
+    return positions; // default valid
+  }
+
+  if (behavior !== 'none') {
```
emilyrohrbough: The logic in this method seems like it could be simplified. I'm thinking something like 
```
if (behavior === 'none') {
   return positions;
}  else if ( behavior === 'flip') {
// get new positions
} else if ( behavior === 'auto') {
// get new positions
}

return isValidPositions() ? newPositions : positions;
```

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,336 @@
...
+   */
+  content: PropTypes.element.isRequired,
+  /**
+   * String pair separated by a space using values of top, middle, bottom, and left, center, right. (ex: 'top middle')
+   */
+  contentAttachment: PropTypes.oneOf(ATTACHMENT_POSITIONS).isRequired,
+  /**
+   * String pair of top and left offset, ie "10px -4px".
+   */
+  contentOffset: PropTypes.string,
```
kschuste: contentOffset to me sounds like it should only expect 1 offset being given to it; however, here you are specifying that two should be given in the string.  Seems like this should be broken out into two separate offsets or change the name to better resemble that its two offets

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,339 @@
...
+   */
+  targetRef: PropTypes.func.isRequired,
+  /**
+   * String pair of top, middle, bottom, and left, center, right. Will mirror content is none provided. (ex: 'bottom middle')
+   */
+  targetAttachment: PropTypes.oneOf(ATTACHMENT_POSITIONS),
+  /**
+   * String pair of top and left offset, ie "10px -4px".
+   */
+  targetOffset: PropTypes.string,
```
kschuste: same as above comment about it being two

packages/terra-hookshot/docs/README.md
dkasper-was-taken: Site page updated.

packages/terra-hookshot/src/_HookshotUtils.js
dkasper-was-taken: I'm trying to think of a way to trim it down, but it's a fallback pattern:
Primary>180>90>-90>Primary
So each step needs to be validated. Auto includes flip as well.

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,339 @@
...
+  /**
+   * Callback function when the content has been positioned.
+   */
+  onPosition: PropTypes.func,
+  /**
+   * Required element that the content will hookshot to.
+   */
+  targetRef: PropTypes.func.isRequired,
+  /**
+   * String pair of top, middle, bottom, and left, center, right. Will mirror content is none provided. (ex: 'bottom middle')
```
mjhenkes: Typo:
> Will mirror content is none provided

Will mirror content **if** none  provided

packages/terra-hookshot/src/_HookshotUtils.js
```diff
@@ -0,0 +1,558 @@
...
+  const bounds = getBounds(boundingElement);
+  const style = getComputedStyle(boundingElement);
+  const rect = {
+    top: bounds.top,
+    bottom: bounds.top + bounds.height,
+    left: bounds.left,
+    right: bounds.left + bounds.width,
+  };
+
+  // Account any parent Frames scroll offset
```
mjhenkes: Nit: Account for ...

packages/terra-hookshot/themeable-variables.json
```diff
@@ -0,0 +1 @@
...rra-table-tfoot-tr-border-bottom":"1px solid #c8cacb","--terra-table-tfoot-tr-border-top":"0","--terra-table-subheader-background-color":"#c8cacb","--terra-table-subheader-color":"#1c1f21","--terra-table-subheader-font-size":"1rem","--terra-table-subheader-font-weight":"bold","--terra-table-sort-indicator-color":"#1c1f21","--terra-table-cell-min-width-tiny":"3rem","--terra-table-cell-min-width-small":"4.25rem","--terra-table-cell-min-width-medium":"5.75rem","--terra-table-cell-min-width-large":"8.5rem","--terra-table-cell-min-width-huge":"10.5rem","--terra-table-cell-padding-bottom":"0.5em","--terra-table-cell-padding-left":"0.75em","--terra-table-cell-padding-right":"0.75em","--terra-table-cell-padding-top":"0.5em","--terra-table-row-striped-background-color":"#f4f4f4","--terra-table-row-striped-selected-background-color":"#d4ecfb","--terra-table-row-striped-hover-background-color":"#e2f2fc","--terra-table-row-striped-selected-hover-background-color":"#bde1f8","--terra-table-row-selected-background-color":"#ebf6fd","--terra-table-row-selected-border-color":"#0092e0","--terra-table-row-hover-background-color":"#ebf6fd","--terra-table-row-selected-hover-background-color":"#d4ecfb"}}
```
mjhenkes: doesn't the new version that generates this string create a more readable output?

packages/terra-hookshot/src/Hookshot.jsx
dkasper-was-taken: I mainly abbreviated them to display a difference between the parsed and unparsed values. If you have a recommendation I'd be fine changing them.

mjhenkes:
@nickpith We will add your request to our war room discussion and come to a consensus among ourselves.

nickpith:
Thanks. Another example that drives me crazy when I see it is ButtonGroup. It has a `buttons` propy and a `children` prop. I don't even know what the children prop does but I should not have to set `buttons`. I should be able to just do

```
<ButtonGroup>
  <Button />
  <Button />
</ButtonGroup>
```

mjhenkes:
@nickpith Button group actually takes both, You can do exactly what you requested.
```
<ButtonGroup variant="default">
  <ButtonGroup.Button text="Button A" />
  <ButtonGroup.Button text="Button B" />
  <ButtonGroup.Button text="Button C" />
</ButtonGroup>```

File: packages/terra-hookshot/src/Hookshot.jsx (hunk 13 above)
Comment (ryanthemanuel) [reply packages/terra-hookshot/src/Hookshot.jsx]: I'm either ok with this or I am wondering if we like something like:

```
this.content = {
  offset: <blah>
  attachment: <blah>
}
```

File: packages/terra-hookshot/src/HookshotContent.jsx [hunk 27]
```diff
@@ -0,0 +1,129 @@
...
+   */
+  onResize: PropTypes.func,
+  /**
+   * The function returning the frame html reference.
+   */
+  refCallback: PropTypes.func,
+};
+
+const defaultProps = {
+  onEsc: null,
```
JakeLaCombe: Should these functions default to undefined as opposed to null? I feel like having null handlers can cause issues.

packages/terra-site/src/examples/hookshot/HookshotExample.jsx
```diff
@@ -0,0 +1,114 @@
...
+          attachmentMargin={this.state.hookshotAttachmentMargin}
+          content={hookshotContent}
+          contentAttachment={this.state.hookshotContentAttachment}
+          isEnabled
+          isOpen={this.state.isOpen}
+          targetAttachment={this.state.hookshotTargetAttachment}
+          targetRef={() => document.getElementById('hookshot-standard-button')}
+        />
+        <Button id="hookshot-standard-button" text="Hookshot Example" onClick={this.handleButtonClick} />
+        <br />
```
JakeLaCombe: Do we need all these br's? One thing about documentation websites is that developers will copy everything from them, even practices such as putting several br's in a row just for spacing.

packages/terra-hookshot/tests/nightwatch/HookshotTemplate.jsx
```diff
@@ -0,0 +1,100 @@
...
+  isOpen: false,
+  contentAttachment: 'middle right',
+  targetAttachment: 'middle left',
+  buttonMarginTop: '55px',
+  buttonMarginLeft: '220px',
+  buttonMarginRight: '220px',
+  boundingWidth: '540px',
+};
+
+class HookshotTemplate extends React.Component {
```
JakeLaCombe: Picky, but should this be renamed to TestingTemplate? HookshotTemplate makes it seems like a component we can extract out of this folder and move to the src folder.

packages/terra-hookshot/tests/nightwatch/HookshotContentTemplate.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import Hookshot from '../../lib/Hookshot';
+
+const HookshotContent = props => (
+  <Hookshot.Content
+    onEsc
```
JakeLaCombe: Do these need to be provided if they are null or undefined by default?

dkasper-was-taken:
@nickpith Updated to use children.

packages/terra-hookshot/themeable-variables.json
dkasper-was-taken: I think it should, but for some reason it isn't. Anyone have any thoughts?

packages/terra-hookshot/src/Hookshot.jsx
dkasper-was-taken: Still debating this one, are other input?

packages/terra-hookshot/themeable-variables.json
bjankord: Its odd to me this file is even being generated in this directory. I thought @JakeLaCombe had an update to only generate this in the terra-site directory.

packages/terra-hookshot/tests/nightwatch/components/DefaultHookshot.jsx
```diff
@@ -34,11 +34,12 @@ class HookshotStandard extends React.Component {
       <div>
         <Hookshot
           content={hookshotContent}
```
emilyrohrbough: Oops. Removed contentAttachment instead of content

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,335 @@
...
+    if (this.state.isEnabled) {
+      if (!this.listenersAdded) {
+        this.enableListeners();
+      }
+      this.update();
+    }
+  }
+
+  componentWillReceiveProps(newProps) {
+    if (newProps.isEnabled !== this.props.isEnabled || newProps.isOpen !== this.props.isOpen) {
```
tbiethman: Since you're not implementing `shouldComponentUpdate`, you can probably ditch the prop checks and just update the state every time. Render is going to get called regardless.

packages/terra-hookshot/src/Hookshot.jsx
ryanthemanuel: Could it be an object itself?  contentOffset.top and contentOffset.left?

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,335 @@
...
+
+    if (this.props.onPosition) {
+      this.props.onPosition(
+        event,
+        result.positions,
+      );
+    }
+  }
+
+  destroy() {
```
tbiethman: Can this be implemented directly in `componentWillUnmount`?

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,335 @@
...
+    this.disableListeners();
+    this.contentNode = null;
```
tbiethman: You shouldn't need to nullify the ref.

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,335 @@
...
+    this.content = {
+      offset: HookshotUtils.parseOffset(contentOffset),
+      attachment: HookshotUtils.parseAttachment(contentAttachment),
+    };
+    this.target = {
+      offset: HookshotUtils.parseOffset(targetOffset),
+      attachment: targetAttachment ? HookshotUtils.parseAttachment(targetAttachment) : HookshotUtils.mirrorAttachment(this.content.attachment),
+    };
+
+    if (document.getElementsByTagName('html')[0].getAttribute('dir') === 'rtl') {
```
tbiethman: Taking the directionality as a prop instead would make this more flexible (unnecessarily so?)

packages/terra-hookshot/src/HookshotContent.jsx
```diff
@@ -0,0 +1,122 @@
...
+  ESCAPE: 27,
+};
+
+const propTypes = {
+  /**
+   * The children to be displayed as content within the content.
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * The function callback when am escape keydown event occurs..
```
tbiethman: You have an extra period here.

packages/terra-hookshot/src/HookshotContent.jsx
```diff
@@ -0,0 +1,122 @@
...
+class HookshotContent extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleClickOutside = this.handleClickOutside.bind(this);
+    this.handleKeydown = this.handleKeydown.bind(this);
+    this.handleResize = this.debounce(this.handleResize.bind(this), 100);
+  }
+
+  componentDidMount() {
+    if (this.props.onEsc) {
```
tbiethman: Is it possible for a HookshotContent instance to be mounted initially without onEsc/onResize but later receive them through a `componentWillRecieveProps` call? The event listeners would not be added in that case.

packages/terra-hookshot/src/HookshotContent.jsx
bjankord: typo: `am` => `an`
wink

packages/terra-hookshot/themeable-variables.json
bjankord: We should hard code the [file path](https://github.com/cerner/terra-core/blob/master/packages/terra-site/theming-plugin.js#L36) the theming file is output to to always output in terra-site instead of rely on the `__dir` variable.

packages/terra-hookshot/src/Hookshot.jsx
bjankord: What happens if someone provides a string with only one value like `-10px`? Does the left offset default to `0` or would one need to pass in `-10px 0`?

Two props may provide a clean api for defaults, the single prop taking an object could also provide a default object. I lean towards the two props as I think its more friendly to the use case of providing a single direction offset without hindering multi-direction offset.

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,335 @@
...
+const ATTACHMENT_BEHAVIORS = [
+  'auto',
+  'flip',
+  'none',
+];
+
+const propTypes = {
+  /**
+   * How the content should be positioned when the primary attachment is not available.
+   * Valid values: 'auto', 'flip', and 'none'.
```
bjankord: Is the behavior of `auto`, `flip`, `none` defined somewhere. How do consumers know what these do? The answer could just be we have examples of the behavior to highlight it, but I didn't see any written doc on what these do.

packages/terra-hookshot/src/Hookshot.jsx
```diff
@@ -0,0 +1,335 @@
...
+  attachmentMargin: 0,
+  attachmentBehavior: 'auto',
+  contentOffset: '0 0',
+  isEnabled: false,
+  isOpen: false,
+  targetOffset: '0 0',
+};
+
+class Hookshot extends React.Component {
+  static now() {
```
bjankord: In one of our prior war rooms we discussed using static properties, the consensus was to hold off on using them. If we want to revisit that decision, we'd need to look at adding the [class properties plugin](https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-class-properties)

With static props, there is a caveat in IE10. Classes that use static props can't be inherited in that browser. More info: https://babeljs.io/docs/usage/caveats/#internet-explorer-classes-10-and-below-

packages/terra-hookshot/tests/nightwatch/components/DefaultHookshot.jsx
```diff
@@ -0,0 +1,49 @@
...
+  handleButtonClick() {
+    this.setState({ open: true });
+  }
+
+  handleRequestClose() {
+    this.setState({ open: false });
+  }
+
+  render() {
+    const hookshotContent = (
```
bjankord: Can we update the styling on this so it presents a bit better. 
Current display when open:
screen shot 2017-10-10 at 4 21 41 pm


I'm also fine with removing the default all together, the other tests cover everything I think this covers.

packages/terra-hookshot/src/HookshotContent.jsx
```diff
@@ -0,0 +1,122 @@
...
+      onOutsideClick,
+      onResize,
+      refCallback,
+      ...customProps
+    } = this.props;
+
+    // Delete the disableOnClickOutside and enableOnClickOutside prop that comes from react-onclickoutside.
+    delete customProps.disableOnClickOutside;
+    delete customProps.enableOnClickOutside;
+    delete customProps.closePortal;
```
emilyrohrbough: > // Delete the closePortal prop that comes from react-portal.

packages/terra-hookshot/tests/nightwatch/HookshotTestTemplate.jsx
```diff
@@ -0,0 +1,121 @@
...
+  contentAttachment: 'middle end',
+  targetAttachment: 'middle start',
+  buttonMarginTop: '55px',
+  buttonMarginLeft: '220px',
+  buttonMarginRight: '220px',
+  boundingWidth: '540px',
+};
+
+class HookshotTemplate extends React.Component {
+  static attachmentValues(attachment) {
```
bjankord: Same comment as https://github.com/cerner/terra-core/pull/879/files/27ace3aad46fcb0bd77ee0fb305dc219efc0c542#r143846129

packages/terra-site/src/examples/hookshot/BoundedHookshotExample.jsx
```diff
@@ -0,0 +1,162 @@
...
+  'middle start',
+  'middle center',
+  'middle end',
+  'bottom start',
+  'bottom center',
+  'bottom end',
+];
+
+class HookshotStandard extends React.Component {
+  static generateOptions(values) {
```
bjankord: Same comment as https://github.com/cerner/terra-core/pull/879/files/27ace3aad46fcb0bd77ee0fb305dc219efc0c542#r143846129

packages/terra-site/src/examples/hookshot/HookshotExample.jsx
```diff
@@ -0,0 +1,142 @@
...
+  'middle start',
+  'middle center',
+  'middle end',
+  'bottom start',
+  'bottom center',
+  'bottom end',
+];
+
+class HookshotStandard extends React.Component {
+  static generateOptions(values) {
```
bjankord: Same comment as https://github.com/cerner/terra-core/pull/879/files/27ace3aad46fcb0bd77ee0fb305dc219efc0c542#r143846129

packages/terra-hookshot/tests/nightwatch/components/HookshotContentAttachmentBR.jsx
```diff
@@ -5,7 +5,7 @@ class HookshotExample extends React.Component {
   constructor(props) {
```
emilyrohrbough: Sooo these test files were named with the first letters of the content attachments. BR for 'bottom right'. Should those change?

dkasper-was-taken: Hrmmm.... you are correct....

packages/terra-hookshot/tests/nightwatch/components/DefaultHookshot.jsx
dkasper-was-taken: What browser shows that? I'm not seeing it displayed that way.

bjankord: Chrome, it was [this url](https://terra-core-deployed-pr-879.herokuapp.com/static/#/tests/hookshot-tests/default) but it looks fixed now: 
screen shot 2017-10-10 at 8 34 21 pm

PR #883, opened 2017-10-04
I18n plugin filesystem

### LINKED ISSUE #878: When using create-react-app and terra components, npm run build command fails to minify
# Issue Description
When building an app with create-react-app and using terra components, running `npm run build` in the react app errors out with message around failing to minify the code in relation to terra translation files.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

## Expected Behavior
Terra translation files are compatible with create-react-apps with little config on the users part.

## Current Behavior
Terra translation files cause compatibility issues with create-react-apps when trying to minify assets for production builds.

## Steps to Reproduce

1. Run `create-react-app my-app`
2. Eject from create-react-app
3. Add terra-base component to one of the app pages
4. Add i18n-plugin setup required for webpack
5. Run `npm run build`
6. The following output is then generated and the compiler errors out.

```sh
Creating an optimized production build...
Failed to compile.

Failed to minify the code from this file:

 	./config/aggregated-translations/pt.js:12

Read more here: http://bit.ly/2tRViJ9
```

## Environment

* Component Version: terra-i18n-plugin v1.0.0

### LINKED ISSUE #882: terra-i18n-plugin should use the compiler filesystem to create translation files
# Issue Description
When starting the webpack-dev-server, the translations for the terra-site package are generated by the terra-i18n-plugin.  We can use the compiler's input and output filesystems to create these files and remove the dependencies on fs and mkdirp.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

## Expected Behavior
The translation files are generated using the compiler's filesystems.

## Current Behavior
The translation files are generated with the fs and mkdirp dependencies.

kschuste:
### Summary
Resolves #882 and Resolves #878 

Change to use the compilers filesystems to create the translation files once the environment has been set up for the compiler.  Removed the short-hand syntax for creation of the message and set it back to a key-value pair to fix compilation of minified assets.

### Additional Details
This removes the use of fs and mkdirp dependencies.


packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -1,83 +1,86 @@
-import fs from 'fs';
 import path from 'path';
-import mkdirp from 'mkdirp';
 
 let supportedLocales;
 
 function generateTranslationFile(language, messages) {
   return `import { addLocaleData } from 'react-intl';
-import localeData from 'react-intl/locale-data/${language.split('-')[0]}';
+  import localeData from 'react-intl/locale-data/${language.split('-')[0]}';
```
mjhenkes: I don't think you want the added spacing here. I think it would result in a generated translation file that wouldn't pass the linter.

bjankord: +1, I'd leave spacing as is.

.eslintrc
```diff
@@ -12,7 +12,8 @@
   "rules": {
     "max-len": "off", // Remove warnings on max line length exceeding 100 characters
     "react/require-default-props": "off", // Disabled the requirement to default all non-required props
-    "no-multi-assign": "off" // TODO: Decide if we want to turn this on. It was turned on in eslint-config-airbnb-base v14.1.0
+    "no-multi-assign": "off", // TODO: Decide if we want to turn this on. It was turned on in eslint-config-airbnb-base v14.1.0
+    "no-underscore-dangle": "off"
```
kschuste: added this to get around a linter error with accessing '_plugins' from the compiler.  Was not able to access them any other way without having another lint error occur.

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -15,42 +14,48 @@ const messages = ${JSON.stringify(messages, null, 2)};
...
-      } else {
-        // eslint-disable-next-line no-console
+      try {
+        Object.assign(languageMessages[language], JSON.parse(compiler.inputFileSystem.readFileSync(translationFile, 'utf8')));
+      } catch (e) {
         console.warn(`Translation file ${language}.json not found for ${translationsDirectory}`);
       }
     });
+  } catch (e) {
+    console.log(e.message);
```
kschuste: the linter would not allow me to have no catch block or an empty catch block; previously we were not logging out anything here, but I put in the log as nothing else really to do here to get the linter error from occurring.  This causes many many logs to appear when starting the webpack-dev-server that previously did not appear.  It also seems to be causing the Travis CI job to fail as the output becomes too large.  Therefore, should we allow the linter to ignore not having a catch block or possibly do something else here?

mjhenkes: For reference the existsSync call was removed because the default input file system doesn't support it. I believe webpack itself just try-catches file reads. I'm ok with disabling the linter in this specific plugin instance.

emilyrohrbough:
We should update the changelog.

.eslintrc
mjhenkes: Instead of turning it off completely just allow '_plugins' to be accessed in the test file by adding this to the top of the file:
```/* eslint no-underscore-dangle: ["error", { "allow": ["_plugins"] }] */```

File: packages/terra-i18n-plugin/package.json [hunk 4]
```diff
@@ -28,7 +28,9 @@
     "react-dom": "^15.4.2"
   },
   "dependencies": {
-    "mkdirp": "^0.5.1"
+    "enhanced-resolve": "^3.4.1",
+    "memory-fs": "^0.4.1",
+    "webpack": "^3.6.0"
```
mjhenkes: these should be development dependencies

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -65,20 +70,52 @@ function aggregateTranslations(options) {
...
-    } else {
-      throw new Error(`Translation file found for ${language}.json, but translations were not loaded correctly. Please check that your translated modules were installed correctly.`);
-    }
-  });
+  languageMessages = aggregateDirectory(languageMessages, options.baseDirectory, compiler);
+  return languageMessages;
+}
+
+function aggregateTranslations(options, compiler) {
+  if (options.outputFileSystem) {
```
mjhenkes: You should be able to DRY this code out by setting a filesystem variable to either be the passed in file system or the standard file system. You've got alot of repeated code here.

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -15,42 +14,48 @@ const messages = ${JSON.stringify(messages, null, 2)};
 module.exports = {
   areTranslationsLoaded: true,
   locale: '${language}',
-  messages,
+  messages: messages
```
emilyrohrbough: We will want to keep the shorthand notation, updating to `messages: messages` will result in eslint error `expected shorthand property`.

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -15,42 +14,48 @@ const messages = ${JSON.stringify(messages, null, 2)};
...
   const nodeMoudlesPath = path.resolve(currentDirectory, 'node_modules');
-  if (fs.existsSync(nodeMoudlesPath)) {
-    getDirectories(nodeMoudlesPath).forEach((module) => {
-      aggregateDirectory(languageMessages, path.resolve(nodeMoudlesPath, module));
+  try {
+    getDirectories(nodeMoudlesPath, compiler).forEach((module) => {
+      aggregateDirectory(languageMessages, path.resolve(nodeMoudlesPath, module), compiler);
     });
+  } catch (e) {
+    // eslint-disable-line no-use-before-defin
```
emilyrohrbough: Should we output information in these catch's?

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -15,42 +14,48 @@ const messages = ${JSON.stringify(messages, null, 2)};
...
         console.warn(`Translation file ${language}.json not found for ${translationsDirectory}`);
       }
     });
+  } catch (e) {
+    // eslint-disable-line no-use-before-defin
   }
 
   // Check the directory's node_modules for translation files
   const nodeMoudlesPath = path.resolve(currentDirectory, 'node_modules');
-  if (fs.existsSync(nodeMoudlesPath)) {
```
emilyrohrbough: We should probably keep this. If I remember correctly, the `.bin` directory is an example of a node_modules dir that does not contain a node_modules itself which was the original reason for this check.

emilyrohrbough: I wonder if this is the reason for the excessive console logs you were seeing.

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
mjhenkes: It's to resolve this issue: https://github.com/cerner/terra-core/issues/878

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
kschuste: It was decided to ignore outputting anything here, I was outputting the error here, however that provided many more logs that previously were not logged out.  They were also causing the CI job to fail as the output became to large.  Before, we were using existsSynch and did nothing in the case when the directory didn't exist.  Now we try to read from it instead of using existsSynch to determine if the directory is there.

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
mjhenkes: The Try catch is replacing the exists call because the default inputfile system provided by node does not implement existsSync. This is also why we aren't outputting information in the catch blocks, because it's not really an error that the file doesn't exist.

kschuste: The goal of this is to remove as much of the fs as possible and use the compilers inputFileSystem; hence, no existsSynch exists anymore.  Therefore, we use the try catch.  We attempt to read from the directory instead inside getDirectories, which will throw the exception that gets caught.

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
kschuste: The shorthand syntax is what was causing issue #878, therefore changed it back to this.  I've ran the linter and didn't receive any errors as it is inside ``.

emilyrohrbough: ^ The error would have been seen in the resulting file, however I just checked, the `.eslintignore` file actually ignores the aggregated-translations dir so this won't cause an issue either way.

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -15,42 +14,50 @@ const messages = ${JSON.stringify(messages, null, 2)};
...
-    getDirectories(nodeMoudlesPath).forEach((module) => {
-      aggregateDirectory(languageMessages, path.resolve(nodeMoudlesPath, module));
+  try {
+    getDirectories(nodeMoudlesPath, compiler).forEach((module) => {
+      aggregateDirectory(languageMessages, path.resolve(nodeMoudlesPath, module), compiler);
     });
+  } catch (e) {
+    // eslint-disable-line no-use-before-defin
```
bjankord: Is the line to disable `no-use-before-define` needed any more?

packages/terra-i18n-plugin/package.json
bjankord: I think we should update our own usage of webpack in terra-site to webpack v3 before we have plugins using webpack v3. Thoughts @mjhenkes @emilyrohrbough ?

packages/terra-i18n-plugin/package.json
```diff
@@ -21,15 +21,15 @@
   },
   "homepage": "https://github.com/cerner/terra-core#readme",
   "devDependencies": {
-    "babel-jest": "^19.0.0"
+    "babel-jest": "^19.0.0",
+    "enhanced-resolve": "^3.4.1",
+    "memory-fs": "^0.4.1",
+    "webpack": "^3.6.0"
```
bjankord: I'm concerned about pulling in / using webpack 3 in the i18n-plugin. I think we'll want to pull in webpack v2 in the i18n-plugin or update toolkit / terra-site to use webpack 3.

mjhenkes: you should be able to drop this back to the supported webpack version in terra-core

kschuste: updated it to match the webpack version listed under terra-site

emilyrohrbough: Why is webpack a devDependency at all?

kschuste: I use its compiler in the tests
const Compiler = require('webpack/lib/Compiler');

PR #888, opened 2017-10-05
Moved postcss config to its own file. Updated webpack dependencies

mhemesath:
### Summary
This is a prep change to make storybook integration easier and simpler. I moved postcss config to its own file and removed it from webpack. This cleans up the webpack build and makes it slightly easier to understand. 

I also noticed webpack errors during complication and it seems it was a version issue. I took the liberty to update all the dependencies as well which I noticed resolves a nasty issue we ran into with terra-consumer where [CSS files are being concatenated in an indeterministic-order.](https://github.com/webpack/webpack/pull/5225). This makes using CSS custom properties more predictable when expecting `:root{}` declarations to override previously imported components props.


postcss.config.js
```diff
@@ -0,0 +1,3 @@
+const config = require('./packages/terra-site/postcss.config');
```
mhemesath: this file doesn't do anything right now, but it will be used by storybook once its added and will ensure storybook uses the same postcss config as terra-site.

postcss.config.js
```diff
@@ -0,0 +1,3 @@
...
+
+module.exports = config;
```
mhemesath: this file doesn't do anything right now, but it will be used by storybook once its added and will ensure storybook uses the same postcss config as terra-site.

packages/terra-site/postcss.config.js
```diff
@@ -0,0 +1,23 @@
...
+module.exports = {
+  plugins() {
+    return [
+      Autoprefixer({
+        browsers: [
+          'ie >= 10',
+          'last 2 versions',
+          'last 2 android versions',
+          'last 2 and_chr versions',
+          'iOS >= 8',
```
mhemesath: should this be bumped to 10?

mjhenkes: Yes, our currently minimum supported version is 10.2

packages/terra-site/package.json
```diff
@@ -88,25 +88,25 @@
...
     "raw-loader": "^0.5.1",
     "react": "^15.4.2",
     "react-dom": "^15.4.2",
-    "sass-loader": "^6.0.3",
-    "style-loader": "^0.16.1",
-    "webpack": "^2.6.1",
-    "webpack-dev-server": "^2.4.5"
+    "sass-loader": "^6.0.6",
+    "style-loader": "^0.19.0",
+    "webpack": "^3.6.0",
```
emilyrohrbough: I'm not sure if we will be able to bump webpack versions until terra-toolkit's webpack version is bumped. I when I made changes in toolkit for nightwatch, I ran across webpack chunk issues in consumer when trying to run nightwatch tests. 

Do the nightwatch tests run without error?

bjankord: We'll see after travisCI runs.

bjankord: It appears there are nightwatch test failures. @mhemesath It would be good to sync with @emilyrohrbough and @ryanthemanuel on this.

bjankord: It appears there are nightwatch tests failures after updating webpack. We'll want to get this resolved, feel free to reach out to @emilyrohrbough and @ryanthemanuel as there the most familiar with this.

mhemesath: This PR should address those failures https://github.com/cerner/terra-toolkit/pull/14/files

bjankord: Gonna clear out travis CI's node_modules cache and rerun the build. This *should* pick up the toolkit v2.1.0 release

mhemesath: oh, i explicitly updated to it

bjankord: Might need to update the version in terra-site as well: https://github.com/cerner/terra-core/blob/master/packages/terra-site/package.json#L109-L110

mhemesath: i don't think it actually uses that version

bjankord: Ok once tests pass, lets get this merged down.

mhemesath:
Hurray!

PR #896, opened 2017-10-06
Fix Site Bidi

emilyrohrbough:
### Summary
Oops. When we pulled out the postCss configuration to its own file, we forgot to include the configuration in our wepback configuration.


mhemesath:
I'm confident I checked this and verified it worked.....

PR #898, opened 2017-10-06
Update Toggle Docs

### LINKED ISSUE #676: Update Toggle Documentation & Examples
# Issue Description
The toggle component should recommend or make reference to the use of the toggle button in its documentation given they are separate components. 

Additionally, on the site examples, a button is used to toggle displays in and out. This should be updated to use the ToggleButton component. It might also be nice for consumers to provide an example that does not use a button to toggle information .

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

yuderekyu:
### Summary
Resolves #676.
* Button examples have been refactored to icons - doesn't make much sense to provide toggle examples with buttons when we have a specific toggle-button component.

### DIFF
--- a/packages/terra-site/CHANGELOG.md
+++ b/packages/terra-site/CHANGELOG.md
@@ -8,6 +8,7 @@ Unreleased
 ------------------
 ### Changed
 * Minor version bump
+* Updated toggle docs
 
 1.12.0 - (September 26, 2017)
 ------------------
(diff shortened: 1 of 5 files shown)


packages/terra-site/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Unreleased
 ------------------
 ### Changed
 * Minor version bump
+* Updated toggle docs
```
emilyrohrbough: Updated terra-toggle _examples_

PR #899, opened 2017-10-06
DO NOT MERGE - Figure out what's breaking IE

jmsv6d:
### Summary
Figuring out what's breaking IE

### Additional Details
If you have anything else that you think may be relevant to this issue, list it here. Additional information can help us better understand your changes and speed up the review process.

Please add your name to the [CONTRIBUTORS.md] file. Adding your name to the [CONTRIBUTORS.md] file signifies agreement to all rights and reservations provided by the [License].

Thanks for contributing to Terra. 
@cerner/terra

[CONTRIBUTORS.md]: ../blob/master/CONTRIBUTORS.md
[License]: ../blob/master/LICENSE

PR #901, opened 2017-10-08
Set webpack dev server at 2.7.1 for IE10 compatibility

mhemesath:
### Summary
Set webpack dev server at 2.7.1 for IE10 compatability. see https://github.com/webpack/webpack-dev-server#caveats

This should fix https://github.com/cerner/terra-core/pull/899


mhemesath:
@emilyrohrbough @mjhenkes should I be adding all these `DEPENDENCIES.md` changes to the PR?

mjhenkes:
In my opinion we should rollback the dependencies.md generation. I'm not convinced it adds value and it adds confusion.

mhemesath:
Sounds good, but for this pull request should I include them or leave them out?

mjhenkes:
Include them for now.

bjankord:
> In my opinion we should rollback the dependencies.md generation. I'm not convinced it adds value and it adds confusion.

Agreed, lets get a PR to pull them out of the post install script and we can figure out how we want to display that info in the future.

bjankord:
I've tested the deployment URL in IE10 VM, IE 11 VM, and MS Edge VM and all the doc pages loaded as expected.

PR #902, opened 2017-10-09
Cleanup babel and webpack config:

mhemesath:
* Switch to use [module.rules](https://webpack.js.org/configuration/module/#module-rules) in webpack
* Remove [json-loader as its included now](https://github.com/webpack-contrib/json-loader)
* [Update babel to use babel-preset-env](https://babeljs.io/env/) instead of the 2015 plugin
* Added [transform runtime plugin](https://babeljs.io/docs/plugins/transform-runtime/) which decreases dev build size by 40K
* Stop transpiling modules in babel so [webpack can take advantage of tree shaking](https://medium.com/@johnstew/webpack-tree-shaking-20914b7a9ca5).
* Updated all code loaded through webpack to use import/export. Change code loaded in webpack itself to use require.


mhemesath:
Hold off on merging this until https://github.com/cerner/terra-core/pull/901 is completed so I can integrate and test these all together.

mhemesath:
With preset-env, consumer (which supports IE11+) doesn't need to transpile `const` and `let` for example. At most this will give IE a performance bump, at the very least hopefully transpiling doesn't take as long. I think it also makes it clearer in babel config which browsers are supported.

.babelrc
```diff
@@ -1,4 +1,8 @@
 {
-  "presets": ["es2015", "react"],
+  "presets": ["react", ["env", {
+    "targets": {
```
bjankord: Should this mirror what we use for autoprefixer: https://github.com/cerner/terra-core/blob/master/packages/terra-site/postcss.config.js#L12-L16

mhemesath: Potentially. Its slightly different because whereas CSS uses prefixes  JavaScript its either ther e or it isn't. That means that being we support IE10 its a pretty strong bet that will include anything chrome, safari and firefox is missing that we do support.   Once we no longer support IE10, then we'll probably want to be more nuanced about it. Because supporting IE10 basically means everythign has to be transpiled.

bjankord: Lets update this to `"browsers": ["last 2 versions", "iOS >= 10"]` 
IE 10 & 11 will be captured by the `last 2 versions` string.

This config matches the same results as our current autoprefixer setup. 

screen shot 2017-10-09 at 9 49 04 am

We should also update autoprefixer to use the same config.

If we drop the `iOS >= 10` part and go with `"browsers": ["last 2 versions"]`, we'd drop iOS 10.0-10.2 and iOS 10 "support" would start at `10.3`. I'd defer to @mjhenkes on what level of iOS 10 support we need.

screen shot 2017-10-09 at 9 49 21 am

mjhenkes: We currently support iOS 10.2 and up.

bjankord: Ok sounds like we need `"browsers": ["last 2 versions", "iOS >= 10"]`

mhemesath:
Before we merge this I need to thoroughly test these settings and make sure the transform-runtime plugin integrates well with projects downstream.

.babelrc
```diff
@@ -1,4 +1,13 @@
 {
-  "presets": ["es2015", "react"],
-  "plugins": ["transform-object-assign", "transform-object-rest-spread"],
+  "presets": ["react", ["env", {
+    "targets": {
+      "browsers": ["last 2 versions", "iOS >= 10"]
+    },
+    "modules": false
```
mhemesath: this stops babel from transpiling the import/export statements. This enables webpack to process those imports and do proper tree shaking

packages/terra-base/src/Base.jsx
```diff
@@ -48,14 +48,14 @@ class Base extends React.Component {
 
   componentDidMount() {
     if (this.props.locale !== undefined) {
-      i18nLoader(this.props.locale, this.setState, this);
+      i18nLoader(this.props.locale).then(i18n => this.setState(i18n.default));
```
mhemesath: I updated this code to use promises... much cleaner and works better with dynamic imports.

packages/terra-demographics-banner/src/_sharedObjects.jsx
```diff
@@ -72,7 +72,7 @@ const applicationIdentifiers = (props) => {
   return null;
 };
 
-module.exports = {
+export {
```
mhemesath: Now that webpack is processing the imports, everything it processes much be using import/export syntax

packages/terra-i18n-plugin/.babelrc
```diff
@@ -0,0 +1,4 @@
+{
```
mhemesath: This file is odd. Basically things processed by webpack we don't want modules compiled, but things not processed by webpack (e.g. plugins imported into webpack config) need modules compiled because they are executed by node, not webpack.

packages/terra-i18n/src/i18nLoader.js
```diff
@@ -15,15 +15,8 @@ const permitParams = (locale, callback) => {
...
-    require.ensure([], (require) => {
-      require('intl');
-      intlLoaders[locale]();
-      translationLoaders[locale](callback, scope);
-    }, 'intl-polyfill');
-  } else {
-    translationLoaders[locale](callback, scope);
-  }
+export default (locale) => {
+  return import(/* webpackChunkName: "intl-polyfill" */'intl')
```
mhemesath: update to use promises and dynamic imports

packages/terra-site/webpack.config.js
```diff
@@ -11,18 +11,13 @@ const i18nSupportedLocales = require('terra-i18n/lib/i18nSupportedLocales');
 
 module.exports = {
   entry: {
-    'babel-polyfill': 'babel-polyfill',
```
mhemesath: babel-transform-runtime does the same thing as babel-polyfill but scoped (rather than global). I need to verify it works as expected and if not I'll revert babel-polyfill.

packages/terra-site/webpack.config.js
```diff
@@ -63,6 +64,7 @@ module.exports = {
       supportedLocales: i18nSupportedLocales,
     }),
     new webpack.NamedChunksPlugin(),
+    new UglifyJSPlugin(),
```
mjhenkes: does this have an effect on compilation speed? I think we noticed this slowed kaiju down quite a bit. Should we have a prod/dev webpack config?

mhemesath: Yeah, its really slow. I'm just pushing it up to ensure it deploys correctly then I'll move it to the prod config.

mhemesath:
closing this PR. I'm going to submit this as a series of smaller PRs. I talked to @bjankord and tree shaking and dynamic imports is a larger initiative to tackle later

mhemesath:
please leave this branch open as a reference as we may use it for i18n changes/etc.

PR #904, opened 2017-10-09
Update Site Header to use Collapsible Menu View

### LINKED ISSUE #838: Uplift site header to use collapsible menu view
# Issue Description
Currently the header will break at smaller view port sizes. Uplifting to use the collapsible menu view will collapse the options that cannot fit into a menu. 

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Header content should not overlap each other when the view port is small

## Current Behavior
screen shot 2017-09-20 at 1 47 16 pm

emilyrohrbough:
### Summary
Uplifted terra-site to use the collapsible menu view such that the options collapse at smaller view port sizes. Closes #838.

### DIFF
--- a/packages/terra-site/src/App.jsx
+++ b/packages/terra-site/src/App.jsx
@@ -3,13 +3,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { Link } from 'react-router';
 import Base from 'terra-base';
-import Button from 'terra-button';
-import ButtonGroup from 'terra-button-group';
 import SlidePanel from 'terra-slide-panel';
 import ContentContainer from 'terra-content-container';
 import List from 'terra-list';
 import IconMenu from 'terra-icon/lib/icon/IconMenu';
 import ThemeProvider from 'terra-theme-provider';
+import CollapsibleMenuView from 'terra-collapsible-menu-view';
 import styles from './site.scss';
 
 import FormComponentNavigation from './examples/form/FormComponentNavigation';
@@ -22,26 +21,39 @@ const propTypes = {
 
 const locale = document.getElementsByTagName('html')[0].getAttribute('lang');
 
+const themes = {
+  'Defualt Theme': '',
+  'Consumer Theme': 'cerner-consumer-theme',
+  'Mock Theme': 'cerner-mock-theme',
+};
+
 class App extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
       isOpen: window.innerWidth >= 768,
+      isLtr: true,
       locale,
-      theme: '',
+      theme: 'Default Theme',
     };
+    this.handleBidiChange = this.handleBidiChange.bind(this);
     this.handleThemeChange = this.handleThemeChange.bind(this);
     this.handleLocaleChange = this.handleLocaleChange.bind(this);
     this.handleToggleClick = this.handleToggleClick.bind(this);
     this.handleResetScroll = this.handleResetScroll.bind(this);
   }
 
+  handleBidiChange(e) {
+    document.getElementsByTagName('html')[0].setAttribute('dir', e.currentTarget.id);
+    this.setState({ isLtr: e.currentTarget.id === 'ltr' });
+  }
+
   handleLocaleChange(e) {
-    this.setState({ locale: e.target.value });
+    this.setState({ locale: e.currentTarget.id });
   }
 
   handleThemeChange(e) {
-    this.setState({ theme: e.target.value });
+    this.setState({ theme: e.currentTarget.id });
   }
 
   handleToggleClick() {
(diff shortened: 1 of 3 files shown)


dkasper-was-taken:
Should we add the CollapsibleMenuViewDivider to separate the rtl elements from the other drill in items?

packages/terra-site/src/site.scss
```diff
@@ -64,33 +64,19 @@
...
     width: 100%;
   }
 
-  .site-utility {
-    display: flex;
-    flex: 1 1 auto;
-    justify-content: flex-end;
-    position: relative;
-    width: 100%;
+  .site-header > div:first-child{
```
StephenEsser: Nit: This bracket should have spacing

packages/terra-site/src/site.scss
```diff
@@ -64,33 +64,19 @@
...
-    padding-right: 1.25rem;
-    position: relative;
-  }
-
-  .site-bidi {
-    align-self: center;
-    flex: 0 0 auto;
-    position: relative;
+  .site-header > div[class*="divider"][class*="face-up-item"]{
+    border-right: none;
```
StephenEsser: Why are we hiding this?

emilyrohrbough: We don't necessarily have not. Here is what it looks like without styling:

screen shot 2017-10-09 at 2 16 17 pm

vs with:

screen shot 2017-10-09 at 2 16 53 pm

emilyrohrbough: This has no affect on the rolled up divider

StephenEsser: It feels like we are breaking our own rules if we are adding custom CSS to change the behavior / styles of our own Component. 

If there is value in hiding the divider here and showing it in the menu it might be worth talking with UX.

packages/terra-site/src/App.jsx
```diff
@@ -60,36 +72,33 @@ class App extends React.Component {
...
+        shouldCloseOnClick={false}
+        subMenuItems={[
+          <CollapsibleMenuView.Item id="en" text="en" key="en" onClick={this.handleLocaleChange} />,
+          <CollapsibleMenuView.Item id="en-GB" text="en-GB" key="en-GB" onClick={this.handleLocaleChange} />,
+          <CollapsibleMenuView.Item id="en-US" text="en-US" key="en-US" onClick={this.handleLocaleChange} />,
+          <CollapsibleMenuView.Item id="de" text="de" key="de" onClick={this.handleLocaleChange} />,
+          <CollapsibleMenuView.Item id="es" text="es" key="es" onClick={this.handleLocaleChange} />,
+          <CollapsibleMenuView.Item id="fr" text="fr" key="fr" onClick={this.handleLocaleChange} />,
+          <CollapsibleMenuView.Item id="pt" text="pt" key="pt"onClick={this.handleLocaleChange} />,
+          <CollapsibleMenuView.Item id="fi-FI" text="fi-FI" key="fi-FI" onClick={this.handleLocaleChange} />,
```
StephenEsser: We should store the selected state to checkmark the selected option. 

image

packages/terra-site/src/App.jsx
```diff
@@ -22,26 +21,39 @@ const propTypes = {
...
+  'Consumer Theme': 'cerner-consumer-theme',
+  'Mock Theme': 'cerner-mock-theme',
+};
+
 class App extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
       isOpen: window.innerWidth >= 768,
+      isLtr: true,
```
StephenEsser: This is completely up to preference. An alternative to `isLtr` would be `dir`. Which would read "The state of directionality is left to right / right to left" instead of "The state of left to right is true / false"

packages/terra-site/src/App.jsx
```diff
@@ -22,26 +21,39 @@ const propTypes = {
 
 const locale = document.getElementsByTagName('html')[0].getAttribute('lang');
 
+const themes = {
+  'Defualt Theme': '',
```
yuderekyu: Misspelled `Default Theme`

packages/terra-site/src/App.jsx
emilyrohrbough: The [current implementation](https://github.com/cerner/terra-core/blob/master/packages/terra-collapsible-menu-view/src/CollapsibleMenuViewItem.jsx#L139) of CollapsibleMenu Subgroup Items does not carry forward a selection state. I think the limitation here is we are wanting select dropdown functionality via  submenu component.

jmsv6d: @emilyrohrbough Selection state would be set and maintained on the items passed through. If it's not working then it needs to be logged as a bug.
Also, I would have expected these options to be put in a selectable item group.

jmsv6d: `subMenuItems={[`
&nbsp;&nbsp;&nbsp;`<CollapsibleMenuView.ItemGroup isSelectable key="locale">`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`<CollapsibleMenuView.Item id="en" text="en" key="en" onClick={this.handleLocaleChange} isSelected={this.state.locale === 'en'} />`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`<CollapsibleMenuView.Item id="en-GB" text="en-GB" key="en-GB" onClick={this.handleLocaleChange} isSelected={this.state.locale === 'en-GB'} />`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`<CollapsibleMenuView.Item id="en-US" text="en-US" key="en-US" onClick={this.handleLocaleChange} isSelected={this.state.locale === 'en-US'} />`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`<CollapsibleMenuView.Item id="de" text="de" key="de" onClick={this.handleLocaleChange} isSelected={this.state.locale === 'de'} />`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`<CollapsibleMenuView.Item id="es" text="es" key="es" onClick={this.handleLocaleChange} isSelected={this.state.locale === 'es'} />`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`<CollapsibleMenuView.Item id="fr" text="fr" key="fr" onClick={this.handleLocaleChange} isSelected={this.state.locale === 'fr'} />`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`<CollapsibleMenuView.Item id="pt" text="pt" key="pt" onClick={this.handleLocaleChange} isSelected={this.state.locale === 'pt'} />`
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; `<CollapsibleMenuView.Item id="fi-FI" text="fi-FI" key="fi-FI" onClick={this.handleLocaleChange} isSelected={this.state.locale === 'fi-FI'} />`
&nbsp;&nbsp;&nbsp;`</CollapsibleMenuView.ItemGroup>,`
`]}`

jmsv6d: The CollapsibleMenuView.Items are not selectable unless they are in an ItemGroup, otherwise you would need to use the CollapsibleMenuView.Toggle to get the checkmark styling.

emilyrohrbough: Ah. I did not consider placing these subitems into a single ItemGroup subitem

packages/terra-site/src/App.jsx
```diff
@@ -60,36 +72,35 @@ class App extends React.Component {
...
+            <CollapsibleMenuView.Item id="en-GB" text="en-GB" key="en-GB" isSelected={this.state.locale === 'en-GB'} onClick={this.handleLocaleChange} />
+            <CollapsibleMenuView.Item id="en-US" text="en-US" key="en-US" isSelected={this.state.locale === 'en-US'} onClick={this.handleLocaleChange} />
+            <CollapsibleMenuView.Item id="de" text="de" key="de" isSelected={this.state.locale === 'de'} onClick={this.handleLocaleChange} />
+            <CollapsibleMenuView.Item id="es" text="es" key="es" isSelected={this.state.locale === 'es'} onClick={this.handleLocaleChange} />
+            <CollapsibleMenuView.Item id="fr" text="fr" key="fr" isSelected={this.state.locale === 'fr'} onClick={this.handleLocaleChange} />
+            <CollapsibleMenuView.Item id="pt" text="pt" key="pt" isSelected={this.state.locale === 'pt'} onClick={this.handleLocaleChange} />
+            <CollapsibleMenuView.Item id="fi-FI" text="fi-FI" key="fi-FI" isSelected={this.state.locale === 'fi-FI'} onClick={this.handleLocaleChange} />
+          </CollapsibleMenuView.ItemGroup>,
```
jmsv6d: You could simplify this a little more by using the onChange for the group instead of the onClick for each item.

emilyrohrbough: 🙈 You're right. Good call.

mmalaker:
Collapsible menu works as expected for the site header bar. Only issue is that in IE and Edge, instead of the Theme button, there is a small, gray button artifact present. it is clickable but not functional (I understand that dynamic theming not enabled in IE/Edge).

emilyrohrbough:
@mmalaker it should be fixed now

mmalaker:
Button artifact is fixed now, but I found a couple other issues: IE/Edge shows ellipse menu button even if the header bar is not collapsed. Edge doesn't collapse toolbar into the menu at all.

emilyrohrbough:
@mmalaker the menu appears to be collapsing correctly if the screen is small enough when testing on both 11 & Edge.

@jmsv6d do you have any insight into why the menu button is always displaying on IE/Edge? The menu example does not seem to have this issue.

jmsv6d:
My first guess would be that the way you have the toggleContent in the CollapsibleMenuView and modify the styling so that it will be left aligned is messing with the width calculation logic. I would create a div that has the header styling, then pull the toggleContent out of the CollapsibleMenuView, making the toggleContent left aligned and the CollapsibleMenuView right aligned and see if that fixes the issue.

emilyrohrbough:
@mmalaker I've fixed the header such that the ellipse menu button only appears when expected.

mmalaker:
Functional verification completed. The site header components now display correctly and responsive behavior functions as expected.

Issue #881, opened 2017-10-03
Adjust Type Check in SingleSelectableRows of terra-table


dv297:
# Issue Description
[In SingleSelectableRows](https://github.com/cerner/terra-core/blob/terra-table%401.11.0/packages/terra-table/src/SingleSelectableRows.jsx#L119), there is a line that checks if the child element is a "TableRow". This type check is included to ensure that TableHeaders are not given any "selection" related props when wrapped by SingleSelectableRows 

However, many consumers of component libraries will create consumption layers above the component. This gives the consumer the flexibility to compose the components however they desire and separate out concerns of a Row to its own component, rather than cluttering the Table component with the display of the Row. 

For example, a consumer may create a "FooTable" which displays information about foo in "FooRow"'s. FooRow will use TableRows internally, but because of this type check, the FooRow won't get the new props normally given to TableRows by SingleSelectableRows.

```
const FooRow = ({ foo }) => {
  const fooDate = foo.date.format(desiredFormat);
  const { a, b, c } = foo;
  
  return (
    <Table.Row>
      <Table.Cell key={'COLUMN_0'} content={fooDate} />
      <Table.Cell key={'COLUMN_1'} content={a} />
      <Table.Cell key={'COLUMN_2'} content={b} />
      <Table.Cell key={'COLUMN_3'} content={c} />
    </Table.Row>
  );
};
```

When wrapped by SingleSelectableRows, FooRow cannot be selected. This type check could be adjust to check if something is a TableHeader, preserving the intended behavior but opening up a more desired behavior of allowing custom components that use TableRow internally. Other solutions may also be desired, as there might be a use case for people to create their own TableHeader, although this seems less likely  than creating their own custom TableRow. 

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Any component that renders TableRow and is consumed by SingleSelectableRows should be able to be selected.

## Current Behavior
Any component that renders TableRow and is consumed by SingleSelectableRows cannot be selected.

emilyrohrbough:
Resolved in #906

PR #906, opened 2017-10-10
Update Single Selectable Rows typecheck

yuderekyu:
### Summary
Addresses #881. Now, `terra-table-header` and `terra-table-subheader` cannot be passed selection props.


emilyrohrbough:
You dropped your changes somewhere with your merges

yuderekyu:
Yeah, I forgot I didn't move the change into the new `SelectableTableRows`. Commit coming up.

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -130,7 +131,6 @@ class SelectableTableRows extends React.Component {
 
 SelectableTableRows.propTypes = propTypes;
 SelectableTableRows.defaultProps = defaultProps;
-SelectableTableRows.Row = TableRow;
```
kschuste: probably want to keep this in here for passivity

kschuste: @bjankord @mjhenkes Should we keep this line here for passivity reasons since this is part of what is being exported?

emilyrohrbough: Yes. We will want to keep it. 

SelectableTableRows.Row gives consumers access to the TableRow component directly through the SelectableTable rows. We provide this type of consumption for several packages that use multiple components to compose the full structure. Here are a examples:

- [ButtonGroup.Button](https://github.com/cerner/terra-core/blob/master/packages/terra-button-group/src/ButtonGroup.jsx#L136)

- [Menu.Item or Menu.ItemGroup or Menu.Divider](https://github.com/cerner/terra-core/blob/master/packages/terra-menu/src/Menu.jsx#L190)

yuderekyu: Makes sense. All three table nightwatch tests fail on the initial table render with this export line included -- looking more into it.

emilyrohrbough: What are your test failures?

yuderekyu: screen shot 2017-10-11 at 12 07 56 pm

yuderekyu: Meh, thinking something is only wrong on my local build. Pushing to see what Travis says.

Issue #909, opened 2017-10-11
Add disableCloseOnEsc Prop to Overlay


emilyrohrbough:
# Issue Description
Provide an opt-out of the defaulted close on escape behavior.

Proposal to add `disableCloseOnEsc` prop. Default value `false`. 

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Allow consumers to decide if close on escape is desired.

## Current Behavior
When an onRequestClose function is provided, overlay always closes on escape.

bjankord:
@emilyrohrbough What's the logic around disabling the ability to close via `esc` key? This feature is built in for accessibility reasons.

emilyrohrbough:
I came across this prop in Modal. It made me consider the possibility that consumers may want to disable close on escape for overlays as well, in particular overlays relative to container.

emilyrohrbough:
Holding off on adding this functionality until a specific need is in place.

PR #911, opened 2017-10-11
Added disableCloseOnEsc prop to support on close behavior

emilyrohrbough:
### Summary
Addresses #909

Provide an opt-out of the defaulted close on escape behavior.


emilyrohrbough:
Holding off on adding this functionality until a specific need is in place.

Issue #897, opened 2017-10-06
Uplift Menu to consume popup with hookshot


jmsv6d:
# Issue Description
Uplift Menu to consume popup with hookshot.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
This should allow a menu to be dynamically sized by it's content. Also, the menu should place first focus on the first menu item.

## Current Behavior
Menu is sized with a best guess of height based on number of items and given one of the limited popup height options. The menu also currently will focus on the popup container first on tab, before focusing on the first menu item.

jmsv6d:
Moving to blocked until Popup changes make it through PR.

Issue #797, opened 2017-09-12
Popup closes when keyboard opens on android


hboege:
# Issue Description
When the keyboard displays on a tablet, it causes the view to resize, thus closing any open popups.  If the popup contains a text field, there is no way for a tablet user to enter text.

## Issue Type
- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior
A popup that has a text box should not close when the keyboard opens.

## Current Behavior
The popup closes when the view resizes.

## Steps to Reproduce
We will post a video with a specific example, but here is the general idea.
1. On a tablet, display a modal.  Display a popup, with a text box, from the header in the modal.
2. Click in the text box, so it has focus and the keyboard displays. 
3. Notice the popup disappears.

## Environment
* Component Version: terra-popup 1.8.0
* Browser Name and Version: Chrome
* Operating System and version (desktop or mobile): mobile Samsung tablet

dkasper-was-taken:
We will need to discuss possible solutions, as exposing the ability to disable resize dismissal may cause vulnerabilities, so we need to evaluate concerns.

hboege:
David had done some more testing, and found that the keyboard caused the popup to close on an Android device, but not iOS device.

neilpfeiffer:
Adding to this issue a similar bug discovered by Message Center. A summary of the issue they've reported:

### Issue Description/Current Behavior
Message Center has a custom toolbar with a compose message box (similar to a text messaging input) and a "send" button and "more" (plus/"+") button. When the keyboard is showing, pressing the "more" (plus/"+") button closes both the keyboard and the popup (not the intended use).

### Expected Behavior
Keyboard stays displayed and the popup displays over the page content (or if we need a new control that provide an action sheet or new row of actions) so user can add a photo or a quick message. _(See attached screenshot for expected behavior.)_

### Steps to Reproduce
1. The compose message text box is in focus and keyboard is displayed…
2. The user selects the "more" (plus/"+") button to display the quick message/picture message popup
3. The popup flashes and disappears, and the keyboard is removed. _(Once the keyboard is removed and the “more” button is selected it works just fine, but that’s obviously not intuitive to a user.)_

![image](https://user-images.githubusercontent.com/1355485/30817412-6a953552-a1de-11e7-9d37-d203d7230571.png)

laurunning:
[...]

PR #912, opened 2017-10-11
Terra popup with hookshot

dkasper-was-taken:
### Summary
Uplift terra-popup to consume terra-hookshot, and introduce dynamic sizing behavior.
Address: #897 #797 

@cerner/terra


packages/terra-popup/package.json
```diff
@@ -31,18 +31,19 @@
...
     "focus-trap-react": "^3.0.2",
     "prop-types": "^15.5.8",
-    "react-onclickoutside": "^5.11.1",
     "react-portal": "^3.0.0",
     "terra-base": "^2.6.0",
     "terra-content-container": "^1.11.0",
-    "terra-icon": "^1.12.0",
-    "tether": "^1.4.0"
+    "terra-hookshot": "^0.x",
+    "terra-icon": "^1.11.0"
```
emilyrohrbough: keep  "terra-icon": "^1.12.0"

packages/terra-popup/src/_PopupContent.jsx
```diff
@@ -123,21 +77,37 @@ const propTypes = {
...
+  isWidthDynamic: false,
 };
 
 class PopupContent extends React.Component {
-  static getContentStyle(arrow, arrowPosition, height, maxHeight, width, maxWidth) {
-    const validHeight = maxHeight <= 0 || height <= maxHeight ? height : maxHeight;
-    const validWidth = maxWidth <= 0 || width <= maxWidth ? width : maxWidth;
-    return { height: `${validHeight.toString()}px`, width: `${validWidth.toString()}px` };
+  static getDimensionStyle(value, maxValue, isDynamic) {
+    if (value > 0) {
```
yuderekyu: Could we use a guard clause here? 
```
if (value <= 0) {
  return null;
}
....
```

packages/terra-popup/src/_PopupUtils.js
```diff
@@ -10,124 +19,270 @@ const MIRROR_TB = {
...
   const offset = { vertical: 0, horizontal: 0 };
+
   if (targetNode) {
-    if (isVerticalAttachment(attachment) && targetNode.clientWidth <= (arrowOffset * 2) + cornerOffset) {
-      if (attachment.horizontal === 'left') {
-        offset.horizontal = (arrowOffset + cornerOffset) - (targetNode.clientWidth / 2);
-      } else if (attachment.horizontal === 'right') {
-        offset.horizontal = -((arrowOffset + cornerOffset) - (targetNode.clientWidth / 2));
+    const segment = arrowOffset + cornerOffset;
+    if (isVerticalAttachment(cAttachment)) {
```
yuderekyu: Can we also use a guard clause here? It's clearer on what's happening without more nesting.
```
if (isVerticalAttachment(cAttachment)) == false) {
  return offset;
}
if (cAttachment.horizontal ! == tAttachment.horizontal) {
....
return offset;
```

packages/terra-popup/src/_PopupContent.jsx
dkasper-was-taken: The return null also acts as an "else" if we aren't constrained and we are dynamic to allow for the initial layout.

packages/terra-popup/src/Popup.jsx
```diff
@@ -212,57 +245,57 @@ class Popup extends React.Component {
...
+      </div>
     );
   }
 }
 
 Popup.propTypes = propTypes;
 Popup.defaultProps = defaultProps;
+Popup.attachmentPositions = ATTACHMENT_POSITIONS;
+Popup.heights = PopupHeights;
+Popup.widths = PopupWidths;
```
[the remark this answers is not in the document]
dkasper-was-taken: Updated.

packages/terra-popup/src/_PopupUtils.js
```diff
@@ -10,124 +19,270 @@ const MIRROR_TB = {
...
+ */
+const doesArrowFitVertical = (contentRect, targetRect, arrowOffset, cornerOffset) => {
+  const fitsTopBound = calculateBottomPosition(contentRect) - arrowOffset - cornerOffset >= targetRect.top;
+  const fitsBottomBound = contentRect.top + arrowOffset + cornerOffset <= calculateBottomPosition(targetRect);
+
+  return fitsTopBound && fitsBottomBound;
+};
+
+/**
+ *
```
kschuste: missing description for this function

packages/terra-popup/docs/README.md
```diff
@@ -1,7 +1,6 @@
 # Terra Popup
 
-The Terra Popup is higher order component that launches tether positioned content with the ability to display a dynamic arrow.
-- Additional tether documentation can be found on their site [TetherJS](http://tether.io/)
+The Terra Popup is higher order component that launches terra-hookshot positioned content with the ability to display a dynamic arrow.
```
emilyrohrbough: Add link to terra-hookshot

packages/terra-popup/src/Popup.jsx
```diff
@@ -40,17 +56,19 @@ const propTypes = {
    */
   classNameOverlay: PropTypes.string,
   /**
-   * Attachment point for the popup, this will be mirrored to the target.
+   * Attachment point for the popup, this will be mirrored to the target. Options include: 'top left',
+   * 'top center', 'top right', 'middle left', 'middle center', 'middle right', 'bottom left',
+   * 'bottom center', or 'bottom right'.
    */
-  contentAttachment: PropTypes.oneOf(TetherComponent.attachmentPositions),
+  contentAttachment: PropTypes.oneOf(ATTACHMENT_POSITIONS),
```
emilyrohrbough: Should we make this a shape to match hookshot's prop?

```
  contentAttachment: PropTypes.shape({
    horizontal: PropTypes.oneOf(Hookshot.HORIZONTAL_ATTACHMENTS),
    vertical: PropTypes.oneOf(Hookshot.VERTICAL_ATTACHMENTS),
  }),
```

dkasper-was-taken: It wouldn't be bad, but I know we were hoping to keep the API passive. If we plan on doing a 2.0 and are fine with breaking passivity we could.

dkasper-was-taken: After discussion for the time being we'll leave it as a minor version bump.

packages/terra-popup/src/Popup.jsx
```diff
@@ -84,64 +112,92 @@ const defaultProps = {
   contentHeight: '80',
   contentWidth: '240',
```
StephenEsser: This might be non-passive so not a change we can make. But I would expect the default behavior of these to be auto / dynamic

packages/terra-popup/src/Popup.jsx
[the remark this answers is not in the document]
StephenEsser: Should these to follow the opts pattern?

https://github.com/cerner/terra-core/blob/master/packages/terra-alert/src/Alert.jsx#L205-L206
https://github.com/cerner/terra-core/blob/master/packages/terra-signature/src/Signature.jsx#L204-L205

packages/terra-popup/src/Popup.jsx
```diff
@@ -40,24 +56,30 @@ const propTypes = {
...
   /**
    * A string representation of the height in px, limited to:
-   * 40, 80, 120, 160, 240, 320, 400, 480, 560, 640, 720, 800, 880
+   * 40, 80, 120, 160, 240, 320, 400, 480, 560, 640, 720, 800, 880 or dynamic.
    */
   contentHeight: PropTypes.oneOf(Object.keys(PopupHeights)),
   /**
    * A string representation of the width in px, limited to:
-   * 160, 240, 320, 640, 960, 1280, 1760
+   * 160, 240, 320, 640, 960, 1280, 1760 or dynamic.
```
StephenEsser: What are your thoughts on using "auto" in place of dynamic? It feels like it matches closer to the expected css behavior of height / width.

packages/terra-popup/src/Popup.jsx
dkasper-was-taken: It would be non-passive, and in general I don't know if we want to encourage the consumption of the least perfomant option.

packages/terra-popup/src/Popup.jsx
dkasper-was-taken: Any other votes on this one?

kschuste: I like auto

zhongyn: auto +1

packages/terra-popup/src/Popup.jsx
```diff
@@ -1,17 +1,33 @@
...
 import Portal from 'react-portal';
 import PopupContent from './_PopupContent';
 import PopupArrow from './_PopupArrow';
 import PopupOverlay from './_PopupOverlay';
-import TetherComponent from './_TetherComponent';
 import PopupUtils from './_PopupUtils';
-import PopupHeights from './_PopupHeights';
-import PopupWidths from './_PopupWidths';
-import './Popup.scss';
+import PopupHeights from './PopupHeights';
```
bjankord: What's the logic on removing the underscore prefix from the file name? Should we do that for the other files with underscore prefixes that imported above?

dkasper-was-taken: For better consumption by more HOC, example Menu, we are exposing the height and width files as exports, and as such are removing the private denoting underscores.

dkasper-was-taken: Utilities are still remaining private.

bjankord: Ok that makes sense.

packages/terra-menu/tests/nightwatch/menu/BoundedMenu.jsx
```diff
@@ -13,6 +13,10 @@ class BoundedMenu extends React.Component {
     this.state = { open: true };
   }
 
+  componentDidMount() {
+    this.forceUpdate();
```
bjankord: What is forceUpdate used for in the componentDidMount lifecycles in nightwatch examples?

dkasper-was-taken: That is the only way to to launch a popup on page load, as there will be no targetRef available before the content has been rendered once. So you have to force a second render after the component mounts. This previously worked in the tests, because tether would place the component on screen, even though it wouldn't position it, and the menu tests were only testing menu functionality. Hookshot will leave the component hidden if there is no targetRef.

dkasper-was-taken: That could be updated in the menu PR following this to use a click to launch the menu, but an early pass at this made me realize it was too large in scope for this PR.

bjankord: Ok I'm cool with updating in the menu PR.

dkasper-was-taken:
Removing UX label, as no visual changes were implemented.

mmalaker:
Functional verification completed. Hookshot examples and popups display and function as expected.

SInce Hookshot functionality is specific to popups and menus, let's add some explanatory text and links back to those components.

PR #916, opened 2017-10-12
Update I18n plugin to generate ES6 modules. Fixes #914

### LINKED ISSUE #914: Terra i18n generated translation files cause script errors
# Issue Description
The js files generated in the aggregated-translations folder use a mix of ES6 'import' statements and module.exports. 

## Issue Type
- [x] Bug

## Expected Behavior
the plugin should generate code that uses es6 imports with es6 exports. The code should be generated as follows:
```javascript
import { addLocaleData } from 'react-intl';
import localeData from 'react-intl/locale-data/en';

addLocaleData(localeData);

// aggregated messages ...

let i18nData = {
  areTranslationsLoaded: true,
  locale: 'en',
  messages,
};

export default i18nData;
```

## Current Behavior
ES6 imports are used with module.exports in the generated code. This may have been acceptable on older versions of webpack. It is currently causing a script error. The generated code looks as follows:
```javascript
import { addLocaleData } from 'react-intl';
import localeData from 'react-intl/locale-data/en';

addLocaleData(localeData);

// messages

module.exports = {
  areTranslationsLoaded: true,
  locale: 'en-GB',
  messages,
};
```

## Steps to Reproduce

1. install terra i18n plugin
2. Use the plugin to generate aggregated-translations
3. Examine the generated files.
4. 

## Environment

* Component Version: "terra-i18n": "^1.6.0",
    "terra-i18n-plugin": "^1.2.0",
    "webpack": "3.6.0",

* Browser Name and  @Version: Chrome Version 61.0.3163.100 (Official Build) (32-bit)
* Operating System and version (desktop or mobile): Desktop Windows 8.1

mhemesath:
### Summary
Updates i18n plugin to generate pure ES6 modules. This is required for anyone wanting to use webpack to optimizations such as tree shaking that has turned module transpiling off on babel.

### Additional Details
This should address https://github.com/cerner/terra-core/issues/914

### DIFF
--- a/packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
+++ b/packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
@@ -10,11 +10,12 @@ import localeData from 'react-intl/locale-data/${language.split('-')[0]}';
 addLocaleData(localeData);
 
 const messages = ${JSON.stringify(messages, null, 2)};
-
-module.exports = {
-  areTranslationsLoaded: true,
-  locale: '${language}',
-  messages: messages
+const areTranslationsLoaded = true;
+const locale = '${language}'
+export {
+  areTranslationsLoaded,
+  locale,
+  messages
 };`;
 }
 
--- a/packages/terra-i18n-plugin/CHANGELOG.md
+++ b/packages/terra-i18n-plugin/CHANGELOG.md
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Generated translations are now ES6 modules.
 
 1.6.0 - (October 6, 2017)
 ------------------
--- a/packages/terra-modal/docs/DEPENDENCIES.md
+++ b/packages/terra-modal/docs/DEPENDENCIES.md
@@ -4,7 +4,7 @@
 | Dependency | Version | React Version | Description |
 |-|-|-|-|
 | classnames | ^2.2.5 | -- | A simple utility for conditionally joining classNames together |
-| focus-trap-react | ^3.0.2 | 0.14.x \|\| ^15.0.0 \|\| ^16.0.0 | A React component that traps focus. |
+| focus-trap-react | ^3.0.2 | 0.14.x \|\| ^15.0.0 | A React component that traps focus. |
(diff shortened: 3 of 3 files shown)


packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -10,11 +10,12 @@ import localeData from 'react-intl/locale-data/${language.split('-')[0]}';
...
-module.exports = {
-  areTranslationsLoaded: true,
-  locale: '${language}',
-  messages: messages
+const areTranslationsLoaded = true;
+const locale = '${language}'
+export {
+  areTranslationsLoaded,
+  locale,
+  messages
```
StephenEsser: This may have been fixed with some of the other changes but is this shorthand syntax still causing the same errors mentioned here? 

https://github.com/cerner/terra-core/issues/878

mhemesath: Its worth testing and I think this will work. This isn't object shorthand anymore... its ES6 export syntax. What you referenced is object shorthand in an object to`module.exports`

StephenEsser: Ah, I missed that. That makes sense.

kushanson: After the code genrates, I can modify it and test in browsers. Here is the code that worked in chrome and IE10:
```javascript
export default {
  areTranslationsLoaded: true,
  locale: 'en',
  messages: messages
};
```
IE 10 did not like the short hand on the 'messages' property. Both browsers did not like it when I left out the 'default' keyword.

kushanson: I'll retract my previous comment. I had a copy/paste error. The code works good. looks like there is missing semi-colon for the declaration of 'locale'.

bjankord: Tested in this change in the Healthet@Cerner site which is using create-react-app and the build compiled successfully. So no regression on #878.

bjankord:
@mhemesath Do you wanna send out another PR to add it after locale.

PR #917, opened 2017-10-12
Updated webpack config

### LINKED ISSUE #921: Update webpack config to use v3 recommend syntax
# Issue Description
With webpack v3, its documneted in examples to use rules instead of loaders
We can also remove JSON loader as its now included in webpack v3.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
webpack config is converted to v3 syntax for loaders

## Current Behavior
Webpack config uses v2 syntax for loaders

mhemesath:
Resolves #921 
* Use rules instead of loaders
* Remove JSON loader as its now included in webpack


bjankord:
For functional verification on this, spot checking is needed. We should also look at translation examples in IE 10 / IE 11 to make sure changes in #916 work as well.

emilyrohrbough:
Should we remove `json-loader` from the listed dev-dependencies?

mmalaker:
Functional verification completed. Site examples display and function correctly after Webpack config update.

Issue #699, opened 2017-08-03
Arrange does not properly expand it's fit end and start aligned elements.


dkasper-was-taken:
# Issue Description
Arrange fit elements lack "flex: 0 0 auto" styling, so end elements do not properly expand with various content.

This works correctly on IE 10, but does not work on Firefox, Chrome, and Safari.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

![screen shot 2017-08-03 at 1 06 03 pm](https://user-images.githubusercontent.com/22619500/28936776-5e6ae51a-784e-11e7-969e-7239455bd4fa.png)

cwalten:
We are having similar issues, only in IE, where items are pushed off of the viewable window if the elements in 'fill' get too wide.

<img width="559" alt="ie arrange not fitting on screen" src="https://user-images.githubusercontent.com/19825717/30223366-346771e4-9490-11e7-8be5-b43b66301155.png">

emilyrohrbough:
Resolved in #926.

PR #926, opened 2017-10-17
Arrange expand content

kschuste:
### Summary
This fixes issue #699.  This applies flex-box styling to the arrange component to prevent content from being pushed off-screen when the containers are too large to fit on the screen.
https://github.com/cerner/terra-clinical/pull/203/ 

Have added props to allow for the consumer to pass in their own attributes for the fit and fill containers

Deployment link: https://terra-core-deployed-pr-926.herokuapp.com/static/#/site/arrange


packages/terra-arrange/src/Arrange.scss
```diff
@@ -7,21 +7,22 @@
...
+    flex: 1 1 auto;
     min-width: 0;
   }
 
   // Styles specific to align items independently of each other
   .fit {
     align-self: flex-start;
-    display: flex;
+    flex: 0 0 auto;
+    max-width: 40%;
```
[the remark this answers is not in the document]
kschuste: updated the summary of this PR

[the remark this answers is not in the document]
jmsv6d: This seems pretty opinionated for such a generic component. If a max width is really necessary, I feel like it should be customizable via a prop.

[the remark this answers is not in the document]
kschuste: without the max-width, then content can be pushed offscreen if are bigger than the screen size, which is why this issue was introduced for when have content that does not wrap.  Seems odd for the consumer to have to supply this when this container can handle this internally.  However, I will look at adding it as a prop in the meantime.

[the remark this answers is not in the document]
kschuste: Have the thought of defaulting the prop to be 40% if not given, else use the prop to set the max-width?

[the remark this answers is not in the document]
jmsv6d: I think that would be fine.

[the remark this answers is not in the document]
bjankord: I think if we go with a max-width prop, we should have 3, one for the start, middle, and end containers.

packages/terra-arrange/src/Arrange.jsx
```diff
@@ -47,21 +55,25 @@ const Arrange = ({
...
   alignFill,
   alignFitEnd,
+  fitStartMaxWidth,
+  fitEndMaxWidth,
   ...customProps
   }) => {
   if (fitStart === undefined && fitEnd === undefined) {
     throw new Error('At least one of the props: [fitStart, fitEnd] should be supplied.');
   }
+  const fitStartProps = (fitStartMaxWidth !== undefined) ? { maxWidth: fitStartMaxWidth } : { maxWidth: '40%' };
```
jmsv6d: Instead of doing this you can use default props to set the default for these to '40%'

kschuste: +1, forgot about the defaultProps, updated now

packages/terra-arrange/src/Arrange.jsx
```diff
@@ -36,8 +36,20 @@ const propTypes = {
...
+   */
+  fitStartMaxWidth: PropTypes.string,
+  /**
+   * The max-width to be set on the fitEnd element
+   */
+  fitEndMaxWidth: PropTypes.string,
 };
 
+const defaultProps = {
+  fitStartMaxWidth: '40%',
```
mhemesath: the default should be no max width

mhemesath:
I'M wondering if we trying to make arrange do too much and what we actually need is a flex component that allows for this level of customization.... Arrange was really only meant for simple media alignment in its inception

packages/terra-arrange/src/Arrange.scss
```diff
@@ -7,21 +7,21 @@
   // Vertically aligns arrange content to the top by default
   .arrange {
     display: flex;
+    word-wrap: break-word;
```
jmsv6d: This feels too opinionated for this component as well. Shouldn't the content elements be responsible for how they wrap their content?

zhongyn: The content should handle its own style. The arrange is all about structure.

kschuste: I added this for when there is large text in the end containers, specifically as in one of the examples on the issue where the text is one really really long word.  With the styling of flex: 0 0 auto, the second 0 is say to not shrink on the end caps; therefore if have text that is much longer, it won't wrap when hit the width and would just overflow into the fill element.

I attempted to add it to the content I gave to the arrange component; however, it would only work at this level; hence I added it here as putting it on the content given didn't solve the issue.

If I set it to use flex: 0 1 auto for the start/end containers, this will handle the wrapping as now the content knows it has to shrink; however, this causes issues with other use cases as then the fill content (middle) will squish down the end containers if the middle content is too large.

packages/terra-arrange/src/Arrange.jsx
kschuste: Without defaulting a max-width; then the start and end containers will not shrink and wrap and will overflow into the fill (center) container and off the screen.  I can set this the '' and then set the start and end containers to flex: 0 1 auto; however, that will cause other issues with other use cases as then the fill content (middle) will squish down the end containers if the middle content is too large.

mhemesath: Why can’t the fit container content set their own max width?

jmsv6d: That doesn't work if consumers want the max width to be a percentage.

jmsv6d: I am wondering now, though, if it would be better to just have `fitStartAttributes` and `fitEndAttributes` props instead for these types on needs.

kschuste: I agree with @jmsv6d applying the maxWidth and wrapping on the content given doesn't exactly word.  The content can calculate the width of the fit container before applying any text wrapping.
Using firstStart and fitEnd attributes should solve the needs

mhemesath: At this point... is it better to just let consumers render their own arrange structure...
```jsx
<Arrange>
  <Arrange.Fit style={{ maxWidth: "40%" }}/>
  <Arrange.Fill/>
</Arrange>
```

mhemesath: Also relating to percentage... when would you want that? Generally the fit container is a fixed width and the fill is a width that grows. That means that you  wouldn't want 40%  as that changes with the component parents available width. At this point its sounding more like you want a structural grid than an arrange.

jmsv6d: I would be okay with that.

kschuste:
Updated this to the following:  the arrange component to use flex layout with the startFit and endFit elements to not shrink to display their contents and the fill element that is able to shrink to fill the remaining space available.  I have added props for each element to allow consumers to apply their own styling for cases where the widths may need to be different to allow text to wrap appropriately.

packages/terra-arrange/src/Arrange.jsx
```diff
@@ -55,13 +69,13 @@ const Arrange = ({
 
   return (
     <div {...customProps} className={cx('arrange', customProps.className)} >
-      <div className={cx('fit', align || alignFitStart)}>
+      <div {...fitStartAttributes} className={cx('fit', align || alignFitStart)}>
```
jmsv6d: You should probably add `fitStartAttributes.className` to  the className here. Same for the others.

packages/terra-arrange/src/Arrange.scss
```diff
@@ -12,16 +12,15 @@
   // Makes sure fill expands to fill the remaining space of its parent container
   // Display set to ensure correct appreance if applied to inline elements like span
```
bjankord: This comment about display can be removed.

packages/terra-arrange/tests/nightwatch/ArrangeTests.jsx
```diff
@@ -12,6 +12,7 @@ const BaseTests = () => (
       <li><Link to="/tests/arrange-tests/aligned-individually-fit-start">Individually Aligned FitStart</Link></li>
       <li><Link to="/tests/arrange-tests/aligned-individually-fill">Individually Aligned Fill</Link></li>
       <li><Link to="/tests/arrange-tests/aligned-individually-fit-end">Individually Aligned FitEnd</Link></li>
+      <li><Link to="/tests/arrange-tests/aligned-all-arrange-long-word">Aligned Array with LongWords</Link></li>
```
bjankord: The `to` prop in the Link component should match the Route component's `path` prop. Need to add an `s` after `long-word`

packages/terra-arrange/tests/nightwatch/AlignedAllLongWords.jsx
```diff
@@ -0,0 +1,61 @@
...
+    </ArrangeWrapper>
+
+    <ArrangeWrapper >
+      <Arrange
+        id="stretch"
+        align="stretch"
+        fitStart={longWordText}
+        fill={longWordText}
+        fitEnd={longWordText}
+        fitStartAttributes={{ className: 'test1', style: { maxWidth: '10px', wordWrap: 'break-word' } }}
```
bjankord: When I initially looked at this test page, the examples looked broken, took me a bit to make sense of what I was looking at. 
screen shot 2017-10-20 at 6 03 11 am

I realized you went with smaller max-widths to likely test the word-wrap behavior. I'd add in some longer words, and increase the max-width on these examples to more realistic widths. The words can be whatever crazy long combinations to test wrapping.

packages/terra-arrange/src/Arrange.scss
kschuste: removed the comment

packages/terra-arrange/src/Arrange.scss
```diff
@@ -10,18 +10,16 @@
...
+    flex: 1 1 auto;
     min-width: 0;
   }
 
   // Styles specific to align items independently of each other
   .fit {
     align-self: flex-start;
-    display: flex;
+    flex: 0 0 auto;
+    overflow: auto;
```
dkasper-was-taken: Is there a reason we are using overflow:auto here? Are we expecting content to scroll if it's beyond it's max width?

kschuste: I put it in here for when say you have content (say images) that is too wide in the start and end containers.  This would allow them to shrink down if given a max-width rather than push the end-content off the screen.  However, now since we are passing in the props for the fit containers, this isn't really needed anymore.

mmalaker:
Looks good except in IE, the [LongWords](https://terra-core-deployed-pr-926.herokuapp.com/static/#/tests/arrange-tests/aligned-all-arrange-long-words) example breaks the arrange and text overflows into adjacent areas.

kschuste:
@mmalaker updated the text to work across browsers, it was to make sure word-wrap styling happens even for when there is no spacing in text

kschuste:
@mmalaker added additional site example using the new props and setting smaller widths on the start and end containers.
screen shot 2017-10-25 at 11 57 46 am

mmalaker:
Functional verification completed. The arrange examples display and function as expected.

No need to change in this PR, but I logged issue #968 to simplify and improve the arrange example page.

Issue #832, opened 2017-09-19
Signature: Right clicking Signature Component causes "sticky" draw


StephenEsser:
# Issue Description

Right clicking inside of a Signature Component can cause a "sticky" effect that continuously draws line segments without the mouse being pressed and requires an additional click to deactivate.

![stickysignature](https://user-images.githubusercontent.com/6720991/30609208-5c92d010-9d40-11e7-9f2e-d77c941b791c.gif)

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior

Right click should not trigger drawing

## Current Behavior

Right click causes sticky drawing

## Steps to Reproduce

1. Navigate to http://engineering.cerner.com/terra-core/#/site/signature
2. Right click and release inside of the Signature component and continue moving the mouse
3. Observe the lines will draw as the mouse moves

## Environment

* Component Version:
* Browser Name and Version:
* Operating System and version (desktop or mobile):

dkasper-was-taken:
This can also be reproduced using the left click, and leaving the confines of the browsers. I believe we'd need to investigate a mouseout event in addition to the current mouseup.

PR #929, opened 2017-10-17
Fix Signature Sticky Draw

yuderekyu:
### Summary
Addresses #832. 

#### Changes
- Only left-click drawing is allowed when using a mouse.
- Drawing is disabled if the pointer leaves the confines of the browser.
- Removed typos in readme.


packages/terra-signature/README.md
```diff
@@ -5,7 +5,7 @@
 [![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
 
 The terra-signature component provides users a way to define a signature with a series of line segments.
-The component can be define the line width, line color and optionally display a signature if the line segments have 
+The component can define line width, line color and optionally display a signature if the line segments have
```
kschuste: nit: need to a comma after line color as well

packages/terra-signature/docs/README.md
```diff
@@ -1,7 +1,7 @@
 # Terra Signature
 
 The terra-signature component provides users a way to define a signature with a series of line segments.
-The component can be define the line width, line color and optionally display a signature if the line segments have 
+The component can define line width, line color and optionally display a signature if the line segments have
```
kschuste: nit, add comma after 'line color'

packages/terra-signature/README.md
yuderekyu: https://www.youtube.com/watch?v=P_i1xk07o4g

packages/terra-signature/src/Signature.jsx
```diff
@@ -3,18 +3,23 @@ import PropTypes from 'prop-types';
...
+
+const MOUSECODES = {
+  LEFTCLICK: 0,
+};
+
 const propTypes = {
   /**
    * The line width to use when drawing the signature on the canvas.
-   * One of LineWidths.EXTRA_FINE, LineWidths.FINE, LineWidths.MEDIUM, LineWidths.HEAVY.
+   * One of LINEWIDTHS.EXTRA_FINE, LINEWIDTHS.FINE, LINEWIDTHS.MEDIUM, LINEWIDTHS.HEAVY.
```
jmsv6d: This comment isn't actually accurate, consumers would be using Signature.Opts.Width.EXTRA_FINE etc.

packages/terra-signature/src/Signature.jsx
```diff
@@ -122,6 +137,15 @@ class Signature extends React.Component {
...
   }
 
+  mouseOut(event) {
+    if (this.state.painting) {
+      this.setState({ painting: false });
+      if (this.props.onChange) {
+        this.props.onChange(event, this.state.lineSegments, this.state.lineSegments[this.state.lineSegments.length - 1]);
+      }
+    }
+  }
```
StephenEsser: Canceling the signature on mouse out adds a more crucial role of precision to the user and frequently cancels signatures midway. It would be worth exploring alternatives and getting opinions on how to handle this.

signature

emilyrohrbough:
We should add a test case or two to verify this fix works as expected.

packages/terra-signature/src/Signature.jsx
yuderekyu: Right, I talked to @dkasper-was-taken about this yesterday - the other alternative would be to add a timer, similar in hookshoot, where the signature can continue within a certain timeframe.

yuderekyu: Actually, this is a simpler solution - utilize `mouseleave` instead of `mouseout` ([differences](https://developer.mozilla.org/en-US/docs/Web/Events/mouseleave)) and attach that to the document.body. We can take advantage of `mouseleave` only being triggered by leaving the document window, so a signature can still be drawn when exiting the canvas element.

packages/terra-signature/src/Signature.jsx
```diff
@@ -3,18 +3,23 @@ import PropTypes from 'prop-types';
...
+
+const MOUSECODES = {
+  LEFTCLICK: 0,
+};
+
 const propTypes = {
   /**
    * The line width to use when drawing the signature on the canvas.
-   * One of LineWidths.EXTRA_FINE, LineWidths.FINE, LineWidths.MEDIUM, LineWidths.HEAVY.
+   * One of Signature.Opts.Width.EXTRA_FINE, Signature.Opts.Width.FINE, Signature.Opts.Width.MEDIUM, Signature.Opts.Width.HEAVY.
```
kschuste: reason why the comment here is not using the LINEWIDTHS const that the code is using?  should the code be using Signature.Opts?

emilyrohrbough: Singature.Opts are the options obj we export to the consumer. [See here](https://github.com/cerner/terra-core/pull/929/files/cf952e026eef88ead8a22f24cd40683bb8775b11#diff-439e726998c8c3153688d80b4864792dR229)

packages/terra-signature/src/Signature.jsx
```diff
@@ -3,18 +3,23 @@ import PropTypes from 'prop-types';
...
 
-const LineWidths = {
+const LINEWIDTHS = {
   EXTRA_FINE: 1,
   FINE: 2,
   MEDIUM: 4,
   HEAVY: 6,
 };
+
+const MOUSECODES = {
```
StephenEsser: Should we underscore Line widths and Mouse codes? 

```
LINE_WIDTHS
MOUSE_CODES
```

packages/terra-signature/src/Signature.jsx
```diff
@@ -86,17 +94,24 @@ class Signature extends React.Component {
...
+      window.removeEventListener('mouseleave', this.mouseLeave);
       window.removeEventListener('mouseup', this.mouseUp);
     }
 
     window.removeEventListener('resize', this.updateDimensions);
   }
 
   mouseDown(event) {
+    if ('button' in event) {
+      if (event.button !== MOUSECODES.LEFTCLICK) {
```
StephenEsser: These can probably be flattened to reduce nesting.
```
if ('button' in event && event.button !== MOUSECODES.LEFTCLICK)
```

packages/terra-signature/src/Signature.jsx
```diff
@@ -86,17 +94,24 @@ class Signature extends React.Component {
...
+        return;
+      }
+    }
```
StephenEsser: These can probably be flattened to reduce nesting.
```
if ('button' in event && event.button !== MOUSECODES.LEFTCLICK)
```

packages/terra-signature/src/Signature.jsx
yuderekyu: I was going with consistency here with constants in other packages like hookshot, which don't use underscores.

packages/terra-signature/src/Signature.jsx
```diff
@@ -86,17 +94,22 @@ class Signature extends React.Component {
...
+      window.removeEventListener('mouseleave', this.mouseLeave);
       window.removeEventListener('mouseup', this.mouseUp);
     }
 
     window.removeEventListener('resize', this.updateDimensions);
   }
 
   mouseDown(event) {
+    if ('button' in event && event.button !== MOUSECODES.LEFTCLICK) {
+      return;
```
StephenEsser: This might disable touch devices, we should double check

packages/terra-signature/src/Signature.jsx
bjankord: I'm cool with no underscore all caps.

mjhenkes:
@yuderekyu We need functional verification on this issue before we merge to master. I've reverted your commit.

yuderekyu:
nooooooo

yuderekyu:
nooooooo_luke_skywalker

emilyrohrbough:
Do we need a new PR for this? Not seeing it come up

PR #934, opened 2017-10-18
Added webdriver.io POC

mhemesath:
### Summary
This is a proof-of-concept pull request using webdriver.io.  Webdriver.io comes with [built in support for regression testing](http://webdriver.io/guide/services/visual-regression.html) which eliminates a lot of custom code we need to write to achieve this.

My recommendation is that we use webdriver.io for visual regression testing and slowly phase out nightwatch. Teams can use nightwatch if they like, but UI tests should be done with webdriver.io

benefits
* Tests can be written using our babel configuration and mocha
* Includes service to auto-start selenium
* Includes service to auto-start static file server and webpack dev server
* First class integartion support with saucelabs and browsermob
* Comes with a variety of test reporters
* Tests are synchronous by default (with async option)
* A lot more granularity around before/after hooks for tests see [config](http://webdriver.io/guide/testrunner/configurationfile.html)
* [Command reference](http://webdriver.io/api.html) is very clear and extensive
* [Explicit viewport sizing](http://webdriver.io/api/window/setViewportSize.html) command which we had issues with in past 
* Multiple core contributors with activity from 3years -> 8hrs ago
* We can phase in wdio while we phase out nightwatch


packages/terra-badge/tests/browser/badge-spec.js
```diff
@@ -0,0 +1,32 @@
+/* global browser, it, viewport */
+describe('badge', () => {
+  const viewports = viewport('tiny', 'huge');
+
+  it('defaults', () => {
+    browser.url('/#/tests/badge-tests/default');
+    browser
```
mhemesath: I think we can consolidate this all in a plugin so its no so verbose.

mhemesath: EU did a good job of creating a command to do this in their repo https://github.com/ec-europa/ecl-toolkit/blob/master/packages/ecl-qa/wdio/assertions/matchReference.js

PR #936, opened 2017-10-18
Fix Build: Stop checking against the computed value for font weight.

mjhenkes:
### Summary
For an unknown as of now reason, normal and bold font weights are not returning the computed value but the default value (ie 700 instead of bold). I'm assuming this is a change in the chrome driver we've pulled in, but I'm not sure yet.

@cerner/terra

### DIFF
--- a/packages/terra-heading/tests/nightwatch/heading-spec.js
+++ b/packages/terra-heading/tests/nightwatch/heading-spec.js
@@ -42,8 +42,8 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
 
   'Displays a heading component with a set weight': (browser) => {
     browser.url(`${browser.launchUrl}/#/tests/heading-tests/weight`);
-    browser.assert.cssProperty('#heading-weight-700', 'font-weight', '700'); // Browser computes 700 as bold
-    browser.assert.cssProperty('#heading-weight-400', 'font-weight', '400'); // Browser computes 400 as normal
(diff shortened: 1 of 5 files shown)


emilyrohrbough:
Should probably update the Heading tests as well: https://github.com/cerner/terra-core/blob/master/packages/terra-heading/tests/nightwatch/heading-spec.js#L45

emilyrohrbough:
I just checked my browser, I think this is may be an update with Chrome itself. I know we recently had [this PR](https://github.com/cerner/terra-core/pull/858) that resulted in failing builds due to browser specific css compilation.

packages/terra-heading/tests/nightwatch/heading-spec.js
```diff
@@ -42,8 +42,8 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
 
   'Displays a heading component with a set weight': (browser) => {
     browser.url(`${browser.launchUrl}/#/tests/heading-tests/weight`);
-    browser.assert.cssProperty('#heading-weight-700', 'font-weight', 'bold'); // Browser computes 700 as bold
-    browser.assert.cssProperty('#heading-weight-400', 'font-weight', 'normal'); // Browser computes 400 as normal
+    browser.assert.cssProperty('#heading-weight-700', 'font-weight', '700'); // Browser computes 700 as bold
```
jmsv6d: These inline comments aren't really necessary anymore with these changes.

Issue #927, opened 2017-10-17
Allow the embedded content consumer to communicate with its iframe content


rasthana:
# Issue Description
The EmbeddedContentConsumer component allows embedding iframe content in order to present non-react content within react applications. The content in these embedded frames use XFC's messaging system to communicate with the application. Currently, the component provides a mechanism for the content (Provider) to communicate with it containing frame. However, there is no mechanism in place for the container to communicate with the content.

I would like to enhance the embedded content consumer to provide a standard to allow the parents of the EmbeddedContentConsumer to communicate with its content.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The EmbeddedContentConsumer wraps the XFC frame and exposes communication with the provider through function props. Theoretically, if the consumer provides access to the XFC frame then its parents will be able to trigger handlers defined by their providers.

The XFC frame, created after the Consumer has been successfully mounted, can be exposed to the parent component an onMount function prop.
- `onMount` - Notifies the component that the container has been mounted. Provides a frame to trigger events on the provider.

Another approach would expose a function on the EmbeddedContentConsumer that allows triggering events on the provider. The parent would obtain a reference to the EmbeddedContentConsumer and invoke this "trigger" function.

**Note**
The EmbeddedContentConsumer currently notifies the container via the onLaunch and onAuthorize hooks. Providing a hook to monitor the "Mount" status would complete all the stages of an application life cycle provided by XFC.

## Current Behavior
There is no standard that allows the container to communicate with its content.

rasthana:
# Technical Design
Allow EmbeddedContentConsumer to communicate with its content.

## Possible Solutions
- [ ] EmbeddedContentConsumer provides an onMount hook and provides the xfc frame to allow its parents to invoke methods available on the frame.
`onMount(this.xfcFrame)`
- [ ] EmbeddedContentConsumer provides an onMount hook and provides itself to allow its parents to invoke instance methods.
`onMount(this)`
- [ ] Parents of EmbeddedContentConsumer declare a ref and invoke instance methods.
`
<EmbeddedContentConsumer
    ref={(element) => { this.embeddedContentConsumer = element; }}
/>
`

emilyrohrbough:
Resolved in #938.

PR #938, opened 2017-10-18
Allow EmbeddedContentConsumer to trigger events on its application

rasthana:
### Summary
Allow the parent components of EmbeddedContentConsumer to communicate with its application. EmbeddedContentConsumer exposes the onMount function prop that is invoked after the Consumer frame has been mounted, if provided. The onMount function passes a reference to instance of the EmbeddedContentConsumer to allow the parent to trigger an event on the application.

Addresses: #927


packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -35,6 +40,11 @@ class EmbeddedContentConsumer extends React.Component {
     // Mount the provided source as the application into the content wrapper.
     this.xfcFrame = Consumer.mount(this.embeddedContentWrapper, this.props.src, this.props.options);
 
+    // Notify that the consumer frame has mounted.
+    if (this.props.onMount) {
+      this.props.onMount(this);
```
mhemesath: I'd recommend just passing back the xfcFrame

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
```diff
@@ -53,9 +63,21 @@ class EmbeddedContentConsumer extends React.Component {
     (customEvents || []).forEach(e => this.addEventListener(e.key, e.handler));
   }
 
+  /**
+  * Triggers an event in the application. Assumes this consumer has been mounted.
+  * @param {string} event - The event name to trigger.
+  * @param {object} detail - The data context to send with the event.
+  */
+  trigger(event, detail) {
```
mhemesath: if you pass back the xfc frame this isn't needed

packages/terra-embedded-content-consumer/tests/nightwatch/providers/MessagingProvider.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+class MessagingProvider extends React.Component {
+  constructor(props) {
+    super(props);
+    Provider.init({
+      acls: ['*'],
+      secret: 'OAuth Secret',
+      onReady: () => {
+        Provider.on('SHOW_STATUSES', () => {
+          const lifeCycleStatuses = ['Mounted', 'Launched', 'Authorized'];
```
mhemesath: can we simplify this example... really all you need to do is acknowledge the event was received

packages/terra-embedded-content-consumer/tests/nightwatch/consumers/MessagingConsumer.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
```
mhemesath: This really should just be a jest test to ensure XFC is passed back on mount. We have thorough tests around XFC messaging in that project.

packages/terra-embedded-content-consumer/tests/nightwatch/providers/MessagingProvider.jsx
```diff
@@ -0,0 +1,46 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
mhemesath: This is testing XFC... which we have thorough tests for. I recommend removing this test file

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
rasthana: @mhemesath can you provide your reasoning for this? I understand that the xfcFrame would be indirectly exposed via the embeddedContentConsumer Instance. However, the component already provides first class support to handle events from the provider. It felt odd to expose the xfc frame and not abstract it away.

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
rasthana: Agreed this would not be needed if the xfcFrame is passed when mounted. Follow the [above discussion](https://github.com/cerner/terra-core/pull/938#discussion_r145528274).

packages/terra-embedded-content-consumer/src/EmbeddedContentConsumer.jsx
mhemesath: IMO the lifecycle events are first-class events that many consumers of XFC want to know about and the props are convenience methods to set them.  If someone needs access to custom events IMO it makes sense to just hand over XFC as its cleaner and allows that object to be passed up to a higher level so that it can be more directly integrated into the state manager (rather than be confined at the component level).

mhemesath:
You may consider updating XFC as well... it no longer requires a polyfill

rasthana:
> You may consider updating XFC as well... it no longer requires a polyfill
XFC is a peer dependency and bumping EmbeddedContentConsumer to require XFC ^1.1.0 probably constitutes a major version bump. @mhemesath If you feel strongly about this then we can discuss this further.

PR #947, opened 2017-10-19
Add space between icon and text in a Button

doogthemushroom:
### Summary
Adds a small amount of space between an icon and the text in a Button. Was off putting to me that they were sitting right next to each other.

### Additional Details
I used 0.714em for the space to match the padding on the side of the buttons, but could see a smaller size being appropriate as well.


packages/terra-button/src/Button.scss
```diff
@@ -117,4 +117,12 @@
...
     padding-top: var(--terra-button-padding-top-compact, 0);
   }
+
+  .text-with-icon {
+    padding-left: 0.714em;
+  }
+
+  .text-with-icon-reverse {
+    padding-right: 0.714em;
+  }
```
StephenEsser: I think these should be margins to push space between content

mhemesath: IMO we should have a component to manage this. Badges do basically the exact same thing.

bjankord: I'm working the OCS clinical/consumer teams to create a component for this. The consumer UX team started design work on this today.

packages/terra-button/src/Button.jsx
```diff
@@ -80,7 +80,10 @@ const Button = ({
   ...customProps
   }) => {
   const attributes = Object.assign({}, customProps);
-  const buttonText = text ? <span className={styles.text}>{text}</span> : null;
+
+  const iconWithTextStyle = isReversed ? cx(['text-with-icon-reverse']) : cx(['text-with-icon']);
+  const textStyle = icon ? iconWithTextStyle : null;
+  const buttonText = text ? <span className={textStyle} >{text}</span> : null;
```
JakeLaCombe: Extra space before >

packages/terra-button/src/Button.jsx
```diff
@@ -80,7 +80,10 @@ const Button = ({
   ...customProps
   }) => {
   const attributes = Object.assign({}, customProps);
-  const buttonText = text ? <span className={styles.text}>{text}</span> : null;
+
+  const iconWithTextStyle = isReversed ? cx(['text-with-icon-reverse']) : cx(['text-with-icon']);
```
JakeLaCombe: I think would be less confusing if we renamed it to iconTextStyleClass

StephenEsser: Assuming the css for the Badge is robust we can just mimic the same css styling and keep this class as text and only make changes to Button.scss

https://github.com/cerner/terra-core/blob/master/packages/terra-badge/src/Badge.scss#L41-L47

bjankord:
New OCS button designs had indicated spacing between icon and text. I'll follow up with the UX team and figure out what the values should be.

bjankord:
Talked to UX and `0.714em` is the value they want for the spacing, nice work @doogthemushroom

mjhenkes:
Actual footage of @doogthemushroom choosing the default button icon spacing.

neilpfeiffer:
Tested: PR #947
 
**Reviewed:**
- Spacing between icon and button text looks good - ready to merge.

 
**Pass** |
-----------|
UX Review|

packages/terra-button/src/Button.jsx
doogthemushroom: Removed space

packages/terra-button/src/Button.scss
doogthemushroom: Changed to margins.

Looking forward to the component.

packages/terra-button/src/Button.jsx
doogthemushroom: Ah that is some solid css. I've brought it in and removed the changes to the jsx.

Issue #945, opened 2017-10-19
tests fail on windows because of file paths


Singleton06:
# Issue Description

There are a decent number of tests that seem to be performing checks that rely on path separators specific to the OS.  Those tests have the path with a forward slash, but when running on windows, it seems to have the error log happen with two backslashes.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
The tests would pass on windows

## Current Behavior

The tests fail on windows.

## Steps to Reproduce

1. Clone terra-core on a windows machine
2. run npm install at the root 
3. run npm start (@emilyrohrbough had indicated this solves the issue of jest failing due to missing traslations because of i18n aggregation that webpack does)
4. run npm test

## Additional Details
[Here](https://gist.github.com/Singleton06/ac5b214e934081cd325a2d22d7ad2f85) is a build that shows the issue.  Specifically, here is an example of the issue I'm seeing.

```javascript
  ? i18n-aggregator-plugin > when searching directories for translation messages with options outputFileSystem given > should fill the tranlsation files with the expected information for each locales

    expect(string).toContain(value)

    Expected string:
      "c:\\installs\\workspaces\\dustin-terra\\terra-core\\packages\\terra-i18n-plugin\\tests\\fixtures\\aggregated-translations\\en.js"
    To contain value:
      "aggregated-translations/en.js"

      at packages/terra-i18n-plugin/tests/i18n-plugin-spec.test.js:151:42
      at Array.forEach (native)
      at Object.<anonymous> (packages/terra-i18n-plugin/tests/i18n-plugin-spec.test.js:150:24)
          at Promise (<anonymous>)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:169:7)
```

## Environment

* Browser Name and Version: N/A
* Operating System and version (desktop or mobile): Windows 10

bjankord:
Talked with @Singleton06, we should update these to use `path.join` any time we are using paths. 
`path.join` will use `/` on unix and `\\` on windows.

For any `import` or `require` statements, we don't need to use `path.join`.

Singleton06:
Just to make sure no one might be duplicating effort, I've got the code finished to handle this. I'll try to get out a PR tomorrow

Singleton06:
This can be closed now.

Issue #839, opened 2017-09-20
Add *.idea* to .gitignore


dv297:
# Issue Description
JetBrains IDE's (RubyMine, WebStorm) create a hidden directory ".idea" whenever you open a directory as a project. These files are workspace specific and should be ignored.

Because this is a monorepo, opening the whole entire repo in the IDE usually just crashes the IDE on slower machines. So I've been opening just the particular package I'm working on in WebStorm, which means the .idea directory is in a subdirectory of the project. The expression "*.idea*" should solve this.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [X] Other

## Expected Behavior
* After opening the project in any JetBrains IDE, running "git status" should not show any .idea directory.
* After opening a subdirectory of the project in any JetBrains IDE, running "git status" should not show any .idea directory.

## Current Behavior

After opening the project in any JetBrains IDE, running "git status" currently shows the .idea directory.

## Steps to Reproduce
1. Open project in any JetBrains IDE
2. Run `git status`

StephenEsser:
I don't think it is inherently bad to blacklist the .idea file but this seems to be an artifact of the IDE and is a problem for every project that opens in that IDE. 

I'd recommend adding .idea to your local .gitignore_global file on your machine rather than every project individually.

RubyMine attempts to index the node_modules directory and simply isn't built to handle the size of the mono repos web assets.

The IDE switch can be difficult, but I'd recommend migrating over to Atom or Sublime if you can get a license. Atom is free, open source, and doesn't crash loading the project like RubyMine does.

dv297:
That's fine. Honestly, this is the first time I've heard of a global
gitignore, so thanks for that! I wouldn't be surprised if others didn't
know about a global one either, so the minimal cost of adding an entry to
the project's gitignore vs the overhead a dev has to know to not be annoyed
seems justified to me. But I understand that opting to not add an entry for
every IDE and editor can just be a project preference thing. I've just seen
projects opt otherwise.

I'm good with either addressing it or closing this with the global
gitignore as the preferred route.

StephenEsser:
I think there is some value in adding the change. The global ignore is pretty nifty though

PR #949, opened 2017-10-20
use path.join in place of hard coded paths

Singleton06:
### Summary
Tests were failing on windows because paths were being used for comparisons that were hard coded with the forward slash.  Using node's path.join, it is now able to handle builds on windows.  

I was able to build nightwatch and jest tests both from the root directory.

### Additional Details
One important note (marked [here](https://github.com/cerner/terra-core/pull/949/files#r145879687)) is that any paths that show up in the import statements in generated files do not seem to get converted to the backslash, so the path.join is not required (and actually would break the build on windows if it is used)

### Unrelated Config Change
Just a warning, I added the .idea folder to the .gitignore file.  If that's an issue, I can remove it, just figured that I would call it out as it is not technically related to the issue this PR addresses.

### Addresses
#945 #839

### DIFF
--- a/packages/terra-i18n-plugin/tests/i18n-plugin-spec.test.js
+++ b/packages/terra-i18n-plugin/tests/i18n-plugin-spec.test.js
@@ -2,6 +2,7 @@
 /* eslint-disable global-require */
 /* eslint no-underscore-dangle: ["error", { "allow": ["_plugins"] }] */
 import fs from 'fs';
+import path from 'path';
 import I18nAggregatorPlugin from '../src/I18nAggregatorPlugin';
 
 const Compiler = require('webpack/lib/Compiler');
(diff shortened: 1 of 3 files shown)


packages/terra-i18n-plugin/tests/i18n-plugin-spec.test.js
```diff
@@ -142,13 +143,13 @@ describe('i18n-aggregator-plugin', () => {
...
-      expect(consoleMessage).toContain('/tests/fixtures/node_modules/fixtures2/translations');
+      expect(consoleMessage).toContain(path.join('tests', 'fixtures', 'node_modules', 'fixtures2', 'translations'));
       expect(SpyOnConsoleWarn).toHaveBeenCalled();
     });
 
     it('should fill the tranlsation files with the expected information for each locales', () => {
       supportedLocales.forEach((locale, index) => {
-        expect(outputtedFileName[index]).toContain(`aggregated-translations/${locale}.js`);
+        expect(outputtedFileName[index]).toContain(path.join('aggregated-translations', `${locale}.js`));
         expect(outputtedFileContent[index]).toContain(`react-intl/locale-data/${locale}`);
```
Singleton06: This is the line that will fail if we were to do path.join because this will be part of the import statement.

PR #954, opened 2017-10-23
Fix signature sticky draw

yuderekyu:
### Summary
Re-opens https://github.com/cerner/terra-core/pull/929. Functional verification needed.


mmalaker:
Functional verification completed. The signature component now functions as expected without the sticky draw issue.

PR #959, opened 2017-10-23
Delete stats.json file from terra-site

tbiethman:
### Summary
This 23 MB file doesn't seem to be doing much. We could add it to the .gitignore too.

It's being generated by one of our scripts:
`"compile:prod-stats": "webpack --config webpack.prod.config -p --json > stats.json"`


mjhenkes:
Should we stop generating the file? I'm not sure what value is added by saving off the webpack stats?

bjankord:
stats.json can be used with tools to analyze your webpack bundle. We are not currently doing this so I'm fine with just removing the command that generates stats.json and adding it to gitignore if we ever manually generate it to test things with bundle analyzers

Issue #956, opened 2017-10-23
Allow for inputFileSystem to be passed in as param to i18nPlugin


kschuste:
# Issue Description
Would like to add the ability to pass in an inputFileSystem to be used in place of the compiler's default inputFileSystem.  This is currently being done with the outputFileSystem.  

## Issue Type
- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
An inputFileSystem should be allowed to be passed in through the options the same way as the outputFileSystems.  Then can be used instead of the compilers inputFileSystem.

mjhenkes:
+1

bjankord:
Resolved in #961

PR #961, opened 2017-10-24
I18n input file system

kschuste:
### Summary
This resolves issue #956.  Can pass in an inputFileSystem through the options rather than use the compiler's inputFileSystem to read the translations.

### Additional Details
In the options given to the i18nPlugin, can pass in an inputFileSystem to be used instead of the compiler's inputFileSystem.  Updated the tests as well.

### DIFF
--- a/packages/terra-i18n-plugin/tests/i18n-plugin-spec.test.js
+++ b/packages/terra-i18n-plugin/tests/i18n-plugin-spec.test.js
@@ -10,9 +10,13 @@ const NodeJsInputFileSystem = require('enhanced-resolve/lib/NodeJsInputFileSyste
 const CachedInputFileSystem = require('enhanced-resolve/lib/CachedInputFileSystem');
 const MemoryFileSystem = require('memory-fs');
 
+function createInputFileSystem() {
+  return new CachedInputFileSystem(new NodeJsInputFileSystem(), 60000);
+}
+
 function createCompiler() {
   const compiler = new Compiler();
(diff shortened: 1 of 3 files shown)


packages/terra-i18n-plugin/tests/i18n-plugin-spec.test.js
```diff
@@ -103,6 +107,45 @@ describe('i18n-aggregator-plugin', () => {
...
+        inputReadFileSyncPaths.push(filePath);
+        inputReadFileEncondings.push(encoding);
+      });
+
+      new I18nAggregatorPlugin({
+        baseDirectory,
+        supportedLocales,
+        inputFileSystem,
+      }).apply(compiler);
+      compiler._plugins['after-environment'][0]();
```
emilyrohrbough: Is this line a test specific line or does this require additional configuration to the I18nAggregatorPlugin?

kschuste: This is needed to have the test hit the 'after-environment' hook in the code.  Else, it would skip over the following line of code.  Is how I solved it the when I updated this to be able to provide your own outputFileSystem by calling the plugin directly.
"compiler.plugin('after-environment', () => {"

emilyrohrbough: Okay. I assumed that was the case but wanted to know for sure!

Issue #953, opened 2017-10-20
Search Field: disabled state & fits container


neilpfeiffer:
# Related Issues:
[#943 - Search Field: clear field](https://github.com/cerner/terra-core/issues/943)
[#952 - Search Field: autosearch & minLength change](https://github.com/cerner/terra-core/issues/952)

----

# Issue Description
New features to add to the search field component:
1. Allow Search Field to be disabled (prop to pass disabled attribute to both input and button).
2. Add prop to allow for display to switch from fixed size to a resizable fluid/expanding style input (with button fixed at the end) that fits its parent container.

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
- items listed above

## Current Behavior
- no disabled state and fixed size

neilpfeiffer:
cc: @Singleton06 @bjankord @mjhenkes

Singleton06:
@neilpfeiffer, I'm going to be working on this today.  For the disabled state when the character count is less than specified, could we move this to a separate enhancement if it is not ready?  I wouldn't want that to block progress on the rest of the features that we will for sure need.

Singleton06:
# Tech Design for Search Field disabled and fits container
## Summary:
The search field needs to be able to be disabled and fit the width of the container.  

## Requirements Added:
* Allow Search Field to be disabled (prop to pass disabled attribute to both input and button).
* Add prop to allow for display to switch from fixed size to a resizable fluid/expanding style input (with button fixed at the end) that fits its parent container.

## Bidirectionality
These changes should make no changes to the existing ltr/rtl functionality.

## React Props:
| Prop     | Type | Default | Description  |
|----------|------|---------|--------------|
| disabled | bool | false    | Indicates whether or not the search control should be disabled. |
| sizing | string | `fixed` | Indicates what sizing style should be used.  Should be `fixed` or `stretch`.  |

## Example:
```jsx
// with new fields
<SearchField
  onSearch={this.handleSearch}
  disabled
  sizing='stretch' 
/>

// with defaults
<SearchField
  onSearch={this.handleSearch}
/>
```

## CSS Classes:
| Selector         | Description |
|------------------|-------------|
| .stretch | This will be applied to the container of the search field to stretch it to it's parent container.  |

## Notes
[...]

PR #965, opened 2017-10-25
Added search field disabling and stretch sizing

Singleton06:
### Summary
Added code changes that allow for the terra-search-field to be disabled and to be stretched to fit its parent container. 

Addresses #953 

Below is a screenshot of the new functionality:

image

### DIFF
--- a/packages/terra-search-field/src/SearchField.jsx
+++ b/packages/terra-search-field/src/SearchField.jsx
@@ -34,12 +34,24 @@ const propTypes = {
    * A callback to indicate an invalid search. Sends parameter {String} searchText.
    */
   onInvalidSearch: PropTypes.func,
+
+  /**
+   * Indicates whether or not the search control should be disabled.
+   */
+  isDisabled: PropTypes.bool,
+
+  /**
+   * Indicates what sizing style should be used. Should be `fixed` or `stretch`.
+   */
+  sizing: PropTypes.oneOf(['fixed', 'stretch']),
 };
 
 const defaultProps = {
   placeholder: '',
   minimumSearchTextLength: 2,
   searchDelay: 250,
+  isDisabled: false,
+  sizing: 'fixed',
 };
 
 class SearchField extends React.Component {
@@ -86,10 +98,21 @@ class SearchField extends React.Component {
   }
 
   render() {
-    const { placeholder, searchDelay, minimumSearchTextLength, onSearch, onInvalidSearch, ...customProps } = this.props;
+    const {
+      placeholder,
+      searchDelay,
+      minimumSearchTextLength,
+      onSearch,
+      onInvalidSearch,
+      isDisabled,
+      sizing,
+      ...customProps
+    } = this.props;
+
     const searchFieldClassNames = cx([
       'searchfield',
       customProps.className,
+      { stretch: sizing === 'stretch' },
     ]);
 
     return (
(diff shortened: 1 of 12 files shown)


packages/terra-search-field/src/SearchField.jsx
```diff
@@ -34,12 +34,24 @@ const propTypes = {
...
+
+  /**
+   * Indicates whether or not the search control should be disabled.
+   */
+  isDisabled: PropTypes.bool,
+
+  /**
+   * Indicates what sizing style should be used. Should be `fixed` or `stretch`.
+   */
+  sizing: PropTypes.string,
```
kschuste: could we set this to only allow for fixed or stretch like so?
sizing: PropTypes.oneOf(['fixed', 'stretch']),

Singleton06: Definitely.... I should have thought of that.  Would you envision that the test for 'renders a search field with no stretch class specified if sizing invalid' is still needed with this specified?

Singleton06: Actually, I went ahead and made the change and jest fails in situations where there's an invalid prop.  I'll remove the test as well.

Singleton06: This has been addressed

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -86,10 +98,20 @@ class SearchField extends React.Component {
...
+      searchDelay,
+      minimumSearchTextLength,
+      onSearch,
+      onInvalidSearch,
+      isDisabled,
+      sizing,
+      ...customProps
+    } = this.props;
+
+    const searchFieldClassNames = cx(['searchfield',
```
jmsv6d: Nit: move the searchfield string to next line.

packages/terra-search-field/tests/nightwatch/search-field-spec.js
```diff
@@ -74,4 +74,19 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
       });
     });
   },
+
+  'Displays the search button disabled': (browser) => {
+    browser.url(`${browser.launchUrl}/#/tests/search-field-tests/disabled`);
+
+    browser
+      .assert.elementPresent('#disabledsearchfield')
```
jmsv6d: Use expect syntax instead.

packages/terra-search-field/src/SearchField.jsx
Singleton06: Addressed

packages/terra-search-field/tests/nightwatch/search-field-spec.js
Singleton06: Addressed

packages/terra-search-field/tests/nightwatch/search-field-spec.js
```diff
@@ -74,4 +75,15 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
       });
     });
   },
+
+  'Displays the search button disabled': (browser) => {
+    browser.url(`${browser.launchUrl}/#/tests/search-field-tests/disabled`);
+
+    browser.assert.expect.element('#disabledsearchfield').to.be.present;
```
jmsv6d: Remove the asserts here.

Singleton06: Yeah... :( I was getting some other really weird errors and I assumed it was something windows related, but now realized that I goofed.

Singleton06:
Just a heads up to everyone, I am wanting to get a few additional fixes pushed in for the search field that would be on top of these changes.  Instead of forcing multiple rounds of ux review, etc. I think I am just going to bunch them together because I don't have a large amount of time.  All of the changes are fairly small but will be in the same code.

For that reason, I'm going to go ahead and close this PR and I'll be creating a new one with the other changes added.

PR #969, opened 2017-10-25
Field OSC Uplift - Phase 1

### LINKED ISSUE #889: Uplift Field Component for OCS
# Issue Description
Uplift Field component.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other
#

emilyrohrbough:
### Summary
Closes #889.

#### New props include:
- `errorIcon`
- `hideRequired`
- `isInvalid`
- `showOptional`

#### Added/Changed Behavior
- If component `isInvalid`, the provided error icon and error text will show. Previous error text always show when provided.
- If `hideRequired` prop is used, and the component `isInvalid` the required indicated will show 
- If `showOptional` prop is used, the label '(optional)' is appended to the provide label text. Translations are provided. (Fl-fi still needs to be updated)

### Additional Details
There will be an additional PR to uplift NumberField, SelectField, TextField and TextAreaField to expose the new props.

### Deployment Links
- https://terra-core-deployed-pr-969.herokuapp.com/static/#/site/form/field
- https://terra-core-deployed-pr-969.herokuapp.com/static/#/tests/form-tests


packages/terra-form/src/Field.jsx
```diff
@@ -13,18 +13,30 @@ const propTypes = {
    */
   children: PropTypes.node,
   /**
-   * Error message for when the input is invalid.
+   * Error message for when the input is invalid on validation.
```
JakeLaCombe: Minor:

"on validation" feels a little extraneous.

packages/terra-form/src/Field.jsx
```diff
@@ -13,18 +13,30 @@ const propTypes = {
...
   /**
+   * Determines whether or not to hide the required indicator on the label.
+   */
+  hideRequired: PropTypes.bool,
+  /**
    * The htmlFor attribute on the field label.
    */
   htmlFor: PropTypes.string,
   /**
+   * Determines whether the field is in error or not.
```
JakeLaCombe: perhaps update this to say "invalid or not"?

packages/terra-form/src/Field.jsx
```diff
@@ -40,24 +52,56 @@ const propTypes = {
...
   labelAttrs: {},
   required: false,
+  showOptional: false,
+};
+
+const contextTypes = {
+  /* eslint-disable consistent-return */
+  intl: (context) => {
+    if (context.intl === undefined) {
+      return new Error('Please add locale prop to Base component to load translations');
```
JakeLaCombe: It feels like the error her is actually not wrapping the field inside terra-i18n or terra-base.

packages/terra-form/src/Field.scss
```diff
@@ -1,62 +1,80 @@
...
-    &::before {
-      color: #e50000;
-      content: '*';
-      display: inline-block;
-      padding-right: 0.357em;
-    }
+  .label-group {
+    font-family: var(--terra-form-field-label-font-family, 'Helvetica Neue');
+    font-size: var(--terra-form-field-label-font-size, 14px);
+    line-height: 1.29;
```
JakeLaCombe: It feels as though line-height should be themed as well, since it goes hand to hand with font-size. Did UX have thoughts on this?

packages/terra-form/tests/jest/Field.test.jsx
```diff
@@ -1,29 +1,140 @@
...
+      label="Field Label"
+      required
+      hideRequired
+    />
+  );
+  const wrapper = shallow(field, intlContexts.shallowContext);
+  expect(wrapper).toMatchSnapshot();
+});
+
+it('should render a feild error message', () => {
```
yuderekyu: *field

packages/terra-form/tests/jest/Field.test.jsx
```diff
@@ -1,29 +1,140 @@
...
+    <Field
+      label="Field Label"
+      required
+    />
+  );
+  const wrapper = shallow(field, intlContexts.shallowContext);
+  expect(wrapper).toMatchSnapshot();
+});
+
+it('should render a required feild label with required hidden', () => {
```
yuderekyu: *field

packages/terra-form/src/Field.scss
emilyrohrbough: @neilpfeiffer

packages/terra-form/src/Field.jsx
emilyrohrbough: What do you mean? I'm not sure I understand.

packages/terra-form/src/Field.jsx
```diff
@@ -40,24 +52,56 @@ const propTypes = {
...
+  /**
+   * Determines whether or not to append the 'optional' label to a non-required field label.
+   */
+  showOptional: PropTypes.bool,
 };
 
 const defaultProps = {
   children: null,
   error: null,
+  errorIcon: <IconError />,
```
kschuste: do we want to give them the option to pass in their own error icon?  Shouldn't we just lock this down like we have done with the error and no data views.

packages/terra-form/src/Field.scss
```diff
@@ -1,62 +1,80 @@
...
+    margin-bottom: var(--terra-form-field-label-margin-bottom, 0.357em);
+    text-align: left;
```
yuderekyu: Seems like we could have a themeable variable associated with `text-align` to account for label text placement. Thoughts?

packages/terra-form/src/Fieldset.scss
```diff
@@ -0,0 +1,59 @@
...
+  .fieldset-required > .legend:first-child {
+    &::before {
+      color: #e50000;
+      content: '*';
+      display: inline-block;
+      padding-right: 0.357em;
+    }
+  }
+
+  .label {
```
yuderekyu: Same as the previous comment for text-align

packages/terra-form/src/Field.jsx
JakeLaCombe: Sorry for vaguness.

I think this error comes not from providing a locale to the base component, but by not wrapping this field inside a terra-i18n or react-intl object. For example, if I call on FormattedMessage from react-intl but don't wrap the component inside the IntlProvider, I get this error message

"[React Intl] Could not find required `intl` object. <IntlProvider> needs to exist in the component ancestry."

I feel that the error message should direct users to ensuring they have wrapped this component in terra-base or the terra-i18n-provider. (IMO, I would direct them to ensuring this is wrapped in terra-base, and not even mention the terra-i18n-provider that we have)

packages/terra-form/src/Field.jsx
JakeLaCombe: I think for themability, we could be limited if we lockdown this icon.

packages/terra-form/src/Field.scss
```diff
@@ -1,62 +1,80 @@
...
-  .fieldset {
+  .field {
     border: 0;
-    margin-bottom: 1.428em;
+    margin-bottom: var(--terra-form-field-margin-bottom, 1.428em);
+    max-width: 1020px;
     padding: 0;
   }
 
+  // Remove this??
```
yuderekyu: Remove this (hah meta)

packages/terra-form/src/Field.scss
emilyrohrbough: Base on the current mocks I've seen, text-alignment at this point will be left aligned. In the future there would be potential for right alignment, but I foreseeing any logic regarding that would need to be handled by the component itself.

packages/terra-form/src/Field.scss
emilyrohrbough: Ah right. This was a reminded. With the newest uplift should we remove screen width specific stylings and leave responsiveness to a HOC using the Field?

packages/terra-form/src/Field.jsx
emilyrohrbough: Given there are multiple sets of icons available with their own version of an Error Icon, this is needed for themability. It was specified in the implementation wireframe.

packages/terra-form/src/Field.jsx
```diff
@@ -13,18 +13,30 @@ const propTypes = {
...
+   */
+  hideRequired: PropTypes.bool,
+  /**
    * The htmlFor attribute on the field label.
    */
   htmlFor: PropTypes.string,
   /**
+   * Determines whether the field is in error or not.
+   */
+  isInvalid: PropTypes.bool,
```
mjhenkes: The isInvalid change is non passive.

packages/terra-form/src/Field.jsx
jmsv6d: That's probably something that should be logged as a separate issue seeing as this string has been copy and pasted in all of our components that have translations.

emilyrohrbough:
It was decided to create a new npm package for OSC uplift of Field. Closing this PR. The new PR will address the comments made here and I will add comments with a screenshot of any unanswered questions.

PR #970, opened 2017-10-25
Create Multiple Row Select Table

### LINKED ISSUE #835: Create Multiple Selectable Table Component
# Issue Description
We want a table where multiple rows of a terra-table can be selected. Currently, there is only a SingleSelectableRows component. A parallel MultipleSelectableRows component should be made.

## Issue Type

- [X] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior

* Initial Render
  * The table is rendered with any preselected row selected
    * Honors the `isSelected` prop
* Clicking on an unselected row will select the row unless `isSelectable` is `false`. \* 
* Clicking on a selected row will deselect the row unless `isSelectable` is `false`. \* 
* Multiple rows can be selected
* A callback function can be provided to the component that is called after any row is selected or deselected.

\* - isSelectable could be renamed/ a new prop with different name could be used to something else to better convey this functionality, especially with deselect.

## Current Behavior

N/A

dv297:
### Summary
This pull request adds a new component called `MultiSelectableRows` to terra-table. This allows the user to select multiple rows from a table. Closes #835

Deployment link: https://terra-core-deployed-pr-970.herokuapp.com/static/#/site/table/multi-select


packages/terra-table/src/SelectableUtils.js
```diff
@@ -1,7 +1,20 @@
 /**
-* Returns the index of the first selected row for SingleSelectableRows.
-* To be used in the constructor, to set initial state.
-*/
+ * Returns a validated max count for selection based on the rows and max provided.
+ */
+const validatedMaxCount = (rows, maxSelectionCount) => {
```
kschuste: would rename this to be validatedMaxCountSelection to be more representative of what returning

packages/terra-table/src/SelectableUtils.js
```diff
@@ -16,21 +29,76 @@ const initialSingleSelectRowIndex = (rows) => {
...
+ */
+const initialMultiSelectRowIndexes = (rows, maxSelectionCount) => {
+  if (!rows || !rows.length) {
+    return [];
+  }
+  // Find the rows which are selected and are selectable
+  const selectedIndexes = [];
+  const validatedMaxSelectionCount = validatedMaxCount(rows, maxSelectionCount);
+  for (let i = 0; i < rows.length; i += 1) {
+    if (selectedIndexes.length >= validatedMaxSelectionCount) {
```
kschuste: could this get moved into the isSelected if statement to after you push on the selected index,

this way it wouldn't get checked every loop and would still be able to break out early after finding all the indexes

packages/terra-table/src/SelectableUtils.js
```diff
@@ -16,21 +29,76 @@ const initialSingleSelectRowIndex = (rows) => {
...
+  }
+  return selectedIndexes;
+};
+
+/**
+ * Returns a new array, updated with the newIndex being added or removed from the existing.
+ */
+const updatedMultiSelectedIndexes = (currentIndexes, newIndex) => {
+  let newIndexes = [];
+  if (currentIndexes.length) {
```
kschuste: want to check if currentIndexes is not undefined first

packages/terra-table/src/SelectableUtils.js
dv297: Changed to validatedMaxCountSelection

packages/terra-table/src/SelectableUtils.js
dv297: Moved this check to inside the if statement after the push

packages/terra-table/src/SelectableUtils.js
dv297: Set to if (currentIndexes && currentIndexes.length)

packages/terra-table/src/SelectableUtils.js
```diff
@@ -16,21 +29,77 @@ const initialSingleSelectRowIndex = (rows) => {
...
+  }
+  return selectedIndexes;
+};
+
+/**
+ * Returns a new array, updated with the newIndex being added or removed from the existing.
+ */
+const updatedMultiSelectedIndexes = (currentIndexes, newIndex) => {
+  let newIndexes = [];
+  if (currentIndexes && currentIndexes.length) {
```
yuderekyu: Thinking we could rewrite this so it's not as nested. Something like:
```javascript
if(!Array.isArray(currentIndexes) || !currentIndexes.length ){
  newIndexes.push(newIndex);
  return newIndexes;
}
//same conditionals you had after the first if statement
```
Thoughts?

dkasper-was-taken: If guard statements result in duplicate return statements I'd prefer documentation instead of additional lines, as the uglified code will still be smaller.

yuderekyu: To me, it's easier to follow the logic with a guard clause that is included at the top, even if there are multiple return statements. Guess it comes down to personal preference though, +1 either way.

emilyrohrbough: In this instance, I think its fine & is consistent with patterns seen in other components.

dv297: I'm fine with making any changes, but I do want to bring up that I basically drew most of my coding and styling considerations from terra-list, which already has a MultiSelect component published. This function parallels [this function](https://github.com/cerner/terra-core/blob/master/packages/terra-list/src/SelectableUtils.js#L55) for terra-list. So I tried to not introduce my own stylistic tendencies here and opted to follow code that already made it through PR.

I'm fine with making the change, but since @yuderekyu already mentioned he's +1 either way, I'll leave it as is unless someone explicitly says to go ahead and change it.

packages/terra-table/src/SelectableUtils.js
```diff
@@ -16,21 +29,77 @@ const initialSingleSelectRowIndex = (rows) => {
...
+ * To be used in the constructor, to set initial state.
+ */
+const initialMultiSelectRowIndexes = (rows, maxSelectionCount) => {
+  if (!rows || !rows.length) {
+    return [];
+  }
+  // Find the rows which are selected and are selectable
+  const selectedIndexes = [];
+  const validatedMaxSelectionCount = validatedMaxCountSelection(rows, maxSelectionCount);
+  for (let i = 0; i < rows.length; i += 1) {
```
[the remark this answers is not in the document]
dv297: Maybe I'm not seeing how you would write this while loop, but there is no guarantee that selectedIndexes.length would be >= validatedMaxSelectionCount. We only push to the array if the row's isSelected prop is true. If any row isn't initially selected, then the while loop would never terminate.

Also, same reasoning as comment below, but this function parallels [this function](https://github.com/cerner/terra-core/blob/master/packages/terra-list/src/SelectableUtils.js#L36) in terra-list.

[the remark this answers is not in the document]
yuderekyu: I didn't realize that break statement is an edge case. Go ahead and leave it 👍

packages/terra-table/docs/README.md
```diff
@@ -16,7 +16,8 @@ Terra-table provides several different components for building tables accessible
 - HeaderCell: Renders a table header cell.
 - TableRows: Wrapper to apply onClick and onKeyDown function on each child row.
 - SelectableTableRows: Wrapper that allows for selection state management of each child row. Use in place of TableRows.
-- SingleSelectableRows: Wrapper that handles single selection state management of each child row. Use in place of TableRows.
+- SingleSelectableRows: Wrapper that handles single selection state mange of each child row. Use in place of TableRows.
+- MultiSelectableRows: Wrapper that handles multiple selection state mange of each child row. Use in place of TableRows.
```
jmsv6d: Typo *magement

packages/terra-table/docs/README.md
```diff
@@ -83,5 +84,36 @@ import Table from 'terra-table';
...
+      <Table.Cell content="Jane Smith" key="NAME" />
+      <Table.Cell content="321 Drive Street" key="ADDRESS" />
+      <Table.Cell content="111-222-3333" key="PHONE_NUMBER" />
+    </Table.Row>
+    <Table.Row key="PERSON_2">
+      <Table.Cell content="Dave Smith" key="NAME" />
+      <Table.Cell content="213 Raymond Road" key="ADDRESS" />
+      <Table.Cell content="111-222-3333" key="PHONE_NUMBER" />
+    </Table.Row>
+  </Table.SingleSelectableRows>
```
jmsv6d: Update to MultiSelectableRows

packages/terra-table/docs/README.md
dv297: Changed to management

packages/terra-table/docs/README.md
dv297: Updated to MultiSelectableRows

packages/terra-site/src/TestLinks.jsx
```diff
@@ -37,6 +37,7 @@ const TestLinks = () => (
       <li><Link to="/tests/modal-tests">Modal Tests</Link></li>
       <li><Link to="/tests/modal-manager-tests">Modal Manager Tests</Link></li>
       <li><Link to="/tests/multi-select-list-tests">MultiSelectList Tests</Link></li>
+      <li><Link to="/tests/multi-select-table-tests">MultiSelectTable Tests</Link></li>
       <li><Link to="/tests/overlay-tests">Overlay Tests</Link></li>
```
emilyrohrbough: I noticed that SelectableTable Tests come after the Signature Tests, when it should come after Selectable List Tests. Would you mind sneaking this in?

packages/terra-table/src/SelectableUtils.js
```diff
@@ -1,7 +1,20 @@
...
+ */
+const validatedMaxCountSelection = (rows, maxSelectionCount) => {
+  if (maxSelectionCount !== undefined) {
+    return maxSelectionCount;
+  }
+  if (rows) {
+    return rows.length;
+  }
+  return 0;
+};
```
emilyrohrbough: Since rows are actually React children, we can use `return React.Children.count(rows);`

packages/terra-table/src/SelectableUtils.js
```diff
@@ -1,7 +1,20 @@
 /**
-* Returns the index of the first selected row for SingleSelectableRows.
-* To be used in the constructor, to set initial state.
-*/
+ * Returns a validated max count for selection based on the rows and max provided.
```
emilyrohrbough: nit: 'Returns a validated max count for selection. Validates the max count prop, and if undefined returns a max of the count of children.'

packages/terra-table/src/SelectableUtils.js
```diff
@@ -16,21 +29,77 @@ const initialSingleSelectRowIndex = (rows) => {
 };
 
 /**
-* Returns if the selected row index is already selected for SingleSelectableRows.
-*/
+ * Returns the indexes of the selected rows for MultiSelectableRows.
+ * To be used in the constructor, to set initial state.
+ */
+const initialMultiSelectRowIndexes = (rows, maxSelectionCount) => {
+  if (!rows || !rows.length) {
```
emilyrohrbough: Since rows are actually React children, we don't need this if check. We can clean this up to be 

`const childArray = React.Children.toArray(row);`

If there are no children, react handles it for us.

packages/terra-table/src/SelectableUtils.js
```diff
@@ -16,21 +29,77 @@ const initialSingleSelectRowIndex = (rows) => {
...
+      if (selectedIndexes.length >= validatedMaxSelectionCount) {
+        break;
+      }
+    }
+  }
+  return selectedIndexes;
+};
+
+/**
+ * Returns a new array, updated with the newIndex being added or removed from the existing.
```
emilyrohrbough: nit: returns a new array _of the selected indexes_...

packages/terra-table/tests/jest/SingleSelectableRows.test.jsx
```diff
@@ -29,12 +29,6 @@ it('should render SingleSelectableRows with one row', () => {
   expect(tableRows).toMatchSnapshot();
 });
 
-it('should render SingleSelectableRows with maximum height set', () => {
```
emilyrohrbough: Why did we remove this test?

packages/terra-table/tests/jest/TableHeaderContent.test.jsx
```diff
@@ -19,9 +19,3 @@ it('should render table header content tag with descending sort indicator', () =
   const tableHeaderContent = shallow(tableHeaderContentTag);
   expect(tableHeaderContent).toMatchSnapshot();
 });
-
-it('should render table header content tag with maximum height fixed', () => {
```
emilyrohrbough: Why did we remove this test?

packages/terra-table/tests/jest/TableRowContent.test.jsx
```diff
@@ -7,9 +7,3 @@ it('should render a default table row content', () => {
   const tableRowContent = shallow(defaultTableRowContent);
   expect(tableRowContent).toMatchSnapshot();
 });
-
-it('should render a table row content with maximum height', () => {
```
emilyrohrbough: Why did we remove this test?

packages/terra-table/tests/nightwatch/multi-select-table/MultiSelectTableTestRoutes.jsx
```diff
@@ -0,0 +1,27 @@
...
+import TableMaxCount from './components/MultiRowSelectableTableMaxCount';
+import TableMaxCountInitiallyMet from './components/MultiRowSelectableTableMaxCountInitiallyMet';
+import TableMaxCountInitiallyExceeded from './components/MultiRowSelectableTableMaxCountInitiallyExceeded';
+
+const routes = (
+  <div>
+    <Route path="/tests/multi-select-table-tests" component={MultiSelectTableTests} />
+    <Route path="/tests/multi-select-table-tests/default" component={Table} />
+    <Route path="/tests/multi-select-table-tests/onchange" component={TableOnChange} />
+    <Route path="/tests/multi-select-table-tests/preselected" component={TablePreselected} />
```
emilyrohrbough: The nice part of having the SelectableTableRows component it is handling the logic needed to wrap the table rows with the correct selected and selectable props. This being said, the preselected and non-selectable tests can be removed.

packages/terra-table/tests/nightwatch/multi-select-table/multi-select-table-spec.js
```diff
@@ -0,0 +1,286 @@
...
+const resizeTo = require('terra-toolkit/lib/nightwatch/responsive-helpers').resizeTo;
+
+/**
+ * Create NightWatch expects against the row element at the 0-indexed position.
+ */
+function expectRowForBrowserAtIndex(browser, index) {
+  return browser.expect.element(`[class*="row"]:nth-child(${index + 1})`);
+}
+
+/**
```
emilyrohrbough: These helper functions are a good idea!! 

Lets update the names to be a bit more readable, maybe something like: `expectRowIsSelected` and `expectRowIsNotSelected`. If anything, this needs to be updated from the use of the term 'assert'. The assert and expect syntax in nightwatch are different and our team is in the process of moving to use only expect. Plus, assert is misleading given these helpers actually do use expect by calling `expectRowForBrowswerAtIndex`.

packages/terra-site/src/examples/table/examples/MultiSelect.jsx
```diff
@@ -0,0 +1,112 @@
...
+          <Table.Cell content="111-222-3333" key="PHONE_NUMBER" />
+        </Table.Row>
+        <Table.Row key={'PERSON_2'}>
+          <Table.Cell content="Dave Smith" key="NAME" />
+          <Table.Cell content="213 Raymond Road" key="ADDRESS" />
+          <Table.Cell content="111-222-3333" key="PHONE_NUMBER" />
+        </Table.Row>
+      </Table.MultiSelectableRows>
+    </Table>
+    <h3>Multiple Selectable Rows (First Row Initially Selected)</h3>
```
emilyrohrbough: We probably don't need this site example. The Table site page shows initially shows a pre-selected table row which provides an example of this behavior. If you felt it was still important to show within mulit-select rows, maybe update an one of the other examples?

packages/terra-table/tests/jest/SingleSelectableRows.test.jsx
dv297: It was actually your recommendation. It is a stale test left by [this pull request removing max height tests](https://github.com/cerner/terra-core/pull/238)

packages/terra-table/tests/jest/TableHeaderContent.test.jsx
dv297: Same as above

It was actually your recommendation. It is a stale test left by [this pull request removing max height tests](https://github.com/cerner/terra-core/pull/238)

packages/terra-table/tests/jest/TableRowContent.test.jsx
dv297: Same as above

It was actually your recommendation. It is a stale test left by [this pull request removing max height tests](https://github.com/cerner/terra-core/pull/238)

packages/terra-table/tests/jest/TableHeaderContent.test.jsx
emilyrohrbough: Opps forgot! 👍

packages/terra-table/tests/nightwatch/multi-select-table/MultiSelectTableTestRoutes.jsx
dv297: It might be beneficial to leave them for several reasons. We know that the internal implementation of MultiSelectableRows uses the SelectableTableRows and there is a certain level of trust we can offload onto the SelectableTableRows. But it's still the responsibility/ acceptance criteria of the MultiSelectableRows component to have these behaviors. So these tests can be mentally viewed as an integration test of these two components.

In our Jest tests, we shallow render and assert against the props that MSR passes to SelectableTableRows. So we are offloading the responsibility to the contract of the SelectableRows component. But since it's shallow rendered, the actual behavior of the MSR component isn't actually asserted against. So Nightwatch tests cover that.

I'm fine with removing it, but perhaps it would be good for any consumer/wrapper around SelectableTableRows to have a base suite of tests like this to have and ensure the intended behavior is preserved/ any changes to the original behavior is called out.

packages/terra-table/src/SelectableUtils.js
dv297: I think it cleans up 3 physical lines but perhaps forces the reader to mentally read all of the lines? With the existing code, we break out of the function immediately on edge case, and the reader doesn't have to read the rest of the function. Whereas with the new suggestion, React.children.toArray returns an empty function but we can't return just yet. We go to the for loop and then have to know toArray gave us an empty array, and then we skip to return an emptySelectedIndexes.

Code style preference, but explicitly bringing out the edge case seems helpful here.

packages/terra-table/tests/nightwatch/multi-select-table/multi-select-table-spec.js
dv297: Thanks! =) I left the original functions with ForBrowserAtIndex for clarity but I adjusted them all to use expect and for the functions used in the actual tests, I removed AtIndex as per your suggestion.

packages/terra-site/src/examples/table/examples/MultiSelect.jsx
dv297: I just added it to the default MultiSelectableRows example. The other examples showed that behavior before so seemed like it would work here =)

packages/terra-table/src/SelectableUtils.js
jmsv6d: This will likely also trigger in the case that there is only 1 row. React children only become arrays if there is more than one child element. If there is only 1 then it is passed in as type element (so `!rows.length` would be true). I would recommend removing this conditional and just using `React.Children.forEach` for your loop below, or using `toArray` as @emilyrohrbough  suggested. These helper functions will handle all of these edge cases for us.

packages/terra-table/tests/nightwatch/multi-select-table/MultiSelectTableTestRoutes.jsx
emilyrohrbough: I see value in both sides. Whatever other's think.

packages/terra-site/src/TestLinks.jsx
dv297: Moved it to be after MultiSelectList

packages/terra-site/src/examples/table/examples/MultiRowSelectableTable.jsx
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable import/no-extraneous-dependencies */
+import React from 'react';
+import Table from 'terra-table';
+
+const MultiRowSelectableTable = () => (
+  <Table isStriped={false}>
+    <Table.Header>
+      <Table.HeaderCell content={'Name'} key={'NAME'} minWidth={'small'} />
```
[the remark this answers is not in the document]
dv297: Looked in these 3 places, changed to use double quotes rather than braces+single.

packages/terra-table/docs/MultiSelectableRows.md
```diff
@@ -0,0 +1,36 @@
...
+import Table from 'terra-table';
+
+<Table isStriped={false}>
+  <Table.Header>
+    <Table.HeaderCell content={'Name'} key={'NAME'} minWidth={'small'} />
+    <Table.HeaderCell content={'Address'} key={'ADDRESS'} minWidth={'medium'} />
+    <Table.HeaderCell content={'Phone Number'} key={'PHONE_NUMBER'} minWidth={'large'} />
+  </Table.Header>
+  <Table.MultiSelectableRows maxSelectionCount={2}>
+    <Table.Row key={'PERSON_0'} isSelected={true}>
```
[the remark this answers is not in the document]
dv297: Changed to isSelected

packages/terra-table/src/SelectableUtils.js
dv297: Updated

packages/terra-table/src/SelectableUtils.js
dv297: Good point. Learned a bit there! Have adjusted to reflect @emilyrohrbough 's original suggestion =)

packages/terra-site/src/examples/table/examples/MultiSelect.jsx
```diff
@@ -0,0 +1,87 @@
...
+
+import Table from 'terra-table';
+
+const table = () => (
+  <div>
+    <h2>Examples</h2>
+    <h3>Multiple Selectable Rows</h3>
+    <Table isStriped={false}>
+      <Table.Header>
+        <Table.HeaderCell content={'Name'} key={'NAME'} minWidth={'small'} />
```
jmsv6d: This example file still has some single quotes being used. Also, this file has several examples, they should be broken out into separate files.

packages/terra-site/src/examples/table/examples/MultiRowSelectableTable.jsx
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
jmsv6d: It doesn't look like this file is being used anywhere...

packages/terra-site/src/examples/table/examples/MultiSelect.jsx
dv297: Addressed the single quotes in this file. Also, I'm fine with splitting them out, I just copied this style from [MultiSelectList](https://github.com/cerner/terra-core/blob/master/packages/terra-site/src/examples/list/examples/MultiSelect.jsx). Is there a preferred way of organizing these separated files? Should there be another directory under "examples" that keeps these MultiSelectableRows example? Or should I keep them like this to be consistent with List?

packages/terra-site/src/examples/table/examples/MultiRowSelectableTable.jsx
dv297: You're right. [Deleted this file](https://github.com/cerner/terra-core/pull/970/files/84e8501009018d54655e3b677f2e26a1465beab9..b75a40549d698748c69fd3e2638f5aac789464be#diff-acdeacd3cc8c3c673be8201657fcf4fdL1)

packages/terra-site/src/examples/table/examples/MultiSelect.jsx
jmsv6d: All of our other components have the examples split out, I'm not sure why that one got done differently. It should be fine to keep them all under the examples directory.

dv297: I just assumed it was because MultiSelect and SingleSelect List had their own documentation page separate from the List component. But I went ahead and [moved all the examples for table to be directly under example](https://github.com/cerner/terra-core/pull/970/commits/213dd233bc14cb7c06b5b59958d338fcbd0b932b#diff-049c31c4f88c4154e188a37bfde6a7b4R21)

packages/terra-site/themeable-variables.json
```diff
@@ -1,5 +1,5 @@
 {
-  "Alert": {
+  "C:\\workspace\\dv045327_personal_repos\\terra-core\\packages\\terra-alert\\lib\\Alert": {
```
emilyrohrbough: We'll want to revert these changes in themeable-variables.

packages/terra-site/themeable-variables.json
```diff
@@ -1,5 +1,5 @@
 {
-  "C:\\workspace\\dv045327_personal_repos\\terra-core\\packages\\terra-alert\\lib\\Alert": {
```
emilyrohrbough: Looks like you already got it 😄

dv297: Yep! Definite "oh no" moment! Just curious, seems like these full paths are auto generated on Windows but don't seem to be an issue for others, are there any issues logged to terra-core about this? I haven't done too much investigation on what this file does or what causes this path to get here.

emilyrohrbough: I'm not sure if there is an issue logged, but this behavior is something we are aware of. Our themeable-variables plugin needs updated to handle windows file paths. If I remember its checking for '/' to split the path and since windows is '\\' it returns the full thing.

mmalaker:
We seem to be missing a deployment link

mjhenkes:
@mmalaker Weird, Here is the link https://terra-core-deployed-pr-970.herokuapp.com/

mmalaker:
Functional verification completed. The multi-select table row examples display and function correctly.

PR #975, opened 2017-10-27
Update hook shot and popup to enable positioning on resize.

dkasper-was-taken:
### Summary
Use resize observer to add ability to position with inner content changes size while using automatic height and widths.

@cerner/terra


packages/terra-hookshot/src/HookshotContent.jsx
```diff
@@ -122,6 +135,24 @@ class HookshotContent extends React.Component {
     }
   }
 
+  enableContentResizeListener() {
+    if (!this.contentResizeListenerAdded) {
```
StephenEsser: Would you be able to check for the presence of `this.resizeObserver` here and set it to null in disable to remove the additional booleans?

packages/terra-hookshot/src/HookshotContent.jsx
```diff
@@ -122,6 +135,24 @@ class HookshotContent extends React.Component {
...
+        this.props.onContentResize(entries[0].contentRect);
+      });
+      this.resizeObserver.observe(this.contentNode);
+      this.contentResizeListenerAdded = true;
+    }
+  }
+
+  disableContentResizeListener() {
+    if (this.contentResizeListenerAdded) {
+      this.resizeObserver.disconnect(this.contentNode);
```
StephenEsser: To be safe we should disconnect the resize observer when the component unmounts

packages/terra-hookshot/src/HookshotContent.jsx
```diff
@@ -150,7 +182,7 @@ class HookshotContent extends React.Component {
     delete customProps.closePortal;
 
     return (
-      <div {...customProps} className={cx(['content', customProps.className])} ref={refCallback}>
+      <div {...customProps} className={cx(['content', customProps.className])} ref={(element) => { this.contentNode = element; this.props.refCallback(element); }}>
```
StephenEsser: It looks like refCallback was previously using the spread variable from above but was prefixed with `this.props` after the change

packages/terra-hookshot/src/HookshotContent.jsx
dkasper-was-taken: Good catch, I had missed that one.

packages/terra-hookshot/src/HookshotContent.jsx
dkasper-was-taken: I could, but I'd like to match the other listeners for this.

packages/terra-popup/tests/nightwatch/popup-spec.js
```diff
@@ -307,4 +307,11 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
     browser.expect.element('.test-content').to.have.attribute('data-terra-popup-automatic-width');
     browser.expect.element('.test-content').to.have.attribute('style').which.contains('width: 200px;');
   },
+
+  'Displays a popup with automatic dimensions resizing': (browser) => {
+    browser.url(`${browser.launchUrl}/#/tests/popup-tests/resize-content`);
+    browser.expect.element('.test-content').to.have.attribute('data-terra-popup-automatic-height');
+    browser.expect.element('.test-content').to.have.attribute('data-terra-popup-automatic-width');
+    browser.click('#resize-content');
+  },
```
emilyrohrbough: Do we need to verify the content's new attachment position remained the same and the content shifted?

packages/terra-popup/tests/nightwatch/PopupTests.jsx
```diff
@@ -30,6 +30,7 @@ const PopupTests = () => (
       <li><Link to="/tests/popup-tests/vertical-right-offset">Popup - Vertical Right Attachment -- Arrow Adjustment</Link></li>
       <li><Link to="/tests/popup-tests/vertical-right-target-offset">Popup - Vertical Right Attachment -- Target Smaller than Arrow Offset</Link></li>
       <li><Link to="/tests/popup-tests/popup-inside-modal">Popup - Popup Inside Modal</Link></li>
+      <li><Link to="/tests/popup-tests/resize-content">Popup - Resize Cotent</Link></li>
```
emilyrohrbough: typo: Content, you have Cotent

packages/terra-hookshot/src/HookshotContent.jsx
```diff
@@ -92,6 +100,12 @@ class HookshotContent extends React.Component {
     } else {
       this.disableResizeListener();
     }
+
+    if (this.props.onContentResize) {
+      this.enableContentResizeListener();
+    } else {
```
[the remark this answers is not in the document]
dkasper-was-taken: This is kind of an edge case, if the HOC updates the prop and no longer passes a callback function, we will need to remove it.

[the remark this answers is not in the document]
emilyrohrbough: Gottchya. Makes sense

packages/terra-hookshot/src/HookshotContent.jsx
```diff
@@ -122,6 +136,24 @@ class HookshotContent extends React.Component {
...
   }
 
+  enableContentResizeListener() {
+    if (!this.contentResizeListenerAdded) {
+      this.resizeObserver = new ResizeObserver((entries) => {
+        this.props.onContentResize(entries[0].contentRect);
+      });
+      this.resizeObserver.observe(this.contentNode);
+      this.contentResizeListenerAdded = true;
+    }
```
StephenEsser: I don't know if this can happen with this implementation, but with the resize observers I believe it's possible to enter an infinite resizing loop depending on how the content resizes when the bounding rect resizes. I've never seen it happen with the ResponsiveElement so I think it's unlikely it will happen here, just thought it might be worth mention. The resize observer might internally debounce slightly to avoid this

dkasper-was-taken:
Closing, changes will be held off until this functionality is needed.

alex-bezek:
Discussed IRL a bit. 
* Originally UX had said pop ups should not resize while open, so this was not needed anymore and we were going to look at making it so it wouldn't resize when contents resized
* It was discussed in the war room and decided that pop ups should still resize when contents size changes (largely due to ajax loaded content)

Sounds like we went full circle and we are back to our original use case for this, so it is still needed I think. There was the idea to use modal's instead of the pop up for our use case, but this is just intended for mobile, not desktop.

cc @mhemesath  @bjankord

dkasper-was-taken:
Reopening.

PR #978, opened 2017-10-30
Terra select ocs uplift

### LINKED ISSUE #880: Uplift Select component for OCS
# Issue Description
Uplift select component to support ocs expected behavior.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

jmsv6d:
### Summary
Resolves #880. First round of changes to uplift select for ocs. Props have been updated to follow our current conventions for children and to align with new props being added in field and text area uplifts. Changes include:

- Created new package `terra-form-select`
- Created custom select component utilizing the auto sized popup. 
  - Utilizes the auto sized popup for menu drop down sizing.
  - Can navigate options in menu using arrow keys.
  - Will wrap text in menu option if there is not enough space available
  - Will truncate text in the input if there is not enough space available
- Created select.option component for building out options. 
- Removed props:
  - `options`
  - `choices`
- Added props:
  - `children`
  - `disabled`
  - `isInvalid`

#### Future enhancements
This is the first iteration of OCS changes. Some of the OCS functionality that has not been included and will need to be added in future iterations include:
- Desired fullscreen functionality (currently using the default popup functionality)
- Themeable variables relating to: gradient border, punch through effect, ~menu background transparency and blur~ (themeable variables for transparency and blur have been added).

### Deployment Link
https://terra-core-deployed-pr-978.herokuapp.com/static/#/site/select

### Additional Details
If you have anything else that you think may be relevant to this issue, list it here. Additional information can help us better understand your changes and speed up the review process.

@cerner/terra


packages/terra-form-select/translations/ar.json
```diff
@@ -0,0 +1,3 @@
+{
+  "Terra.form.select.defaultDisplay": "- Select -"
```
jmsv6d: Still waiting on arabic translation.

packages/terra-form-select/docs/MigrationGuide.md
```diff
@@ -0,0 +1,87 @@
+# Migrating from terra-form
```
emilyrohrbough: Were we including the Migration Guide in the docs section or release notes?

(just realized I need to write this too...)

packages/terra-form-select/docs/MigrationGuide.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Getting Started
+
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-form-select`
+
+
+## Props Changes
+
+Terra Forms Select has switched from using props to supply options to using an Options component that can be supplied to the select as children.
```
emilyrohrbough: For clarification, is this terra-form's Select or terra-form-select

packages/terra-select/docs/README.md
```diff
@@ -0,0 +1,34 @@
+# Terra Select
+
+The select component is a form input with a drop down list of options that allows for single selection.
+
+## Getting Started
+
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-select`
```
msaikaushik1992: Can't we also do this with 'yarn add terra-select' ?

packages/terra-select/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "terra-base": "^1.0.0",
```
msaikaushik1992: The terra-base version should be "^2.6.1"

packages/terra-form-select/src/SelectOption.scss
```diff
@@ -0,0 +1,38 @@
+:local {
+  .option {
```
StephenEsser: These options have a bulleted disc, is this intentional?

image

dkasper-was-taken:
We might need to pass in the props to allow this to function in a modal:
```
releaseFocus: PropTypes.func,
requestFocus: PropTypes.func,
```

packages/terra-form-select/src/Select.jsx
```diff
@@ -0,0 +1,285 @@
...
+          }
+        }
+        nextFocus.focus();
+      }
+    };
+  }
+
+  handleSelection(event, option) {
+    if (this.props.onChange && option.props.value !== this.state.value) {
+      this.props.onChange(event, option.props.value);
```
StephenEsser: There was a short discussion a while back about flipping the order of these props: https://github.com/cerner/terra-core/pull/768#discussion_r143198019

With this being a new package it might be worth discussing if we want to make this change or continue exposing the event as the first param

dkasper-was-taken:
On site page, looks like we are missing arrow key accessibility compared to the previous browser select.
It appears that when focused, left and right cycle the selected option without opening, with up and down opening the dropdown and changing the selection.

packages/terra-form-select/src/Select.jsx
```diff
@@ -0,0 +1,285 @@
...
+   * Indicates if the select should be disabled.
+   */
+  disabled: PropTypes.bool,
+};
+
+const defaultProps = {
+  onChange: undefined,
+  name: null,
+  required: false,
+  defaultValue: undefined,
```
kschuste: do these need to be defaulted to undefined?

packages/terra-site/package.json
```diff
@@ -50,6 +50,7 @@
     "terra-dynamic-grid": "^1.6.1",
     "terra-embedded-content-consumer": "^1.6.1",
     "terra-form": "^1.12.0",
+    "terra-form-select": "^0.x",
```
kschuste: should this be ^0.0.0?

jmsv6d: This is usually how we reference components before the v1 release.

packages/terra-form-select/src/Select.jsx
jmsv6d: I don't think so, I hadn't updated these when it was working on the select in the form package so it just got copied over. I can remove.

packages/terra-form-select/src/Select.jsx
```diff
@@ -0,0 +1,285 @@
...
+          isOpen={this.state.isOpen}
+          targetRef={this.getTargetRef}
+          classNameContent={cx('select-menu')}
+        >
+          <SelectMenu>
+            {defaultOption}
+            {clonedChildren}
+          </SelectMenu>
+        </Popup>
+        {/* eslint-disable jsx-a11y/no-static-element-interactions */}
```
yuderekyu: Do we have a better alternative for attaching on click handlers to some interactive html element, instead of non interactive containers to satisfy eslint? I'm running into this same issue with the control components being wrapped in container elements.

emilyrohrbough: I workaround to possibly quiting the eslint could be creating an object that contained the onClick and onkeyDown like 

```cont callbacks = { onClick: () => {}, onKeyDown: ()=>{} }``` 
and then spreading them on the div like 
```<div {...callbacks}``` 
I remember @dkasper-was-taken mentioning this before.

Comment (jmsv6d) [reply packages/terra-form-select/src/Select.jsx]: Disabling this eslint rule here is only needed because there is a bug in the version of eslint that we're using. Adding an appropriate role to the element is supposed to satisfy the rule, but it doesn't pick it up. https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/272

I really don't think that we should be using work arounds to trick eslint. If we don't think the rule is appropriate we should really just turn off the rule.

File: packages/terra-form-select/src/Select.jsx (hunk 7 above)
Comment (jmsv6d) [reply packages/terra-form-select/src/Select.jsx]: We talked about this in scrum this morning and there was a consensus that we should continue with our current pattern of passing the event as the first argument.

File: packages/terra-form-select/src/Select.jsx [hunk 11]
```diff
@@ -0,0 +1,297 @@
...
+   */
+  defaultValue: PropTypes.string,
+
+  /**
+   * The value of the select element. Use this to create a controlled input.
+   */
+  value: PropTypes.string,
+
+  /**
+   * Select options
```
emilyrohrbough: nit: missing `.`

packages/terra-form-select/src/Select.jsx
```diff
@@ -0,0 +1,297 @@
...
+   */
+  releaseFocus: PropTypes.func,
+
+  /**
+   * A callback function to request focus from the containing component (e.g. modal).
+   */
+  requestFocus: PropTypes.func,
+};
+
+const defaultProps = {
```
emilyrohrbough: Do we need to default onChange, requestFocus, and releaseFocus to undefined?

packages/terra-form-select/src/Select.jsx
```diff
@@ -0,0 +1,297 @@
...
+      isOpen: false,
+    };
+  }
+
+  getInitialState() {
+    if (!this.props.value) {
+      const childArray = React.Children.toArray(this.props.children);
+      for (let i = 0; i < childArray.length; i += 1) {
+        const child = childArray[i];
+        if (child.props.isSelected || child.props.value === this.props.defaultValue) {
```
emilyrohrbough: It looks like we honor the the `isSelected` prop over the `value` prop over. Would this be the expected behavior?

packages/terra-form-select/src/Select.jsx
```diff
@@ -0,0 +1,297 @@
...
+  getTargetRef() {
+    return this.targetRef;
+  }
+
+  wrapOnClick(option) {
+    return (event) => {
+      this.handleSelection(event, option);
+
+      if (option.props.onClick) {
+        option.props.onClick(event);
```
emilyrohrbough: It seems like returning the option in the onClick would be valuable.

packages/terra-form-select/src/Select.jsx
```diff
@@ -0,0 +1,297 @@
...
+      }
+    };
+  }
+
+  wrapOnKeyDown(option) {
+    return (event) => {
+      if (option.props.onKeyDown) {
```
emilyrohrbough: Should we handle selection first before calling the onKeyDown event?

packages/terra-form-select/src/Select.jsx
```diff
@@ -0,0 +1,297 @@
...
+    };
+  }
+
+  handleSelection(event, option) {
+    if (this.props.onChange && option.props.value !== this.state.value) {
+      this.props.onChange(event, option.props.value);
+    }
+
+    let newValue = null;
+    if (!this.props.value) {
```
emilyrohrbough: Whats the logic on checking the Select's value? I'm probably missing it, but it seems this prop would remain consistent while selection would be occurring.

packages/terra-form-select/src/Select.jsx
```diff
@@ -0,0 +1,297 @@
...
+    attributes['aria-disabled'] = disabled;
+
+    let selectedValue;
+    let display;
+    let defaultOption;
+    let isDefaultDisplay = false;
+    const clonedChildren = React.Children.map(children, (child) => {
+      let isSelected = false;
+
+      if (child.props.value === value || child.props.value === this.state.selectedValue) {
```
emilyrohrbough: Could we add an inline comment saying something like 'check if the option is the selected option'. It took me a second to realize why this wasn't a separate function to wrap the children and that this helped determine if the defaultDisplay should be created.

packages/terra-form-select/tests/jest/SelectOption.test.jsx
```diff
@@ -0,0 +1,32 @@
...
+  expect(wrapper).toMatchSnapshot();
+});
+
+it('should render a disabled Option', () => {
+  const option = <Option value="a" display="Apple" disabled />;
+  const wrapper = shallow(option);
+  expect(wrapper).toMatchSnapshot();
+});
+
+it('should render an Option with children', () => {
```
emilyrohrbough: Maybe add a test that has children & a display to confirm only display is rendered as described in the children description?

packages/terra-form-select/tests/nightwatch/select/SelectTests.jsx
```diff
@@ -0,0 +1,18 @@
...
+
+const SelectTests = () => (
+  <div>
+    <ul>
+      <li><Link to="/tests/select-tests/default">Select - Default</Link></li>
+      <li><Link to="/tests/select-tests/disabled">Select - Disabled</Link></li>
+      <li><Link to="/tests/select-tests/invalid">Select - Invalid</Link></li>
+      <li><Link to="/tests/select-tests/long-text">Select - Long Text</Link></li>
+      <li><Link to="/tests/select-tests/controlled">Select - Controlled</Link></li>
+    </ul>
```
emilyrohrbough: It might be good to add a test that has a preselected option for a controlled select.

emilyrohrbough: Also a test with select within a modal to verify onRequestFocust/onReleaseFocus

packages/terra-form-select/tests/nightwatch/select/select-spec.js
```diff
@@ -0,0 +1,63 @@
...
+  },
+  'Displays select with truncated text when appropriate': (browser) => {
+    browser.url(`${browser.launchUrl}/#/tests/select-tests/long-text`);
+    browser.expect.element('#longTextSelect').text.to.contain('Lorem ipsum');
+  },
+  'Displays invalid select': (browser) => {
+    browser.url(`${browser.launchUrl}/#/tests/select-tests/invalid`);
+    browser.expect.element('#invalidSelect').to.have.attribute('class').which.contains('invalid');
+  },
+});
```
emilyrohrbough: Add a test step to verify options within a menu are disabled correctly.

packages/terra-form-select/src/Select.jsx
jmsv6d: Not really. I prefer not to set a defaultProp if it's going to be undefined since that's what it would be anyway if the consumer doesn't pass a value.

packages/terra-form-select/src/Select.jsx
jmsv6d: I'm checking the `value` prop at the top of the the function. If the `value` prop is set I'm not setting state at all since it would then be a controlled component. I didn't really have an opinion of honoring one over the other between `isSelected` on the child and `defaultValue` on the parent since they were both serving the same purpose, but I can switch them if anyone has a strong opinion.

packages/terra-form-select/src/Select.jsx
jmsv6d: I'm not sure if I want to be passing helper info on the onClick since I would rather consumers didn't use the onClick at all, they _should_ be using onChange instead. That's why I didn't even expose onClick as a first class prop on the Option component.

packages/terra-form-select/src/Select.jsx
jmsv6d: Setting the value prop is supposed to make this a controlled component. The consumer wants be handling the selection state in that case so we shouldn't do it.

packages/terra-form-select/src/Select.jsx
emilyrohrbough: Okay I see what you mean. 

If we'd prefer consumers use OnChange and use the information returned their to perform other logic, should we even honor the onClick & onKeyDown they provided? I could see people expecting the same information the onChange would provide. 

Either way doesn't matter, just seems a bit odd to add the logic if the props aren't available & wrap our own functions for these events.

packages/terra-form-select/tests/nightwatch/select/select-spec.js
jmsv6d: Disabled options are tested here: https://github.com/cerner/terra-core/pull/978/files/6f3e94a053877dc998683ad048adac44f5681ee8#diff-9d1e674af92b70e3829dd7f9f28f5ef3R24

packages/terra-form-select/src/Select.jsx
jmsv6d: Fixed: https://github.com/cerner/terra-core/pull/978/commits/70812139c85b71e4035c0388406c73685de32e1f

packages/terra-form-select/src/Select.jsx
jmsv6d: Updated: https://github.com/cerner/terra-core/pull/978/commits/70812139c85b71e4035c0388406c73685de32e1f

packages/terra-form-select/src/Select.jsx
jmsv6d: Added comments: https://github.com/cerner/terra-core/pull/978/commits/70812139c85b71e4035c0388406c73685de32e1f

packages/terra-form-select/tests/jest/SelectOption.test.jsx
jmsv6d: Added test: https://github.com/cerner/terra-core/pull/978/commits/70812139c85b71e4035c0388406c73685de32e1f

packages/terra-form-select/tests/nightwatch/select/SelectTests.jsx
jmsv6d: The controlled select test does have a preselected option. 
Added jest tests for requestFocus and releaseFocus, since those are just passed through to the popup component and select doesn't do anything with them: https://github.com/cerner/terra-core/pull/978/commits/70812139c85b71e4035c0388406c73685de32e1f

packages/terra-form-select/src/Select.jsx
```diff
@@ -240,10 +247,19 @@ class Select extends React.Component {
       });
     });
 
+    // If there is no selected option, or none of the options match the selected value, create the default option.
     if (!selectedValue) {
       display = intl.formatMessage({ id: 'Terra.form.select.defaultDisplay' });
       selectedValue = '';
-      defaultOption = <SelectOption value={selectedValue} display={display} tabIndex="-1" isSelected className={cx('default-option')} />;
+      defaultOption = (
```
emilyrohrbough: Did you mean to drop `tabIndex="-1"`?

jmsv6d: Yeah, not allowing focus made keyboard navigation a little weird and wasn't necessary for functionality so I dropped it.

packages/terra-form-select/src/Select.jsx
jmsv6d: I could be convinced either way, so I'll see if anyone else has opinions.

mjhenkes:
The default theme is broken on safari. It’s totally transparent.

packages/terra-form-select/src/Select.scss
```diff
@@ -0,0 +1,118 @@
...
+    border-radius: 3px;
+    border-style: solid;
+    border-width: 1px;
+    cursor: pointer;
+    font-size: var(--terra-select-font-size, 1em);
+    line-height: var(--terra-select-line-height, 1.07);
+    padding: var(--terra-select-input-padding, 0.571em 0.5em);
+    transition-duration: var(--terra-select-focus-transition-duration, 0s);
+    transition-property: background-size;
+    transition-timing-function: ease;
```
JakeLaCombe: For form-textarea, we wanted to theme the timing-function. Should we do the same here? https://github.com/cerner/terra-core/blob/a6033d7e3670cea80497408167673168fd5d38f9/packages/terra-form-textarea/src/Textarea.scss#L25

packages/terra-form-select/src/Select.scss
```diff
@@ -0,0 +1,118 @@
...
+
+    &:hover {
+      border-color: var(--terra-select-input-hover-border-color, #4cb2e9);
+    }
+
+    &:focus,
+    &.open {
+      background-size: var(--terra-select-focus-animation-size, 0 0);
+      border-color: var(--terra-select-input-focus-border-color, #4cb2e9);
+      box-shadow: var(--terra-select-focus-box-shadow, 0 0 7px 4px rgba(76, 178, 233, 0.35), 0 0 1px 3px rgba(76, 178, 233, 0.5), 0 0 1px 0 #fff);
```
JakeLaCombe: Do we anticipate box-shadows that appear at places other than 0 0? If so, we should encapsulate this with directional attribute values like we have here. https://github.com/cerner/terra-core/blob/a6033d7e3670cea80497408167673168fd5d38f9/packages/terra-form-textarea/src/Textarea.scss#L57

tm4027:
+1 IE10 and Firefox

mmalaker:
One issue found: the disabled select field should not show the focused style.

JakeLaCombe:
+1 on new changes

jmsv6d:
@mmalaker The issue with the disabled focus style has been resolved.

packages/terra-form-select/src/Select.scss
```diff
@@ -0,0 +1,149 @@
...
+        /* stylelint-enable max-nesting-depth */
+      }
+
+      &:hover {
+        border-color: var(--terra-select-disabled-border-color, #dedfe0);
+      }
+    }
+
+    &:hover {
+      border-color: var(--terra-select-input-hover-border-color, #4cb2e9);
```
neilpfeiffer: minor default theme update:
`border-color: var(--terra-select-input-hover-border-color, #dedfe0);`

packages/terra-form-select/src/Select.scss
```diff
@@ -0,0 +1,149 @@
...
+  .is-disabled .select-arrow {
+    color: var(--terra-select-disabled-arrow-color, #b2b5b6);
+  }
+
+  .select-arrow {
+    align-self: center;
+    color: var(--terra-select-arrow-color, auto);
+    display: flex;
+    height: 0.5em;
+    width: 0.857em;
```
neilpfeiffer: size update for icon:
`height: 0.715em;`
`width: 1em;`

packages/terra-form-select/src/Select.scss
```diff
@@ -0,0 +1,149 @@
...
+    background-image: var(--terra-select-focus-animation-background, none);
+    background-position: 0 100%;
+    background-repeat: no-repeat;
+    background-size: var(--terra-select-focus-background-start, 0 0);
+    border-color: var(--terra-select-input-border-color, #dedfe0);
+    border-radius: 3px;
+    border-style: solid;
+    border-width: 1px;
+    cursor: pointer;
+    font-size: var(--terra-select-font-size, 1em);
```
neilpfeiffer: minor default theme update:
`var(--terra-select-font-size, 1.143rem);`

packages/terra-form-select/src/Select.scss
```diff
@@ -0,0 +1,149 @@
...
+    line-height: var(--terra-select-line-height, 1.07);
+    padding: var(--terra-select-input-padding, 0.571em 0.5em);
```
neilpfeiffer: minor default theme update:
var(--terra-select-input-padding, 0.393rem 0.429rem);

neilpfeiffer:
A few interaction issues we found:
- the disabled select could still be clicked on and show focus state (looks to be fixed already)
- the default "- Select -" should still be available to be selected from the list even after picking an option.

packages/terra-form-select/src/Select.scss
```diff
@@ -0,0 +1,149 @@
...
+
+  .display {
+    color: var(--terra-select-input-text-color, #2d3539);
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+
+  .is-disabled .display {
+    color: var(--terra-select-disabled-font-color, #b2b5b6);
```
neilpfeiffer: disabled text color update:
`color: var(--terra-form-textarea-disabled-color,rgba(0,0,0,.25));`

packages/terra-form-select/src/Select.scss
```diff
@@ -0,0 +1,149 @@
...
+    font-size: var(--terra-select-font-size, 1em);
+    line-height: var(--terra-select-line-height, 1.07);
+    padding: var(--terra-select-input-padding, 0.571em 0.5em);
+    transition-duration: var(--terra-select-focus-transition-duration, 0s);
+    transition-property: background-size;
+    transition-timing-function: var(--terra-select-focus-transition-timing, ease);
+    width: 100%;
+
+    &.is-disabled {
+      background-color: var(--terra-select-disabled-background-color, #f4f4f4);
```
neilpfeiffer: disabled background-color update:
`background-color: var(--terra-select-disabled-background-color, rgba(111,116,119, 0.5));`

packages/terra-form-select/src/Select.scss
neilpfeiffer: new correction - adjusted line-height so now padding had to adjust slightly:
`padding: var(--terra-select-input-padding, 0.3572rem 0.429rem);`

packages/terra-form-select/src/SelectOption.scss
```diff
@@ -0,0 +1,38 @@
...
+    background-color: rgba(0, 0, 0, 0);
+    color: var(--terra-select-option-font-color, rgba(0, 0, 0, 0.9));
+    cursor: pointer;
+    font-size: var(--terra-select-option-font-size, 1em);
```
neilpfeiffer: font-size update: 
`font-size: var(--terra-select-option-font-size, 1.143rem);`

packages/terra-form-select/src/SelectOption.scss
```diff
@@ -0,0 +1,38 @@
...
+    line-height: var(--terra-select-option-line-height, 1.07);
```
neilpfeiffer: `line-height: var(--terra-select-option-line-height, 1.125);`

packages/terra-form-select/src/SelectOption.scss
```diff
@@ -0,0 +1,38 @@
...
+
+      &:hover {
+        background-color: rgba(0, 0, 0, 0);
+      }
+    }
+
+    &.is-selected {
+      background-color: var(--terra-select-option-selected-color, #007cc3);
+      color: var(--terra-select-option-selected-font-color, #fff);
+      font-weight: var(--terra-select-option-selected-font-weight, bold);
```
neilpfeiffer: `font-weight: var(--terra-select-option-selected-font-weight, normal);`

neilpfeiffer:
(this should be the end of the style/theme updates)

neilpfeiffer:
Updates look great @jmsv6d - I'll get team to review and validate.

mmalaker:
Functional verification completed. Select examples display and function correctly.

neilpfeiffer:
Tested: PR #978
Resolves: Issue #880

**Reviewed:**
- Follows pattern considerations documented in Cerner Design Standards: [Drop-down List]( https://wiki.ucern.com/x/jGsVYQ)
- Follows UX defined design implementation for Terra-Form-Select

Verified by OCS initial review, Ready for Release.

**Pass** |
-----------|
UX Review|

PR #979, opened 2017-10-30
Field OCS Uplift - Phase 1

### LINKED ISSUE #889: Uplift Field Component for OCS

### LINKED ISSUE #973: Make labelText a required prop on form fields
# Issue Description
We need to make `labelText` required on all form fields. All form inputs should include a label for accessibility. Though I know UX has designs which do not display a visual label. For this case, we should add a prop, `isLabelHidden` which defaults for `false` which can be used to visually hide the label, while maintaining the label for screen readers. This is a change we need to make across the board, but for any new form components, we should be including this prop change by default.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Form fields require labels, labels can be visually hidden to match designs.

## Current Behavior
Labels are optional on form fields.

emilyrohrbough:
### Summary
Created terra-form-field npm package to address the uplift of the Field Component to OCS. This PR replaces #969. Comments on # 969 are addressed here & questions have been brought forward. 

Closes #889 & Closes #973.

#### New props include:
- `errorIcon`
- `hideRequired`
- `isInvalid`
- `showOptional`
- `isLabelHidden`

#### Added/Changed Behavior
- If component `isInvalid`, the provided error icon and error text will show. Previously the error text always showed when provided.
- If `hideRequired` prop is used, and the component `isInvalid` the required indicated will show 
- If `showOptional` prop is used, the label '(optional)' is appended to the provide label text. Translations are provided. (Fl-fi still needs to be updated)
- If `isLabelHidden` is used, the label is created on the dom for accessibility, but the label is not visible.

### Deployment Links
- https://terra-core-deployed-pr-979.herokuapp.com/static/#/site/form-field
- https://terra-core-deployed-pr-979.herokuapp.com/static/#/tests/form-field-tests


packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,80 @@
...
+      display: inline-block;
+      margin-right: var(--terra-form-field-inline-margin-right, 0.7143em);
+      vertical-align: top;
+    }
+  }
+
+  .label-group {
+    font-family: var(--terra-form-field-label-font-family, 'Helvetica Neue');
+    font-size: var(--terra-form-field-label-font-size, 14px);
+    line-height: 1.29;
```
emilyrohrbough: screen shot 2017-10-30 at 10 27 47 am

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,80 @@
...
+
+:local {
+  .field {
+    border: 0;
+    margin-bottom: var(--terra-form-field-margin-bottom, 1.428em);
+    max-width: 1020px;
+    padding: 0;
+  }
+
+  // Remove this??
```
emilyrohrbough: @neilpfeiffer 
screen shot 2017-10-30 at 10 28 12 am

packages/terra-form-field/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "terra-form-field",
+  "main": "lib/Field.js",
+  "version": "0.0.0",
+  "description": "The Form Field component handles the layout of a standard form field.",
```
StephenEsser: Nit: This seems like it's saying "a form field is a form field". Can we use a variation the second part of the description above? Something that indicates the form field handles the text / node layout of a label, helper, and error?

packages/terra-form-field/src/Field.jsx
```diff
@@ -0,0 +1,146 @@
...
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+   * Children the Field contains.
+   */
+  children: PropTypes.node,
+  /**
+   * Error message for when the input is invalid on validation.
+   */
```
StephenEsser: Nit: This sounds like the Field is doing validation, it might be more direct to say something like `Error message for when isInvalid is true`

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,84 @@
+@import '~terra-mixins';
+@import './cerner-mock-theme';
+
+:local {
+  .field {
+    border: 0;
+    margin-bottom: var(--terra-form-field-margin-bottom, 1.428em);
+    max-width: 1020px;
```
StephenEsser: This seems oddly specific, is there an animation or overflow this is enabling or is this a hard cutoff for field widths?

packages/terra-site/src/examples/form-field/FieldExamples.jsx
```diff
@@ -0,0 +1,66 @@
...
+          error="Now the required indicator is showing because the field is in error."
+          isInvalid={this.state.isInvalid}
+          isInline
+          required
+          hideRequired
+        >
+          <div style={{ border: 'dashed 1px lightGrey' }}>Control Placeholder</div>
+        </Field>
+        <hr />
+        <p>If a feild is invalid, an error icon will be displayed. <button onClick={this.toggleIsInvalid}>Toggle Invalid State</button></p>
```
StephenEsser: Typo: Field

packages/terra-form-field/tests/nightwatch/form-field-spec.js
```diff
@@ -0,0 +1,154 @@
...
+  'Displays an invalid required field with a label, help text and error text': (browser) => {
+    browser.url(`${browser.launchUrl}/#/tests/form-field-tests/combinations`);
+    browser.expect.element('#required-invalid').to.be.present;
+    browser.expect.element('#required-invalid > div[class*="label-group"] > div[class*="error-icon"]').to.be.present;
+    browser.expect.element('#required-invalid > div[class*="label-group"] > label > div[class*="required_"]').to.be.present;
+    browser.expect.element('#required-invalid > div[class*="error-text"]').to.be.present;
+    browser.expect.element('#required-invalid > div[class*="help-text"]').to.be.present;
+  },
+
+  'Label remains hidden when Feild is invalid': (browser) => {
```
StephenEsser: Typo: Field

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,84 @@
...
+
+    .required-hidden {
+      visibility: hidden;
+    }
+
+    .optional {
+      display: inline-block;
+      font-style: var(--terra-form-field-label-optional-font-style, italic);
+      padding-left: var(--terra-form-field-label-optional-padding-left, 0.357em);
+    }
```
StephenEsser: Do these classes need to be nested?

packages/terra-form-field/src/Field.scss
emilyrohrbough: This is the hard cutoff with widths. This value was 
> derived from Design Standard max of 60 characters  using default size 14pt all capital W’s.

packages/terra-form-field/src/Field.scss
emilyrohrbough: No they wouldn't.

packages/terra-form-field/docs/MigrationGuide.md
```diff
@@ -0,0 +1,76 @@
+# Migrating from terra-form's Field Component to terra-form-field
+
+The following sections describe how to migrate from the Field component provided in the terra-form package to the Field component provided in the terra-form-field package. Terra-form-field provides enhanced API and layout to begin the first phase of the OSC uplift.
+
+## Getting Started
+
+- Install with [npmjs](https://www.npmjs.com):
```
JakeLaCombe: Do you need to mention getting started twice? Looks like it's specified in your README.md doc

packages/terra-form-field/docs/MigrationGuide.md
```diff
@@ -0,0 +1,76 @@
...
+  - `npm install terra-form-field`
+
+## Prop Changes
+#### Props Added
+- `errorIcon` - icon to indicates the Field is invalid.
+- `hideRequired` - indicates a Field is required, but hides the required indicator.
+- `isInvalid` - indicates whether or not a field is valid or invalid.
```
JakeLaCombe: I think can be made a little simpler.

`indicates whether or not a field is invalid`

packages/terra-form-field/docs/MigrationGuide.md
```diff
@@ -0,0 +1,76 @@
+# Migrating from terra-form's Field Component to terra-form-field
```
JakeLaCombe: Since you included this migration guide, is it worth adding it to terra-site as well so that it's easy to see? Or at least provide a link to it?

StephenEsser: Adding on to this are we considering the original Field deprecated? Do we need to post a banner or message on the original Field doc page indicating this or will we do this at a later time? 

I imagine all development on the original Field will stop when this is released to prevent inconsistency and technical debt

JakeLaCombe:
Do we need to put a deprecation message on the field component inside terra-form?

packages/terra-form-field/src/Field.jsx
JakeLaCombe: I would go further and just say `when the input is invalid`

JakeLaCombe:
From Looking at https://github.com/cerner/terra-core/pull/982, I think we are ok to just log a new issue for the deprecation warnings, and take care of it there.

emilyrohrbough:
@JakeLaCombe  I assume deprecation would come with the deprecation of the terra-form in full.

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,84 @@
+@import '~terra-mixins';
+@import './cerner-mock-theme';
+
+:local {
```
JakeLaCombe: I'm getting this warning on my textarea code review

` text: 'Custom property ignored: not scoped to the top-level :root element `

Should this be updated here?

JakeLaCombe: I think this maybe more related to https://github.com/cerner/terra-core/issues/986 than it is this stylesheet in particular.

packages/terra-form-field/docs/README.md
```diff
@@ -0,0 +1,35 @@
+# Terra Form Field
+
+The Form Field component by handles the layout of the label, help text and error text associated with a form control Additionally, it provides required, optional, and invalid indicators.
```
JakeLaCombe: I think it would be better to say form element. control could potentially have confusion with our control elements of radio and checkbox.

JakeLaCombe:
For all comments related to deprecation, see https://github.com/cerner/terra-core/issues/984. We can take care of everything in there.

packages/terra-form-field/docs/MigrationGuide.md
```diff
@@ -0,0 +1,76 @@
...
++ import Field from 'terra-form-field';
+
+  <Field
+    label="Field Label"
+    labelAttrs={{ className: 'field-label' }}
+    error="This field is required"
+    help="This is a required field"
++   isInvalid={this.state.isInvalid}  // Needed to display error icon and error text
+    isInline
+    required
```
kschuste: this example here, I would make it take booleans here to show it matching the api

isInvalid={true}
isInline={false}
required={true}

packages/terra-form-field/docs/MigrationGuide.md
emilyrohrbough: @bjankord @mjhenkes

packages/terra-form-field/docs/MigrationGuide.md
emilyrohrbough: The default value of each of these boolean props are false, passing the name evaluates it to true. The convention with naming boolean props is to provide this shorthand prop passing. The example provided is how we'd expect one to use these props, unless they are toggling state like `isInvalid`.

packages/terra-form-field/docs/README.md
emilyrohrbough: Good call. I struggled with the term to use and chose control based on the zep naming. Also I missed the `.` between my sentences. 🙈

packages/terra-form-field/docs/MigrationGuide.md
StephenEsser: > I assume deprecation would come with the deprecation of the terra-form in full.

There's going to be an unknown period of time where we have released form packages such as `terra-form-field` but have not released every broken out package from terra-form. During this time there will be two components ( `terra-form-field` / `terra-form/lib/field` ). 

My vote is as soon as we have a component we want to encourage people to use we should to deprecate the old component.

emilyrohrbough:
Since the Migration Guide will be attached to the release notes, I removed from the PR and made it available [here](https://gist.github.com/emilyrohrbough/618c9f56765fdf028c7cb57cbeee402a).

packages/terra-form-field/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Form Field
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-form-field.svg)](https://www.npmjs.org/package/terra-form-field)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+The Form Field component by handles the layout of the label, help text and error text associated with a form element. Additionally, it provides required, optional, and invalid indicators.
```
StephenEsser: > The Form Field component by handles

I think there may be an extra or missing word here?

packages/terra-form-field/src/Field.scss
emilyrohrbough: @JakeLaCombe Talked with Neil. For now we are going to leave this as static and make it themeable later if we determine its needed.

packages/terra-form-field/src/Field.scss
emilyrohrbough: @yuderekyu Talked with Neil & we removed it. Thanks for pointing it out though! I needed the reminder.

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,75 @@
+@import '~terra-mixins';
+@import './cerner-mock-theme';
```
JakeLaCombe: I'd like to see what others think of moving this into terra-site instead of leaving it here. https://github.com/cerner/terra-core/issues/995

Note: Please don't feel obligated to resolve this right now. I see components such as terra-badge have mock theme inside their package, and I think we should stick to that solution for the moment. I just want others to think about moving this to terra-site, and when we get the themable components delivered, we can consider taking up this issue.

bjankord: +1 for moving this into terra-site as part of work for #995

packages/terra-form-field/src/Field.scss
bjankord: We should set this value in rems so it scales if the user adjusts the default font-size. With 14px being our computed root font size, the rem value for 1020px is `72.857rem`

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,75 @@
...
+:local {
+  .field {
+    border: 0;
+    margin-bottom: var(--terra-form-field-margin-bottom, 1.428em);
+    max-width: 1020px;
+    padding: 0;
+  }
+
+  .label-group {
+    font-family: var(--terra-form-field-label-font-family, 'Helvetica Neue');
```
bjankord: Why is font-family being set here? This be inherited from font-family being set on the [HTML element](https://github.com/cerner/terra-core/blob/master/packages/terra-base/src/Base.scss#L8)?

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,75 @@
...
+    font-size: var(--terra-form-field-label-font-size, 14px);
```
bjankord: We should set this value in rems so it scales if the user adjusts the default font-size. Set to `1rem`

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,75 @@
...
+
+  .optional {
+    display: inline-block;
+    font-style: var(--terra-form-field-label-optional-font-style, italic);
+    padding-left: var(--terra-form-field-label-optional-padding-left, 0.357em);
+  }
+
+  .help-text {
+    color: var(--terra-form-field-help-text-color, rgba(0, 0, 0, 0.5));
+    font-family: var(--terra-form-field-help-text-font-family, 'Helvetica Neue');
```
bjankord: Why is font-family being set here? This be inherited from font-family being set on the [HTML element](https://github.com/cerner/terra-core/blob/master/packages/terra-base/src/Base.scss#L8)?

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,75 @@
...
+    font-size: var(--terra-form-field-help-text-font-size, 12px);
+    font-weight: var(--terra-form-field-help-text-font-weight, normal);
+    line-height: 1.25;
+    margin-top: var(--terra-form-field-help-text-margin-top, 0.357em);
+    text-align: left;
+  }
+
+  .error-text {
+    color: var(--terra-form-field-error-text-color, #e50000);
+    font-family: var(--terra-form-field-error-text-font-family, 'Helvetica Neue');
```
bjankord: Why is font-family being set here? This be inherited from font-family being set on the [HTML element](https://github.com/cerner/terra-core/blob/master/packages/terra-base/src/Base.scss#L8)?

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,75 @@
...
+    font-size: var(--terra-form-field-error-text-font-size, 12px);
```
bjankord: Set to `0.857rem`

packages/terra-form-field/src/Field.jsx
```diff
@@ -0,0 +1,146 @@
...
+    label,
+    labelAttrs,
+    required,
+    showOptional,
+    ...customProps
+  } = props;
+
+  const fieldClasses = cx([
+    'field',
+    { 'field-inline': isInline },
```
bjankord: I'm not seeing a `field-inline` class defined in the SCSS file. I'm guessing that class would have styles to set field to display as an inline field. Is this something we are working on at a later phase?

packages/terra-form-field/src/Field.scss
mjhenkes: +1 from me too

packages/terra-form-field/src/Field.scss
emilyrohrbough: Ah totally missed converting these.

packages/terra-form-field/src/Field.scss
emilyrohrbough: Good point, updated to inherit.

packages/terra-form-field/src/Field.scss
emilyrohrbough: Sounds good. Updated.

packages/terra-form-field/src/Field.jsx
emilyrohrbough: Oh shoot. Originally this styling was incorporated with a [media query](https://github.com/cerner/terra-core/pull/979/commits/13b62562cf02115c4fc0a5dd2fdacbcf8aefbceb), but Neil said lets remove that until we are certain this is the behavior we wanted. I replaced the .inline styling.

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,78 @@
...
+
+  .optional {
+    display: inline-block;
+    font-style: var(--terra-form-field-label-optional-font-style, italic);
+    padding-left: var(--terra-form-field-label-optional-padding-left, 0.357em);
+  }
+
+  .help-text {
+    color: var(--terra-form-field-help-text-color, rgba(0, 0, 0, 0.5));
+    font-family: var(--terra-form-field-help-text-font-family, inherit);
```
JakeLaCombe: Should this be a themeable variable? I think the original request wanted this to use the value of --terra-base-font-family, which inherit would take care of?

emilyrohrbough: Though maybe this is an all encompassing theme for base then.

emilyrohrbough: @bjankord Did I miss-interpret your request?

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  .field-inline {
+    display: inline-block;
+    margin-right: var(--terra-form-field-inline-margin-right, 0.7143em);
+    vertical-align: top;
+  }
+
+  .label-group {
+    font-family: var(--terra-form-field-label-font-family, inherit);
```
bjankord: font-family can be removed from this instance. The label font family should be the same font family as what is set on the html element and the label-group div / label element will inherit that font through the cascade without us needing to set any styles.

packages/terra-form-field/src/Field.scss
bjankord: Reminder about updating this to `72.857rem`

packages/terra-form-field/src/Field.scss
bjankord: font-family can be removed from this instance. The help-text font family should be the same font family as what is set on the html element and the help-text div will inherit that font through the cascade without us needing to set any styles.

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,78 @@
...
+    font-size: var(--terra-form-field-help-text-font-size, 0.857rem);
+    font-weight: var(--terra-form-field-help-text-font-weight, normal);
+    line-height: 1.25;
+    margin-top: var(--terra-form-field-help-text-margin-top, 0.357em);
+    text-align: left;
+  }
+
+  .error-text {
+    color: var(--terra-form-field-error-text-color, #e50000);
+    font-family: var(--terra-form-field-error-text-font-family, inherit);
```
bjankord: font-family can be removed from this instance. The error-text font family should be the same font family as what is set on the html element and the error-text div will inherit that font through the cascade without us needing to set any styles.

JakeLaCombe: If you do have issues with the font not inheriting, you may potentially have to set `font-family: inherit`. I ran into that issue with the textarea

packages/terra-form-field/src/Field.scss
emilyrohrbough: I had the change locally & didn't have it pushed. Should be good to go now.

packages/terra-form-field/src/Field.scss
```diff
@@ -0,0 +1,76 @@
+:local {
+  .field {
+    border: 0;
+    font-family: var(--terra-form-field-font-family, inherit);
```
bjankord: I think we can remove this font-family property here too, unless UX wants the ability to make field a different font than what we set in [base](https://github.com/cerner/terra-core/pull/988/files#diff-cf7e9b185f60dd895f755538bcd63363).

emilyrohrbough: Yes the theme font-family for Field are different for both clinical & consumer

mmalaker:
Functional verification completed. Form field examples display and function correctly with OCS changes.

packages/terra-form-field/docs/MigrationGuide.md
mjhenkes: >My vote is as soon as we release a component we want to encourage people to use we should to deprecate the old component.

Agreed

neilpfeiffer:
Tested: PR #979
Resolves: Issue #889

**Reviewed:**
- Follows pattern considerations documented in Cerner Design Standards: [Form-Field](https://wiki.ucern.com/x/MFcVYQ)
- Follows UX defined design implementation for Form-Field

Verified by OCS initial review, Ready for Release.

**Pass** |
-----------|
UX Review|

Issue #964, opened 2017-10-25
Create new Flexbox container component


kschuste:
# Issue Description
Create a new container component that is specific to using flexbox to align its content.  

## Issue Type

- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Should be able to give the container a list of children and to be able to set the flexbox styles such as flex-wrap, order, and align-self as properties along with other css stylings such as margin and padding.

kschuste:
# Tech Design for Flex Container
## Summary:
This is a new component that will allow for the consumer to apply flex-box stylings to the container and drop in any number of elements into it.

## Bidirectionality
These changes should make no changes to the existing ltr/rtl functionality.

## React Props:
| Prop     | Type | Default | Description  |
|----------|------|---------|--------------|
| alignContent | bool |  | Aligns the container's lines within when there is extra space in the cross-axis.  Should be one of `center`, `flex-end`, `flex-start`, `space-around`, `space-between`, or `stretch`. |
| alignItems | string |  | Defines the default behavior for how flex items are laid out along the cross axis on the current line.  Should be  one of `baseline`, `center`, `flex-end`, `flex-start`, or `stretch`.  |
| alignSelf | string |  | Allows the default alignment to be overridden for individual flex items.  Should be  one of `baseline`, `center`, `flex-end`, `flex-start`, or `stretch`.  |
| children | node |  | Child nodes to be displayed in the container.  |
| display | string | `flex` | Enables flex context for all its direct children either inline or block.  Should be `flex` or `inline-flex`.  |
| element | string | `div` | Indicates what type of element the flex container can be.  Should be one of `article`, `aside`, `div`, `figure`, `footer`, `header`, `main`,  `nav`, or `section`.  |
| flex | string |  | Shorthand for flex-grow, flex-shrink and flex-basis combined.  Example `0 1 auto`.  |
| flexBasis | string |  | Defines the default size of an element before the remaining space is distributed.  Can be a length (20%) or a keyword (auto).  |
| flexDirection | string |  | Defines the direction flex items are placed in the container.  Should be one of `column`, `column-reverse`, `row`, or `row-reverse`.  |
| flexGrow | number |  | Defines the ability for a flex item to grow if necessary. It accepts a unitless value that serves as a proportion.  |
| flexShrink | number |  | Defines the ability for a flex item to shrink if necessary.  |
[...]

PR #985, opened 2017-11-01
Terra flexbox container

kschuste:
### Summary
This PR fixes issue #964.  This creates a new flexbox container element that allows users to pass in the flex-box stylings via the props.

### Additional Details
The component contains props for both parent and child flex items rather than having separate parent and child flex components.  This allows for this component to be the flex parent container and also pass it in for it's children components.


packages/terra-flex-box/src/FlexBox.jsx
```diff
@@ -0,0 +1,155 @@
...
+  maxHeight,
+  maxWidth,
+  minHeight,
+  minWidth,
+  order,
+  width,
+  ...props
+}) => {
+  const customProps = props;
+  if (!customProps.style) { customProps.style = {}; }
```
bjankord: How are vendor prefixes and different flexbox syntaxes for IE handles with this? I don't believe this will work in IE unless we account for that.

mhemesath: This goes back to my original question of what is the value proposition for this component. It seems to provide no simplification or advantage to just using flexbox html/css directly

mhemesath: This goes back to my original question of what is the value proposition for this component. It seems to provide no simplification or advantage to just using flexbox html/css directly

mhemesath: This goes back to my original question of what is the value proposition for this component. It seems to provide no simplification or advantage to just using flexbox html/css directly

bjankord: I see the value proposition as reducing duplicative styles. This component would reduce CSS needed every time we need CSS for flex properties with known values like display, align-content, align-items. We could write that CSS once and not need to repeat it. For properties which do not have a known set of values and can contain any value like width and height, theres no real trival way to prevent duplication there unless we provide a scale of "approved" values, but that seems limiting.

I do have concerns that this API may actually be too simple. My concern comes from working with flexbox and coming across issues where, to resolve a flexbox bug, you need to add additional styles, maybe not necessarily flexbox related. The [flexbug repo](https://github.com/philipwalton/flexbugs) has a list of various bugs where having the ability to add additional styles beyond flexbox based ones is required to achieve cross-browser functionality.

So I think in addition to providing vendor prefixes and different flexbox syntaxes for IE handled, we'd want this component to abstract common [flexbox bugs](https://github.com/philipwalton/flexbugs) as an implementation detail.

Whats the use case for this @mjhenkes @kschuste? My thought was this was something @mjhenkes and @mhemesath wanted.

mhemesath: I had originally wanted one but upon seeing the tech design I'd rather prefer we create a subset of purposed "grid" components that solve reusable scenarios we encounter with flexbox.

packages/terra-flex-box/src/FlexBox.jsx
```diff
@@ -0,0 +1,155 @@
...
+  if (flexShrink !== undefined) { customProps.style.flexShrink = flexShrink; }
+  if (flexWrap) { customProps.style.flexWrap = flexWrap; }
+  if (height) { customProps.style.height = height; }
+  if (justifyContent) { customProps.style.justifyContent = justifyContent; }
+  if (maxHeight) { customProps.style.maxHeight = maxHeight; }
+  if (maxWidth) { customProps.style.maxWidth = maxWidth; }
+  if (minHeight) { customProps.style.minHeight = minHeight; }
+  if (minWidth) { customProps.style.minWidth = minWidth; }
+  if (order !== undefined) { customProps.style.order = order; }
+  if (width) { customProps.style.width = width; }
```
mhemesath: FWIW this can all be cleaned up if you use [aphrodite](https://github.com/Khan/aphrodite). Its the library that generates the CSS for dynamic grid.

packages/terra-flex-box/src/FlexBox.jsx
mjhenkes: I'm looking for some basic tools for building simple layouts without having to duplicate the flexbox css and it's gotchas. I also like the idea of non devs (maybe) being able to use this in Kaiju, but if the API is to complex that may not work out.

mhemesath: >  without having to duplicate the flexbox css

But now we've just pushed the problem up where instead of duplicating HTML/CSS we'll now duplicate implementations of this component.

Also, if we have duplications I'd much rather see us identify those and build dedicated components to solve those use cases.

packages/terra-flex-box/src/FlexBox.jsx
kschuste: updated to use aphrodite and cleaned up the code a bit more

bjankord:
Discussed in scrum this morning w/ @mjhenkes and @kschuste on holding off on creating this component.

PR #988, opened 2017-11-02
Adjust chevron positioning

dkasper-was-taken:
### Summary
Correct chevron positioning with new arrange changes.

@cerner/terra


packages/terra-list/src/List.scss
```diff
@@ -27,11 +27,14 @@
   // ========================================================
 
   .chevron {
+    display: flex;
     padding-left: 10px;
     padding-right: 10px;
 
     > svg {
```
emilyrohrbough: previously would this scale with font-size?

packages/terra-list/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Chevron positioning
```
bjankord: More details on this fix would be helpful if we every need to modify this in the future. Like what browsers it was an issue in, etc.

kschuste: could include that you applied it's using flexbox now as well

mmalaker:
Need more detail to functionally verify. Where was the positioning problem occuring? What components are affected?

StephenEsser:
@mmalaker 

There may be more areas to verify, the one area I know of that was an issue was the Single Select Lists chevron position that can be seen on the doc site. 

Before Change:
http://engineering.cerner.com/terra-core/#/site/list/single-select

After Change:
https://terra-core-deployed-pr-988.herokuapp.com/static/#/site/list/single-select

mmalaker:
Functional verification completed. The examples with chevron icons now display vertically centered.

bjankord:
This has been released in terra-list 1.16.0

PR #990, opened 2017-11-02
Fix issues with rtl plugin and css-custom-properties

### LINKED ISSUE #986: Theming approach is having issues with directional styles
# Issue Description
To implement our theming, we use a combination of the rtl mixin, and the postcss-custom-properties mixin. When both of these are used, the following warning is generated.

     text: 'Custom property ignored: not scoped to the top-level :root element ([dir=ltr]:root { ... --terra-form-field-inline-margin-right: ... })',

This is preventing themeable stylesheets from being used in IE10, 11 and other browsers that do not support css custom variables.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior
Themeable styles work correctly with dir styles

## Current Behavior
Dir styles are causing issues with the application.

JakeLaCombe:
### Summary
Resolves #986 

The rtl and postcss-custom-properties are conflicting with each other, and causing issues with producing the correct styles. To resolve this, we run rtl before we run postcss-custom-properties.

### Additional Details
This is the first part of the long term solution. We will need to eventually create better mixins for left and right functionality, as styles such as box-shadow, and background-position causes issue. If you need these styles themed, you need to wrap them individually in your components with the proper [dir=ltr] & [dir=rtl] attribute selectors.

### DIFF
--- a/packages/terra-site/webpack.config.js
+++ b/packages/terra-site/webpack.config.js
@@ -5,6 +5,7 @@ const webpack = require('webpack');
 const postCssConfig = require('./postcss.config');
 const PostCSSAssetsPlugin = require('postcss-assets-webpack-plugin');
 const PostCSSCustomProperties = require('postcss-custom-properties');
+const rtl = require('postcss-rtl');
 const path = require('path');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
@@ -70,6 +71,7 @@ module.exports = {
       log: false,
       plugins: [
         PostCSSCustomProperties({ preserve: true }),
+        rtl(),
       ],
     }),
     new webpack.NamedChunksPlugin(),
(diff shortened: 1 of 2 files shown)


packages/terra-site/webpack.config.js
```diff
@@ -70,6 +71,7 @@ module.exports = {
       log: false,
       plugins: [
         PostCSSCustomProperties({ preserve: true }),
+        rtl(),
```
mhemesath: id' rather we move PostCSSCustomProperties into `postcss.config.js` if possible

mhemesath: Thanks for ignoring my statement. The reasoning behind a consolidated postcss config is to make it reusable for storybook in the future as well as easier for consumers to see/reuse that config.

mhemesath: I apologize, I realized this was merged before my comment. I do suggest we consolidate/simplify the config

PR #991, opened 2017-11-03
terra-signature - Fixes for removing listeners and lastLineSegments callback param

vikonava:
### Summary
Two fixes were included in this PR:
* Listeners were attached to document.body on mount, while the listeners were being removed from window on unmount; which caused that the listener wasn't really removed and it stayed even if the component didn't existed.
* onChange callback last param was returning the last segment instead of the last line segments. Created a variable for that to maintain the last line segment state and return it onChange.


packages/terra-signature/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Removing event listeners from the right element during component unmount
+* Fix onChange, it was returning the last segment instead of the array of the whole last line segments
```
StephenEsser: I don't know how difficult this would be to test, but it feels like this change should have broken one or had a test written to make sure the API stays passive

StephenEsser: This would technically be a non-passive change if it wasn't considered a bug

Issue #997, opened 2017-11-06
Terra Button has incorrect font-family style


JakeLaCombe:
# Issue Description
Terra button does not have the correct styles, and is instead hard coded into the css file as helvetica. This is limiting the abilities of teams to change the text display when their themes call for a different base font.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
terra-button should have the style `font-family: inherit`

## Current Behavior
terra-button hard codes the font to be "Helvetica"

PR #1001, opened 2017-11-08
Fix font-family defect in terra-button

dylan-sftwr:
### Summary
Prior to this change, the terra-button component had it’s own font-family set explicitly within its stylesheet. Now, the font-family being applied to terra-button can be set using the ‘--terra-base-font-family’ CSS custom prop. Fixes issue #997.

### Additional Details
This fix is needed for the new Healthe@Cerner site themed-buttons.

### DIFF
--- a/packages/terra-button/src/Button.scss
+++ b/packages/terra-button/src/Button.scss
@@ -33,7 +33,6 @@
     border-width: var(--terra-button-border-width, 1px);
     cursor: pointer;
     display: inline-block;
-    font-family: inherit;
     font-size: map-get($terra-button-font-sizes, medium);
     font-weight: var(--terra-button-font-weight, 400);
     line-height: 1.429; // 20px when html font size is computed to 14px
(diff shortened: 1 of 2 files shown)


packages/terra-button/src/Button.scss
```diff
@@ -33,7 +33,7 @@
     border-width: var(--terra-button-border-width, 1px);
     cursor: pointer;
     display: inline-block;
-    font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
+    font-family: inherit;
```
mhemesath: could probably flat out delete the property...

JakeLaCombe: Try that out in see. I know with textarea is was inheriting from system-ui, but if that doesn't happen with button, I'd remove this style.

dylan-sftwr:
@mhemesath you were right, the rule isn't needed.

Issue #952, opened 2017-10-20
Search Field: autosearch & minLength change


neilpfeiffer:
# Related Issues:
[#943 - Search Field: clear field](https://github.com/cerner/terra-core/issues/943)
[#953 - Search Field: disabled state & fits container](https://github.com/cerner/terra-core/issues/953)

----

# Issue Description
New features to add to the search field component:
1. Add keyboard interaction - listen for `enter` key to execute search when input has tab focus.
2. add a new prop to disable autosearch, so that search is only on click/enter synchronous execution.

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
- items listed above

## Current Behavior
- default search is executed at 250ms and only after 2 minimum characters

neilpfeiffer:
cc: @Singleton06 @bjankord @mjhenkes

Singleton06:
There are a couple of ambiguities that I would like cleared up before creating a tech design.
* Will the enter key to search always be enabled or are we thinking that it will be configurable so that consumers can choose not to have it if they want?  Will this be on by default or no?
* Could you clarify a bit more what you mean by indicating that we need to make an adjustment to the `searchDelay` and `minimumSearchTextLength` such that there is no default? Are you indicating that this is only in the situation where enter is able to be used?  Are we talking about literally just changing the default values? Our particular use case was just to be able to disable these.  This could be by simply providing a flag that turns of auto-searching.

mjhenkes:
- Enter to search should always be available. 
- Removing the default value for` searchDelay` and `minimumSearchTextLength` would be a non passive change and should be avoided. To disable it there will have to be an additional prop passed in.

Singleton06:
# Tech design for the search field disabling auto search and enter to search
## Summary:
Upgrade the search field control to be able to press enter to force a search at any time (without waiting). Additionally, for some consumers, it is desired to disable the auto-search functionality.

## Requirements Added
* Allow the enter key to trigger search, either valid (calling the onSearch prop) or invalid (calling the onInvalidSearch function).
  * The search is considered valid if it meets the minimumSearchTextLength and onSearch has been specified
  * The search is considered invalid if it does not meet the minimumSearchTextLength and onInvalidSearch has been specified.


## Responsiveness
No change should occur to the responsiveness.

## Bidirectionality
No change should occur to the bidirectionality.

## Accessibility
This might help improve accessibility because the enter button can be used to search.

## React Props
[...]

PR #1004, opened 2017-11-09
Search field disable autoSearching, use ENTER keypress to trigger search

kschuste:
### Summary
This fixes issue #952.
Adding the ability to disable the auto-search functionality via a prop and allow for the searching to be triggered via the ENTER keypress.

### Additional Details
Added a new disableAutoSearch prop that is defaulted to false.  Handled onKeyDown event to add the logic for the ENTER keypress.


packages/terra-search-field/tests/nightwatch/search-field-spec.js
```diff
@@ -74,4 +74,38 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
...
+      .expect.element('#search-callback-text').text.to.equal('').after(250);
+    browser
+      .setValue('input[type=search]', 'e')
+      .expect.element('#search-callback-text').text.to.equal('').after(250);
+    browser
+      .setValue('input[type=search]', 'a')
+      .expect.element('#search-callback-text').text.to.equal('').after(250);
+  },
+
+  'Displays the search field and handles ENTER keypress to search ': (browser) => {
```
emilyrohrbough: Could we add a test that searches on Enter with autoSearch enabled?

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -63,11 +74,17 @@ class SearchField extends React.Component {
   handleTextChange(event) {
     this.setState({ searchText: event.target.value });
 
-    if (!this.searchTimeout) {
+    if (!this.searchTimeout && !this.props.disableAutoSearch) {
       this.searchTimeout = setTimeout(this.handleSearch, this.props.searchDelay);
     }
   }
 
+  handleKeyDown(event) {
```
emilyrohrbough: I feel like this logic should move into the handleTextChange function.  and the searchTimeout logic should be broken into a resetSearchTimeout function.

emilyrohrbough: That being if we keep this,  handleKeyDown should be updated to something a bit more descriptive, like handleSearch

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -63,11 +74,17 @@ class SearchField extends React.Component {
...
+    if (event.nativeEvent.keyCode === KEYCODES.ENTER) {
```
benbcai: Is there any difference between event.nativeEvent.keyCode and event.keyCode? I see that some components are using event.nativeEvent.keyCode while others are using event.keyCode

yuderekyu: `event.nativeEvent` references the browser's native event, while `event` is a React cross browser synthetic event. Seems that React prefers synthetic events to standardize handling events across browsers.

packages/terra-search-field/src/SearchField.jsx
kschuste: handleTextChange is only used for when the value inside the search field actually does change, handling the ENTER key does not trigger the onChange event as the value stays the same.  Therefore, added this function to handle the keyPress of ENTER.
We already have a handleSearch function that both handleKeyDown and handleTextChange both call so I think handleKeyDown is descriptive and makes the most sense in that it is handing the down press of the ENTER key

PaulSolDev:

neilpfeiffer:
+1 : Marking as Verified, ready for Release.

Issue #998, opened 2017-11-07
Enable i18n plugin to check translation files in custom path


zhongyn:
# Issue Description
Current i18n plugin only [checks node_modules for translation files](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js#L47) and aggregate them into one folder, which makes the usage of plugin very restricted. When introducing storybook into terra-consumer, we want to customize the search path so that it can load translations from the packages source folder directly instead of npm-installing the package and loading it from node_modules.

Assume the project structure looks like:
- root
  - storybook
     - webpack.config.js
  - packages

New plugin config in webpack.config.js will looks like:
```
  plugins: [
    new I18nAggregatorPlugin({
      baseDirectory: path.resolve(__dirname, '../'),
      translationsDirectoryRouters: ['packages'],
      supportedLocales: i18nSupportedLocales,
    }),
  ],
```
With the above config, i18n will search both the packages folder and root node_modules. `translationsDirectoryRouters` is optional. The default is node_modules.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

PR #1005, opened 2017-11-09
add custom dir option for searching and aggregating translation messages

zhongyn:
### Summary
This is an enhancement for the issue: https://github.com/cerner/terra-core/issues/998

It allows an optional i18n plugin option `customTransDir`, e.g.
```
plugins: [
    new I18nAggregatorPlugin({
      baseDirectory: path.resolve(__dirname, '../'),
      translationsDirectory: path.resolve(__dirname, '../packages'),
      supportedLocales: i18nSupportedLocales,
    }),
],
```

It will search and aggregate translations under any descendant dir of `customTransDir` and the default node_modules.

@cerner/terra


packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -23,9 +23,16 @@ function getDirectories(srcPath, inputFileSystem) {
...
 
-function aggregateDirectory(languageMessages, currentDirectory, inputFileSystem) {
-  // Check the directory for translations
-  const translationsDirectory = path.resolve(currentDirectory, 'translations');
+// Find all descendant directories by name under a root directory and return an array
+function findDirectories(dirName, rootDir, inputFileSystem) {
+  const result = [];
+  const subDirs = getDirectories(rootDir, inputFileSystem);
+  result.push(subDirs.filter(dir => dir === dirName).map(dir => path.join(rootDir, dir)));
+  subDirs.forEach(dir => result.push(findDirectories(dirName, path.join(rootDir, dir), inputFileSystem)));
```
bjankord: Using findDirectories within the findDirectories function seems odd to me. I think adding some more doc around what all this is doing would make it more clear and easier to maintain in the future.

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -42,7 +49,19 @@ function aggregateDirectory(languageMessages, currentDirectory, inputFileSystem)
   } catch (e) {
     // not outputting anything here as the catching of the directory not existing is not an error in this case
   }
+}
 
+// Aggregate all translation messages under a custom directory
+function aggregateCustomDirectory(languageMessages, rootDir, inputFileSystem) {
```
emilyrohrbough: This function is written quite differently than our current setup. I'd prefer it follows the aggregation setup we have for consistency. Additionally, this seems to assume we would provided the root dir of the translations directory, but I would assume the translation directory path would be the path provided.

 I would expect something like
```
function aggregateCustomDirectory(languageMessages, customTranslationsDirectory, inputFileSystem) {
  // Check the directory for translations
  readTranslations(languageMessages, customTranslationsDirectory, inputFileSystem);   
  // Check sub directories for translations
    try {
    getDirectories(customTranslationsDirectory, inputFileSystem).forEach((subDirectory) => {
      aggregateDirectory(languageMessages, path.resolve(customTranslationsDirectory, subDirectory), inputFileSystem);
    });
  } catch (e) {
    // not outputting anything here as the catching of the directories not existing is not an error in this case
  }

  return languageMessages;
}
```

emilyrohrbough: This makes me think we could possibly just reuse `aggregateDirectory` and update [this line]( https://github.com/cerner/terra-core/blob/6b248786e32c373ff25e03fba693855d2fb1ef7f/packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js#L66) to check if is customDir

emilyrohrbough:
IMO the term `customTransDir` is sorta confusing. We should be spelling it out to be consistent but I would think `translationsDirectory` or something similar would be fine.

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
zhongyn: > This function is written quite differently than our current setup. I'd prefer it follows the aggregation setup we have for consistency. Additionally, this seems to assume we would provided the root dir of the translations directory, but I would assume the translation directory path would be the path provided.

This function does two things:
- Find all paths to translation folders under the custom directory. Updated the param `rootDir` to `customTranslationsDirectory` to avoid ambiguous. 
- Iterate the array of target paths and read translation messages from them

The current setup reads the translation messages when searching the target folder recursively. I was trying to reuse the same setup. But the searching pattern is a bit different:
- The current algorithm assumes the searching path follows a specific directory structure, for example:
   - node_modules
      - packages 1
        - node_modules
        - folder 11
        - folder 12
        - translations
      - packages 2
        - node_modules
        - folder 21
        - folder 22
        - translations


 The search paths include:
  - node_modules/packages1
  - node_modules/packages1/node_modules...
  - node_modules/packages2
  - node_modules/packages2/node_modules...

It is very efficient since it doesn't go into folder 11/12/21/22. 

For the custom directory searching, it will look for all possible paths, e.g.:
   - customDir
      - folder 1
        - folder 11
        - folder 12
        - translations
      - folder 2
        - folder 21
        - folder 22
        - translations
     - translations

 The search paths include:
  - customDir
  - customDir/folder 1/
  - customDir/folder 1/folder 11
  - customDir/folder 1/folder 12
  - customDir/folder 1/translations
  - customDir/folder 2/
  - customDir/folder 2/folder 21
  - customDir/folder 2/folder 22
  - customDir/folder 2/translations
  - customDir/translations

I prefer to decouple the path searching and files reading, which make it easy to follow the code.

zhongyn:
@emilyrohrbough updated the option name `customTransDir` and function param `rootDir` to `translationsDirectory` to make it more readable

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
zhongyn: Updated with more docs

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
zhongyn: Also since this option is an optional, decoupling it from node_modules searching would preserve the default performance if the custom directory is not specified

emilyrohrbough: With the suggestions I made, I would expect this structure to be handled by adding an additional bool param to be passed to aggregateDirectory() to indicate `searchNodeModules`. 
Then I'd expect this this line:

```
const nodeMoudlesPath = path.resolve(currentDirectory, 'node_modules');
```

to be updated to 

```js
const nodeMoudlesPath = searchNodeModules ? path.resolve(currentDirectory, 'node_modules') : path.resolve(currentDirectory);
```

This should accomplish the folder structor you have indicated.

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -44,10 +45,10 @@ function aggregateDirectory(languageMessages, currentDirectory, inputFileSystem)
   }
 
   // Check the directory's node_modules for translation files
```
emilyrohrbough: Update inline comment

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -70,8 +71,13 @@ function aggregateTranslationMessages(options, inputFileSystem) {
   let languageMessages = {};
   supportedLocales.forEach((language) => { languageMessages[language] = {}; });
 
-  // Aggregate translation messages for the directory
-  languageMessages = aggregateDirectory(languageMessages, options.baseDirectory, inputFileSystem);
+  // Aggregate translation messages for node_modules
```
emilyrohrbough: Nit: I know its a bit cleaner to just pass true & false to aggregrateDirectory, but it might improve readability to add a variable `searchNodeModules` so it indicates why true & false are passed

packages/terra-i18n-plugin/docs/README.md
```diff
@@ -32,3 +32,12 @@ module.exports = {
 ```
 
 Then all of the messages for a given language across all dependencies will be aggregated into an overall translation file for that language and placed into an aggregated-translations directory.  In order for these resulting files to be able to be loaded by terra-i18n, that directory then needs to be added to the list of directories that the webpack resolver looks into.
+
+By default it will search translation folders in the root node_modules. If an optional option `translationsDirectory` is specified:
```
emilyrohrbough: ....in the root node_modules. _The i18n-plugin will also search and aggregate translation files under a custom file path when the optional key `translationsDirectory` is specified:_

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
zhongyn: +1

packages/terra-i18n-plugin/docs/README.md
```diff
@@ -32,3 +32,10 @@ module.exports = {
 ```
 
 Then all of the messages for a given language across all dependencies will be aggregated into an overall translation file for that language and placed into an aggregated-translations directory.  In order for these resulting files to be able to be loaded by terra-i18n, that directory then needs to be added to the list of directories that the webpack resolver looks into.
+
+By default it will search translation folders in the root node_modules. The i18n-plugin will also search and aggregate translation files under a custom file path when the optional key `translationsDirectory` is specified:
+```js
+new I18nAggregatorPlugin({
+  translationsDirectory: 'customPath',
```
mjhenkes: What if we made this an array ( `translationsDirectories` ) that defaulted to `['node_modules']` but could be overridden to whatever you wanted `['customPath', 'node_modules']`. I think this could clean up some of the messy 'searchNodeModules' code below.

zhongyn: That would be a non-passive change. But I am open to it since it aligns with some common webpack config format.

mjhenkes: We could keep it passive by assuming `translationsDirectories = ['node_modules']` if translationsDirectories are not provided. Am I missing something?

emilyrohrbough: @mjhenkes That's a good idea, then a consumer could provide multiple paths for 'custom' translation directories

zhongyn: @mjhenkes the reason for adding `searchNodeModules` is that there are two different search strategies. For node_modules, it only searches the packages under node_modules, and then search the node_modules under each package. For customPath, it searches every possible path under it. So it has a larger scope. Is your idea of considering node_modules as a default customPath?

emilyrohrbough: @zhongyn utilizing this approach we'd have the ability to indicate more than one search pathern. For example say we specify 
```['node_modules', 'custom_translations', 'verified_translations']```
this would allow for searches for any `node_modules` folders, `custom_translations` folders and `verified_tranlsations `folders. This approach would allow for multi translation directories patterns to be search for and aggregated to a single file.

Comment (zhongyn) [reply packages/terra-i18n-plugin/docs/README.md]: updated config to support multi custom translation dirs

**Comment by bjankord** [comment]
This is currently erroring out on TravisCI with the following error message:

`FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory`

**Comment by zhongyn** [comment]
@emilyrohrbough @mjhenkes there is a performance issue when searching node_modules exhaustively as a custom path. It ran out of memory when building terra-site with i18n-plugin. See the travis-ci build for more details. The existing search algorithm looks like:
- baseDirectory
  - translations
  - node_modules
      - packages 1
        - node_modules
        - folder 1
        - folder 2
        - translations
      - packages 2
        - ...
1. check if a transltions folder under baseDirectory
2. go to node_modules and loop through the packages
3. in each package, check if a transltions folder exist
4. repeat step 2, 3

The above algorithm avoids going to folder 1,2 which reduce a lot of search work. If consider it as a custom path, it would be:

1. check if a transltions folder under baseDirectory
2. go to node_modules 
3. loop through child folders
4. in each child folder, check if a transltions folder exist
5. repeat 3, 4

So i suggest keeping the exsiting node_modules search algorithm

**Comment by emilyrohrbough** [comment]
@zhongyn I would have expected the same output at what currently exists with the custom path.

**Comment by zhongyn** [comment]
@emilyrohrbough correct, the custom path is the enry point of searching, the target folder name is still `translations`

**Comment by ryanthemanuel** [comment]
@zhongyn I think we should get together in person to make sure we're all on the same page on what is happening with this PR.  I'll set something up.

**Comment by zhongyn** [comment]
@ryanthemanuel sure, will be there on Friday

**Comment by emilyrohrbough** [comment]
Discussed offline.

Work to be done is:
1. Update `translationDirectory` to `translationsDirectories` (or something similar) that accepts an array of absolute paths to a directory containing translation files.
2. Add additional param that accepts an array of search patterns with the default of ['node_modules'] that will recursively search from the base directory for the 'pattern + /translation' path specified.

File: packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js [hunk 7]
```diff
@@ -70,13 +71,17 @@ function aggregateTranslationMessages(options, inputFileSystem) {
...
-  translationsDirectory.forEach(dir =>
-    aggregateDirectory(languageMessages, path.resolve(options.baseDirectory, dir), inputFileSystem),
+  // Read translations directly from specific paths
+  if (options.translationsDirectories) {
+    options.translationsDirectories.forEach(dir =>
+      readTranslations(languageMessages, path.resolve(options.baseDirectory, dir), inputFileSystem),
+    );
+  }
+  // Set translationsDirectoryRouters to node_modules if not specified
+  const translationsDirectoryRouters = options.translationsDirectoryRouters || ['node_modules'];
```
emilyrohrbough: So I am wondering after reading the documentation above if it would make sense to always concat 'node_modules' given this is the purpose of the plugin and it removes the responsibility of the consumer to specify it themselves. IMO consumer would expect this aggregation by default.

```
const translationsDirectoryRouters = options.translationsDirectoryRouters ?options.translationsDirectoryRouters.concat(['node_modules']) || ['node_modules'];
```

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
```diff
@@ -70,13 +71,17 @@ function aggregateTranslationMessages(options, inputFileSystem) {
...
-  readTranslations(languageMessages, options.baseDirectory, inputFileSystem);
-  // Set translationsDirectory to root node_modules if not specified
-  const translationsDirectory = options.translationsDirectory || ['node_modules'];
-  // Search and aggregate translation messages under the custom directories
-  translationsDirectory.forEach(dir =>
-    aggregateDirectory(languageMessages, path.resolve(options.baseDirectory, dir), inputFileSystem),
+  // Read translations directly from specific paths
+  if (options.translationsDirectories) {
+    options.translationsDirectories.forEach(dir =>
+      readTranslations(languageMessages, path.resolve(options.baseDirectory, dir), inputFileSystem),
```
emilyrohrbough: Since we are providing absolute paths now, would it make since to remove the baseDirectory restriction and just have 

 ```
readTranslations(languageMessages, dir, inputFileSystem),
```

File: packages/terra-i18n-plugin/CHANGELOG.md [hunk 9]
```diff
@@ -3,6 +3,7 @@ Changelog
 
 Unreleased
 ----------
+* Update to allow custom translations directory to be passed in through the options
```
emilyrohrbough: Update change log to reflect new keys and their use

packages/terra-i18n-plugin/src/I18nAggregatorPlugin.js
zhongyn: This option expects an array of paths relative to baseDirectory, e.g.
```
[folder1/folder2, ../folder3]
```
The goal is to make the config simple instead of
```
[path.resolve(__dirname, folder1/folder/2), path.resolve(__dirname, ../folder3)]
```

bjankord:
Need some more eyes on this @mhemesath, @ryanthemanuel, @mjhenkes

PR #1012, opened 2017-11-12
Update to enzyme v3

### LINKED ISSUE #871: Update Enzyme to v3.x
# Issue Description
This is the most up-to-date version, plus v3.x provides support for > React 16! 🎉  They authors have even provided a v2 to v3 [migration guide](https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md). 

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

bjankord:
### Summary
* Updated enzyme to latest stable release
* Updated enzyme-to-json to latest stable release
* Updated babel-jest to latest stable release

### Additional Details
* Replaced use of `wrapper.unrendered` with `wrapper.instance()`. 
  * `unrendered` was a [private enzyme method](https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#private-properties-and-methods-have-been-removed) which has been removed.
* Replaced use of `wrapper.node` with `wrapper.getElement()`. 
  * `node` was a [private enzyme method](https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#private-properties-and-methods-have-been-removed) which has been removed.
* Removed jest tests for Base component which hit code that used `require.ensure`.
  * With updating to enzyme v3 and babel-jest v21, code that uses `require.ensure` throws errors in testing. 
  * When searching around for this issue others have recommended shimming `require.ensure` when testing.
  * I tried adding the following code in various ways to shim the `require.ensure` calls in terra-i18n with no luck.

```
if (typeof require.ensure !== 'function') {
  require.ensure = (deps, cb) => cb(require);
}
```
* Added nightwatch tests to cover the jest test which were removed for Base
  * With removing the jest tests that hit `require.ensure`, I added tests to nightwatch which cover the same functionality that was being tested before in jest for Base component

Resolves #871


emilyrohrbough:
I am curious, did you try mocking require.ensure to load the translation data?

packages/terra-base/tests/jest/Base.test.jsx
```diff
@@ -8,6 +8,7 @@ it('should support rendering a string without translation', () => {
   expect(base).toMatchSnapshot();
 });
 
+
```
kschuste: nit: extra line got added here

packages/terra-base/tests/nightwatch/base-spec.js
```diff
@@ -9,6 +9,17 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
     browser.expect.element('div').text.to.contain('en').before(waitInms);
   },
 
+  // new
```
kschuste: this comment could be a bit more descriptive

packages/terra-base/tests/nightwatch/ArrayBase.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
```
kschuste: nit:  have an extra line here between imports don't really need, also could remove an extra line after the imports as well

packages/terra-base/tests/jest/Base.test.jsx
kschuste: nit:  added an extra line here not needed

packages/terra-embedded-content-consumer/tests/jest/EmbeddedContentConsumer.test.jsx
```diff
@@ -8,7 +8,7 @@ it('should render the embedded content consumer container', () => {
     src="https://www.google.com/"
   />);
 
-  const wrapper = shallow(embeddedContentConsumer);
```
yuderekyu: What's the reason we're switching from shallow rendering to rendering the static html?

packages/terra-base/tests/jest/Base.test.jsx
bjankord: Removed in this commit: https://github.com/cerner/terra-core/pull/1012/commits/de92c42b10d391c997e2d2edd3ee4b606e679213

packages/terra-base/tests/nightwatch/ArrayBase.jsx
bjankord: Removed in this commit: https://github.com/cerner/terra-core/pull/1012/commits/de92c42b10d391c997e2d2edd3ee4b606e679213

packages/terra-base/tests/nightwatch/base-spec.js
bjankord: Removed in this commit: https://github.com/cerner/terra-core/pull/1012/commits/de92c42b10d391c997e2d2edd3ee4b606e679213

bjankord:
@emilyrohrbough I looked around for a [few](https://stackoverflow.com/questions/44114737/how-do-i-stub-webpacks-require-ensure) [ways](https://stackoverflow.com/questions/33070482/writing-tests-for-javascript-module-using-webpacks-require-ensure-function) to mock `require.ensure`, yet I got caught up on how exactly I might go about mocking the `.ensure` method so I ended up moving these tests from jest to nightwatch.

packages/terra-embedded-content-consumer/tests/jest/EmbeddedContentConsumer.test.jsx
bjankord: The event listeners in the EmbeddedContentConsumer component were throwing errors after updating to enzyme v3 using the `shallow` method. My understanding on why this is an issue now is a bit rough but here's what I'm thinking. Attaching an event listener to the document with `addEventListener` means the event is not being handled by React's synthetic event system and I believe enyzme's `shallow` and `mount` methods use React's synthetic event system while `render` uses another implementation which supports `addEventListener`.

emilyrohrbough:
@bjankord I think moving them to nightwatch is totally fine. I did a little digging myself since I was curious and it seem like the general suggestion is to move away from using require.ensure and actually use [System.import](https://webpack.js.org/guides/code-splitting/) (introduced with webpack v2) to do code splitting and dynamical loading. Might be something we should keep in mind for using/updating in the future.

packages/terra-demographics-banner/tests/jest/__snapshots__/DemographicsBannerDisplay.test.jsx.snap
```diff
@@ -3,7 +3,7 @@
 exports[`renders a banner that has basic information 1`] = `
 <ResponsiveElement
   defaultElement={
-    <Component
+    <Unknown
```
ryanthemanuel: What is "Unknown"?

bjankord: Not entirely sure, I tested this with shallow, mount, and render and all output the `string` unknown here. Looks like jest has some logic to determine the type of component for react elements / components which are passed in as props [here](https://github.com/facebook/jest/blob/master/packages/pretty-format/src/plugins/react_element.js#L34-L42).

With how [_LargeDemographicsBannerDisplay](https://github.com/cerner/terra-core/blob/master/packages/terra-demographics-banner/src/_LargeDemographicsBannerDisplay.jsx#L10) and [_SmallDemographicsBannerDisplay](https://github.com/cerner/terra-core/blob/master/packages/terra-demographics-banner/src/_SmallDemographicsBannerDisplay.jsx#L10) are exported, it seems jest is unable to determine what type they are and falls back to using the string `Unknown` where before it would use the string `Component`.

PR #1014, opened 2017-11-13
Search field - add disabled state and stretch sizing

### LINKED ISSUE #953: Search Field: disabled state & fits container

kschuste:
### Summary
This fixes #953.  This allows for the search field to be disabled and have it stretch to 100% of its containers width.

### Additional Details
Two new props have been added.  isDisabled is for disabling the field and button so you cannot click on it or tab to it.  sizing has been added that will allow you to pass in 'fixed' or 'stretch' for it to work the current way (fixed) or to have it fit its container (stretch).

### Deployment Link
https://terra-core-deployed-pr-1014.herokuapp.com/static/#/site/search-field


packages/terra-search-field/src/SearchField.jsx
```diff
@@ -86,9 +98,19 @@ class SearchField extends React.Component {
...
+      minimumSearchTextLength,
+      searchDelay,
+      onSearch,
+      onInvalidSearch,
+      sizing,
+      ...customProps
+    } = this.props;
     const searchFieldClassNames = cx([
       'searchfield',
+      (sizing === 'stretch') ? 'stretch' : '',
```
emilyrohrbough: use  { 'stretch': sizing === 'stretch' }

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -34,12 +39,19 @@ const propTypes = {
    * A callback to indicate an invalid search. Sends parameter {String} searchText.
    */
   onInvalidSearch: PropTypes.func,
+
+  /**
+   * How the component should display from a fixed size to stretching to fill its container
+   */
+  sizing: PropTypes.oneOf(['fixed', 'stretch']),
```
emilyrohrbough: This goes back to discussion on the issue @StephenEsser & @Singleton06 had regarding this prop. Was it ever determined if there would be more than one sizing option other than 'stretch'? 'Fixed' doesn't apply any special logic, and IMO the behavior 'fixed' is representing here is actual 'auto'. To me, 'fixed' would indicate an max-width was attached to the search field.

kschuste:
@emilyrohrbough updated to just use a boolean prop for the stretching, I used the string prop before in case in the future we needed another sizing attribute as discussed in tech design, but agree there should not be any other cases

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -86,9 +98,19 @@ class SearchField extends React.Component {
...
+      placeholder,
+      minimumSearchTextLength,
+      searchDelay,
+      onSearch,
+      onInvalidSearch,
+      ...customProps
+    } = this.props;
     const searchFieldClassNames = cx([
       'searchfield',
+      { stretch: isStretched === true },
```
jmsv6d: You don't need the `=== true` here you can just have `isStretched`

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -11,6 +11,16 @@ const cx = classNames.bind(styles);
 
 const propTypes = {
   /**
+   * A boolean that when true will disable the field
```
jmsv6d: Nit. You don't need the prop type in the description.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -11,6 +11,16 @@ const cx = classNames.bind(styles);
...
+   */
+  isDisabled: PropTypes.bool,
+
+  /**
+   * A boolean that when true will stretch the field to fill its container
+   */
+  isStretched: PropTypes.bool,
```
mjhenkes: What do folks think about changing this to `isBlock` to be more consistent with other apis, namely button. Would describing this behavior as block be an abuse of the term?

kschuste: isBlock would make me believe how it will display next to its peers, currently it is display: inline-flex, which makes it atomic to its peers and always displays inline

I have the width set to 100% to match with the select stying; however, setting it to be display: flex allows it to fill its container as well

If changing the display attribute to be flex, then I think isBlock makes sense as the 'flex' style makes it a block-level element

After thinking more on this, I am more in favor of using isBlock and using display: flex and let it size itself next to its peers rather than always be 100% that maybe could cause display issues that I have been experiencing with the collapsible menu view

kschuste:
mjhenkes 39 minutes ago  Owner
What do folks think about changing this to isBlock to be more consistent with other apis, namely button. Would describing this behavior as block be an abuse of the term?

@mjhenkes isBlock would make me believe how it will display next to its peers, currently it is display: inline-flex, which makes it atomic to its peers and always displays inline

I have the width set to 100% to match with the select stying; however, setting it to be display: flex allows it to fill its container as well

If changing the display attribute to be flex, then I think isBlock makes sense as the 'flex' style makes it a block-level element

After thinking more on this, I am more in favor of using isBlock and using display: flex and let it size itself next to its peers rather than always be 100% that maybe could cause display issues that I have been experiencing with the collapsible menu view.

tbiethman:
@kschuste I prefer isBlock as well

packages/terra-search-field/src/SearchField.jsx
tbiethman: 👍

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -11,6 +11,16 @@ const cx = classNames.bind(styles);
 
 const propTypes = {
   /**
+   * When true, will disable the field
```
emilyrohrbough: nit(s): missing '.' in descriptions.

packages/terra-search-field/tests/nightwatch/SearchFieldBlock.jsx
```diff
@@ -0,0 +1,4 @@
+import React from 'react';
+import SearchField from '../../src/SearchField';
```
emilyrohrbough: Update 'src' to 'lib'

packages/terra-search-field/tests/nightwatch/SearchFieldDisabled.jsx
```diff
@@ -0,0 +1,4 @@
+import React from 'react';
+import SearchField from '../../src/SearchField';
```
emilyrohrbough: Update 'src' to 'lib'

packages/terra-search-field/tests/nightwatch/search-field-spec.js
```diff
@@ -74,4 +74,26 @@ module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous
       });
     });
   },
+
+  'Displays a search field that is disabled': (browser) => {
+    browser
+      .url(`${browser.launchUrl}/#/tests/search-field-tests/disabled`)
+      .assert.elementPresent('#searchfield')
```
emilyrohrbough: Want to use 'expect' syntax in these test steps.

packages/terra-search-field/tests/nightwatch/search-field-spec.js
```diff
@@ -3,19 +3,19 @@ const resizeTo = require('terra-toolkit/lib/nightwatch/responsive-helpers').resi
...
 module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous'], {
   'Displays a search field with search button': (browser) => {
-    browser
-      .url(`${browser.launchUrl}/#/tests/search-field-tests/default`)
-      .assert.elementPresent('#searchfield')
-      .assert.elementPresent('#searchfield button')
-      .assert.attributeEquals('#searchfield input', 'placeholder', '');
+    browser.url(`${browser.launchUrl}/#/tests/search-field-tests/default`);
+
+    browser.expect.element('#searchfield').to.be.present.to.equal(true);
```
yuderekyu: Doesn't `to.be.present` already assert if the element exists in the dom? `to.equal(true)` seems redundant

kschuste: it is needed to be done to get around the no-unused-expressions linter rule, else I would need to turn that off

PaulSolDev:
Functional issue complete.  No issues identified.

mmalaker:
Functional verification completed. The search examples display and function as expected, including the new properties.

neilpfeiffer:
+1 : Marking as Verified, ready for Release.

- Disabled search has funny very minor space/styling  issue to watch for in the future, but all looks great otherwise.

PaulSolDev:
JIRA created

PR #1022, opened 2017-11-16
Cleaned up travis file

mhemesath:
* Removed redundant timezone environment variable
* Update docker container to reference host memory per selenium guidance

See https://github.com/SeleniumHQ/docker-selenium#running-the-images for more details


mjhenkes:
You could remove the env section too, or we could do it later. That was for sauce labs setup that we aren't using.

PR #1023, opened 2017-11-16
Add webdroverio tests for badge

mhemesath:
### Summary
The hello world of tests, badge :).  This PR adds support for webdriverio and creates tests for badges.

A few notes:
* I added terra-site/dist to gitignore as it is generated while running tests
* There are some test helpers that we can discuss about potentially moving to toolkit as it makes sense.


wdio.conf.js
```diff
@@ -0,0 +1,37 @@
...
+  ],
+  staticServerPort,
+  staticServerLog: false,
+  staticServerFolders: [
+    { mount: '/', path: path.join(__dirname, 'packages', 'terra-site', 'dist') },
+  ],
+
+  suites: {
+    badge: [
+      path.join(__dirname, 'packages', 'terra-badge', 'tests', 'specs', '**'),
```
mhemesath: I think we can create a function that pulls all the packages and auto-generates suite names...

packages/terra-badge/tests/specs/badge.js
```diff
@@ -0,0 +1,39 @@
+/* global browser, describe, it, expect, before, Terra */
+const viewports = Terra.viewports('tiny', 'medium');
+
+const shouldBeAccessible = () => {
```
mhemesath: this is probably a candidate to move to toolkit... i just learned about these shared assertions here https://github.com/mochajs/mocha/wiki/Shared-Behaviours

packages/terra-badge/tests/nightwatch/badge-spec.js
```diff
@@ -1,57 +0,0 @@
-/* eslint-disable no-unused-expressions */
```
emilyrohrbough: I agree, nothing needs tested in nightwatch now. With the removal of this spec, I think we make our first step towards our testing migration and move the nightwatch examples and routes into their own 'examples' directory. since both site, wdio tests and nightwatch tests will be sharing them. 

My reasoning why now is keeping them in nightwatch could be confusing later on since we are removing this file.

packages/terra-badge/tests/specs/badge.js
emilyrohrbough: My vote is to also move these to toolkit.

mhemesath: Do we want to set this on the Terra helper object?

wdio.conf.js
```diff
@@ -0,0 +1,37 @@
...
+    ],
+  },
+
+  seleniumDocker: {
+    enabled: !process.env.TRAVIS,
+    cleanup: false,
+  },
+  // eslint-disable-next-line global-require
```
ryanthemanuel: Rather than disabling eslint here can we move the require to the top of the file and then reference resulting const here?

wdio.conf.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-disable import/no-extraneous-dependencies */
+const wdioConf = require('terra-toolkit/lib/wdio/conf');
+const localIP = require('ip');
+const path = require('path');
+
+const staticServerPort = 8080;
+
```
emilyrohrbough: I have a few ideas for this file: I have included my changes in [this gist](https://gist.github.com/emilyrohrbough/db688ee72304c69d8be43f9d6de76c28)

This will also keep the package level scripts to `npm run test:wdio ../../wdio.conf.js` and remove the need to 
> create a function that pulls all the package directories and auto-generates suite names...

wdio.conf.js
mhemesath: yeah this might change anyway pending @emilyrohrbough adding a webpack service

wdio.conf.js
```diff
@@ -0,0 +1,37 @@
...
+  seleniumDocker: {
+    enabled: !process.env.TRAVIS,
+    cleanup: false,
+  },
+  // eslint-disable-next-line global-require
+  webpackConfig: require('./packages/terra-site/webpack.config.js'),
+};
+
+
+config.services = wdioConf.config.services.concat(['webpack', 'static-server']);
```
ryanthemanuel: I know @emilyrohrbough was looking into getting webpack-dev-server working rather than using webpack to generate the site.  I think I'd ideally like to use webpack-dev-server if possible.  Should we wait for that?

packages/terra-badge/tests/specs/badge.js
```diff
@@ -0,0 +1,39 @@
...
+  it('is accessible', () => expect(browser.axe()).to.be.accessible());
+};
+
+const shouldBeVisuallyConsistent = () => {
+  it('[0] checks visual comparison', () => {
+    const screenshots = browser.checkElement('[data-reactroot]', { viewports });
```
emilyrohrbough: Moving this to toolkit, I'd like to see this updated to pass in the desired checkElement with the fallback to '[data-reactroot]'

packages/terra-badge/tests/specs/badge.js
ryanthemanuel: I don't have a strong preference but I would lean towards a Terra helper.

packages/terra-badge/tests/specs/badge.js
```diff
@@ -0,0 +1,39 @@
+/* global browser, describe, it, expect, before, Terra */
+const viewports = Terra.viewports('tiny', 'medium');
+
+const shouldBeAccessible = () => {
+  it('is accessible', () => expect(browser.axe()).to.be.accessible());
```
emilyrohrbough: Update to check viewports specified

package.json
```diff
@@ -102,6 +104,11 @@
     "stylelint-scss": "^2.0.0",
     "stylelint-suitcss": "^1.0.0",
     "terra-toolkit": "^2.2.0",
+    "wdio-mocha-framework": "^0.5.11",
+    "wdio-static-server-service": "^1.0.0",
```
emilyrohrbough: I'm adding tests then putting out a PR for WebPackDevService so we can drop the wdio-static-server-service & continue to use the webpack-dev-server as expected. Then we can drop the `packages/terra-site/dist/` from .gitIgnore

wdio.conf.js
mhemesath: 👍 I'm awaiting her pull request!!!

calebmeyer:
Not sure if this is your branch, but running npm install on this branch in a non-elevated prompt failed:
https://gist.github.com/calebmeyer/947aa9d7fa71590df3250fc7f5827940

This was the first time I attempted to install terra-core, I'll try again from an elevated prompt.

packages/terra-badge/tests/nightwatch/badge-spec.js
mhemesath: 👍

packages/terra-badge/tests/specs/badge.js
```diff
@@ -0,0 +1,39 @@
+/* global browser, describe, it, expect, before, Terra */
```
emilyrohrbough: Two opinions on this:

1. I'd prefer we put wdio testing in 'wdio' dir
1. I'd prefer we update these test files to use '-spec.js' syntax so we can filter on 'tests/wdio/**/*-spec.js'

mhemesath: 👍

calebmeyer:
I see 4 test failures and 1 lint when I run the tests: https://gist.github.com/calebmeyer/da5beb916bf8ba386e98ac3d908b8950

mhemesath:
> I see 4 test failures and 1 lint when I run the tests: 

Interesting, those are jest failures. Does master fail for you too? My changes wouldn't have touched that functionality.

calebmeyer:
I'll check it out and comment back.

calebmeyer:
Okay, testing on master passes all the jest tests, but fails on a docker pull before the nightwatch tests start: https://gist.github.com/calebmeyer/1326603a411641e42a649fb9fd1589ae

emilyrohrbough:
I'm not seeing these failures locally.

calebmeyer:
Okay, must be a race condition or something. I moved the master branch to the root of my hard drive and now I see 5 jest failures instead of 4. I don't think my machine is very good as a testbed. Carry on

package.json
bjankord: @emilyrohrbough What's the status on this?

mhemesath: We’re close to pushing a release of toolkit which will clean up this or a bunch

packages/terra-badge/tests/specs/badge.js
bjankord: Related PR: https://github.com/cerner/terra-toolkit/pull/29/files

wdio.conf.js
bjankord: @emilyrohrbough What's the status on this?

wdio.conf.js
bjankord: What's the status on this?

packages/terra-badge/tests/wdio/badge-spec.js
```diff
@@ -0,0 +1,28 @@
+/* global browser, describe, it, expect, before, Terra */
```
bjankord: Curious, what is this comment for?

mhemesath: This tells the linter these are globals

.gitignore
```diff
@@ -55,3 +55,6 @@ target
 **/aggregated-translations/*.*
 lib
 props-table
+**/__snapshots__/screen
+**/__snapshots__/diff
+errorShots
```
emilyrohrbough: This was updated to [errorScreenshots ](https://github.com/cerner/terra-toolkit/blob/master/src/wdio/conf.js#L22)

wdio.conf.js
```diff
@@ -0,0 +1,37 @@
...
+if (__dirname === process.cwd()) {
+  specs = path.join('packages', '**', specs);
+}
+
+const config = {
+  ...wdioConf.config,
+
+  baseUrl: `http://${localIP.address()}:${webpackPort}`,
+  specs,
+  suites: {
```
emilyrohrbough: Think we can drop the suites config. The specs paths will handle mono-repo level vs package level testing and this will become quite large as we add more wdio-specs.

packages/terra-badge/tests/nightwatch/badge-spec.js
emilyrohrbough: Still think it would be good to make our first step towards our testing migration and move the nightwatch examples and routes into their own 'examples' directory. 

Other than that, this PR looks good to me!

mhemesath: I'm not sure its worth the work... we have to move the files and update the terra-site doc. Being @bjankord is starting work to move components to storybook this is a short lived effort..

emilyrohrbough: sure

wdio.conf.js
```diff
@@ -0,0 +1,45 @@
...
+
+  seleniumDocker: {
+    enabled: !process.env.TRAVIS,
+  },
+
+  webpackPort,
+  webpackConfig,
+  before() {
+    // eslint-disable-next-line prefer-arrow-callback, func-names
+    browser.addCommand('setCSSCustomProps', function (properties) {
```
mhemesath: this needs to move to toolkit... but i'd like to push this through and do a PR this weekend to put it there if possible. I'd like to wrap up the badge PR unless anyone feels strongly about it

packages/terra-badge/tests/wdio/badge-spec.js
```diff
@@ -0,0 +1,81 @@
...
+const viewports = Terra.viewports('tiny', 'medium');
+
+const shouldTheme = (customProperties) => {
```
mhemesath: this should move to toolkit

wdio.conf.js
```diff
@@ -0,0 +1,45 @@
...
+  baseUrl: `http://${localIP.address()}:${webpackPort}`,
+  specs,
+
+  seleniumDocker: {
+    enabled: !process.env.TRAVIS,
+  },
+
+  webpackPort,
+  webpackConfig,
+  before() {
```
emilyrohrbough: I like the idea of this as a candidate for toolkit. @bjankord @ryanthemanuel what are your thoughts? This before enables us to ensure the themeable variables are correctly applied. 

If we want to wait on this as a toolkit helper, I would be in favor of breaking this before into its own file. At minimal documentation describe the before func.

mhemesath: If we broke it out, where would we put it?

wdio.conf.js
bjankord: I'm fine with moving this to toolkit in a follow-up PR.

Issue #925, opened 2017-10-13
Forms: Checkbox/Radio long text wrapping


neilpfeiffer:
# Issue Description
Correct the text wrapping for long labels for Checkboxes and Radio buttons.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior
Multiword and single word (non-breaking character labels) should maintain text alignment and not wrap underneath the input.

![image](https://user-images.githubusercontent.com/1355485/31566604-ff14e4ec-b030-11e7-9c8d-32cc0253fa4f.png)

![image](https://user-images.githubusercontent.com/1355485/31566606-02c4873c-b031-11e7-8e91-be99ef79c188.png)

## Current Behavior
Multiword and single word (non-breaking character labels) wrap underneath the input. 

![image](https://user-images.githubusercontent.com/1355485/31566243-57253670-b02f-11e7-8f16-48c2f65dd8b8.png)

![image](https://user-images.githubusercontent.com/1355485/31566252-5e992510-b02f-11e7-942a-58aa262435a6.png)

neilpfeiffer:
Can be resolved when radios and checkboxes uplifted to new OCS/themes - meets same design input guidelines.

skalyanmoguloju:
@neilpfeiffer: is this issue open? If so can I get assigned to me?

jmsv6d:
@skalyanmoguloju This should be getting resolved with this issue: https://github.com/cerner/terra-core/issues/960 that's currently being worked on.

bjankord:
Resolved in #1135

PR #1024, opened 2017-11-16
[WIP] Checkbox OCS

### LINKED ISSUE #999: Checkbox component for OCS
# Issue Description
Create a checkbox component for OCS. Tech design here: #960

## Issue Type

- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

yuderekyu:
### Summary
Creates a checkbox component with themeable variables for OCS. The component is responsive and allows for text wrapping. Focus can be triggered with tab; toggle can be triggered with the space bar. Defaults to terra open source theme.

### Closes
#999 
#925 (for checkboxes)

### Deployment Links
* https://terra-core-deployed-pr-1024.herokuapp.com/static/#/site/form-checkbox
* https://terra-core-deployed-pr-1024.herokuapp.com/static/#/tests/form-checkbox-tests


packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,161 @@
...
+      'label-text',
+      labelTextAttrs.className,
+    ]);
+
+    const inputClasses = cx([
+      inputAttrs.className,
+    ]);
+
+    return (
+      <div className={`${controlClasses} ${this.getFocusClass()}`} >
```
jmsv6d: Instead of dong this and having the getFocusClass function you can just add
`{ focus: this.state.focus }`  in the controlClasses cx array.

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,161 @@
...
+          type="checkbox"
+          id={id}
+          disabled={isDisabled}
+          name={name}
+          value={value}
+          onChange={onChange}
+          onFocus={this.onFocus}
+          onBlur={this.onBlur}
+          className={inputClasses}
+          {...controlInputAttrs}
```
jmsv6d: Spread the controlInputAttrs first so consumers can't override your props.

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,161 @@
...
+  constructor(props, context) {
+    super(props, context);
+    this.state = { focus: false };
+    this.onFocus = this.onFocus.bind(this);
+    this.onBlur = this.onBlur.bind(this);
+    this.getFocusClass = this.getFocusClass.bind(this);
+  }
+
+  onFocus() {
+    this.setState({ focus: true });
```
jmsv6d: I'm trying to think if there's a better way to accomplish your focus styling than having focus as state. But if you keep it this way you'll need to call the onFocus and onBlur functions passed in through the inputAttrs props, if they exist.

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,161 @@
...
+      labelTextAttrs,
+      name,
+      onChange,
+      value,
+      ...customProps
+    } = this.props;
+
+    const controlInputAttrs = Object.assign({}, inputAttrs);
+
+    if (checked !== undefined) {
```
jmsv6d: Why do you need this check? Shouldn't the input handle this logic if you pass the given props down?

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,161 @@
...
+      controlInputAttrs.checked = checked;
+    } else {
+      controlInputAttrs.defaultChecked = defaultChecked;
+    }
+
+    const controlClasses = cx([
+      'control',
```
jmsv6d: Update this to be checkbox classes

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,116 @@
+:local {
+  .control {
```
jmsv6d: Change this to class to .checkbox

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,116 @@
...
+    margin-right: var(--terra-form-checkbox-margin-right, 2em);
+    margin-top: var(--terra-form-checkbox-margin-top, 0.625em);
+    vertical-align: middle;
+
+    .label-text {
+      font-size: var(--terra-form-checkbox-font-size, 1rem);
+    }
+
+    /* Hide actual input element */
+    input:not(:checked),
```
jmsv6d: I'm confused by the need for these selectors. Why wouldn't you just add a class to your input and use that?

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,116 @@
...
+    input:not(:checked) + .label-text,
+    input:checked + .label-text {
+      cursor: pointer;
+      display: flex;
+      padding-left: var(--terra-form-checkbox-padding-left, 1.6em);
+      position: relative;
+    }
+
+    /* Checkbox aspect */
+    input:not(:checked) + .label-text::before,
```
jmsv6d: All of your selectors seem to be overly complex. You should be able to just use `.label-test::before` here.

packages/terra-site/src/cerner-mock-theme.scss
```diff
@@ -3,6 +3,34 @@
...
+  --terra-form-checkbox-checked-border-radius: 2px;
+  --terra-form-checkbox-checked-border-style: solid;
+  --terra-form-checkbox-checked-border-width: 0.0625em;
+  --terra-form-checkbox-checked-height: 1.285714285714286em;
+  --terra-form-checkbox-checked-width: 1.285714285714286em;
+  --terra-form-checkbox-disabled-background-color: gba(134, 138, 140, 0.1);
+  --terra-form-checkbox-disabled-border-color: gba(134, 138, 140, 0.4);
+  --terra-form-checkbox-disabled-font-color: #868a8c;
+  --terra-form-checkbox-focus-box-shadow: none;
+  --terra-form-checkbox-container-focus-box-shadow: 0 0 0 1 #4a90e2;
```
yuderekyu: Updating this to a valid value.

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,116 @@
...
+    input:checked + .label-text::before {
+      background-color: var(--terra-form-checkbox-background-color, #f6f6f6);
+      background-image: var(--terra-form-checkbox-background-image, linear-gradient(316deg, rgba(227, 227, 227, 0.08), #a9a9a9));
+      background-position: center center;
+      background-repeat: no-repeat;
+      border-color: var(--terra-form-checkbox-border-color, #979797);
+      border-radius: var(--terra-form-checkbox-border-radius, none);
+      border-style: var(--terra-form-checkbox-border-style, solid);
+      border-width: var(--terra-form-checkbox-border-width, 0.0625em);
+      box-shadow: var(--terra-form-checkbox-box-shadow, inset 0 0 0 0.09375em #f6f6f6);
```
jmsv6d: You're going to need to define ltr and rtl values for box-shadows.

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,116 @@
...
+    margin-top: var(--terra-form-checkbox-margin-top, 0.625em);
+    vertical-align: middle;
+
+    .label-text {
+      font-size: var(--terra-form-checkbox-font-size, 1rem);
+    }
+
+    /* Hide actual input element */
+    input:not(:checked),
+    input:checked {
```
jmsv6d: You can just use `display: none` to hide the input.

packages/terra-form-checkbox/src/Checkbox.jsx
yuderekyu: This logic is based on discussion here within Control.jsx: https://github.com/cerner/terra-core/pull/448/files. 

I do have `defaultChecked` defaulted to `false` - I'll change it to undefined.

packages/terra-form-checkbox/src/Checkbox.scss
yuderekyu: The custom css checkbox depends on interaction with the actual input element, so I opted to go with the hidden positioning route found in html5-boilerplat: https://github.com/h5bp/html5-boilerplate/blob/c3d52666d0d297e50f965af4ad214702a683e02f/src/css/main.css#L126-L135

jmsv6d: If your goal is to allow the consumer to interact with the actual input, this approach doesn't work either. With a 1px height and width the click point is too small, and it is placed behind the label so all click events will hit the label, not the input. The reason your controls work now is that all your examples have an id which allows interacting with the label to affect the state of the input. If you had an example without an id it wouldn't work. (Note: This also means that id probably needs to be a required prop).

jmsv6d: Also note, for visually hidden elements you can just use `visibility: hidden`

yuderekyu: Anyone have thoughts on making `id` required? Another route is changing the implementation back to the old Control, with `input` inside `label`. `label` will automatically handle `input` toggling.
``` html
<label>
  <input></input>
  <span></span>
</label>
```

yuderekyu:
@jmsv6d Made changes based on your feedback - check em out 🎉

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,162 @@
...
+    */
+  checked: PropTypes.bool,
+  /**
+    * The checked property of the Input element. Use this to generate an uncontrolled Checkbox Element.
+    */
+  defaultChecked: PropTypes.bool,
+  /**
+    * The id of the input field.
+    */
+  id: PropTypes.string,
```
yuderekyu: Based on previous discussion: should we have this required, change the implementation and encase `input` within `label`, or something else?

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,132 @@
+:local {
+  .checkbox {
+    color: var(--terra-form-checkbox-font-color, rgb(0, 0, 0));
+    display: block;
+    font-family: inherit;
+    margin-right: var(--terra-form-checkbox-container-margin-right, 2em);
+    margin-top: var(--terra-form-checkbox-container-margin-top, 0.625em);
+    
+    /* Hide actual input element */
+    .hidden-input {
```
yuderekyu: On second thought, I *should* be able to hide the custom checkbox and show the actual input when using the default theme, instead of always using the custom css checkbox across themes. This would stop me from trying to recreate the windows checkbox style for some themes. Most of the themeable variables would be set to none. I'll give another update after I've tinkered with this.

yuderekyu: ^^ I'll close this pr for now and open another after this update.

PR #1027, opened 2017-11-19
Implement Time Input for Theming

### LINKED ISSUE #1008: Uplift terra-time-input for theming
# Issue Description
terra-time-input needs to be changed to allow for theming. In order to incorporate this, additional changes must be made to the component.

- Allow for the inputing of seconds
- Allow for the input of a meridiem.

These are the request mockups for the input.

screen shot 2017-11-10 at 9 21 27 am

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

JakeLaCombe:
### Summary
Time Input has been uplifted to allow for theming. In addition, new functionality was added to the time input.
- Meridiem Support has been added
- Mobile has it's own specific time picker.

Closes #1008 

## Deployment Link
https://terra-core-deployed-pr-1027.herokuapp.com/static/#/site/time-input

### Additional Details
I've included the translations I've seen from hi-format into this pull request. I'll log a more formal translations Jira as I get more feedback on the pull request.


packages/terra-time-input/docs/README.md
```diff
@@ -2,7 +2,15 @@
...
 
-The currently supported time format is the 24-hour format (hh:mm). The time input enforces the entry that masks to the format. The hour input only accepts values between 00 and 23 and the minute input only accepts values between 00 and 59. For example, a time of 25:65 cannot be entered. A 0 will automatically be prepended to the hour if the entered hour is greater than 2. Likewise, a 0 will automatically be prepended to the minute if the entered minute is greater than 5.
+The currently supported time format is the 24-hour format (hh:mm). The time input enforces the entry that masks to the format. The hour input only accepts values between 00 and 23 and the minute input only accepts values between 00 and 59. For example, a time of 25:65 cannot be entered. Any times that are not in the format hh:mm will be ignored when initialized a terra-time-input component.
```
bjankord: nit: Might rework the last sentence to "Time values provided to the component when it is initialized which are not in the format of `hh:mm` will be ignored".

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -30,39 +69,46 @@ const propTypes = {
...
-      hour: TimeUtil.splitHour(props.value),
-      minute: TimeUtil.splitMinute(props.value),
-      isFocused: false,
-    };
+    let value = this.props.value;
+
+    if (value && !validDateRegex.test(value)) {
+      if (process.env !== 'production') {
+        // eslint-disable-next-line
+        console.warn(`Invalid time ${value} pass to terra-time-picker and has been ignored. Format of time must be hh:mm.`);
```
bjankord: nit: Might reword to `An invalid time value, ${value}, has been passed to the terra-time-picker. This value has been ignored and will not be rendered. Time values must be in hh:mm format.`

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -71,19 +117,93 @@ class TimeInput extends React.Component {
...
+
     this.setState({
-      hour: TimeUtil.splitHour(nextProps.value),
+      hour,
       minute: TimeUtil.splitMinute(nextProps.value),
+      meridiem,
     });
   }
 
+  parseTimeTwelveHourTime(hour) {
```
bjankord: What do you think about using doc here to note what the intentions of this function are. I'm thinking some JSDoc format to describe it, the argument passed to it, and what it returns.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -146,32 +263,171 @@ class TimeInput extends React.Component {
...
+        return;
+      }
+    }
+
+    let inputValue = event.target.value;
+    const stateValue = this.state.minute;
+    const maxValue = 59;
+
+    // Ignore the entry if the value did not change or it is invalid.
+    // When 'Predictive text' is enabled on Android the maxLength attribute on the input is ignored so we have to the length of inputValue to make sure that it is less then 2.
```
bjankord: nit: Looks like this phrase is missing the word check:  `is ignored so we have to` check `the length of inputValue`

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -146,32 +263,171 @@ class TimeInput extends React.Component {
...
+    if (inputValue === stateValue || inputValue.length > 2 || Number(inputValue) > maxValue) {
+      return;
+    }
+
+    // If the change made was not a deletion of a digit,
+    // then prepend '0' if the input value is a single digit value between 3 and 9 for hour or 6 and 9 for minute.
```
bjankord: With this comment being within `handleMinuteChange`, can probably drop the `3 and 9 for hour or` part.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -146,32 +263,171 @@ class TimeInput extends React.Component {
...
+  }
+
+  meridiemBlur() {
+    this.setState({
+      isFocused: false,
+      meridiemFocused: false,
+    });
+  }
+
+  handleHourInputKeyDown(event) {
```
bjankord: What do you think about using doc here to note what the intentions of this function are. I'm thinking some JSDoc format to describe it, the argument passed to it, and what it returns.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -146,32 +263,171 @@ class TimeInput extends React.Component {
...
+          }
+        } else if (this.props.variant === FORMAT_12_HOUR) {
+          stateValue = '01';
+          updateStateValue = true;
+        }
+      } else {
+        stateValue = this.props.variant === FORMAT_12_HOUR ? '12' : '00';
+        updateStateValue = true;
+      }
+    } else if (event.keyCode === keyCodes.ARROWDOWN) {
```
bjankord: What are your thoughts on breaking this off to its own if statement:

`if (event.keyCode === keyCodes.ARROWDOWN) {` instead of using `else if`

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -146,32 +263,171 @@ class TimeInput extends React.Component {
...
+          }
+        } else if (this.props.variant === FORMAT_12_HOUR) {
+          stateValue = '12';
+          updateStateValue = true;
+        }
+      } else {
+        stateValue = this.props.variant === FORMAT_12_HOUR ? '12' : '00';
+        updateStateValue = true;
+      }
+    } else if (event.keyCode === keyCodes.ARROWRIGHT) {
```
bjankord: What are your thoughts on breaking this off to its own if statement:

`if (event.keyCode === keyCodes.ARROWRIGHT) {` instead of using `else if`

packages/terra-time-input/src/TimeInput.scss
```diff
@@ -1,40 +1,204 @@
...
+    border-color: var(--terra-time-input-focus-keyboard-border-color, #4cb2e9);
+    border-style: var(--terra-time-input-focus-keyboard-border-style, solid);
+    border-width: var(--terra-time-input-focus-keyboard-border-width, 1px);
+    opacity: var(--terra-time-input-focus-keyboard-opacity, 1);
+    outline: none;
+  }
+
+  .time-input {
+    background-color: var(--terra-time-input-background, #fff);
+    background-image: var(--terra-time-input-focus-animation-background, none);
```
bjankord: Looks like there are some CSS custom properties in the `.time-input` selector which include the word `focus`. Not sure if this is what you meant to include. e.g I'd expect `background-image: var(--terra-time-input-focus-animation-background, none);` to actually be `background-image: var(--terra-time-input-animation-background, none);` and things like `border-style: var(--terra-time-input-desktop-focus-border-style, solid);` to actually be `border-style: var(--terra-time-input-desktop-border-style, solid);`

Then you can use `-focus-` in the CSS custom properties within the `is-focused` selector below.

packages/terra-time-input/src/TimeInput.scss
```diff
@@ -1,40 +1,204 @@
...
+
+  .time-input-hour,
+  .time-input-minute {
+    background-color: var(--terra-time-input-background, #fff);
+    background-position: 0 100%;
+    background-repeat: no-repeat;
+    background-size: var(--terra-time-input-focus-background-start, 0 0);
+    border-color: var(--terra-time-input-desktop-focus-border-color, #dedfe0);
+    border-radius: var(--terra-time-input-border-radius, 3px);
+    border-style: var(--terra-time-input-desktop-focus-border-style, solid);
```
bjankord: Same comment as the one about `-focus-` named variables in `.time-input` applies to the `.time-input-hour` and `.time-input-minute` selectors.

packages/terra-time-input/src/TimeInput.jsx
emilyrohrbough: +1 to inline docs since time-input can be pretty complex

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -146,32 +263,171 @@ class TimeInput extends React.Component {
...
+
+  handleHourInputKeyDown(event) {
+    let stateValue = this.state.hour;
+    const maxValue = this.props.variant === FORMAT_12_HOUR ? 12 : 23;
+    const minValue = this.props.variant === FORMAT_12_HOUR ? 1 : 0;
+
+    let meridiem = this.state.meridiem;
+    let updateStateValue = false;
+
+    if (event.keyCode === keyCodes.ARROWUP) {
```
kschuste: this method is quite large, can we break this up
I tend to go with the rule that it there is over 10 ways to get through a method it should be broken apart

maybe could put some of the minute calculation in a private method

packages/terra-time-input/docs/README.md
```diff
@@ -2,7 +2,15 @@
...
+
+There are two types of clocks that are currently supported: A 12-hour clock and a 24-hour clock. This can be controlled through the `variant` prop on the component.
+
+If a 24 hour clock is chosen, the hour displayed will be a value between 0 and 23 inclusively, and the minute displayed will be between 0 and 59 inclusively as well. A 0 will automatically be prepended to the hour if the entered hour is greater than 2. Likewise, a 0 will automatically be prepended to the minute if the entered minute is greater than 5. In addition, the inputs can be changed through the use of the up and down arrow keys on a non-mobile display.
```
emilyrohrbough: This wording is a bit confusing to me.

> A 0 will automatically be prepended to the hour if the entered hour is greater than 2. 

For example, I would expect when entering '1' then tabbing to result in '01', but this reads like only would only occur for entering '3-9' then tabbing to result in '03 - 09'.

Then,

> Likewise, a 0 will automatically be prepended to the minute if the entered minute is greater than 5.

I would expect any value entered '0-9' then tabbing to result in '00-09', but to me this reads time input would  only occur for entering '6-9' then tabbing to result in '06 - 09'.

packages/terra-time-input/docs/README.md
```diff
@@ -2,7 +2,15 @@
...
+
+If a 12 hour clock is chosen, the hour will be displayed between 01 and 12 inclusively, and an additional meridiem will be appended to the time input. The values of the meridiem are internationalized, meaning the time input will automatically initialize the post meridiem and ante meridiem for you. For the behavior of the hour input will inherit the behavior of the 24 hour clock with a few additions. Inputting 00 will automatically change the value to 12, and the hour input doesn't stop at 01 or 12. If the up or down arrows are pressed on the hour and the value goes over 12, it will automatically start at 01. Likewise, switching between 11 and 12 will update the value of the meridiem.
```
emilyrohrbough: >  Inputting 00 will automatically change the value to 12, and the hour input doesn't stop at 01 or 12. If the up or down arrows are pressed on the hour and the value goes over 12, it will automatically start at 01.

I feel like this could be updated to say something like

"Inputting 00 will automatically change the value to 12."
 -I would also include which meridian it applies with '00' I assume AM.

"Then when using the arrow keys to update the hour value, if switching between 11 and 12, the meridian value will update, pressing the up arrow when the value is 12 rolls over to 01 and pressing the down arrow when the value is 01 rolls back the value to 12."

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -3,16 +3,55 @@ import PropTypes from 'prop-types';
...
 import 'terra-base/lib/baseStyles';
 import Input from 'terra-form/lib/Input';
+import ButtonGroup from 'terra-button-group';
+
 import TimeUtil from './TimeUtil';
 import styles from './TimeInput.scss';
 
 const cx = classNames.bind(styles);
 
+const validDateRegex = new RegExp('^(0[0-9]|1[0-9]|2[0-3]):[0-5][0-9]$');
```
emilyrohrbough: @benbcai Didn't moment provide a validate date function?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -30,39 +69,46 @@ const propTypes = {
    * An ISO 8601 string representation of the time value in the input.
    */
   value: PropTypes.string,
+  /**
+   * Type of time input to initialize. Must be '24' or '12'
```
emilyrohrbough: nit: thoughts on updating these variants to be '12-hour' and '24-hour'?

packages/terra-time-input/src/TimeInput.jsx
emilyrohrbough: Also, I feel like this should be moved to the TimeUtils file.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -3,16 +3,55 @@ import PropTypes from 'prop-types';
...
+const keyCodes = {
+  BACKSPACE: 8,
+  ARROWLEFT: 37,
+  ARROWUP: 38,
+  ARROWRIGHT: 39,
+  ARROWDOWN: 40,
+  DELETE: 46,
+};
+
+const isConsideredMobileDevice = () =>
```
emilyrohrbough: I feel like this should be moved to the TimeUtils file.

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -3,16 +3,55 @@ import PropTypes from 'prop-types';
...
+
+const FORMAT_12_HOUR = '12';
+const FORMAT_24_HOUR = '24';
+
+const inputType = {
+  HOUR: 0,
+  MINUTE: 1,
+};
+
+const keyCodes = {
```
emilyrohrbough: Moved to the TimeUtils file. This would align with [menu](https://github.com/cerner/terra-core/blob/master/packages/terra-menu/src/_MenuUtils.js#L18), [table](https://github.com/cerner/terra-core/blob/b2fd3f8eab22d0f0daca643226659a55d301b94f/packages/terra-table/src/SelectableUtils.js#L26), menu etc

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -71,19 +117,93 @@ class TimeInput extends React.Component {
...
+      } else {
+        meridiem = this.anteMeridiem;
+      }
+    }
+
+    this.state = {
+      hour,
+      minute: TimeUtil.splitMinute(value),
+      isFocused: false,
+      meridiem,
```
emilyrohrbough: Should we check if its format before saving the meridian state? Wound't need to track this a 24 clock

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -71,19 +117,93 @@ class TimeInput extends React.Component {
...
   componentWillReceiveProps(nextProps) {
-    if (nextProps.value === this.props.value) {
+    if (
+      nextProps.value === this.props.value &&
+      nextProps.variant === this.props.variant
+    ) {
       return;
     }
 
+    let hour = TimeUtil.splitHour(nextProps.value);
```
emilyrohrbough: lines 159-171 are almost 1-to-1 with the 12-hour logic in the constructor. Maybe we should break this out to a reusable function in Time Utils?

packages/terra-time-input/src/TimeInput.jsx
emilyrohrbough: Can we move this to TimeUtils?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -106,8 +226,13 @@ class TimeInput extends React.Component {
 
     // Prepend a 0 to the value when losing focus and the value is single digit.
     if (stateValue.length === 1) {
-      stateValue = '0'.concat(stateValue);
-      this.handleValueChange(event, type, stateValue);
+      if (this.props.variant === FORMAT_12_HOUR && type === inputType.HOUR && stateValue === '0') {
```
emilyrohrbough: Should there be another check for when input is '00' for 12-hour clock?
```
if (
   stateValue.length === 2
   && this.props.variant === FORMAT_12_HOUR
   && type === inputType.HOUR
   && stateValue === '00' 
) {
    stateValue = '12';
}
```

emilyrohrbough: I see you have it in lines 279-282 in handleHourChange(). I wonder if it would be cleaner to keep this logic handling in the same place?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -146,32 +263,171 @@ class TimeInput extends React.Component {
...
 
-    this.handleValueChange(event, type, inputValue);
+    this.handleValueChange(event, inputType.HOUR, inputValue, this.state.meridiem);
   }
 
-  handleHourInputKeyDown(event) {
-    this.handleInputKeyDown(event, inputType.HOUR);
+  handleMinuteChange(event) {
+    // Ignore the input and don't update the state if the entered value is a non numeric character.
+    if (event.target.value.length > 0) {
```
emilyrohrbough: What is this included for? I notice the hour doesn't check this?

packages/terra-time-input/src/TimeInput.jsx
emilyrohrbough: I agree, breaking these into a smaller function would make this more readable. Possibly add a func to TimeUtils that will break each if into its own sub-function with descriptive names maybe?

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -210,59 +466,206 @@ class TimeInput extends React.Component {
...
+  formatHour(hour, meridiem) {
+    if (!hour) {
+      return hour;
+    }
+
+    let tempHour = parseInt(hour, 10);
+
+    if (this.props.variant === FORMAT_12_HOUR) {
+      if (meridiem === this.postMeridiem && tempHour < 12) {
+        tempHour += 12;
```
emilyrohrbough: Should this be '+= 12' or '=12'

packages/terra-time-input/src/TimeInput.jsx
benbcai: @emilyrohrbough. I think this regex is used to enforce that the hour can only be  between 00 and 23 and the minute be between 00 and 59. Moment does provide an isValid() function to validate that a moment object is initialized with a valid date but I'm not aware of anything for this regex.

benbcai:
What is the plan to update date-time-picker to consume these new changes? It looks like the height on the time-input now is slightly bigger than what it was before and the date-time-picker looks off. Or should the height of the time-input be reduced to what it was before so that the height is consistent with other text input fields?

screen shot 2017-11-20 at 1 45 40 pm

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: I can't move them all to the timeUtils as some deal with individual attributes attached to this react class, but I should be able to create a functions that properly increment and decrement the minutes

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -1,4 +1,39 @@
 class TimeUtil {
+  /**
+   * Finds the equivalent twelve hour time and meridiem for a given hour, anteMeridiem, and postMeridiem
+   * @param {String} hour Hour to parse.
+   * @param {String} anteMeridiem String representation of the anteMeridiem.
+   * @param {String} postMeridiem String representation of the postMeridiem.
+   * @return {{hourString: string., meridiem: string}} Hour and meridiem associated to the passed in variables.
```
bjankord: nit: looks like there is an extra `.` before the `,`

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: Try this refactor approach. https://github.com/cerner/terra-core/pull/1027/commits/45b44e889215a8dec2e1aed334cdb59ae3670902

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: Broken up here. https://github.com/cerner/terra-core/pull/1027/commits/45b44e889215a8dec2e1aed334cdb59ae3670902

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: Done here. https://github.com/cerner/terra-core/pull/1027/commits/45b44e889215a8dec2e1aed334cdb59ae3670902

packages/terra-time-input/docs/README.md
JakeLaCombe: https://github.com/cerner/terra-core/pull/1027/commits/f7ebb972de0ee5e8d275109b1efba37164c3f2da

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: https://github.com/cerner/terra-core/pull/1027/commits/1a87f172650535bec63b29cb9aa8ee3ce6810f0a

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: Done. https://github.com/cerner/terra-core/pull/1027/commits/a3c2caa1e5915e7ba5d87e054fc695248a528bf3

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: https://github.com/cerner/terra-core/pull/1027/commits/8bcb43ee1321548767fbfd8a1bc2e7b93314931a

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: https://github.com/cerner/terra-core/pull/1027/commits/b8514b2a28b8c91127de27ff45ecf9405a80b7bd

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: Done. https://github.com/cerner/terra-core/pull/1027/commits/ae125fdb415052613a9f94bc6ad315615aa1659d

packages/terra-time-input/src/TimeInput.scss
JakeLaCombe: Done. https://github.com/cerner/terra-core/pull/1027/commits/1d59ce24cd77ec2a62eaeb02e086047c01bdf25a

packages/terra-time-input/src/TimeInput.scss
JakeLaCombe: Done. https://github.com/cerner/terra-core/pull/1027/commits/1d59ce24cd77ec2a62eaeb02e086047c01bdf25a

JakeLaCombe: Added the themeable-variables right here. https://github.com/cerner/terra-core/pull/1027/commits/16e498e27d904c67a851918dbfcae1683bb75c64

packages/terra-time-input/docs/README.md
JakeLaCombe: So for these inputs, pressing a single key character will actually append the 0 if the value is between 2 and 9 for the hour on 23 clocks. Likewise, this same behavior will occur for minutes where the key press is between 5 and 9. Tabbing behavior will only occur for when you type in a 0, 1, or 2.

I can see where the confusion lies, and it's in the fact I didn't specify a single key press. Try this doc out and see if it sounds better. https://github.com/cerner/terra-core/pull/1027/commits/7e58ae016b13c4c56b9d2e1a5c5a7b2434f77986

packages/terra-time-input/docs/README.md
JakeLaCombe: The 00 actually applies to both meridiems. I did specify that inside the documentation at this link, and cleaned up the docs of the behavior of the up and down buttons. See how you like these new docs. https://github.com/cerner/terra-core/pull/1027/commits/a6caa92cb9cc898a87fe7825eacaa44f7663bfac

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: Updated here. https://github.com/cerner/terra-core/pull/1027/commits/ebf13a984c6eae72602c245b3b548232da16ff22

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: Done. https://github.com/cerner/terra-core/pull/1027/commits/8fa1c39d46b47d4e9ef34702f0fa8d50220b8e2d

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: Moved. https://github.com/cerner/terra-core/pull/1027/commits/73c3583547a45c29ebd01fde64887c93f443c486

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: It's in both. It was code left over from last time, so it didn't show up in the check. https://github.com/cerner/terra-core/blob/master/packages/terra-time-input/src/TimeInput.jsx#L129

I'm looking at moving it to TimeUtil, but I can't think of a good method name for this code. It's essentially exiting the function if the value is present, but not numeric. I'll think on it more , but given that it's just a few line in 2 locations, I don't think it's too bad to have a little duplication.

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: Not here actually. For this variable, we want to add 12 to it if the time is not 12 and the meridiem is set to the postMeridiem, or set the tempHour to 0 if the time is 12 and the meridiem is anteMeridiem.

packages/terra-time-input/src/TimeUtil.js
JakeLaCombe: Cleaned up here. https://github.com/cerner/terra-core/pull/1027/commits/0ef896ee9d919d30bdd68359160709c20a303ed0

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: This logic actually isn't dealing with input change. This is the blur function, for when a user has inputted just one digit, and then left the input by clicking off or hitting the tab key. Thus we want to make that we only update the value if the value inside this function is only one character long.

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: The hard part is that one of them is the constructor, which doesn't have the props and context set already, and the other is the component update function, which deals with two sets of props and also having the context attached to the actual component. I gave it a refactoring attempt, but felt that if involved too much passing around of parameters and didn't provide a lot of refactoring value. I'll look at more ideas, but I think it reads better as it is atm.

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: The hard part with this is that the component is made so that it can switch between a 12 hour and 24 hour time, meaning we would be switching between setting and deleting this key. After reading around on react from others in the community, it sounds like it's best practice to instead set the key to undefined if you anticipate needing to set and unset it. https://stackoverflow.com/questions/32884780/how-can-i-remove-an-attribute-from-a-reactjs-components-state-object.

I'll change the default value in the constructor to undefined to better reflect this.

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: Set the default to undefined right here. https://github.com/cerner/terra-core/pull/1027/commits/d0d2cadb3ad5a2456ca691b9f7234ac2ac14396d

JakeLaCombe:
@benbcai Try it now. I've made the height the same as that of the date picker.

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -1,11 +1,143 @@
...
+  }
+
+  /**
+   * Increments a minute to it's next value
+   * @param {String} minute Minute to increment
+   * @return {String} Returns a string representation of the value of the passed in minute after it's incremented
+   */
+  static incrementMinute(minute) {
+    if (minute) {
+      let numericMinute = Number(minute);
```
kschuste: what does this return if the value given is not numeric?

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -1,11 +1,143 @@
...
+  }
+
+  /**
+   * Decrements a minute to it's next value
+   * @param {String} minute Minute to decrement
+   * @return {String} Returns a string representation of the value of the passed in minute after it's decremented
+   */
+  static decrementMinute(minute) {
+    if (minute) {
+      let numericMinute = Number(minute);
```
kschuste: same as above comment on checking if is numeric

packages/terra-time-input/src/TimeUtil.js
JakeLaCombe: Right now, the TimeInput doesn't recognize the value that are not in the format hh:mm has a whole, and resets it to blank when it's initialized with an invalid time value, so I think it's unlikely that this scenario will occur. https://github.com/cerner/terra-core/pull/1027/files/36678e4d3b233e232d19fd7d1565e17ff5e649e6#diff-662e6b853064e84accd761052bfc24c7R78

If you do want to have a check, I propose that if the minute is not a number after parsing, that we just reset the value to 00 or 12, as that's usually the default value for a clock.

benbcai:
@JakeLaCombe The height looks good!

packages/terra-time-input/src/TimeUtil.js
kschuste: I'm happy with that it doesn't set the value as that is what I would expect us to do in this case.

packages/terra-time-input/translations/fi-FI.json
```diff
@@ -0,0 +1,6 @@
+{
```
emilyrohrbough: Looks like fr-FR and fr needs filled with mock translations

packages/terra-time-input/tests/nightwatch/time-input-twelve-hour-spec.js
```diff
@@ -0,0 +1,168 @@
...
+
+    browser.click('#timeInput input[name="terra-time-hour-time-input"]');
+
+    browser.keys('0012p');
+    browser.expect.element('#timeInput input[name="terra-time-hour-time-input"]').to.have.attribute('value').equals('12');
+    browser.expect.element('#timeInput input[name="terra-time-minute-time-input"]').to.have.attribute('value').equals('12');
+    browser.expect.element('#timeInput select[name="terra-time-meridiem-time-input"]').to.have.attribute('value').equals('p.m.');
+  },
+
+  'Time input sets the hour to 12 without changing meridiem when up is pressed on the hour input': (browser) => {
```
emilyrohrbough: Is this test for a 24 hour clock?

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -34,4 +170,44 @@ class TimeUtil {
...
+    'ontouchstart' in window ||
+    // eslint-disable-next-line no-undef
+    (window.DocumentTouch && document instanceof DocumentTouch) ||
+    navigator.maxTouchPoints > 0 ||
+    navigator.msMaxTouchPoints > 0
+  );
+
+Object.defineProperty(TimeUtil, 'FORMAT_12_HOUR', {
+  value: '12-hour',
+  configurable: false,
```
emilyrohrbough: Could we document why we define configurable, enumerable, writable? They are the same for '12-hour' & '24-hour'

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -23,6 +155,10 @@ class TimeUtil {
       if (hourAndMinute.length > 1) {
         const minute = Number(hourAndMinute[1]);
         if (minute >= 0 && minute < 60) {
+          if (hourAndMinute[1].length === 1) {
```
emilyrohrbough: Could you use `minute` here instead of `hourAndMinute[1]`?

packages/terra-time-input/src/TimeUtil.js
```diff
@@ -1,11 +1,143 @@
...
+  }
+
   static splitHour(time) {
     if (typeof (time) === 'string') {
       const hourAndMinute = time.split(':');
 
       if (hourAndMinute.length) {
         const hour = Number(hourAndMinute[0]);
         if (hour >= 0 && hour < 24) {
+          if (hourAndMinute[0].length === 1) {
```
emilyrohrbough: Could you use `hour ` here instead of `hourAndMinute[0]`?

emilyrohrbough: Should this be moved before ` const hour = Number(hourAndMinute[0]);`?

packages/terra-time-input/translations/fi-FI.json
JakeLaCombe: I'll be getting a translation Jira logged for this component. I was waiting for feedback on this review before sending it out, but I think I have enough to get going. I'll get mocks added for these individual files in the mean time.

packages/terra-time-input/src/TimeInput.jsx
emilyrohrbough: possible name could be `validateNumericInput`

packages/terra-time-input/src/TimeUtil.js
JakeLaCombe: The one thing I like about it being here is that the string is only concatenated if the hour is between 0 and 23. If we move it up, we could be doing concatenation when it's not necessary.

packages/terra-time-input/src/TimeUtil.js
JakeLaCombe: https://github.com/cerner/terra-core/pull/1027/commits/8d993b43e0232da22d729afb55fc6225dd22b2e6

packages/terra-time-input/tests/nightwatch/TimeInputTests.jsx
```diff
@@ -9,6 +9,13 @@ const TimeInputTests = () => (
       <li><Link to="/tests/time-input-tests/default">TimeInput - Default</Link></li>
       <li><Link to="/tests/time-input-tests/default-time">TimeInput - Default Time</Link></li>
       <li><Link to="/tests/time-input-tests/on-change">TimeInput - OnChange</Link></li>
+      <li><Link to="/tests/time-input-tests/invalid-time">TimeInput - Invalid Time</Link></li>
+      <li><Link to="/tests/time-input-tests/twelve-hour">TimeInput - Twelve Hour</Link></li>
+      <li><Link to="/tests/time-input-tests/twelve-hour-filled-morning">TimeInput - Twelve Hour Filled Morning</Link></li>
```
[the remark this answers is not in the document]
JakeLaCombe: https://github.com/cerner/terra-core/pull/1027/commits/8d993b43e0232da22d729afb55fc6225dd22b2e6

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -116,153 +190,393 @@ class TimeInput extends React.Component {
...
+          name={name}
+          value={timeValue}
+        />
+        <div className={cx('time-input-group')}>
+          <Input
+            {...inputAttributes}
+            {...hourAttributes}
+            ref={(inputRef) => { this.hourInput = inputRef; }}
+            className={cx('time-input-hour')}
+            type="tel"
```
benbcai: What is the reason for making this a tel input type instead of text?

JakeLaCombe: For mobile devices, UX wants to use just the numeric pad.

benbcai: Does setting the pattern prop on lines 517 and 540 not work? This prop was added by this [change](https://github.com/cerner/terra-core/pull/365/commits/336eb59538fd24e378d3041bd35062d8a6a41921) to use the numeric keyboard on mobile devices. I think the pattern prop can be removed if setting the type to tel also works.

JakeLaCombe: The pattern does do the trick. I do see that tel gives us additional functionality on the keyboard, but I don't see the need for +, * and # values. I'll sync with UX on what they like.

packages/terra-time-input/translations/fi-FI.json
JakeLaCombe: Added. https://github.com/cerner/terra-core/pull/1027/commits/5073f9f2f68e92dc3b20b5a431b9e8f026b8b78e

packages/terra-time-input/src/TimeInput.scss
```diff
@@ -1,40 +1,212 @@
...
+
+  .meridiem-display {
+    border-color: transparent;
+    border-radius: var(--terra-time-input-border-radius, 3px);
+    border-style: var(--terra-time-input-keyboard-border-style, solid);
+    border-width: var(--terra-time-input-keyboard-border-width, 1px);
+    display: inline-block;
+    line-height: var(--terra-time-input-meridiem-display-line-height, 1.429);
+    margin-left: var(--terra-time-input-meridiem-display-left-margin, 0.357em);
+
```
yuderekyu: The focus box for the `meridiem-display` has different spacing than the input value's focus box - if that isn't your intention, adding some padding should put some space.

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: I decided to remove the tel for now. It gives us the control we need, and I don't see the need for those other inputs that appear. https://github.com/cerner/terra-core/pull/1027/commits/6369b5318e8d2e8b3deaae1332d0ff0369498f90

packages/terra-time-input/src/TimeInput.scss
JakeLaCombe: Try it now. I just pushed new styles up. https://terra-core-deployed-pr-1027.herokuapp.com/

yuderekyu: Hmm looks like it's still off (currently using chrome):
screen shot 2017-11-30 at 3 33 43 pm
screen shot 2017-11-30 at 3 33 56 pm

JakeLaCombe: Thats why. I didn't push it like I thought I did :(

yuderekyu: Should the default values add some padding in the first place?

bjankord:
@mmalaker This is now ready for functional review.

packages/terra-time-input/src/TimeInput.scss
JakeLaCombe: Paddings are removed. As for the other styles with padding, we do need them so that the components are able to be aligned and sized correctly.

bjankord:
Some of the recent changes seem to have impacted the height of the time input component compared to the date input:

screen shot 2017-12-05 at 1 07 24 pm

cc/ @JakeLaCombe @neilpfeiffer

JakeLaCombe:
@bjankord they should be fixed in the latest commit. It's the best I can get it to ensure that the heights of both date and time pickers are the same

mmalaker:
Functional verification completed. The time input examples display and function correctly with the latest updates.

bjankord:
@neilpfeiffer @mndrsn This is ready for UX verification.

neilpfeiffer:
Tested: PR #1027
Resolves: Issue #1008

**Reviewed:**
- Follows pattern considerations documented in Cerner Design Standards: 
[Time Input](https://wiki.ucern.com/x/EWUVYQ) _(see stated variations†)_
- Follows UX defined design implementation for Terra-Time-Input

\
**Future Enhancements: †Variations from Cerner Design Standards:**
- Does not currently have a quick method for entering or resetting the time to "Now". Can enhance with keyboard "N" key press (T-logic).
- Does not have the ability to enter time in seconds. Will wait to enhance until request needs functionality.

\
**Future Enhancements: Known Development Issues:**
- Safari: clicking/focusing into the input does not highlight/fully-select the input value as it does in the other browsers, instead does just the normal text cursor. Low Risk - Listing as a known issue and try to solve with future enhancment
- IE10: the custom selection highlight color is not being honored, so selecting the hours and minutes values when clicking/focusing into the inputs shows two highlights: one blue background color and the OS default color. Low Risk - Listing as known issue and may not solve and just wait for IE10 to no longer be supported.

\
\

**Pass** |
-----------|
UX Review|

Issue #1003, opened 2017-11-08
Popup overlay doesn't block touches


noahbenham:
# Issue Description

We're using Popup, but touches outside the bounds of the popup content appear to work differently on desktop vs mobile (tested iOS & Android).

[mobile_examples.zip](https://github.com/cerner/terra-core/files/1455544/mobile_examples.zip)

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior

Touches outside the popup content area, or at the least on the triggering button, should dismiss the popup only and not register as a click event. Works as expected on desktop Chrome & Safari.

## Current Behavior

Touches outside the content area both close the Popup and trigger onClick events on the underlying content. Defect is present on mobile Android & iOS.

## Steps to Reproduce

N/A - described above

## Environment

* Component Version: Popup 1.13.0
* Browser Name and Version: iOS 10.3.3 (14G60) Safari
* Operating System and version (desktop or mobile): Mobile

PR #1029, opened 2017-11-20
Prevent Default on Outside Click Events.

emilyrohrbough:
### Summary
Closes https://github.com/cerner/terra-core/issues/1003

### Verification
To verify this is working as expected do the following:

1. Start up a mobile device or the simulator. Go to the http://engineering.cerner.com/terra-core/#/site/popup and click around on the popup attachments buttons. Notice the behavior is to continuously trigger the next popup. 
    - If you check this on a desktop browser, the behavior is the popup will close and an additional click is needed to trigger the next action on the page.
    - This is the bug in # 1003

2. Now on the simulator, go to https://terra-core-deployed-pr-1029.herokuapp.com/static/#/site/popup. The behavior is now consistent with the desktop browser behavior. 
    - If you check the desktop browser, the behavior remains the same.


packages/terra-hookshot/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ Changelog
 
 Unreleased
 ----------
+* Add stopDefault() to onOutsideClick
```
kschuste: could add a 
### Changed or
### Added
header above this

packages/terra-hookshot/src/HookshotContent.jsx
```diff
@@ -77,6 +77,7 @@ class HookshotContent extends React.Component {
 
   handleClickOutside(event) {
     if (this.props.onOutsideClick) {
+      event.preventDefault();
```
kschuste: is there a way to add a test for this?

packages/terra-hookshot/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Add stopDefault() to onOutsideClick
```
bjankord: nit: change `stopDefault` to `preventDefault`

packages/terra-hookshot/src/HookshotContent.jsx
bjankord: Are there cases where we wouldn't want to preventDefault? If so we might want to look at adding a prop to control this. I can't think of any cases at this time.

mhemesath: So if i click outside of an open popup it will prevent default behavior on that click? I actually can't think of a situation when that is desirable?  Example:

* Clicking and focusing on a form input
* Clicking a link
* Clicking on another thing to trigger a popup

In all these cases I'd want hookshot to disappear and the action i triggered to just happen. IMO if we need to prevent outside clicks, that should be done by a overlay that's paired with this.

mhemesath:
If this fix is explicitly for #1003 then I propose the change is to create a conditional overlay for popovers on mobile. This current fix will actually break expected behavior in the referenced PR on desktop screen sizes.

dkasper-was-taken:
@mhemesath Currently clicking outside on the desktop does not trigger additional actions behind the popup, only a dismissal occurs. Mobile being the inconsistency, which could be a difference in behavior with the use of the focus trap.

packages/terra-hookshot/src/HookshotContent.jsx
JakeLaCombe: I think the only thing you could really add is a jest test, and verify that preventDefault was called when handleClickOutside is triggered.

emilyrohrbough: To verify this is working as expected do the following:


1. Start up a mobile device or the simulator. Go to the http://engineering.cerner.com/terra-core/#/site/popup and click around on the popup attachments buttons. Notice the behavior is to continuously trigger the next popup. 
- If you check this on a desktop browser, the behavior is the popup will close and an additional click is needed to trigger the next action on the page.
- This is the bug in # 1003

2. Now on the simulator, go to https://terra-core-deployed-pr-1029.herokuapp.com/static/#/site/popup. The behavior is now consistent with the desktop browser behavior. 
- If you check the desktop browser, the behavior remains the same.

bjankord:
@mmalaker This is now ready for functional verification.

mmalaker:
Functional verification completed. The popup examples no longer fire click events for triggers outside the popup on mobile. Behavior is consistent with desktop browsers, as expected.

Issue #1026, opened 2017-11-17
Collapsible menu will roll up 1 button


hboege:
# Issue Description
Collapsible menu will roll up 1 button with an icon into an ellipsis button.  This is not helpful, as it does not reduce the amount of space the buttons consume, and it requires the user to click twice, instead of once.

![rollupbutton](https://user-images.githubusercontent.com/30238809/32972163-69a56688-cbb6-11e7-8b05-c185104dd90f.jpg)
In the above image, the Left is IE11, right is Chrome

## Issue Type
- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior
If the one and only button to be put into the ellipsis is already icon-only, do not roll up the button.

## Current Behavior
The one and only button being moved into the ellipsis button is icon-only.  No space is saved.

## Environment

* Component Version: 1.5.0
* Browser Name and Version: IE 11 and Chrome
* Operating System and version (desktop or mobile): desktop Microsoft.

PR #1031, opened 2017-11-20
Show First Hidden Child if the Child has isIconOnly Prop

emilyrohrbough:
### Summary
Closes https://github.com/cerner/terra-core/issues/1026.

Updated the ellipse menu to show first hidden child if the child is an icon-only item, otherwise roll up item as expected.

@cerner/terra


packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -91,11 +91,26 @@ class CollapsibleMenuView extends React.Component {
   render() {
     const { children, boundingRef, menuWidth, ...customProps } = this.props;
     const visibleChildren = React.Children.toArray(children);
+
     let hiddenChildren = null;
+    let singleHiddenIconChild = false;
     if (this.state.hiddenStartIndex >= 0) {
       hiddenChildren = visibleChildren.splice(this.state.hiddenStartIndex);
+
+      if (hiddenChildren.length === 1 && hiddenChildren[0].props.isIconOnly) {
```
JakeLaCombe: Nitpick:

I think you can reduce this if statement to the following

````
singleHiddenIconChild  hiddenChildren.length === 1 && hiddenChildren[0].props.isIconOnly.
````

File: packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx [hunk 2]
```diff
@@ -91,11 +91,23 @@ class CollapsibleMenuView extends React.Component {
   render() {
     const { children, boundingRef, menuWidth, ...customProps } = this.props;
     const visibleChildren = React.Children.toArray(children);
+
     let hiddenChildren = null;
+    let singleHiddenIconChild = false;
     if (this.state.hiddenStartIndex >= 0) {
       hiddenChildren = visibleChildren.splice(this.state.hiddenStartIndex);
+      singleHiddenIconChild  = hiddenChildren.length === 1 && hiddenChildren[0].props.isIconOnly;
```
kschuste: look like have an two spaces before the `=` here

emilyrohrbough: Dang. Knew I shouldn't have updated via the browser 😆

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -110,12 +122,7 @@ class CollapsibleMenuView extends React.Component {
       <div {...customProps} className={collapsibleMenuViewClassName} ref={this.setContainer}>
         {visibleChildren}
```
kschuste: wouldn't it make more sense to put it back in the visibleChildren array to be displayed here rather than have it display in the menu div

emilyrohrbough: ...yeah probably

emilyrohrbough: So after taking a look with how the width is calculated, some logic would need to be added to determine if the ellipse menu button should show or not. Given this, I think this will be a cleaner implementation.

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -91,11 +91,23 @@ class CollapsibleMenuView extends React.Component {
   render() {
     const { children, boundingRef, menuWidth, ...customProps } = this.props;
     const visibleChildren = React.Children.toArray(children);
+
     let hiddenChildren = null;
+    let singleHiddenIconChild = false;
     if (this.state.hiddenStartIndex >= 0) {
       hiddenChildren = visibleChildren.splice(this.state.hiddenStartIndex);
+      singleHiddenIconChild = hiddenChildren.length === 1 && hiddenChildren[0].props.isIconOnly;
```
jmsv6d: We shouldn't be basing this logic on the isIconOnly prop. We should just update with width calculations to determine if there is enough room for the last item if the menu is hidden.

emilyrohrbough: That would be better. I updated the changes.

packages/terra-collapsible-menu-view/src/CollapsibleMenuView.jsx
```diff
@@ -77,8 +77,15 @@ class CollapsibleMenuView extends React.Component {
 
     for (let i = 0; i < this.props.children.length; i += 1) {
       const child = this.container.children[i];
-      calcWidth += child.getBoundingClientRect().width;
+      const childWidth = child.getBoundingClientRect().width;
+      calcWidth += childWidth;
+
       if (calcWidth > availableWidth) {
+        // If last child is the same size as the menu button, leave face up
+        if (i === this.props.children.length - 1 && childWidth === menuButtonWidth) {
```
jmsv6d: While this solves the specific scenario that was logged, it doesn't really solve the entire issue. This would still be broken in the scenario where a child button is larger than the menu button, but could stil fit in the available space if the menu was hidden. The logic should be something like:
`if (i === this.props.children.length - 1 && calcWidth <= availableWidth + menuButtonWidth)`

mmalaker:
The [test example](https://terra-core-deployed-pr-1031.herokuapp.com/static/#/tests/collapsible-menu-view-tests/first-item-hidden) does not collapse, since it only has 3 buttons and all display at the smallest breakpoint. What happens if you have 10 buttons in the toolbar and the first one is an icon? Does the toolbar become non-collapsible in this case?

emilyrohrbough:
@mmalaker Sorry Mike! I had my dev tools open when inspecting the test example which was why I was able to shrink the example to be small enough. I updated the example so it will collapse on a browser.

mmalaker:
Functional verification completed. The collapsible menu examples display and function correctly.

PR #1036, opened 2017-11-27
update i18n-plugin readme

zhongyn:
### Summary
`node_modules` is searched by default, so remove it from the doc example

@cerner/terra


kschuste:
@zhongyn please allow someone from the terra team merge the code next time

mjhenkes:
@kschuste @zhongyn is on the Terra team https://github.com/orgs/cerner/teams/terra/members. The terra team is the only team with merge privileges to master. We have some maybe-documented tribal knowledge about waiting for enough reviewers from both terra-core and terra-clinical teams before merging items into terra core, but it's unlikely @zhongyn would have known about that. If we really want to restrict this we could break up the the clinical, consumer, and core teams and restrict merge access at a more granular level.

Issue #1002, opened 2017-11-08
CareAware Connect Call Forwarding issue 


heyvdi:
# Issue Description

<img width="445" alt="screen shot 2017-11-08 at 2 41 58 pm" src="https://user-images.githubusercontent.com/17009526/32573959-63613612-c495-11e7-98f8-6cd8af3c6665.png">
<img width="442" alt="screen shot 2017-11-08 at 2 41 38 pm" src="https://user-images.githubusercontent.com/17009526/32573963-65be8ea0-c495-11e7-849a-fb4a3df569c0.png">
The Terra platform doesn't provide the consistent UI design in the iOS platform (Please see the iOS screenshots above). The check box design and the margins in the form page are off from the original design. 

## Issue Type

- [ ] New Feature
- [x ] Enhancement
- [] Bug
- [ ] Other

## Expected Behavior

The original designs are attached below: 
![5c copy 78](https://user-images.githubusercontent.com/17009526/32574304-67b081ae-c496-11e7-877d-ef6c0ccad3c7.png)
![5c copy 108](https://user-images.githubusercontent.com/17009526/32574307-69d7dd9c-c496-11e7-9126-11830105f101.png)

## Environment

* Component Version:
* Browser Name and Version: iOS platform 
* Operating System and version (desktop or mobile): mobile

JakeLaCombe:
That's a really light gray for the text color. I imagine that color will be changing most likely.

For the checkbox, that is under way in https://github.com/cerner/terra-core/issues/999.

Finally, the field is being redone in https://github.com/cerner/terra-core/issues/889

mjhenkes:
Closing based on @JakeLaCombe feedback.

PR #1037, opened 2017-11-27
Checkbox OCS

yuderekyu:
### Summary
Updated PR of #1024. Creates a checkbox component with themeable variables for OCS. The component is responsive and allows for text wrapping. Focus can be triggered with tab; toggle can be triggered with the space bar. Defaults to terra open source theme, which uses browser native styling for the input. Themes not using browser native styling utilize a css checkbox and hide the input element.

### Addresses
#999 
#925 (for checkboxes)
#1002

### Deployment Links
* https://terra-core-deployed-pr-1037.herokuapp.com/static/#/site/form-checkbox
* https://terra-core-deployed-pr-1037.herokuapp.com/static/#/tests/form-checkbox-tests


packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+  isLabelHidden: false,
+  labelText: null,
+  labelTextAttrs: {},
+  name: null,
+  onChange: undefined,
+  value: undefined,
+};
+
+class Checkbox extends React.Component {
+  constructor(props, context) {
```
jmsv6d: You don't have to include context here since you're not using it.

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+  /**
+    * The checked property of the Input element. Use this to generate a controlled Checkbox Element.
+    */
+  checked: PropTypes.bool,
+  /**
+    * The checked property of the Input element. Use this to generate an uncontrolled Checkbox Element.
+    */
+  defaultChecked: PropTypes.bool,
+  /**
+    * The id of the input field.
```
JakeLaCombe: Might as well say checkbox at this line as this is what this component is.

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+  /**
+   * Whether the checkbox element is disabled or not.
+   */
+  isDisabled: PropTypes.bool,
+  /**
+    * Whether the checkbox element is inline or not.
+    */
+  isInline: PropTypes.bool,
+  /**
+    * Whether the label is hidden or not.
```
JakeLaCombe: Nitpick:

The or not feels extraneous on these comments. It feels like saying /*Whether the label is hidden*/ feels sufficient enough.

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+  onFocus() {
+    this.setState({ focus: true });
+    if (this.inputAttrs !== undefined && this.inputAttrs.onFocus !== undefined) {
+      this.inputAttrs.onFocus();
+    }
+  }
+
+  onBlur() {
+    this.setState({ focus: false });
+    if (this.inputAttrs !== undefined && this.inputAttrs.onBlur !== undefined) {
```
JakeLaCombe: Can we have onBlur as a prop on the component instead of putting it on the inputAttrs?

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+      labelTextAttrs.className,
+    ]);
+
+    const inputClasses = cx([
+      'native-input',
+      inputAttrs.className,
+    ]);
+
+    const innerSpanClasses = cx([
+      'inner-span',
```
jmsv6d: Maybe instead of `inner-span` something like `label-text`

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    const checkboxClasses = cx([
+      'checkbox',
+      { 'checkbox-inline': isInline },
+      customProps.className,
+    ]);
+
+    const innerDiv = cx([
+      'inner-div',
```
jmsv6d: `inner-div` doesn't really tell you anything about what this div is for. Maybe something like `focus-div`?

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+      { focus: this.state.focus },
+    ]);
+
+    const labelClasses = cx([
+      'label',
+      { 'label-disabled': isDisabled },
```
jmsv6d: Change class to `is-disabled`. `label-disabled` makes me think that only the label is getting disabled.

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,153 @@
...
+    /* Hitbox and container focus style will only surround checkbox + label, instead of the div block */
+    /* Padding puts space for container focus style */
+    .inner-div {
+      display: inline-block;
+      padding-left: var(--terra-form-checkbox-inner-div-padding-left, 0.214rem);
+      padding-right: var(--terra-form-checkboc-innter-div-padding-right, 0.214rem);
+    }
+
+    /* Hide label if isLabelHidden true */
+    .label-hidden {
```
jmsv6d: Instead of using css to hide the text we could just not render the text at all and set it as the aria-label attribute on the input.

packages/terra-form-checkbox/docs/README.md
```diff
@@ -0,0 +1,26 @@
+# Terra Form Checkbox
+
+The Terra Form Checkbox is a responsive input component rendered as a box. When activated, a check mark shall appear. Focus can be activated through tabbing and the checked state can be toggled with the space bar.
```
emilyrohrbough: nit: rendered as a box next to the input value? 

Just rendered as a box seems a bit weird

packages/terra-form-checkbox/docs/README.md
```diff
@@ -0,0 +1,26 @@
...
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-form-checkbox`
+
+## Usage
+
+```jsx
+import React from 'react';
+import Checkbox from 'terra-form-checkbox';
+
+<Checkbox {id="checkbox" labelText="Default Checkbox" } />
```
emilyrohrbough: remove the { and }

packages/terra-form-checkbox/docs/README.md
```diff
@@ -0,0 +1,26 @@
...
+```
+
+## Component Features
+ * [Cross-Browser Support](https://github.com/cerner/terra-core/wiki/Component-Features#cross-browser-support)
+ * [Responsive Support](https://github.com/cerner/terra-core/wiki/Component-Features#responsive-support)
+ * [Mobile Support](https://github.com/cerner/terra-core/wiki/Component-Features#mobile-support)
+ * [Internationalization Support](https://github.com/cerner/terra-core/wiki/Component-Features#internationalization-i18n-support)
```
emilyrohrbough: I don't think this supports Internationalization or Localization

packages/terra-form-checkbox/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "terra-mixins": "^1.11.1"
+  },
+  "scripts": {
+    "compile": "npm run compile:clean && npm run compile:build",
+    "compile:clean": "rimraf lib",
+    "compile:build": "babel src --out-dir lib --copy-files",
+    "lint": "npm run lint:js && npm run lint:scss",
+    "lint:js": "eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
+    "lint:scss": "stylelint src/**/*.scss",
+    "props-table": "props-table ./src/DateTimePicker.jsx --out-dir ./docs/props-table",
```
emilyrohrbough: Update props-table script to use `./scr/Checkbox.jsx`

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+              id={id}
+              disabled={isDisabled}
+              name={name}
+              value={value}
+              onChange={onChange}
+              onFocus={this.onFocus}
+              onBlur={this.onBlur}
+              className={inputClasses}
+            />
+            <span className={innerSpanClasses} > {labelText} </span>
```
jmsv6d: Nit: remove spaces around `{labelText}` and before `>`

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+
+    const innerSpanClasses = cx([
+      'inner-span',
+      { 'label-hidden': isLabelHidden },
+    ]);
+
+    return (
+      <div className={checkboxClasses}>
+        <div className={innerDiv}>
+          <label htmlFor={id} className={labelClasses} >
```
jmsv6d: Nit: remove space before `>`

packages/terra-form-checkbox/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "terra-form-checkbox",
+    "FormCheckbox",
+    "UI"
+  ],
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
```
emilyrohrbough: add
```
  "devDependencies": {
    "terra-props-table": "^1.10.1"
  }
```

packages/terra-form-checkbox/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "peerDependencies": {
+    "react": "^15.4.2",
+    "react-dom": "^15.4.2",
+    "terra-base": "^2.6.1"
+  },
+  "dependencies": {
```
emilyrohrbough: Remove react & react-dom from dependencies

packages/terra-form-checkbox/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "terra-form-checkbox",
+  "main": "lib/FormCheckbox.js",
```
emilyrohrbough: looks like this needs updated to `"main": "lib/Checkbox.js",`

packages/terra-form-checkbox/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "description": "The Terra Form Checkbox is a responsive input component rendered as a box. When activated, a check mark shall appear. Focus can be activated through tabbing and the checked state can be toggled with the space bar.",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/cerner/terra-core.git"
+  },
+  "keywords": [
+    "Cerner",
+    "Terra",
+    "terra-form-checkbox",
+    "FormCheckbox",
```
emilyrohrbough: Maybe this should be updated as well to `Checkbox`

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+import 'terra-base/lib/baseStyles';
+import styles from './Checkbox.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+    * The checked property of the Input element. Use this to generate a controlled Checkbox Element.
+    */
+  checked: PropTypes.bool,
```
emilyrohrbough: Is checked a first-hand input attribute? If not, I feel like this should be isChecked.

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import 'terra-base/lib/baseStyles';
+import styles from './Checkbox.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+    * The checked property of the Input element. Use this to generate a controlled Checkbox Element.
```
emilyrohrbough: Nit: Whether or not the checkbox element is checked

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+   */
+  isDisabled: PropTypes.bool,
+  /**
+    * Whether the checkbox element is inline or not.
+    */
+  isInline: PropTypes.bool,
+  /**
+    * Whether the label is hidden or not.
+    */
+  isLabelHidden: PropTypes.bool,
```
emilyrohrbough: I'm curious, what is the desire for this prop? When would one desire a hidden label?

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,153 @@
...
+      border-width: var(--terra-form-checkbox-border-width, none);
+      content: '';
+      cursor: pointer;
+      display: inline-block;
+      height: var(--terra-form-checkbox-height, none);
+      left: 0;
+      margin-top: var(--terra-form-checkbox-margin-top, none);
+      position: absolute;
+      top: 0;
+      visibility: visible;
```
jmsv6d: `visible` is the default value for `visibility`. You shouldn't need to set it here.

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,153 @@
...
+      height: var(--terra-form-checkbox-checked-height, none);
+      left: 0;
+      line-height: 1em;
+      margin-top: var(--terra-form-checkbox-margin-top, none);
+      opacity: 1;
+      position: absolute;
+      text-align: center;
+      top: 0;
+      transform: scale(1);
+      transition: all 0.2s ease all;
```
jmsv6d: Transition should probably be themeable.

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+    */
+  onChange: PropTypes.func,
+  /**
+    * The value of the input element.
+    */
+  value: PropTypes.string,
+};
+
+const defaultProps = {
+  checked: undefined,
```
emilyrohrbough: Since checked and defaultChecked are booleans, these should be false, not undefined

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+      onChange,
+      value,
+      ...customProps
+    } = this.props;
+
+    const controlInputAttrs = Object.assign({}, inputAttrs);
+
+    if (checked !== undefined) {
+      controlInputAttrs.checked = checked;
+    } else {
```
emilyrohrbough: What happens if both are undefined?

packages/terra-form-checkbox/src/Checkbox.jsx
yuderekyu: [Yeah](https://www.w3schools.com/tags/att_input_checked.asp).

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,153 @@
...
+      cursor: default;
+      visibility: hidden;
+    }
+
+    /* stylelint-disable selector-max-compound-selectors */
+    .label {
+      cursor: pointer;
+      display: flex;
+      font-size: var(--terra-form-checkbox-font-size, 1rem);
+      line-height: 1.42em;
```
jmsv6d: Line-height should be themeable as it goes hand-in-hand with font-size.

packages/terra-form-checkbox/src/Checkbox.jsx
yuderekyu: Checkout #973.

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,153 @@
...
+      padding-left: var(--terra-form-checkbox-padding-left, none);
+      position: relative;
+    }
+
+    /* Native checkbox */
```
emilyrohrbough: why do we define styles for native and custom checkbox? Which browser native indicate here?

packages/terra-form-checkbox/src/Checkbox.scss
jmsv6d: Also line-height should be unit-less.

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /* Custom checkbox */
+    .native-input + .inner-span::before {
+      background-color: var(--terra-form-checkbox-background-color, none);
+      background-image: var(--terra-form-checkbox-background-image, none);
+      background-position: center center;
+      background-repeat: no-repeat;
+      border-color: var(--terra-form-checkbox-border-color, none);
+      border-radius: var(--terra-form-checkbox-border-radius, none);
```
emilyrohrbough: Did you by chance check these styles with a border-radius that wasn't none? Did that affect the focus states?

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,153 @@
...
+      background-color: var(--terra-form-checkbox-checked-background-color, none);
+      border-color: var(--terra-form-checkbox-checked-border-color, none);
+      border-radius: var(--terra-form-checkbox-checked-border-radius, none);
+      border-style: var(--terra-form-checkbox-checked-border-style, none);
+      border-width: var(--terra-form-checkbox-checked-border-width, none);
+      color: var(--terra-form-checkbox-symbol-color, none);
+      content: var(--terra-form-checkbox-symbol, none);
+      height: var(--terra-form-checkbox-checked-height, none);
+      left: 0;
+      line-height: 1em;
```
jmsv6d: Line-height of 1 is usually bad, you should check with @neilpfeiffer on this.

packages/terra-form-checkbox/src/Checkbox.jsx
JakeLaCombe: I believe that we need these set to undefined to start with so that we can distinguish between a controlled and an uncontrolled checkbox. defaultChecked maybe ok to set to false, but we will need to make sure checked has an actual value passed in to verify the component should be controlled.

yuderekyu: The logic for this is based on discussion within the old control component, [Control.jsx](https://github.com/cerner/terra-core/pull/448/files), that you were a part of hehe.

packages/terra-form-checkbox/src/Checkbox.jsx
JakeLaCombe: It should be fine in this case, as the checkbo should default to being uncontrolled.

packages/terra-form-checkbox/src/Checkbox.jsx
JakeLaCombe: Make it more generic as well so that a class name doesn't reference actual markup. I'd vote for focus-container.

packages/terra-form-checkbox/src/Checkbox.jsx
yuderekyu: The logic for this is based on discussion within the old control component, [Control.jsx](https://github.com/cerner/terra-core/pull/448/files).

packages/terra-form-checkbox/src/Checkbox.jsx
JakeLaCombe: In addition, it follows better stateful ideas of CSS as well. Here's a great resource on the convention. https://smacss.com/book/type-state

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    .inner-span {
+      margin-left: var(--terra-form-checkbox-inner-span-margin-left, 1.4em);
+    }
+
+    /* Custom checkbox */
+    .native-input + .inner-span::before {
+      background-color: var(--terra-form-checkbox-background-color, none);
+      background-image: var(--terra-form-checkbox-background-image, none);
```
JakeLaCombe: Can't tell from the mockups, but what is the background-image used for? I didn't see the animation bar inside the original markups.

packages/terra-form-checkbox/src/Checkbox.scss
yuderekyu: Most browsers display the `input` element with browser native styling, no matter what styles you apply. So, for the themes that utilize browser native styling,`input` gets displayed. On themes that want custom checkbox styling, `input` gets hidden (but is still access-able), and the checkbox styling gets applied to a css checkbox next to the label text.

packages/terra-form-checkbox/src/Checkbox.scss
yuderekyu: One of the specified styles needed to be themed was`Button Images`, which I took being able display an image within the checkbox, although none of the themes contain an image. @neilpfeiffer thoughts on this requirement? Definitely ok with removing this, we already have background color.

packages/terra-form-checkbox/src/Checkbox.scss
emilyrohrbough: ^ to add to this, line-height should be unitless

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,153 @@
...
+      display: flex;
+      font-size: var(--terra-form-checkbox-font-size, 1rem);
+      line-height: 1.42em;
+      padding-left: var(--terra-form-checkbox-padding-left, none);
+      position: relative;
+    }
+
+    /* Native checkbox */
+    .native-input {
+      border: var(--terra-form-checkbox-native-input-border, none);
```
JakeLaCombe: Can you spread these properties to be their explicit equivalences? (border-color, border-width, border-style)

packages/terra-form-checkbox/src/Checkbox.scss
emilyrohrbough: Could you add an comment indicating this behavior?

packages/terra-form-checkbox/src/Checkbox.jsx
yuderekyu: I agree with you, it does. I'm following the existing comment structure from other components, which have *Whether or not* for boolean props. I'm ok with switching, anyone else have thoughts?

jmsv6d: We don't really have a convention set for that. Personally I go with "Indicates if label is hidden."

packages/terra-form-checkbox/src/Checkbox.jsx
yuderekyu: Going along the same lines, I'm assuming onFocus should also be a prop

packages/terra-form-checkbox/src/Checkbox.scss
yuderekyu: Yeah, box shadow follows the container's shape - if border-radius is set, box shadow reflects the rounded corners.

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+  /**
+    * The id of the input field.
+    */
+  id: PropTypes.string,
+  /**
+    * Additional attributes for the input object.
+    */
+  inputAttrs: PropTypes.object,
+  /**
+   * Whether the checkbox element is disabled or not.
```
StephenEsser: The comment asterisk indentation here is inconsistent

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+    */
+  isInline: PropTypes.bool,
+  /**
+    * Whether the label is hidden or not.
+    */
+  isLabelHidden: PropTypes.bool,
+  /**
+    * Text of the label.
+    */
+  labelText: PropTypes.node.isRequired,
```
StephenEsser: Any particular reason this is `labelText` instead of [`label`](https://github.com/cerner/terra-core/blob/master/packages/terra-form-field/src/Field.jsx#L50) or [`text`](https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L49)?

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,176 @@
...
+  /**
+    * Additional objects for the text object.
+    */
+  labelTextAttrs: PropTypes.object,
```
StephenEsser: If consumers want to pass in additional attributes for the label should they pass in a node? ( Do we need this prop or should we remove it?) 

<Checkbox label={<span style={{ color: blue }} />

If we do keep this prop, and I may have missed it somewhere, it looks like we are only honoring the className passed in through labelTextAttrs, any additional key values pairs are discarded.

packages/terra-form-checkbox/docs/README.md
bjankord: True, we can drop those bullet points from here.

packages/terra-form-checkbox/src/Checkbox.jsx
bjankord: I believe this just came from what we had before for the [control component](https://github.com/cerner/terra-core/blob/master/packages/terra-form/src/Control.jsx#L40)

packages/terra-form-checkbox/src/Checkbox.scss
bjankord: Possibly, I lean more towards `<label>some text visually hidden with css</label>` than `<label aria-label="some text in aria label"></label>`. While both are similar, I think the first will have wider support for communicating the label text in various applications (screen readers, etc).

packages/terra-form-checkbox/src/Checkbox.jsx
yuderekyu: Good catch, I forgot to pass/spread `labelTextAttrs` into the span. Updated

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,211 @@
...
+      'native-input',
+      inputAttrs.className,
+    ]);
+
+    const labelTextClasses = cx([
+      'label-text',
+      { 'is-hidden': isLabelHidden },
+    ]);
+
+    let labelTextContainer = null;
```
yuderekyu: Updated to do a conditional render based on `isLabelHidden`. I still render a span in both cases, because the css checkbox depends on the span with the `before` and `after` pseudo css selectors. The `labelText`itself changes though - it be set to `aria-label` if `isLabelHidden` is true.

packages/terra-form-checkbox/tests/nightwatch/checkbox-spec.js
```diff
@@ -0,0 +1,44 @@
...
+  },
+
+  'Displays a disabled Checkbox': (browser) => {
+    browser.url(`${browser.launchUrl}/#/tests/form-checkbox-tests/disabled`);
+    browser.expect.element('#disabled').to.not.be.enabled;
+    browser.click('#disabled');
+    browser.expect.element('#disabled').to.not.be.selected;
+  },
+
+  'Displays a checkbox with a hidden label': (browser) => {
```
yuderekyu: Added hidden label tests to check the conditional render.

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,162 @@
...
+    .is-disabled {
+      cursor: default;
+
+      > .label-text {
+        color: var(--terra-form-checkbox-disabled-font-color, #9b9fa1);
+        cursor: default;
+        opacity: var(--terra-form-checkbox-disabled-font-opacity, none);
+      }
+
+      > .label-text::before {
```
JakeLaCombe: I think you can put this inside > .label-text on line 118

```scss
> .label-text {
        color: var(--terra-form-checkbox-disabled-font-color, #9b9fa1);
         cursor: default;
         opacity: var(--terra-form-checkbox-disabled-font-opacity, none);

           &::before {
             /* Styles */
          }
}
```

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,162 @@
...
+        box-shadow: var(--terra-form-checkbox-focus-box-shadow, none);
+      }
+
+      [dir=ltr] & {
+        box-shadow: var(--terra-form-checkbox-focus-box-shadow, none);
+      }
+    }
+  }
+
+  /* Focus style surrounding both custom checkbox and label focus. Utilized in Clinical */
```
JakeLaCombe: I'd be careful with comments related to themes. They are proprietary, meaning it's best not to mention their intended use in our comments.

For this comment, I'd just remove the entire comment. The focus class self documents the code.

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,162 @@
...
+    [dir=rtl] & {
+      box-shadow: var(--terra-form-checkbox-container-focus-box-shadow, none);
+    }
+
+    [dir=ltr] & {
+      box-shadow: var(--terra-form-checkbox-container-focus-box-shadow, none);
+    }
+  }
+
+  /* Display checkbox inline if isInline is set */
```
JakeLaCombe: Extraneous comment as well

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,162 @@
...
+          cursor: default;
+        }
+      }
+    }
+    /* stylelint-enable max-nesting-depth */
+
+    /* Focus custom checkbox styling */
+    .native-input:focus + .label-text::before {
+      [dir=rtl] & {
+        box-shadow: var(--terra-form-checkbox-focus-box-shadow, none);
```
kschuste: these theme variables need to be more specific to distinguish between rtl and ltr, so add
-rtl
-ltr
as suffixes

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -0,0 +1,162 @@
...
+      }
+    }
+  }
+
+  /* Focus style surrounding both custom checkbox and label focus. */
+  .focus {
+    border-radius: var(--terra-form-checkbox-container-focus-border-radius, 3px);
+
+    [dir=rtl] & {
+      box-shadow: var(--terra-form-checkbox-container-focus-box-shadow, none);
```
kschuste: these theme variables need to be more specific to distinguish between rtl and ltr, so add
-rtl
-ltr
as suffixes

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -0,0 +1,211 @@
...
+    super(props);
+    this.state = { focus: false };
+    this.onFocus = this.onFocus.bind(this);
+    this.onBlur = this.onBlur.bind(this);
+  }
+
+  onFocus() {
+    this.setState({ focus: true });
+    if (this.onFocus !== undefined) {
+      this.onFocus();
```
emilyrohrbough: I think this should be 
```
if (this.props.onFocus) {
 this.props.onFocus();
}
```

mmalaker:
Functional verification completed. The checkbox examples display and function correctly.

bjankord:
@neilpfeiffer @mndrsn The checkbox is ready for UX review.

packages/terra-site/src/cerner-mock-theme.scss
```diff
@@ -47,10 +48,15 @@
   --terra-form-checkbox-checked-border-style: solid;
   --terra-form-checkbox-checked-border-width: 1px;
   --terra-form-checkbox-symbol-color: rgb(255, 255, 255);
-  --terra-form-checkbox-symbol: '\2713';
+  --terra-form-checkbox-symbol: url('data:image/svg+xml;utf8,<svg data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48"><path d="M18.4 40.9L0 22.5l4.2-4.2 14.2 14.2L43.8 7.1l4.2 4.2z"/></svg>');
```
bjankord: @kschuste added a mixin to help encode SVGs in his PR: https://github.com/cerner/terra-core/blob/ee14d191262dcdfacf2e04ad76d84ad2a28c84a6/packages/terra-mixins/src/Mixins.scss

We should copy that mixin over to your branch and use that for this SVG. Be sure to test this in IE too, we ran into some issues with the utf8 charset that was used for the svg. Using `data:image/svg+xml;charset=US-ASCII,#{url-encode($string)}` which the mixin will provide seemed to work better in IE.

yuderekyu: Mixin added [here](https://github.com/cerner/terra-core/pull/1037/commits/ad2cba45da51dbc5d6e294681b2b326653159b0c#diff-165d975ef87fbc0f0b863c7984cb1834R575). 

There was an IE 10 issue where the label text was being pushed under the checkbox for some custom themes. Fixed [here](https://github.com/cerner/terra-core/pull/1037/commits/ad2cba45da51dbc5d6e294681b2b326653159b0c#diff-222d72b921649ce14186f14e6910d91bR34).

neilpfeiffer:
Tested: PR #1037
Resolves: Issues #999, #925 (for checkboxes), #1002

**Reviewed:**
- Follows pattern considerations documented in Cerner Design Standards: [Checkbox](https://wiki.ucern.com/x/CZUCYQ)
- Follows UX defined design implementation for Terra-Form-Checkbox

\
**Future Enhancements:**
- will look at increasing the hitsize of the checkbox + label from 19px to 30px (by adjusting padding/margins, mainly for when is only one line) for better touch target

\
\

**Pass** |
-----------|
UX Review|

PR #1047, opened 2017-12-04
Added webdriver.io tests for alert

mhemesath:
### Summary
Added webdriver.io tests and visual comparison testing for alert.


wdio.conf.js
```diff
@@ -24,8 +24,41 @@ const config = {
     enabled: !process.env.TRAVIS,
   },
 
+  maxInstances: 1,
   webpackPort,
   webpackConfig,
+
+  beforeHook() {
```
bjankord: We should get terra-toolkit released so we can pull in the helper for this.

mhemesath: I can go back and create a PR to clean this up once we get that released if you are OK with that

bjankord: +1, sounds good to me.

packages/terra-alert/tests/nightwatch/AlertTitle.jsx
```diff
@@ -4,10 +4,9 @@ import Alert from '../../lib/Alert';
 
 const locale = document.getElementsByTagName('html')[0].getAttribute('lang');
 
+// Alerts of the standard types with custom titles
 export default () =>
   <Base locale={locale}>
-    <h2>Alerts of the standard types with custom titles</h2>
```
emilyrohrbough: So in these examples these blurbs are to give context to the people viewing the link, i.e. for functional review. We will want to leave these in here. A possibility would be to add an additional div and then grab that div when taking screenshots instead of the default `[data-reactroot]` element.

packages/terra-alert/tests/wdio/alert-spec.js
```diff
@@ -0,0 +1,103 @@
+/* global browser, Terra */
+const viewports = Terra.viewports('tiny', 'large');
+const ignoreContrast = {
```
emilyrohrbough: Could we include why we need to disable color-contrast here?

packages/terra-alert/tests/wdio/alert-spec.js
```diff
@@ -0,0 +1,103 @@
...
+      expect(browser.checkElement('[data-reactroot]')).to.matchReference();
+    });
+  });
+};
+
+describe('Alert', () => {
+  describe('Default', () => {
+    beforeEach(() => browser.url('/#/tests/alert-tests/default'));
+    Terra.should.matchScreenshot({ viewports });
+    Terra.should.beAccessible({ viewports, rules: ignoreContrast });
```
emilyrohrbough: Do we need to disable ignoreContract on each test?

wdio.conf.js
```diff
@@ -24,8 +24,41 @@ const config = {
     enabled: !process.env.TRAVIS,
   },
 
+  maxInstances: 1,
```
emilyrohrbough: Isn't the default 1?

packages/terra-alert/tests/nightwatch/AlertTitle.jsx
mhemesath: A lot of examples don't have instructional text. IMO these examples should be for automated tests and we should have other examples for manual validation if needed.

mhemesath: Also, if we can validate in chrome that these never change, chrome should be the baseline to compare other browsers against for behavior...

bjankord: I lean towards keeping these examples for both automated tests and manual verification. I want to avoid duplicate examples as much as we can. The automated tests are only run in Chrome currently while our manual testing is done in all our supported browsers. I terms of instruction text, I'm not too concerned one way or the other if we have it or not. I know our solution designer has asked for it and found it helpful for verifying.

packages/terra-alert/tests/wdio/alert-spec.js
mhemesath: I included it on each invalid test

packages/terra-alert/tests/nightwatch/AlertTitle.jsx
mhemesath: Compromise, what if we log out instruction text in the console?

wdio.conf.js
mhemesath: i think toolkit had 3 configured for somereason. Once we cut a release it won't matter though

mhemesath: this change probably isn't needed though

packages/terra-alert/tests/wdio/alert-spec.js
mhemesath: I added an issue for this and put it next to the rule so we know when to remove: https://github.com/cerner/terra-core/issues/1057

packages/terra-alert/tests/nightwatch/AlertTitle.jsx
bjankord: I'm cool with that. :+1:

packages/terra-alert/tests/wdio/alert-spec.js
bjankord: Yeah, we'll want to log issues for all the accessibility-rules we disable to follow-up and investigate them more.

mhemesath:
@emilyrohrbough do you have any more feedback for this review?

Issue #351, opened 2017-05-15
Button outline styles inside of modal


Matt-Butler:
# Issue Description
Buttons inside the modal do not display an outline when active. However, when the modal is full screen, the buttons do display an outline when active.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Buttons inside the modal display an outline when active.

## Current Behavior
Buttons inside the modal do not display an outline when active. However, when the modal is full screen, the buttons do display an outline when active.

## Steps to Reproduce
1. Navigate to the modal examples using safari.
2. Open the modal and tab through the focusable elements (you may need to turn on keyboard navigation for safari)

Safari: The tab key is being pressed but the button does not display as active
![safari_no_focus](https://cloud.githubusercontent.com/assets/8146159/26076877/d891c920-397f-11e7-9ee6-d2d21dc46ef0.gif)

Chrome:
![chrome_focus](https://cloud.githubusercontent.com/assets/8146159/26076921/003fc6e8-3980-11e7-84e6-9a9cccade592.gif)

## Environment

* Component Version: Button v0 and Modal v0
* Browser Name and Version: Safari
* Operating System and version (desktop or mobile): Mac OSX

Matt-Butler:
@bjankord suggested we investigate adding outline styles instead of using the browsers default focus outline

emilyrohrbough:
As of right now, some sort of outline style will need to be applied to selectable list items when an item contains a chevron. Because of this I think it would be a good idea to set our own outline.

bjankord:
I lean towards using the browser's default styles for focus outlines, but in this case, it looks like there is an issue in Safari that is causing the focus outline not to render on the button element in this use case. This causes a usability / accessibility issue making it difficult to visually see which button is in focus within non-fullscreen modals.

I'd suggest we take a [page from Bootstrap](https://github.com/twbs/bootstrap/blob/dd0ce3e4b2a89f245ee608a1acc9f848cf8c833a/scss/_buttons.scss#L22-L26) and remove the outline from `<Button />` components on focus and provide our own focus outline.

emilyrohrbough:
I took a look. It seems the css styling `bottom: auto` is disabling the tab-ability in Safari for modals of fixed size.

neilpfeiffer:
+1 @emilyrohrbough, what we need to adjust so we can remove `bottom: auto`

emilyrohrbough:
[...]

PR #1048, opened 2017-12-04
Pushed demo content in for modal site examples

msaikaushik1992:
### Summary

Added padding to push content of a modal a little into it to make sure button outline (outline of the close button in the modal) is visible when tab key is hit. Focuses more on fixing bug in Safari.

Fixes https://github.com/cerner/terra-core/issues/351

### Deployment Links
https://terra-core-deployed-pr-1048.herokuapp.com/static/#/site/modal
### Note
Added new class styles to site.scss because they are specific to only, demo on the terra site .

### DIFF
--- a/packages/terra-site/src/site.scss
+++ b/packages/terra-site/src/site.scss
@@ -76,6 +76,18 @@
   .site-input-display {
     display: inline-block;
   }
+
+  .modal-opened-content{
+    padding: 15px;
+  }
+
+  .modal-closed-content{
+    padding: 15px;
+  }
+
+  .modal-full-screen-content{
+    padding: 15px;
+  }
 }
 
 #site-body {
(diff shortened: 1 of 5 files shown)


packages/terra-site/src/site.scss
```diff
@@ -76,6 +76,18 @@
   .site-input-display {
     display: inline-block;
   }
+
+  .modal-opened-content{
```
bjankord: Let's make these all 1 class, `demo-modal-content` and use it across all 3 demos.

msaikaushik1992: Updated https://github.com/cerner/terra-core/pull/1048/commits/f06f7a6050a693f3527615fa37892df515eec782

mmalaker:
Functional verification completed. The modal button examples display the focus styles as expected.

PR #1053, opened 2017-12-06
Added webdriver.io tests for base

mhemesath:
### Summary
Added webdriver.io tests for terra-base


wdio.conf.js
```diff
@@ -11,7 +11,7 @@ const webpackPort = 8080;
 // Flex specs search between local pacakge and repo
 let specs = path.join('tests', 'wdio', '**', '*-spec.js');
 if (__dirname === process.cwd()) {
-  specs = path.join('packages', '**', specs);
+  specs = path.join('packages', '*', specs);
```
mhemesath: so this fixed a nasty issue I discovered where a double glob was causing wdio to pull in tests from the node_modules directory of every package. This means that when I pushed base tests, every component that used base reran those tests. Switching to a single glob excludes those node_module directories.

packages/terra-base/tests/nightwatch/SwitchLocaleBase.jsx
```diff
@@ -34,7 +34,7 @@ class SwitchLocaleBase extends React.Component {
         locale={this.state.locale}
         customMessages={customMessages[this.state.locale]}
       >
-        <FormattedMessage id="App.locale" />
+        <div id="message"><FormattedMessage id="App.locale" /></div>
```
mhemesath: added an id to make test more predictable on div fetch

packages/terra-base/tests/nightwatch/ArrayBase.jsx
```diff
@@ -5,14 +5,10 @@ const ArrayBase = () => (
   <div>
     <Base>
       <div id="one">One</div>
```
mhemesath: only one child is needed to verify this works

wdio.conf.js
mhemesath: @emilyrohrbough if any other projects integrated webdriver.io they may want to update this

emilyrohrbough: @mhemesath I'll work on getting the clinical PR updated

packages/terra-base/tests/nightwatch/ArrayBase.jsx
bjankord: This test was to show multiple sibling children can be passed to Base, we'll want to keep the multiple sibling children.

mhemesath: Technically this is a better test for jest IMO

packages/terra-base/tests/nightwatch/base-spec.js
```diff
@@ -1,43 +0,0 @@
...
-  'Switches locale and displays en-US customized locale message': (browser) => {
-    browser
-      .url(`${browser.launchUrl}/#/tests/base-tests/switch`)
-      .waitForElementVisible('button', waitInms)
-      .click('button');
-
-    browser.expect.element('div').text.to.contain('en-US').before(waitInms);
-  },
-
-  // NOTE: This test is designed to assume that tests won't fail on an error.  If it starts failing,
```
emilyrohrbough: Should we move this comment forward to the wdio spec?

Issue #1028, opened 2017-11-20
Status View OCS component


kschuste:
# Issue Description
Create a new Status View component with OCS styling.  Will be similar to the Error View and No Data View components inside terra-clinical, but will be able to handle both of their functionality.

## Issue Type

- [X] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

kschuste:
# Tech Design for Status View
## Summary:
This is a new component that will allow for the consumer to specify which type of status view they wish to display for their use case that will flex to display the content on the containers height.

## Bidirectionality
These changes should make no changes to the existing ltr/rtl functionality.

## React Props:
| Prop     | Type | Default | Description  |
|----------|------|---------|--------------|
| buttons | node | [] | An array of actionable buttons to display. |
| isGlyphHidden | bool | false | Whether or not the status glyph should be displayed.  |
| message | string |  | The descriptive text to display.  |
| title | string |  |The status title to display.  |
| type | string | `error` | Determines the type of icon to display.  Should be one of `nodata`, `nomatchingresults`, `notauthorized`, `error`, `noservice`, or `unknown`.  |

## Example:
```jsx
// with all fields
<StatusView 
	buttons={[<Button text="OK" />, <Button text="Cancel" />]}
	isGlyphHidden={false}
	message=“No data was returned from the service.”
	title=“404”
        type="nodata"
>

// with no defaults
<<StatusView />
```

emilyrohrbough:
Are you expecting there the children prop to be used as well for displaying content? If not, I vote we change the buttons prop to children.

kschuste:
I don't expect to use the children prop, the name children implies that I can pass in any type of components into it and expect them to be displayed inside the status view; however, here I am only specifically wanting buttons.  Therefore, I think that it is best kept as buttons.

mjhenkes:
@kschuste I agree with @emilyrohrbough. We've gone round and round on this topic and have come to the general consensus that the main area to drop elements should be named children most of the time. Your example would transform from:
```jsx
// with all fields
<StatusView 
	buttons={[<Button text="OK" key="1" />, <Button text="Cancel" key="2" />]}
	isGlyphHidden={false}
	message=“No data was returned from the service.”
	title=“404”
        type="nodata"
>
```
To this, which is easier to read though it does have the drawback you mentioned about not being explicit about requesting buttons, though in either case the consumer can drop in whatever they want.
```jsx
// with all fields
<StatusView 
[...]

PR #1059, opened 2017-12-06
Terra status view

kschuste:
### Summary
This is for issue #1028, the implementation of the new status view component.  This is a combination of the previous No Data and Error View inside the terra-clinical repository.

## Deployment Link
https://terra-core-deployed-pr-1059.herokuapp.com/static/#/site/status-view

### Additional Details
The behavior of the status view is that it will display to fit inside it's container.  If the container has a fixed height and view is too large, ONLY the glyph will disappear to allow the status view to keep it's context messages so the user will know what to do next.  It is also set to be centered in the container if the fixed height is larger by 40% of the remaining space on top and 60% on bottom.  It also allows for its content to be aligned at the top of the container and the user can choose to hide the glyph if allowed.


packages/terra-site/src/cerner-mock-theme.scss
```diff
@@ -144,4 +148,27 @@
...
+  --terra-status-view-glyph-custom: inline-svg("<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 48 48'><path fill='#582c83' d='M24.8 47.7l22.9-22.9c.4-.4.4-1.1 0-1.6L24.8.3c-.4-.4-1.1-.4-1.6 0L.3 23.2c-.4.4-.4 1.1 0 1.6l22.9 22.9c.4.4 1.2.4 1.6 0z'/></svg>");
+  --terra-status-view-title-color: #582c83;
+  --terra-status-view-title-font-family: 'Impact';
+  --terra-status-view-title-font-size: 36px;
+  --terra-status-view-message-color: #582c83;
+  --terra-status-view-message-font-family: 'Impact';
+  --terra-status-view-message-font-size: 24px;
+  --terra-status-view-message-padding-top: 10px;
+  --terra-status-view-message-text-align: center;
+  --terra-status-view-divider-display: block;
```
mhemesath: is there any value to this being a custom prop?

packages/terra-status-view/src/StatusView.scss
```diff
@@ -0,0 +1,112 @@
+@import '~terra-mixins';
+
+:local {
+  .status-view {
+    height: 100%;
+    padding-bottom: var(--terra-status-view-padding-bottom, 20px);
+    padding-top: var(--terra-status-view-padding-top, 20px);
+    width: 100%;
```
mhemesath: this shouldn't be needed.. if its a block level compoennt

packages/terra-status-view/tests/nightwatch/status-view-spec.js
```diff
@@ -0,0 +1,99 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+const { resizeTo } = require('terra-toolkit/lib/nightwatch/responsive-helpers');
+
+module.exports = resizeTo(['tiny', 'small', 'medium', 'large', 'huge', 'enormous'], {
```
mhemesath: convert to webdriver.io tests.  Use badge as an example these should be pretty straightforward a11y and screenshot tests.

packages/terra-status-view/src/StatusView.scss
```diff
@@ -0,0 +1,111 @@
...
+    padding-right: 5px;
+    padding-top: var(--terra-status-view-buttons-padding-top, 10px);
+    width: 100%;
+
+    @media screen and (min-width: 768px) {
+      flex-direction: row;
+    }
+  }
+
+  .buttons > button {
```
mhemesath: this is sort of an anti-pattern.. is there anyway around this?

packages/terra-status-view/src/StatusView.scss
```diff
@@ -0,0 +1,111 @@
...
+  }
+
+  .buttons {
+    display: flex;
+    flex: 0 0 auto;
+    flex-direction: var(--terra-status-view-buttons-flex-direction-small-width, row);
+    justify-content: center;
+    padding-left: 5px;
+    padding-right: 5px;
+    padding-top: var(--terra-status-view-buttons-padding-top, 10px);
```
mhemesath: should this be `--terra-status-view-footer-*` instead?

packages/terra-status-view/src/StatusView.scss
```diff
@@ -0,0 +1,111 @@
...
+    font-size: var(--terra-status-view-message-font-size, 14px);
+    padding-top: var(--terra-status-view-message-padding-top, 10px);
+    text-align: var(--terra-status-view-message-text-align, center);
+    width: 100%;
+  }
+
+  .divider {
+    display: var(--terra-status-view-divider-display, block);
+    padding-top: var(--terra-status-view-divider-padding-top, 20px);
+    width: 100%;
```
mhemesath: not needed with block level components

packages/terra-status-view/src/StatusView.scss
```diff
@@ -0,0 +1,111 @@
...
+
+  .glyph > svg {
+    background-repeat: no-repeat;
+    background-size: contain;
+    height: 100px;
+    width: 100px;
+  }
+
+  .glyph > .no-data {
+    background: var(--terra-status-view-glyph-no-data, inline-svg("<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 48 48' class='is-bidi'><path d='M45.8 12c1.2 0 2.2 1 2.2 2.2v26.7c0 1.2-1 2.2-2.2 2.2H2.2C1 43 0 42 0 40.8V7.2C0 6 1 5 2.2 5H15c1.4.1 2.7.6 3.7 1.6l3.8 3.9c1 1 2.3 1.5 3.7 1.6l19.4-.1h.2z'/></svg>"));
```
mhemesath: i feel like we should break SVGs out into their own file and import them here as SASS variables

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,254 @@
...
+const defaultProps = {
+  children: [],
+  isAlignedTop: false,
+  isGlyphHidden: false,
+  message: undefined,
+  title: undefined,
+  variant: variants.ERROR,
+};
+
+const contextTypes = {
```
mhemesath: why is this needed?

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,254 @@
...
+      componentsHeight += this.buttonsNode.getBoundingClientRect().height;
+    }
+    if (this.messageNode) {
+      componentsHeight += this.messageNode.getBoundingClientRect().height;
+    }
+    if (this.dividerNode) {
+      componentsHeight += this.dividerNode.getBoundingClientRect().height;
+    }
+
+    if (this.glyphNode && viewHeight >= componentsHeight + this.glyphNode.getBoundingClientRect().height) {
```
mhemesath: can the distribution of the padding not be handled by flexbox? I get that we need to calculate to determine if the icon will fit

mhemesath:
A few questions comparing the examples to consumer theme:
* Will this component allow for consumer theme where actions align to the bottom?
* Are custom message bodies/glyph variants allowed?

packages/terra-status-view/src/StatusView.jsx
tbiethman: This could use the `injectIntl` HOC factory instead: https://github.com/yahoo/react-intl/wiki/API#injection-api

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,254 @@
...
+  /**
+   * The required status title to display. Status views with variants of type no-data, no-matching-results,
+   * not-authorized, and error will have defaulted values unless overridden with this prop.
+   */
+  title: PropTypes.string,
+
+  /**
+   * Determines the type of icon to display.
+   */
+  variant: PropTypes.oneOf([
```
tbiethman: I'm not sure if `Object.values` is supported by the babel-polyfill, but if it is, you can just do this:
```
variant: PropTypes.oneOf(Object.values(variants)),
```

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  componentDidMount() {
+    // call after all items have rendered in the view to calculate which items
+    // can be shown and center the items inside the view appropriately
+    this.showAndCenterItems();
+  }
+
+  setInnerNode(node) {
+    if (node === null) { return; } // Ref callbacks happen on mount and unmount, element will be null on unmount
```
tbiethman: I don't think these checks are necessary. If the component is being unmounted, you probably don't care that your refs are going away. Or were you seeing issues without these?

packages/terra-status-view/src/StatusView.jsx
kschuste: it was added for internationalization of the default titles.  is what has been done in other components such as DemographicsBanner, Field, Select, Alert, DatePicker

packages/terra-status-view/src/StatusView.jsx
kschuste: yes, we do need to determine if the icon fits and, attempting to set the padding via flexbox was leading to some over-calculated heights and content being pushed well below the bounds

packages/terra-status-view/src/StatusView.jsx
kschuste: was having trouble of nodes being null when runnings tests at first so switched to this patten seen in the code, fixed the tests so will switch back to providing ref inline when creating the component

packages/terra-status-view/src/StatusView.scss
kschuste: this is for the buttons div, which may or may not display depending on if children are passed in; therefore, calling this footer wouldn't be consistent as the descriptive message would be the last item

mhemesath: But if it doesn’t display then isn’t there just not a footer? Also if footer isn’t he right word perhaps actions then?

kschuste: right, by not displaying this then the view will shrink vertically and the message div will be the bottom, if no message then will be the title div, the outer div (.status-view) has padding-bottom applied to it to act as the bottom margin no matter which element is displays last

actions is fine for the name:  --terra-status-view-actions-padding-top, I am good with that

packages/terra-status-view/src/StatusView.jsx
kschuste: seems to be fine 👍

packages/terra-status-view/src/StatusView.scss
kschuste: removed

packages/terra-status-view/src/StatusView.scss
kschuste: nested these broke out declarations underneath their corresponding parent declarations

kschuste:
@mhemesath this component will allow for the consumer theme
they will be able to change the styling to allow for the buttons to display the width of the container and allow them to be stacked verticall
there is a custom glyph variant that will allow consumers to provide their own image via the css and their own title and message as it will not have a default title for that custom variant

packages/terra-status-view/src/StatusView.scss
bjankord: I want to stick to theming through css custom properties as much as possible. If we start doing theming in both Sass and css custom properties, it adds to theming complexity.

packages/terra-status-view/tests/wdio/status-view-spec.js
```diff
@@ -0,0 +1,189 @@
...
+const shouldTheme = (customProperties) => {
+  Object.entries(customProperties).forEach(([key, value]) => {
+    it(`themed [${key}]`, () => {
+      browser.setCSSCustomProps({ [key]: value });
+      expect(browser.checkElement('[data-reactroot]')).to.matchReference();
+    });
+  });
+};
+
+const viewports = Terra.viewports('tiny', 'small', 'medium', 'large', 'huge', 'enormous');
```
mhemesath: We dont' need to test all these viewports, if we need responsive testing I'd do a few things:

1. Have an explicit responsive test
2. Do edge testing (recommended). Example, do tiny and medium breakpoints to verify component flexes well.

packages/terra-status-view/tests/wdio/status-view-spec.js
```diff
@@ -0,0 +1,189 @@
...
+    });
+
+    Terra.should.matchScreenshot({ viewports });
+    Terra.should.beAccessible({ viewports });
+  });
+
+  describe('Displays a no-matching-results status view with the content aligned at the top in its container', () => {
+    before(() => browser.url('/#/tests/status-view-tests/align-top'));
+
+    shouldTheme({
```
mhemesath: I don't think we need to test theme properties for every scenario. Just pick a representative scenario (usually a default one) and test the properties on that.

mhemesath: Also, I don't think we need to test theming properties for responsive variants unless there is a specific responsive scenario to test.

packages/terra-status-view/tests/wdio/status-view-spec.js
```diff
@@ -0,0 +1,189 @@
...
+      '--terra-status-view-title-color': 'purple',
+      '--terra-status-view-title-font-family': 'Impact',
+      '--terra-status-view-title-font-size': '36px',
+    });
+
+    Terra.should.matchScreenshot({ viewports });
+    Terra.should.beAccessible({ viewports });
+  });
+
+  describe('Displays an error status view with all props given', () => {
```
mhemesath: for brevity in the filenames, i'd change these descriptions to this
> error with all props

packages/terra-status-view/tests/wdio/status-view-spec.js
```diff
@@ -0,0 +1,189 @@
...
+      '--terra-status-view-title-color': 'purple',
+      '--terra-status-view-title-font-family': 'Impact',
+      '--terra-status-view-title-font-size': '36px',
+    });
+
+    Terra.should.matchScreenshot({ viewports });
+    Terra.should.beAccessible({ viewports });
+  });
+
+  describe('Displays a no-matching-results status view with the content aligned at the top in its container', () => {
```
mhemesath: for brevity in the filenames, i'd change these descriptions to this
> No Matching Results aligned top

packages/terra-site/package.json
```diff
@@ -78,6 +78,7 @@
     "terra-slide-panel": "^1.13.0",
     "terra-spacer": "^1.0.0",
     "terra-status": "^1.16.0",
+    "terra-status-view": "^1.0.0",
```
jmsv6d: This should be version "^0.0.0"

packages/terra-status-view/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+Changelog
+=========
+
+1.0.0 - (June 28, 2017)
```
jmsv6d: This should be Unreleased

packages/terra-status-view/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "terra-status-view",
+  "main": "lib/StatusView.js",
+  "version": "1.0.0",
```
jmsv6d: This should start at version "0.0.0"

packages/terra-site/src/examples/status-view/Index.jsx
```diff
@@ -0,0 +1,65 @@
...
+    <h2 id="default">Status View:  No props given</h2>
+    <StatusViewExample style={{ width: '500px' }} />
+    <br />
+    <h2 id="default">Status View:  All props given</h2>
+    <StatusViewExample style={{ width: '500px' }} variant={StatusView.variants.NODATA} title="No Data" message="No data was returned">
+      <Button text="OK" key="1" size="medium" />
+      <Button text="Cancel" key="2" size="medium" />
+    </StatusViewExample>
+    <br />
+    <h2 id="blank">Status View:  Text Wrapping</h2>
```
jmsv6d: These headers should have unique ids.

packages/terra-site/src/examples/status-view/Index.jsx
```diff
@@ -0,0 +1,65 @@
...
+const StatusExamples = () => (
+  <div>
+    <div id="version">Version: {version}</div>
+    <Markdown id="readme" src={ReadMe} />
+    <PropsTable id="props" src={StatusViewSrc} />
+    <h2 id="default">Status View:  No props given</h2>
+    <StatusViewExample style={{ width: '500px' }} />
+    <br />
+    <h2 id="default">Status View:  All props given</h2>
+    <StatusViewExample style={{ width: '500px' }} variant={StatusView.variants.NODATA} title="No Data" message="No data was returned">
```
jmsv6d: You should build out these examples in their own files.

packages/terra-site/src/examples/status-view/StatusViewExample.jsx
```diff
@@ -0,0 +1,89 @@
...
+const propTypes = {
+  children: PropTypes.node,
+  isAlignedTop: PropTypes.bool,
+  isGlyphHidden: PropTypes.bool,
+  message: PropTypes.string,
+  title: PropTypes.string,
+  variant: PropTypes.string,
+};
+
+class StatusViewExample extends React.Component {
```
jmsv6d: Nit: Maybe change this to StatusViewTemplate.

packages/terra-status-view/docs/DEPENDENCIES.md
```diff
@@ -0,0 +1,20 @@
+# Dependency Information
```
jmsv6d: The scripts were updated so that this should only be getting generated on publish. You should remove it from the PR.

packages/terra-status-view/docs/README.md
```diff
@@ -0,0 +1,23 @@
...
+  - `npm install terra-status-view`
+  - `yarn add terra-status-view`
+
+## Usage
+
+```jsx
+import React from 'react';
+import StatusView from 'terra-status-view';
+
+<StatusView </StatusView>
```
jmsv6d: Update with actual code example.

packages/terra-status-view/src/StatusView.scss
jmsv6d: +1 for actions

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,211 @@
...
+    const statusViewInnerStyles = {
+      paddingTop: `${this.state.paddingTop}px`,
+      paddingBottom: `${this.state.paddingBottom}px`,
+    };
+
+    return (
+      <div
+        {...customProps}
+        className={statusViewClassNames}
+      >
```
jmsv6d: Why are two wrapping divs needed?

packages/terra-status-view/tests/jest/StatusView.test.jsx
```diff
@@ -0,0 +1,83 @@
...
+import Button from 'terra-button';
+import { IntlProvider } from 'react-intl';
+import StatusView from '../../src/StatusView';
+import messages from '../../translations/en-US.json';
+
+const locale = 'en-US';
+
+// Snapshot Tests
+it('should render an image with a default status view', () => {
+  const defaultRender = (<IntlProvider locale={locale} messages={messages}>
```
jmsv6d: Check out the intl-context-setup file that @emilyrohrbough wrote in terra-form-field: https://github.com/cerner/terra-core/blob/master/packages/terra-form-field/tests/jest/intl-context-setup.js

packages/terra-status-view/src/StatusView.scss
```diff
@@ -0,0 +1,111 @@
...
+    padding-left: 10px;
+    padding-right: 10px;
+    width: 100%;
+  }
+
+  .title {
+    color: var(--terra-status-view-title-color, #000);
+    flex: 0 0 auto;
+    flex-wrap: wrap;
+    font-family: var(--terra-status-view-title-font-family, 'Roboto');
```
jmsv6d: Roboto probably isn't the right font for the open source theme. Also, this should just inherit from Base.

packages/terra-status-view/src/StatusView.scss
```diff
@@ -0,0 +1,111 @@
...
+    font-size: var(--terra-status-view-title-font-size, 24px);
```
jmsv6d: Replace px values with rems

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,211 @@
...
+    if (this.dividerNode) {
+      componentsHeight += this.dividerNode.offsetHeight;
+    }
+
+    if (this.glyphNode && viewHeight >= componentsHeight + this.glyphNode.offsetHeight) {
+      // if have a glyph to show and can fit inside the containers height, distribute remaining padding to center
+      showGlyph = true;
+      if (!this.props.isAlignedTop) {
+        const remainingHeight = viewHeight - (componentsHeight + this.glyphNode.offsetHeight);
+        paddingTop = remainingHeight * 0.4;
```
yuderekyu: Just curious, how did we decide on multiplying the padding with these values?

packages/terra-status-view/tests/wdio/status-view-spec.js
```diff
@@ -0,0 +1,189 @@
+/* global browser, Terra, before */
+const shouldTheme = (customProperties) => {
```
mhemesath: This wont' work anymore as `browser.setCSSCustomProps` has been removed. Use

```Terra.should.themeEachCustomProperty``` instead. See the [badge test](https://github.com/cerner/terra-core/blob/master/packages/terra-badge/tests/wdio/badge-spec.js#L10)s for an example

File: packages/terra-status-view/src/StatusView.jsx (hunk 24 above)
Comment (kschuste) [reply packages/terra-status-view/src/StatusView.jsx]: the outer wrapping will apply the default padding around content alway, the content is inside its own div to have the remaining vertical space in the view distributed if there is any

File: packages/terra-status-view/src/StatusView.jsx (hunk 28 above)
Comment (kschuste) [reply packages/terra-status-view/src/StatusView.jsx]: it is specified in the callouts to distribute the remaining vertical space by 40% of it on top and 60% of the space below

File: packages/terra-status-view/tests/nightwatch/StatusViewNotAuthorized.jsx [hunk 30]
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import Base from 'terra-base';
+import StatusView from '../../lib/StatusView';
+
+const locale = document.getElementsByTagName('html')[0].getAttribute('lang');
+
+export default () =>
+  <Base locale={locale}>
+    <StatusView role="main" id="statusView" variant={StatusView.variants.NOTAUTHORIZED} />
```
bjankord: What's the logic on setting `role="main"` here?

kschuste: It is is allow the tests to pass when checking to make sure they are accessible or not; the rule that was breaking is described here:
https://dequeuniversity.com/rules/axe/3.0/landmark-one-main

I tried turning if off and while that fixed the tests locally; they still failed on Travis CI.  Therefore, I added the role of main to the test samples instead to get the CI build to succeed.

bjankord: Hmm, I have mixed thoughts on this. I lean towards disabling the landmark-one-main accessibility rule for our component tests rather than adding `role="main"` to our component tests to get the linter to pass.

Might try the following:

```
// Before your describe blocks
const ignoredA11yRules = {
  'landmark-one-main': { enabled: false },
};

// In your describe blocks
Terra.should.beAccessible({ viewports, rules: ignoredA11yRules });
```
And then once thats in place we can clear the cache on travis and see if that fixes the issue described of the tests not passing on travis.

jmsv6d: If clearing the cache on travis causes it to pull in the newest version of axe that has this rule, then all of are accessibility tests in all the other components are going to start failing. We should probably log a separate issue for this.

kschuste: @bjankord What you suggested is how I originally had the code; however, the CI build failed, I will revert the change to use role="main" and try again to see if it passes.

bjankord: Ok I can try clearing the cache on travis too and see if that makes any difference.

kschuste:
@bjankord the tests failed again, reverting back to have them pass again, I will log a new issue for the other components that can have this issue occur

bjankord:
@kschuste Did you clean the cache on travis when you reverted? I was gonna test that out but I haven't had a chance yet.

kschuste:
@bjankord I don't see where to do that, if you could clean it, then I'll revert back and we can see for sure then

bjankord:
@kschuste Ok yeah lets revert it to ignore that rule, and then I'll try clearing the cache on travis CI.

bjankord:
Ok let me try clearing the cache and see if that makes any difference. I want to rule that case out.

bjankord:
Hmm, looks like this is timing out. Let's get the changes from toolkit which pin the axe version in place in this branch and see if that gets this branches tests to pass again

package.json
```diff
@@ -88,7 +88,7 @@
     "identity-obj-proxy": "^3.0.0",
     "ip": "^1.1.5",
     "jest": "^21.0.2",
-    "lerna": "^2.1.2",
+    "lerna": "^2.5.1",
```
emilyrohrbough: Should we update `lerna` and `shelljs` in this PR?

packages/terra-site/src/cerner-mock-theme.scss
```diff
@@ -3,6 +3,10 @@
 @import '~terra-mixins';
```
emilyrohrbough: The `themeable-variables.json` file needs to be updated. See this issue: https://github.com/cerner/terra-core/issues/1106.

packages/terra-status-view/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "compile:build": "babel src --out-dir lib --copy-files",
+    "lint": "npm run lint:js && npm run lint:scss",
+    "lint:js": "eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
+    "lint:scss": "stylelint src/**/*.scss",
+    "props-table": "props-table ./src/StatusView.jsx --out-dir ./docs/props-table",
+    "test": "npm run test:jest && npm run test:wdio",
+    "test:jest": "jest ./tests/jest/* --config ../../jestconfig.json",
+    "test:wdio": "wdio ../../wdio.conf.js"
+  },
+  "devDependencies": {
```
emilyrohrbough: You shouldn't need to list terra-button, terra-divider, and terra-mixins in the devDeps since they are also listed as dependencies.

kschuste: removed

package.json
kschuste: set them back to original

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,210 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import 'terra-base/lib/baseStyles';
+import Divider from 'terra-divider';
+import styles from './StatusView.scss';
+
+const cx = classNames.bind(styles);
+
+const variants = {
+  NODATA: 'no-data',
```
emilyrohrbough: We should seperatet these with hyphens to be readable.

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,210 @@
...
+  /**
+   * The descriptive text to display.
+   */
+  message: PropTypes.string,
+
+  /**
+   * The required status title to display. Status views with variants of type no-data, no-matching-results,
+   * not-authorized, and error will have defaulted values unless overridden with this prop.
+   */
+  title: PropTypes.string,
```
emilyrohrbough: The description says title is required. Should probably change to `title: PropTypes.string.isRequired,`

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,210 @@
...
+
+  /**
+   * Determines the type of icon to display.
```
emilyrohrbough: List the variants in the description so people know what they options are. 

Also nit: I would remove the use of 'type of icon' here given other components have both types and variants and it might be confusing to consumers.

packages/terra-status-view/src/StatusView.jsx
kschuste: title is defaulted in some cases, updated the documentation to leave out it being required

packages/terra-status-view/src/StatusView.jsx
kschuste: listed the variants,

changed to use the term glyph

packages/terra-status-view/src/StatusView.jsx
kschuste: I prefer leaving the hypens out for when variables are in all caps

packages/terra-site/themeable-variables.json
```diff
@@ -1,4 +1,4 @@
-{
+x{
```
jmsv6d: Where did this x come from?

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,211 @@
...
+        </div>
+      </div>
+    );
+  }
+}
+
+StatusView.propTypes = propTypes;
+StatusView.contextTypes = contextTypes;
+StatusView.defaultProps = defaultProps;
+StatusView.variants = variants;
```
jmsv6d: This should probably be StatusView.Opts.variants to align with other components.

packages/terra-site/themeable-variables.json
kschuste: not sure, must have been a typo, it's removed

packages/terra-status-view/src/StatusView.jsx
kschuste: done

packages/terra-status-view/tests/wdio/status-view-spec.js
```diff
@@ -0,0 +1,158 @@
...
+    });
+  });
+
+  describe('Displays an error status view with all props given', () => {
+    beforeEach(() => browser.url('/#/tests/status-view-tests/error-all-props'));
+
+    Terra.should.matchScreenshot({ viewports });
+    Terra.should.beAccessible({ viewports });
+    Terra.should.themeEachCustomProperty({
+      '--terra-status-view-padding-bottom': '10px',
```
yuderekyu: Do we need to continually test padding-bottom, padding-top, title-color, font-family, and font-size for each variation if they will always have the same values?

kschuste: I am doing it to test as many theming props as I can for each scenario, may be a bit overkill but wanted to be thorough

mjhenkes:
@neilpfeiffer The icons used here don't look like they're going to blend in with our existing applications very well. I imagine this will lead to engineers using the custom status view.

kschuste:
updated to match the open-source theme

packages/terra-site/package.json
```diff
@@ -80,6 +80,7 @@
     "terra-slide-panel": "^1.14.0",
     "terra-spacer": "^1.1.0",
     "terra-status": "^1.17.0",
+    "terra-status-view": "^0.0.0",
```
emilyrohrbough: I think this should be "terra-status-view": "^x.0",

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,212 @@
...
+            {messageSection}
+          </div>
+          {actionSection}
+        </div>
+      </div>
+    );
+  }
+}
+
+StatusView.propTypes = propTypes;
```
emilyrohrbough: If the children prop should be buttons, we should export the Button component and update the document examples to use it. See comment on docs.

packages/terra-status-view/docs/README.md
```diff
@@ -0,0 +1,33 @@
...
+import StatusView from 'terra-status-view';
+
+<StatusView
+  isAlignedTop={false}
+  isGlyphHidden={false}
+  variant={StatusView.Opts.variants.ERROR}
+  title="A 500 error was thrown from the service."
+  message=message="A null pointer exception was thrown."
+>
+  <Button text="OK" key="1 />
```
emilyrohrbough: Given buttons are the expected children, we should update this to be

`< StatusView.Button text="OK" key="1 />`

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,212 @@
...
+  message: PropTypes.string,
+
+  /**
+   * The status view's title to display. Status views with variants of type no-data, no-matching-results,
+   * not-authorized, and error will have defaulted values unless overridden with this prop.
+   */
+  title: PropTypes.string,
+
+  /**
+   * Determines the glyph to display, one of the following: 'no-data', 'no-matching-results',
```
emilyrohrbough: Should we update these to be the OPT values?

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,212 @@
...
+
+    if (this.glyphNode && viewHeight >= componentsHeight + this.glyphNode.offsetHeight) {
+      // if have a glyph to show and can fit inside the containers height, distribute remaining padding to center
+      showGlyph = true;
+      if (!this.props.isAlignedTop) {
+        const remainingHeight = viewHeight - (componentsHeight + this.glyphNode.offsetHeight);
+        paddingTop = remainingHeight * 0.4;
+        paddingBottom = remainingHeight * 0.6;
+      }
+    } else if (!this.props.isAlignedTop) {
```
emilyrohrbough: This else-if has the same condition as the if

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,212 @@
...
+      // glyph does not exist or does not fit, distribute remaining padding if any to center
+      const remainingHeight = viewHeight - componentsHeight;
+      if (remainingHeight > 0) {
+        paddingTop = remainingHeight * 0.4;
+        paddingBottom = remainingHeight * 0.6;
+      }
+    }
+
+    // set which components can fit into the view and add the additional 20px
```
emilyrohrbough: are we still adding an additional 20px?

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,212 @@
...
+    title,
+    variant,
+    ...customProps
+    } = this.props;
+
+    const { intl } = this.context;
+
+    let glyphSection;
+    if (!isGlyphHidden && this.state.showGlyph) {
+      glyphSection = <div className={cx('glyph')} ref={(element) => { this.glyphNode = element; }}><svg className={cx(variant)} /></div>;
```
emilyrohrbough: nit: can we make this multiple lines to improve readability?

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,212 @@
...
+
+const variants = {
+  NODATA: 'no-data',
+  NOMATCHINGRESULTS: 'no-matching-results',
+  NOTAUTHORIZED: 'not-authorized',
+  ERROR: 'error',
+  CUSTOM: 'custom',
+};
+
+const propTypes = {
```
mjhenkes: Should there be a prop for a glyph for the custom type?

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,212 @@
...
+    let glyphSection;
+    if (!isGlyphHidden && this.state.showGlyph) {
+      glyphSection = <div className={cx('glyph')} ref={(element) => { this.glyphNode = element; }}><svg className={cx(variant)} /></div>;
+    }
+    let messageSection;
+    if (message) {
+      messageSection = <div className={cx('message')} ref={(element) => { this.messageNode = element; }}>{message}</div>;
+    }
+    let actionSection;
+    if (children.length) {
```
emilyrohrbough: If one child is provided, will this be true?

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -0,0 +1,212 @@
...
+          style={{ ...statusViewInnerStyles }}
+          ref={(element) => { this.innerNode = element; }}
+        >
+          {glyphSection}
+          <div className={cx('message-content-group')} >
+            {titleSection}
+            {dividerSection}
+            {messageSection}
+          </div>
+          {actionSection}
```
emilyrohrbough: Should we be rendering the actionSection within this div?

To determine if we need a dividerSection, we are checking for the messageSection or the actionSection. Here we are rendering the divider within the 'message-content-group' div, but are not rendering the action section. 

I could see consumers expecting the divider to span then entire 'status' container without the padding.

packages/terra-status-view/src/StatusView.scss
```diff
@@ -0,0 +1,107 @@
...
+    padding-left: 0.714rem;
+    padding-right: 0.714rem;
+    width: 100%;
+  }
+
+  .title {
+    color: var(--terra-status-view-title-color, #868a8c);
+    flex: 0 0 auto;
+    flex-wrap: wrap;
+    font-family: var(--terra-status-view-title-font-family, 'HelveticaNeue');
```
emilyrohrbough: font-family should be inherited & themed from base

packages/terra-status-view/src/StatusView.jsx
kschuste: to have the icons be themed, they are provided via the css, therefore, to provide a custom icon you pass in the custom variant and have the css property set.

packages/terra-status-view/tests/jest/__snapshots__/StatusView.test.jsx.snap
```diff
@@ -0,0 +1,331 @@
...
+  variant="no-data"
+>
+  <div
+    className="status-view"
+  >
+    <div
+      className="status-view-inner"
+      style={
+        Object {
+          "paddingBottom": "0px",
```
emilyrohrbough: Would you expect these padding values to be 0 here?

emilyrohrbough: Or what is the case these wouldn't be 0? I am curious on getting a screenshot of an instance where the padding is calculated

packages/terra-status-view/.npmignore
```diff
@@ -0,0 +1,6 @@
+*.log
+node_modules
+src
```
bjankord: need to remove `src` and `tests` from this file

packages/terra-site/package.json
tbiethman: 0.x, right?

packages/terra-site/src/examples/status-view/StatusViewTemplate.jsx
```diff
@@ -0,0 +1,88 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import StatusView from 'terra-status-view';
+import _ from 'lodash';
```
bjankord: I'm a bit curious what impact this has on our bundle size. It looks like it is only being used in this file for unique ids. [ShortID](https://github.com/dylang/shortid) is another option that we could use that should result in a smaller bundle.

packages/terra-status-view/docs/README.md
bjankord: I'm cool with leaving this as is using `<Button />` instead of `<StatusView.Button />`

packages/terra-status-view/src/StatusView.jsx
bjankord: Whatever we do we do for other constants we attach to `Opts`, we should be consistent what that.

packages/terra-status-view/tests/jest/__snapshots__/StatusView.test.jsx.snap
kschuste: this is for when the status view has a default height that is larger than what its content is.  It will add padding after mounting to align the content in the middle with 40% on the top and 60% on the bottom of the remaining space.

packages/terra-status-view/src/StatusView.scss
bjankord: Yeah, this property can be removed. Font-family will cascade down to title from terra-base styles set on HTML element.

packages/terra-status-view/src/StatusView.scss
```diff
@@ -0,0 +1,107 @@
...
+    font-size: var(--terra-status-view-title-font-size, 1.714rem);
+    text-align: center;
+    width: 100%;
+  }
+
+  .message {
+    color: var(--terra-status-view-message-color, #1c1f21);
+    flex: 0 0 auto;
+    flex-wrap: wrap;
+    font-family: var(--terra-status-view-message-font-family, 'HelveticaNeue');
```
bjankord: This property can be removed, font-family will cascade down to message from terra-base styles set on HTML element.

packages/terra-status-view/src/StatusView.jsx
kschuste: The actionSection was placed outside the messageSection for vertical padding reasons as it will have the same top padding no matter what is the last component is in the messageContentGroup, whereas those components can have different vertical spacing between them.  The divider does span the entire width of the container already as it's width will be 100% of its container.

packages/terra-status-view/src/StatusView.jsx
kschuste: updated to pass in a prop of type node to allow the consumer to provide the glyph

packages/terra-status-view/src/StatusView.jsx
kschuste: as @bjankord mentioned below, I would like to leave this as just Button

packages/terra-status-view/package.json
```diff
@@ -48,6 +48,7 @@
     "terra-base": "^2.7.0",
     "terra-button": "^1.0.0",
     "terra-divider": "^1.0.0",
+    "terra-icon": "^1.16.0",
```
emilyrohrbough: This can be listed as a devDependency

packages/terra-status-view/tests/jest/__snapshots__/StatusView.test.jsx.snap
emilyrohrbough: I think it would be good to add a test case to verify this is calculated & applied as expected

mmalaker:
Functional verification completed. The status view examples display and function correctly.

packages/terra-status-view/src/StatusView.scss
```diff
@@ -0,0 +1,101 @@
...
+    align-items: center;
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+    width: 100%;
+  }
+
+  .glyph {
+    flex: 0 0 auto;
+    padding-bottom: 0.714rem;
```
neilpfeiffer: @kschuste - just need to add a theme-able variable for the glyph to take a separate color from the text.
For the default/terra-opensource-theme the color is: `rgba(134,138,140,0.15)`

Also validating if the icon for the error case needs to also take a different/special color (red) - was a late mention that I am checking on.

kschuste: updated to use the correct color, was using the hex value of the rgba before but using rgba directly now and can see the color difference

the color for the glyph is provided in the following variables:
--terra-status-view-glyph-no-data
--terra-status-view-glyph-no-matching-results
--terra-status-view-glyph-not-authorized
--terra-status-view-glyph-error

for those variables the svg code is provided and the color is determined by setting the 'fill' attribute in that svg code

neilpfeiffer: Update - no custom color needed for error icon. All icons colored the same unless a non-themeable/colorable custom icon is used.

neilpfeiffer: Current implementation looks great - will add UX validation.

neilpfeiffer:
Tested: PR #1059
Resolves: Issue #1028


**Reviewed:**
- Follows pattern considerations documented in Cerner Design Standards: [Error Views](https://wiki.ucern.com/x/zlYVYQ) and [No Data Patterns](https://wiki.ucern.com/x/10MVYQ)
- Follows UX defined design implementation for Terra-Status-View

\
\

**Pass** |
-----------|
UX Review|

PR #1063, opened 2017-12-08
Added wdio card tests

mhemesath:
### Summary
Added webdriver.io tests for card


packages/terra-card/tests/wdio/card-spec.js
```diff
@@ -0,0 +1,53 @@
...
+describe('Badge', () => {
+  describe('Default', () => {
+    beforeEach(() => browser.url('/#/tests/card-tests/default'));
+
+    Terra.should.beAccessible({ viewports });
+    Terra.should.matchScreenshot({ viewports });
+    shouldTheme({
+      '--terra-card-background-color': 'yellow',
+      '--terra-card-border': '10px dashed purple',
+      '--terra-card-border-radius': '50px',
```
emilyrohrbough: It looks applying a  border-radius cuts the text off when looking at the screenshot. Did we get an issue logged to fix this styling bug?

mhemesath: Well.... its standard browser behavior. The big question is if we want to implement something to prevent it. It would be kinda tricky IMO with theming... meaning the theme would just have to account for its border radius and make sure text isn't clipped @bjankord  thoughts?

bjankord: I think we leave it as is, my hope would be if people apply 50px border-radius, they also apply enough padding to fix clipping issues which result from the larger border-radius.

emilyrohrbough: Either sounds good to me!

PR #1076, opened 2017-12-13
Update terra-toolkit to 2.4

mhemesath:
Updated to toolkit 2.4. This provides a few changes

* Helper to theme custom properties
* Uses docker stacks so tests run 2x faster.
* Cleans up project wdio.conf.js


wdio.conf.js
```diff
@@ -35,47 +34,6 @@ const config = {
     // in order to reset the site. This ensures customProperty tests and any
     // other dom modifications are cleared before starting a test.
     global.browser.refresh();
-
-    // Ensure the mouse starts in upper left corner before every test.
```
mhemesath: toolkit provides this now

Issue #1078, opened 2017-12-14
Remove terra-toolkit from spacer dependencies


zhongyn:
# Issue Description
terra-toolkit is added as spacer dependency, but it is not consumed in any src code.
https://github.com/cerner/terra-core/blob/master/packages/terra-spacer/package.json#L35

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

PR #1079, opened 2017-12-14
Remove toolkit dependency

zhongyn:
### Summary
https://github.com/cerner/terra-core/issues/1078

@cerner/terra


mhemesath:
I'm really curious what changed that all the sudden those linting error exceptions are needed.

zhongyn:
@mhemesath not sure if any change triggers that. But i believe that linting is always required since the nightwatch folder is not specified in the eslintignore: https://github.com/cerner/terra-core/blob/master/.eslintignore

PR #1089, opened 2017-12-18
Import terra-mixins for checkbox

yuderekyu:
### Summary
Import terra-mixins in order for the checkbox stylesheet to use `inline-svg`, a mixins function which will be passed by custom themes to display their custom checkmark symbol.

### DIFF
--- a/packages/terra-form-checkbox/CHANGELOG.md
+++ b/packages/terra-form-checkbox/CHANGELOG.md
@@ -3,4 +3,6 @@ ChangeLog
 
 (Unreleased)
 -----------------
-Initial stable release
+### Added
+* Import terra-mixins into style sheet
+* Initial stable release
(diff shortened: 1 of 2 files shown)


packages/terra-form-checkbox/CHANGELOG.md
```diff
@@ -3,4 +3,6 @@ ChangeLog
 
 (Unreleased)
 -----------------
-Initial stable release
+### Added
+* Import terra-mixins into style sheet
+* Initial stable release
```
emilyrohrbough: Has this been released? If not, I feel like we could tweak the changelog to just mention `Initial Release`

yuderekyu:
Le's just consolidate this with the upcoming change that addresses #1090.

PR #1093, opened 2017-12-18
Remove focus-container from checkbox

yuderekyu:
### Summary
Resolves #1090 and continuation of #1089. Removes focus container management and styles. Checkbox is now a stateless component. Increased hit-size (using padding) for a better touch target.

### Deployment Links
* https://terra-core-deployed-pr-1093.herokuapp.com/static/#/site/form-checkbox
* https://terra-core-deployed-pr-1093.herokuapp.com/static/#/tests/form-checkbox-tests

PR #1096, opened 2017-12-19
Add checkbox focus ring

### LINKED ISSUE #1090: Add Focus Ring to Checkbox
# Issue Description
The checkbox `container-focus` style, specified in several custom themes, has been replaced with a large focus ring surrounding the checkbox.`container-focus` styling should be removed and a `focus-ring` styling should be added.

*Edit: I can still have state management removed and utilize css selectors for the focus ring.

yuderekyu:
### Summary
* Make checkbox a stateless component. 
* Remove focus-container styling and replace with a focus ring. Viewable on the mock theme.
* Add hover styling.
* Add top and bottom padding to select-container to increase hit-area.
* Resolves #1090. Continuation of #1093.

### Deployment Links
* https://terra-core-deployed-pr-1096.herokuapp.com/static/#/site/form-checkbox
* https://terra-core-deployed-pr-1096.herokuapp.com/static/#/tests/form-checkbox-tests


mmalaker:
A couple issues found:
- The mock theme on Android shows the focus rings slightly off center from the checkboxes
- The checkboxes and text are misaligned on default theme on Android

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -171,12 +167,10 @@ const Checkbox = ({
 
   return (
     <div className={checkboxClasses}>
-      <div className={selectContainer}>
```
yuderekyu: The `selectionContainer` div is redundant - the label itself can serve the same functionality, providing an inline-selection-hitbox for both the checkbox and label.

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -83,131 +83,97 @@ const defaultProps = {
...
   }
 
-  onBlur() {
-    this.setState({ focus: false });
-    if (this.props.onBlur) {
-      this.props.onBlur();
-    }
+  const checkboxClasses = cx([
+    'checkbox',
+    { 'checkbox-inline': isInline },
```
JakeLaCombe: Can we be consistent with class names? since you have is-disabled, I think is-inline would be appropriate.

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -83,131 +83,97 @@ const defaultProps = {
...
+    customProps.className,
+  ]);
+
+  const labelClasses = cx([
+    'label',
+    { 'is-disabled': isDisabled },
```
JakeLaCombe: I've been out of office, so I apologize for the late post. For html attributes that have a direct correlation to their child (ex: id to id, name to name),  I believe we would just use the attribute name as it is instead of prefixing it with something else. Can we do that here?

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -83,131 +83,97 @@ const defaultProps = {
...
+  const labelTextClasses = cx([
+    'label-text',
+    { 'is-hidden': isLabelHidden },
+    { 'is-mobile': CheckboxUtil.isConsideredMobileDevice() },
+    { 'is-mobile-focus-ring': CheckboxUtil.isConsideredMobileDevice() },
+  ]);
+
+  let labelTextContainer = null;
+  let inputContainer = null;
+  if (isLabelHidden) {
```
JakeLaCombe: Instead of having the duplicate section of code for this if statement, can we attach aria-label to controlInputAttrs?

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -83,131 +83,97 @@ const defaultProps = {
...
+
+  const labelTextClasses = cx([
+    'label-text',
+    { 'is-hidden': isLabelHidden },
+    { 'is-mobile': CheckboxUtil.isConsideredMobileDevice() },
+    { 'is-mobile-focus-ring': CheckboxUtil.isConsideredMobileDevice() },
+  ]);
+
+  let labelTextContainer = null;
+  let inputContainer = null;
```
JakeLaCombe: I don't know if container is the write variable name here, as it's the actual input element. I think inputElement is a better name for this variable.

packages/terra-form-checkbox/src/Checkbox.jsx
yuderekyu: The prefix is to attach an `is-disabled` css class - instead, are you saying we should change the css class to use a selector: `.native-input:disabled`?

packages/terra-form-checkbox/src/Checkbox.jsx
kschuste: +1

packages/terra-form-checkbox/src/Checkbox.jsx
```diff
@@ -110,10 +110,14 @@ const Checkbox = ({
 
   const checkboxClasses = cx([
     'checkbox',
-    { 'checkbox-inline': isInline },
+    { 'is-inline': isInline },
     customProps.className,
   ]);
 
+  const selectContainer = cx([
```
yuderekyu: Ok, so I was wrong to remove this. If a theme uses a css checkbox, removing `select-container` and adding the `padding-top` &`padding-bottom` to `label` pushes the label text to be misaligned (because the padding applies to the label text, but not the absolutely positioned css checkbox). So having this outer container to increase the hit-box should be fine.

packages/terra-form-checkbox/src/Checkbox.jsx
JakeLaCombe: Sorry. I meant renaming the isDisabled prop to disabled. I'm fine with the classname and the rest of the implementation.

yuderekyu:
@mmalaker @neilpfeiffer 
* Android alignment will be a wontfix - looks like android chrome has inherent issues with alignment. We'll plan on a future enhancement to make open source theme use a custom checkbox, instead of native.
* Focus ring has been removed from mock theme. Focus ring was off center because mock theme resizes checkboxes, but focus ring does not get resized with it. Because the combination of mobile resize and focus ring are not together in any themes, the mock theme has been altered: 
  1. focus ring is replaced with a box-shadow 
  2. mobile resize is kept.
* The Firefox rendering issue with a default-checked checkbox seems to be a theme provider issue interacting with pseudo css selectors. @bjankord and I tested the checkboxes with the site defaulting to mock theme with no issues. The issue occurs on theme swap `default->mock`. A separate issue will be logged for this.

mmalaker:
Functional verification completed. The checkbox and focus ring examples display and function correctly, with the exception of the styling issues noted above.

neilpfeiffer:
Tested: PR #1096
Resolves: Issue #1090 & #1093

**Reviewed:**
- Follows pattern considerations documented in Cerner Design Standards: [Checkbox](https://wiki.ucern.com/x/CZUCYQ)
- Follows UX defined design implementation for Terra-Form-Checkbox

\
**Future Enhancements:**
- Known Issue #1117 (noted above): When swapping themes, a preselected checkbox fails to render on Firefox, appears correctly once something forces a re-render.
- UI idea still on the list for future, no issue logged yet – copied from https://github.com/cerner/terra-core/pull/1037#issuecomment-351745871: Look at increasing the hitsize of the checkbox + label from 19px to 30px (by adjusting padding/margins, mainly for when is only one line) for better touch target

\
\

**Pass** |
-----------|
UX Review|

PR #1115, opened 2018-01-03
Menu accessibility

### LINKED ISSUE #1066: Divided Menu Accessibilty Issue
# Issue Description
Menu components with dividers fail accessibility because the `<hr/>` dividers cannot be direct descendants of `<ul>`.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Passes accessibility tests

## Current Behavior
```
List element has direct children that are not allowed inside <li> elements
```

## Steps to Reproduce

1. Install [Axe Chrome Extension](https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd?hl=en-US)
2. Navigate to a [Large Menu Test](http://engineering.cerner.com/terra-core/#/tests/menu-tests/large) and open the menu by clicking on it.
3. Run axe by going to developer tools and clicking axe tab.

## Environment

* Component Version:
* Browser Name and Version:

jmsv6d:
### Summary
Fixes #1066 Added menu roles and updated the divider to be an li element with the role of separator instead of an hr. This solved accessibility issues with list having children that were not li elements. 

### Additional Details
It appears like the hr element was getting default bottom padding that was not included in the original callouts. Switching to an li removed this padding, which is correct, but the screenshots needed to be updated. 

@cerner/terra


packages/terra-menu/src/MenuItemGroup.jsx
```diff
@@ -31,14 +31,9 @@ class MenuItemGroup extends React.Component {
...
     return (
-      <SingleSelectList.Item
-        {...customProps}
-        content={(
-          <SingleSelectList onChange={onChange} >
-            {children}
-          </SingleSelectList>
-        )}
-      />
+      <SingleSelectList {...customProps} onChange={onChange} role="group">
```
emilyrohrbough: Should `SingleSelectList` be setting the role to group itself?

emilyrohrbough: Actually, nevermind, I would expect List to have the role of list

Issue #1088, opened 2017-12-18
Add fallback placeholder image to terra-image component


bjankord:
# Issue Description
Need to add ability to displays an fallback image while the image is being loaded. If the src image successfully loads, the fallback image is replaced with the src image. Otherwise, the fallback image is left as is. Look at [terra-profile-image for implementation details](https://github.com/cerner/terra-core/blob/master/packages/terra-profile-image/src/ProfileImage.jsx)

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
If the src image successfully loads, the fallback image is replaced with the src image. Otherwise, the fallback image is left as is.

cwalten:
# Tech design for adding a placeholder image to the Image component and refactoring ProfileImage to utilize this
## Summary:
Logic will be moved from ProfileImage to Image which will control loading the src image and providing a placeholder image both while loading and in case of failure. This logic will only be executed should a placeholder prop be provided. Otherwise, nothing will render during loading or in case of failure (current implementation). ProfileImage should then be able to utilize this functionality by providing the ProfileImage svg placeholder as a prop.

## React Props Added
### Image 
| Prop     | Type | Default | Description  |
|----------|------|---------|--------------|
| placeholder | string | undefined | An svg file which will be put in place of a loading or failed src image |

## Examples
```
import placeholderImage from './placeholder.svg';

<Image
  src={'/my/server/media/route'}
  placeholder={placeholderImage}
/>
```

```
// Placeholder image is built into this component
<ProfileImage
  src={'/my/server/media/route'}
/>
```

bjankord:
> This logic will only be executed should a placeholder prop be provided. Otherwise, nothing will render during loading or in case of failure (current implementation).

I think we should make this logic default behavior for Image. To do this, we'd make the placeholder prop have a defaultProp value of string of a generic SVG fallback image.

cwalten:
@tbiethman discussed this with me, and we had decided this was not desired default behavior, especially for backwards compatibility. Maybe he has a comment on this as well?

tbiethman:
[...]

PR #1129, opened 2018-01-05
Refactored ProfileImage and Image to make Image handle placeholder logic

cwalten:
### Summary
Addresses #1088 
* Move placeholder logic out of ProfileImage and into Image
* Placeholder logic is opt-in for Image, if no placeholder prop is provided, the component will work as before
* ProfileImage has been simplified to passing props onto Image and providing a specific loading/failure avatar as the placeholder

### Deployment URL
https://terra-core-deployed-pr-1129.herokuapp.com/static/#/site/image


packages/terra-image/src/Image.jsx
```diff
@@ -47,36 +51,106 @@ const defaultProps = {
...
+
+  render() {
+    const { src, variant, isFluid, alt, placeholder, height, width, onLoad, onError, ...customProps } = this.props;
+    delete customProps.className;
+
+    if (placeholder) {
+      if (this.state.isLoading) {
+        return (
+          <div>
+            <div className={styles.hidden}>{this.createImage(customProps)}</div>
```
yuderekyu: We should localize this css class, `cx('hidden')`.

packages/terra-image/src/Image.jsx
```diff
@@ -23,6 +23,10 @@ const propTypes = {
    * The text content that specifies an alternative text for an image.
    */
   alt: PropTypes.string.isRequired,
+  /**
+   * An image which will be displayed during loading and in case of src load failure
```
kschuste: update this to include that this is the "src string of the image"

currently it makes me thing that I should pass in an Image node and not a string

packages/terra-image/src/Image.jsx
```diff
@@ -47,36 +51,106 @@ const defaultProps = {
...
+
+  constructor(props) {
+    super(props);
+
+    this.state = { isLoading: true, isError: false };
+    const { src, variant, isFluid, alt, placeholder, height, width, onLoad, onError, ...customProps } = this.props;
+
+    this.classes = cx([
+      'image',
+      variant,
```
tbiethman: The `variant`, `isFluid`, and `customProps.className` props aren't update safe. Since you're caching this set of classes in the constructor, you'd need to recreate it in componentWillReceiveProps in case those props have changed values.

It seems like it'd be nice to cache the classnames, but it might be more work than desirable. If you decide to keep this caching, I'd just store it in state directly. It'll make your componentWillReceiveProps implementation more complicated, but it'd be good to sync the classnames with the normal component refresh cycle.

packages/terra-image/src/Image.jsx
JakeLaCombe: How do you feel about making it more specific as well (placeholderSrc)? I vote for this approach as some of the html nodes do have a placeholder attribute.

packages/terra-image/src/Image.jsx
JakeLaCombe: My vote is not to cache the classes personally. Do we get a huge performance gain by caching them?

packages/terra-image/src/Image.jsx
```diff
@@ -47,36 +51,106 @@ const defaultProps = {
...
+      { fluid: isFluid },
+    ]);
+
+    this.handleOnLoad = this.handleOnLoad.bind(this);
+    this.handleOnError = this.handleOnError.bind(this);
+  }
+
+  componentWillReceiveProps(newProps) {
+    // If a new image is being loaded, reset the state to loading
+    if (newProps.src !== this.props.src) {
```
JakeLaCombe: Should we also check the placeholder value as well here?

packages/terra-image/src/Image.jsx
```diff
@@ -47,36 +51,106 @@ const defaultProps = {
...
+      this.setState({ isLoading: true, isError: false });
+    }
+  }
+
+  handleOnLoad() {
+    this.setState({ isLoading: false });
+    const onLoad = this.props.onLoad;
+    if (onLoad !== undefined) {
```
JakeLaCombe: Nitpick. 

For this if statement and the one in handleOnError, can you put a new line before the if statement?

packages/terra-image/src/Image.jsx
```diff
@@ -47,36 +51,106 @@ const defaultProps = {
...
+  }
+
+  createPlaceholderImage(customProps) {
+    const { placeholder, alt, height, width } = this.props;
+    return (
+      <img
+        src={placeholder}
+        alt={alt}
+        height={height}
+        width={width}
```
JakeLaCombe: What if it errors out here? It feels like we should provide some kind of fallback for this image as well.

packages/terra-image/src/Image.jsx
```diff
@@ -47,36 +51,106 @@ const defaultProps = {
...
+        onError={this.handleOnError}
+        className={this.classes}
+        {...customProps}
+      />
+    );
+  }
+
+  render() {
+    const { src, variant, isFluid, alt, placeholder, height, width, onLoad, onError, ...customProps } = this.props;
+    delete customProps.className;
```
JakeLaCombe: This is why I don't recommend caching. You have 3 different places where you have to deal with the custom class logic instead of one.

packages/terra-image/src/Image.jsx
```diff
@@ -47,36 +51,106 @@ const defaultProps = {
...
+          <div>
+            <div className={styles.hidden}>{this.createImage(customProps)}</div>
+            <div>{this.createPlaceholderImage(customProps)}</div>
+          </div>
+        );
+      }
+
+      return (
+        <div>
+          <div>{this.state.isError ? this.createPlaceholderImage(customProps) : this.createImage(customProps)}</div>
```
JakeLaCombe: What if it's the placeholderImage that threw an error here, and not the original image?

jestconfig.json
```diff
@@ -6,6 +6,7 @@
     "./node_modules/enzyme-to-json/serializer"
   ],
   "moduleNameMapper": {
-    "\\.(css|scss|svg)$": "identity-obj-proxy"
+    "\\.(svg)$": "<rootDir>/__mocks__/fileMock.js",
```
emilyrohrbough: Where are you using this file? I don't see a reference in the tests.

packages/terra-image/src/Image.jsx
emilyrohrbough: Here we are toggling `createPlaceholderImage` and `createImage` based on the error state. However, when we call the `createPlaceholderImage`, we do not pass down the `onLoad` or `onError` props. I feel this would be the case where you'd expect onError to be called.

packages/terra-image/src/Image.jsx
tbiethman: It would be safer and more straightforward to implement 'shouldComponentUpdate' to prevent unnecessary updates when props haven't changed (if we're really concerned about this component's performance).

packages/terra-image/src/Image.jsx
kschuste: +1

jestconfig.json
cwalten: it is necessary for Jest to properly import svg files. [ref](https://facebook.github.io/jest/docs/en/webpack.html#handling-static-assets)

packages/terra-image/src/Image.jsx
cwalten: What do others think about this? It was never discussed, having a fallback for the placeholder

packages/terra-image/src/Image.jsx
cwalten: I agree, I have removed the caching. It was modeled after the old ProfileImage, but it is causing more harm than help now

packages/terra-image/src/Image.jsx
cwalten: as of now, the placeholder rendering cannot trigger isError to be set to true I don't think. maybe that is a logic flow that needs to be explored more, with the discussion of a fallback for the placeholder

cwalten: placeholder is not controlling the state of this component at all. It is simply rendered in place of a loading image. If we want it to have some say in state and by extension what is rendered, its area of control needs to defined more explicitly

jestconfig.json
emilyrohrbough: > necessary for Jest to properly handle imports of svg files

Ah this statement here answers my question. I was wondering where it was getting imported at to be used in the tests. I realize it his here.

Give this, I feel we should Map fileMock to it's path, not every .svg file.

mjhenkes:
@cwalten please add an example to site using this functionality.

packages/terra-image/src/Image.jsx
JakeLaCombe: If UX doesn't deem it a priority, I'm fine with logging an issue to enhance this later. I think we need to have a universal fallback or all error scenarios, but since we don't have one right now, I fine if we don't address it at the moment.

packages/terra-image/src/Image.jsx
JakeLaCombe: This may not have been fully established, and since we don't have a universal fallback atm, I would be okay if we log an issue to come back to it.

jestconfig.json
cwalten: Every svg (or any other static assets for that matter) that is used in Jest has to be mocked or Jest will fail to execute. Since the scope of this mock is Jest testing, I'm not sure providing a specific path is advantageous here.

Or am I not understanding correctly what you are recommending? Can you elaborate?

packages/terra-image/src/Image.jsx
```diff
@@ -23,6 +23,10 @@ const propTypes = {
    * The text content that specifies an alternative text for an image.
    */
   alt: PropTypes.string.isRequired,
+  /**
+   * The source string for the image which will be displayed during loading and in case of src load failure.
+   */
+  placeholderSrc: PropTypes.string,
```
mjhenkes: Could we change this to a node for greater flexibility?

packages/terra-image/src/Image.jsx
```diff
@@ -47,36 +51,113 @@ const defaultProps = {
...
+            }
+          </div>
+        </div>
+      );
+    }
+
+    return (
+      <div>
+        <div>{this.createImage(customProps, imageClasses)}</div>
+      </div>
```
StephenEsser: Why are both the error and this render double nested with divs?

cwalten: That was how the old ProfileImage was written, where this code is reimagined from. I have removed them

packages/terra-image/src/Image.jsx
cwalten: updated

packages/terra-image/src/Image.jsx
```diff
@@ -23,6 +23,10 @@ const propTypes = {
    * The text content that specifies an alternative text for an image.
    */
   alt: PropTypes.string.isRequired,
+  /**
+   * The source string for the image which will be displayed during loading and in case of src load failure.
+   */
+  placeholderSrc: PropTypes.node,
```
mjhenkes: Sorry, I should have been more clear. I'd like this to take a react element instead of a source string and just render the react element.

cwalten: ah, okay, my mistake. updates have been pushed

packages/terra-image/src/Image.jsx
```diff
@@ -47,36 +51,93 @@ const defaultProps = {
...
+
+    this.state = { isLoading: true, isError: false };
+
+    this.handleOnLoad = this.handleOnLoad.bind(this);
+    this.handleOnError = this.handleOnError.bind(this);
+  }
+
+  componentWillReceiveProps(newProps) {
+    // If a new image is being loaded, reset the state to loading
+    if (newProps.src !== this.props.src || newProps.placeholder !== this.props.placeholder) {
```
emilyrohrbough: Should we be expecting the placeholder element to be changing? I would have expected it to be static.

jestconfig.json
emilyrohrbough: From what I am understanding, every svg will be mapped to fileMock.js based on this config
``` "\\.(svg)$": "<rootDir>/__mocks__/fileMock.js",```

I was wondering if we should update this to use regex to specify which fileMock. I have been thinking on it though, and if we run into an instance were we would expect a different mock we could update this then.

File: packages/terra-image/src/Image.jsx (hunk 13 above)
Comment (JakeLaCombe) [reply packages/terra-image/src/Image.jsx]: For most of the cases, I imagine the placeholder won't change, but I still think it's a good idea to handle the case when a developer changes it.

File: packages/terra-image/src/Image.jsx [hunk 14]
```diff
@@ -47,36 +51,93 @@ const defaultProps = {
...
+    return (
+      <img
+        src={src}
+        alt={alt}
+        height={height}
+        width={width}
+        onLoad={this.handleOnLoad}
+        onError={this.handleOnError}
+        className={imageClasses}
+        {...customProps}
```
emilyrohrbough: We want to spread customProps first and then apply ours

cwalten: so move this line to 94?

cwalten: updated!

packages/terra-profile-image/tests/jest/ProfileImage.test.jsx
```diff
@@ -1,22 +1,8 @@
 import React from 'react';
-import ProfileImage from '../../lib/ProfileImage';
+import ProfileImage from '../../src/ProfileImage';
```
bjankord: Probably need to switch this back from `src` to `lib`.

packages/terra-site/src/examples/image/ImageDefault.jsx
```diff
@@ -19,6 +19,9 @@ const ImageDefault = () => (
     <br />
     <h2>Fluid</h2>
     <Image src={placeholderPic1000x200} alt="fluid image" isFluid />
+    <br />
+    <h2>Placeholder</h2>
+    <Image src={placeholderPic150x150} placeholderSrc={placeholderPic1000x200} alt="placeholder image" />
```
JakeLaCombe: This example makes it hard to test. Can we use a completely different image to test with?

cwalten: What do you mean? The src and placeholderSrc are different images

mmalaker:
Per @JakeLaCombe 's comment above, the placeholder image needs to look different from the other image examples, otherwise you can't tell that the placeholder functionality is working.

packages/terra-site/src/examples/image/ImageDefault.jsx
JakeLaCombe: The main issue is that it's the same as the fluid image, and it's making it difficult for our solution designer to test that the placeholder is working. 

I suggest making it a completely different image that isn't the space image we use for the rest of the images on that page.

mmalaker:
Functional verification completed. The image examples display and function as expected.

PR #1134, opened 2018-01-09
Enable theme repo setup

bjankord:
### Summary
* Adding `src` and `tests` directories to npm packages so we can access them in the theme repos.
* Updated postinstall script to be able to detect in what situation it is being run in. (npm install in monorepo vs installing monorepo into another project)
* Add alias config for webpack to use in the theme repos. In the future this will be auto-generated.


scripts/post-install/index.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable import/no-extraneous-dependencies */
+const shell = require('shelljs');
+
+/*
+ The postinstall script is run both when installing packages in the monorepo via `npm install`
+and when the monorepo package is installed into other projects, like the theme repo.
+
+The isInstalled variable allows us to detect between to these 2 cases.
+
+When `npm install` is run within monorepo, process.cwd() does not include `node_modules`
```
mjhenkes: Clever detection 👍

webpack-aliases.js
```diff
@@ -0,0 +1,60 @@
+const path = require('path');
+
+// This file should be auto-generated
```
mjhenkes: I do wish we could get away without aliasing.

emilyrohrbough:
With the assumptions made with the terra-site configuration, all site and test examples are assumed to be living in an `examples` folder. In the script `npm run compile` this will compile `src` to `lib` and `examples` to `lib/examples`. With this, I do not think we need to access `tests` or `src`

webpack-aliases.js
```diff
@@ -0,0 +1,60 @@
...
+export default {
```
emilyrohrbough: The way you are referencing the button package [here](https://github.cerner.com/terra/cerner-consumer-theme/pull/8/files#diff-c952e4bc40cbd3318d484d1ef45b4108R4), I would think you only need to alias the repos.

emilyrohrbough: After looking at the webpack config documentation, an options could be to generate an index.js file for terra-core which would map to each package?

bjankord: I've moved this file into the theme-repo. Ran into some file path issues in how `__dirname` was being resolved.

bjankord:
@emilyrohrbough I've included `src` and `tests` so that we can reference them in the theme repo when installing terra-core into the theme repo via a git url, `"terra-core": "git://github.com/cerner/terra-core.git#theme-repo-setup"`.

All of the individual packages like terra-alert, terra-button do not contain `lib` due to `lib` being ignored in our `.gitignore`. However, it also looks like the `.npmignore` file in the individual packages is in play in this scenario as the `src` directory in the individual packages like terra-alert, terra-button is also not included.

This results in terra-button not including any react code when terra-core is installed in this way. Heres what the `theme-repo/node_modules/terra-core/packages/terra-button` directory looks like after installing this way.

```
// theme-repo/node_modules/terra-core/packages/terra-button
|-docs
|- CHANGELOG.md
|- LICENSE
|- NOTICE
|- package.json
|- README.md
```
Because of this, I've added `src` and `tests` directories to the npm package.

emilyrohrbough:
Late to the game, but @bjankord you're comment makes sense. +1.

Issue #1000, opened 2017-11-08
Radio button component for OCS


yuderekyu:
# Issue Description
Create a radio button component for OCS. Tech design here: #960

## Issue Type

- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

bjankord:
Resolved in #1135

Issue #955, opened 2017-10-23
Control Components Have Unexpected Padding


Singleton06:
# Issue Description

When trying to do a UI that has a label and radio buttons on the same line, I am running into spacing issues.  Here in the UI that I am trying to create:
![image](https://user-images.githubusercontent.com/2235110/31890821-269057f4-b7ca-11e7-9829-45e2202958bb.png)

To accomplish this, I am trying to do this using an Arrange where the label is the 'fitStart' prop and the radio buttons (Control component) are surrounded in a div and passed to the 'fill' prop. [Here](https://gist.github.com/Singleton06/09af249e81bca2fd7cefbcebbb3492cb) is the code.

This screenshot shows how the current spacing is working.
![image](https://user-images.githubusercontent.com/2235110/31893720-dda6b01c-b7d1-11e7-9eaf-f6fba1c7d6db.png)

This seems to be an issue caused by the margin on the controls as can be seen here: 
![image](https://user-images.githubusercontent.com/2235110/31893869-4c904204-b7d2-11e7-913e-eb18e3f04c5c.png)

I have found that if I remove the align "center" then it seems to align better, but talking with @emilyrohrbough, she indicated that it still might be something to be concerned about.  The behavior of "center" does seem a bit weird when used with controls that contain margin.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

## Expected Behavior

The Text component and the Control components would be aligned when using 'center' alignment.

## Current Behavior

The Text component and the Control components are not aligned when using 'center' alignment.  See above for more information.

## Steps to Reproduce

1. See code [here](https://gist.github.com/Singleton06/09af249e81bca2fd7cefbcebbb3492cb)

## Environment

* Component Version:
  * terra-form: 1.11.1
  * terra-arrange: 1.12.1
  * terra-text: 1.11.1
* Browser Name and Version: Chrome Version 62.0.3202.62 (Official Build) (32-bit)
* Operating System and version (desktop or mobile): Windows 10

yuderekyu:
`terra-form` will be deprecated, including this control component - this issue should be fixed in the new `terra-form-radio`.

bjankord:
Resolved in #1135

Issue #1106, opened 2017-12-27
ThreadLoader Doesn't Run ThemingPlugin


emilyrohrbough:
# Issue Description
When updating the themeable variables in the button PR I noticed the themeabale-variables.json file was not updating with the new values. The Theming Plugin was only hit when I removed [this style-loader boolean check](https://github.com/cerner/terra-core/blob/master/packages/terra-site/webpack.config.js#L43). I am wondering if the treads prevents the variables from being aggregated.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

## Expected Behavior
Starting the webpack-dev-server locally should use the ThemingPlugin to generate an up-to-date themeable-variables.json file.

## Current Behavior
Starting the webpack-dev-server locally is not using the ThemingPlugin.

emilyrohrbough:
the depoloyed site for Terra-core uses terra-dev-site's webpack configuration, which no longer uses threadloader so this issue is no longer valid.

PR #1135, opened 2018-01-09
Radio OCS

yuderekyu:
### Summary
Radio button for OCS. The component is themeable, responsive, and allows for text wrapping. Focus can be triggered with tab; toggle can be triggered with the space bar. Defaults to terra open source theme, which uses browser native styling for the input. Themes not using browser native styling utilize a css radio button and hide the input element.

### Addresses
#1000 
#925 (for radio buttons)
#955

### Deployment Links
https://terra-core-deployed-pr-1135.herokuapp.com/static/#/site/form-radio
https://terra-core-deployed-pr-1135.herokuapp.com/static/#/tests/form-radio-tests


packages/terra-form-radio/src/Radio.jsx
```diff
@@ -0,0 +1,178 @@
...
+  /**
+    * Whether the label is hidden.
+    */
+  isLabelHidden: PropTypes.bool,
+  /**
+    * Text of the label.
+    */
+  labelText: PropTypes.node.isRequired,
+  /**
+    * Additional objects for the text object.
```
jmsv6d: Additional *attributes

packages/terra-form-radio/src/Radio.jsx
```diff
@@ -0,0 +1,178 @@
...
+    */
+  labelTextAttrs: PropTypes.object,
+  /**
+    * Name of the input attribute.
```
jmsv6d: *Name attribute of the input.

packages/terra-form-radio/src/Radio.scss
```diff
@@ -0,0 +1,207 @@
...
+    .label {
+      cursor: pointer;
+      display: inline-block;
+      font-size: var(--terra-form-radio-font-size, 1rem);
+      line-height: var(--terra-form-radio-line-height, 1.42);
+      padding-left: var(--terra-form-radio-padding-left, none);
+      position: relative;
+    }
+
+    .is-mobile-label {
```
jmsv6d: You should increase the specificity of this. Currently it's the same specificity as the .label styles above. If anyone ever comes and switches the order of these then the mobile styles would not override the normal style.

packages/terra-form-radio/src/Radio.jsx
```diff
@@ -0,0 +1,178 @@
...
+    { 'is-mobile': RadioUtil.isConsideredMobileDevice() },
+  ]);
+
+  const innerRingClasses = cx([
+    'inner-ring',
+  ]);
+
+  let labelTextContainer = null;
+  if (isLabelHidden) {
+    controlInputAttrs['aria-label'] = labelText;
```
JakeLaCombe: Wouldn't we always want the aria-label attribute set in the input element?

yuderekyu: When the input is not hidden, the label text is accessible to everyone including screen readers. In this case, isn't `aria-label` not needed?

packages/terra-form-radio/src/Radio.scss
```diff
@@ -0,0 +1,207 @@
...
+        height: var(--terra-form-radio-inner-ring-is-mobile-checked-height, none);
+        width: var(--terra-form-radio-inner-ring-is-mobile-checked-width, none);
+      }
+    }
+
+    /* Hide label if isLabelHidden true */
+    /* Label still needs to be big enough to house the radio */
+    .is-hidden {
+      cursor: default;
+      height: var(--terra-form-radio-label-is-hidden-height, 14px);
```
kschuste: use rems instead of px here and for the width

packages/terra-site/themeable-variables.json
```diff
@@ -2,16 +2,16 @@
   "Alert": {
     "--terra-alert-icon-font-size": "1.286em",
     "--terra-alert-custom-text-color": "#1c1f21",
-    "--terra-alert-alert-background-color": "#fed1d1",
-    "--terra-alert-error-background-color": "#fed1d1",
+    "--terra-alert-alert-background-color": "#fedcdc",
+    "--terra-alert-error-background-color": "#fedcdc",
```
kschuste: I'm assuming the changes in this file outside of what's underneath Radio were pulled in from master?

yuderekyu: Actually, I locally removed a boolean check specified in #1106 to get these to update, otherwise the theming plugin wasn't updating these values. We should probably look more into this/remove it in a separate PR.

packages/terra-form-radio/.npmignore
```diff
@@ -0,0 +1,6 @@
+*.log
+node_modules
+src
```
bjankord: Need to remove `src` and `tests` from this file. These were removed from all packages in a recent PR. I'll get the generator updated here soon so this won't be an issue any more for new packages.

yuderekyu:
After offline discussion with @neilpfeiffer: margin-top now defaults to 0 and margin-bottom is added as a themeable variable.

kschuste:
@yuderekyu looks like need to update your wdio reference screenshots to get the build to succeed.

packages/terra-form-radio/src/Radio.jsx
```diff
@@ -0,0 +1,178 @@
...
+    'outer-ring',
+    { 'is-mobile': RadioUtil.isConsideredMobileDevice() },
+  ]);
+
+  const innerRingClasses = cx([
+    'inner-ring',
+  ]);
+
+  let labelTextContainer = null;
+  if (isLabelHidden) {
```
emilyrohrbough: Could we add this check in the classNames and apply a style of `display:none` instead of toggling if we apply the content?

packages/terra-form-radio/src/Radio.jsx
emilyrohrbough: One thing to check here would be if an aria-label value was passed in via custom props & if not, apply the labelText.

packages/terra-form-radio/tests/wdio/radio-spec.js
```diff
@@ -0,0 +1,55 @@
+/* eslint-disable no-unused-expressions */
+/* global browser, Terra, before */
+
```
emilyrohrbough: You'll want to specify you're viewport size(s) at the top like 
```const viewports = Terra.viewports('tiny', 'medium');```
And then pass them into the your test steps as so:
```
    Terra.should.beAccessible({ viewports });
    Terra.should.matchScreenshot({ viewports });
```
You could also loop the viewport sizes, but since these are static components this would be the better option.

File: packages/terra-form-radio/tests/wdio/radio-spec.js [hunk 10]
```diff
@@ -0,0 +1,55 @@
...
+describe('Radio', () => {
+  describe('Default Radio', () => {
+    before(() => browser.url('/#/tests/form-radio-tests/default'));
+    Terra.should.matchScreenshot();
+    Terra.should.beAccessible();
+    Terra.should.themeEachCustomProperty({
```
emilyrohrbough: It appears your screenshots are not updated to reflect the custom values passed in. I noticed the values you provided match the default values.

emilyrohrbough:
These look really nice!! Question, on the site page, you have this example: `[Theme Specific] Resize radio buttons on non-desktop touch device`. I am change my screen width and am not noticing a difference in the component? Additionally, why is it a square? Is this the mobile style?

yuderekyu:
^ This toggles a mobile resize for all the radio buttons on the page. It's only triggered on the mock theme. I can see how the layout and documentation is confusing - per your suggestions I'll change the toggle to a button and re-word the text to be more clear.

packages/terra-form-radio/src/Radio.jsx
yuderekyu: Currently, the `.label.is-hidden` css class just makes sure the outer `label` is large enough to contain the radio. Both ways of hiding the label textwork: making the `span` that contains the label text display to none, or not rendering the labeltext in the first place. I know we had some discussion in Checkbox about this. Anyone else have thoughts?

packages/terra-form-radio/tests/wdio/radio-spec.js
yuderekyu: Actually, almost all of the themeable variables are only applied on the non-default themes. For the default (open source theme) browser native styling reigns supreme. So unless we open source the other themes and inject them into the text examples, I'm not sure how we can actually test for those values.

packages/terra-form-radio/src/Radio.jsx
```diff
@@ -0,0 +1,178 @@
...
+    controlInputAttrs['aria-label'] = labelText;
+    labelTextContainer = <span {...labelTextAttrs} className={labelTextClasses} />;
+  } else {
+    labelTextContainer = <span {...labelTextAttrs} className={labelTextClasses}>{labelText}</span>;
+  }
+
+  return (
+    <div className={radioClasses}>
```
StephenEsser: We should spread custom props here

packages/terra-form-radio/src/Radio.jsx
```diff
@@ -0,0 +1,178 @@
...
+  name: PropTypes.string,
+  /**
+   * Function to trigger when focus is lost from the radio button.
+   */
+  onBlur: PropTypes.func,
+  /**
+    * Function to trigger when user clicks on the radio button. Provide a function to create a controlled input.
+    */
+  onChange: PropTypes.func,
+    /**
```
StephenEsser: Small formatting issue

mmalaker:
Issue found in functional verification. Tab navigation does not work with multiple radio examples. Tab skips over non-preselected radio options, regardless of theme or browser type. Mouse select still functions as expected. See: https://terra-core-deployed-pr-1135.herokuapp.com/static/#/tests/form-radio-tests/multiple

yuderekyu:
@mmalaker This is actually built in functionality of a radio button group, regardless of browser. A radio button group acts like a single control component. Tabbing will switch between radio button groups, but consumers need to use the arrow keys to move between radio buttons of the same group ([Source](https://www.w3.org/wiki/RadioButton)).

mmalaker:
@yuderekyu OK. Still seems weird, but I'll defer to W3C!

packages/terra-form-radio/tests/wdio/radio-spec.js
```diff
@@ -0,0 +1,171 @@
...
+  describe('Disabled Custom Radio', () => {
+    beforeEach(() => {
+      browser.url('/#/tests/form-radio-tests/disabled');
+    });
+
+    Terra.should.matchScreenshot();
+    Terra.should.beAccessible();
+
+    Terra.should.themeEachCustomProperty({
+      '--terra-form-radio-disabled-font-color': '#9b9fa1',
```
emilyrohrbough: These themeable variables should be testing non-default values. In the stylesheet, you define the disabled-font-color to be
> color: var(--terra-form-radio-disabled-font-color, #9b9fa1);

Here you want to verify the theming variable works with a new value for the screenshot comparison.

yuderekyu: Yeah I forgot to change that one, I'll change it 👍

mmalaker:
Functional verification completed. The radio button examples display and function as expected with theming changes applied.

neilpfeiffer:
Tested: PR #1135
Resolves: Issue #1000 #925 #955


**Reviewed:**
- Follows pattern considerations documented in Cerner Design Standards: [Radio](https://wiki.ucern.com/x/wvPzXw)
- Follows UX defined design implementation for Terra-Form-Radio

\
**Future Enhancements:**
- idea still on the list for future: (no issue logged yet – copied from https://github.com/cerner/terra-core/pull/1037#issuecomment-351745871): Look at increasing the hitsize of the input + label from 19px to 30px (by adjusting padding/margins, mainly for when is only one line) for better touch target.

\
\

**Pass** |
-----------|
UX Review|

Issue #1136, opened 2018-01-10
Change checkbox focus style


yuderekyu:
# Issue Description
Currently, form-checkbox uses box-shadow as a focus style. This should be changed to an increased border-width and a different border-color to better align with OCS and form-radio.

Issue #1130, opened 2018-01-05
Increase padding between checkbox and label-text on mobile resize


yuderekyu:
# Issue Description
Verified with @neilpfeiffer: on checkbox mobile resize, padding should be increased to keep spacing between the checkbox and label-text consistent with the desktop render.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

Issue #1126, opened 2018-01-04
Remove select-container from Checkbox


yuderekyu:
# Issue Description
Initially, the `select-container` css class was added to increase the hit-box of the checkbox. In actuality, it is only increasing the top and bottom spacing between each component. Until @neilpfeiffer creates an enhancement issue to increase the hit-box, we should remove `select-container` so the top/bottom spacing better matches the open source theme.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

PR #1137, opened 2018-01-11
Checkbox style changes

yuderekyu:
### Summary
Addresses #1126, #1130, #1136
* Removes select-container css class.
* Increase padding for labeltext when checkbox is resized for mobile devices.
* Changes focus style from `box-shadow` to `border-width`/`border-color`
* Adds `transition` to the unchecked state. This allows animations for hover and focus when the checkbox is in an unchecked state.
* Increase specificity of mobile, disabled, hidden, and inline css classes.
* Spread custom props  

### Deployment Link
https://terra-core-deployed-pr-1137.herokuapp.com/static/#/site/form-checkbox


packages/terra-form-checkbox/CHANGELOG.md
```diff
@@ -3,6 +3,15 @@ ChangeLog
 
 (Unreleased)
 ----------
+### Added
+* Increase padding on mobile resize
```
StephenEsser: Increasing the padding might belong in the `Changed` section

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -120,19 +115,12 @@
     .label:hover {
       .label-text::before {
         border-color: var(--terra-form-checkbox-hover-border-color, none);
-
-        [dir=rtl] & {
```
bjankord: What's the logic around removing the dir selectors? I thought these were needed to fix some issues where initial value is non-bidirectional, e.g. `none` and then the themed value includes styles which would be bidirectional?

yuderekyu:
After offline discussion with @neilpfeiffer: margin-top now defaults to 0 and margin-bottom is added as a themeable variable.

mmalaker:
Functional verification completed. The checkbox examples display and function correctly with the new style changes.

PR #1144, opened 2018-01-15
Move selectable functions to utils and convert to functional component.

dkasper-was-taken:
### Summary
Move selectable functions to utils, add refCallback, and convert to functional component.

@cerner/terra


packages/terra-list/src/SelectableUtils.js
```diff
@@ -86,6 +86,76 @@ const shouldHandleMultiSelect = (children, maxSelectionCount, currentIndexes, ne
  */
 const shouldHandleSingleSelect = (currentIndex, newIndex) => newIndex !== currentIndex;
 
+const wrappedOnClickForItem = (item, index, handleOnChange) => {
+  const initialOnClick = item.props.onClick;
+
+  return (event) => {
+    // The default isSelectable attribute is either undefined or true, unless the consumer specifies the item isSelectable attribute as false.
+    if (item.props.isSelectable !== false && handleOnChange) {
+      handleOnChange(event, index);
```
emilyrohrbough: IMO it'd be less confusing to just call `handleOnChange` `onChange` given this method is now outside of the selectable components and we directly pass the onChange provided to the selectableList.

packages/terra-list/src/SelectableUtils.js
```diff
@@ -86,6 +86,76 @@ const shouldHandleMultiSelect = (children, maxSelectionCount, currentIndexes, ne
  */
 const shouldHandleSingleSelect = (currentIndex, newIndex) => newIndex !== currentIndex;
 
+const wrappedOnClickForItem = (item, index, handleOnChange) => {
```
emilyrohrbough: Maybe add some docs indicating what is these methods are doing.

packages/terra-list/src/SelectableUtils.js
```diff
@@ -118,6 +124,9 @@ const wrappedOnKeyDownForItem = (item, index, handleOnChange) => {
   };
 };
 
+/**
+ * Returns an objet containing accessiblity and selectable projects.
```
emilyrohrbough: selectable ~projects~. properties?

dkasper-was-taken: good catch, i think the bbq is clouding my judgement

emilyrohrbough: Hey, its clouding mine too

PR #1148, opened 2018-01-16
Theming Repo Setup

emilyrohrbough:
### Summary
To correctly install terra-core within theming repos, we need the follow packages listed as dependencies to correctly run install, clean and compile scripts.


emilyrohrbough:
Talked with @mjhenkes. Have a few more ideas to try before using this approach

PR #1155, opened 2018-01-17
Uplift ButtonGroup styles

### LINKED ISSUE #1132: Uplift Button Group Styles
# Issue Description
Uplift styles for the Button Group component and consume Button 2.0

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

benbcai:
### Summary
Closes #1132. 

### Deployment URL
https://terra-core-deployed-pr-1155.herokuapp.com/

#### New prop
- `selectedKeys`

#### Removed props
- `buttons`, `isCompact`, `isSelectable`, `size`, `variant`

#### New props
- `isDisabled`, `onBlur`, `onFocus`, `onKeyDown`, `onKeyUp`

#### Removed props
- `children`, `isReversed`, `isSelected `

#### Changed prop
- `text`


packages/terra-button-group/src/ButtonGroupButton.scss
```diff
@@ -1,14 +1,55 @@
 :local {
   .button-group-button {
+    background-color: var(--terra-button-group-background-color, #fff);
```
bjankord: I'm on the fence of using`--terra-button-group` prefix vs `--terra-button-group-button` prefix.

The `--terra-button-group-button` prefix for these button styles is nice as it frees up the `--terra-button-group` prefix if we ever need to add theming styles to the div all the button-groups sit within and its explicit as to what the themable styles will be applied to. On the flip side, `--terra-button-group-button` is a long prefix and I understand why it was shortened. 

Anyone else have thoughts on this one way or the other?

packages/terra-button-group/tests/wdio/button-group-spec.js
```diff
@@ -1,109 +1,153 @@
...
+      '--terra-button-group-active-and-hover-background-color': 'white',
+      '--terra-button-group-active-and-focus-background-color': 'white',
+    });
 
-  describe('Compact', () => {
-    beforeEach(() => browser.url('/#/tests/button-group-tests/compact-button-group'));
     Terra.should.beAccessible();
-    Terra.should.matchScreenshot();
-  });
+    Terra.should.matchScreenshot('0');
```
bjankord: I noticed that a few of the wdio tests for button-group pass a number to `Terra.should.matchScreenshot`, what is it used for?

packages/terra-tabs/src/Tabs.jsx
```diff
@@ -189,6 +189,8 @@ class Tabs extends React.Component {
       </CollapsedTabs>
     );
 
+    debugger;
```
bjankord: Looks like there is a debugger statement that needs removed.

packages/terra-tabs/tests/nightwatch/Tabs/ControlledTabs.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import Base from 'terra-base';
-import Tabs from '../../../lib/Tabs';
+import Tabs from '../../../src/Tabs';
```
bjankord: We should probably switch this back to `lib`

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -3,92 +3,50 @@ import PropTypes from 'prop-types';
...
-   * Sets the button size. One of `tiny`, `small`, `medium`, `large`, `huge`
-   */
-  size: PropTypes.oneOf(['tiny', 'small', 'medium', 'large', 'huge']),
-
-  /**
-   * Sets the button group style variation. One of `defaut` or `secondary`
+   * A list of keys of the buttons that should be selected.
    */
-  variant: PropTypes.oneOf(['default', 'secondary']),
+  selectedKeys: PropTypes.arrayOf(PropTypes.string),
```
StephenEsser: This is probably worth getting a few opinions on. Does anyone have any thoughts on the selectedKeys api vs an explicit selection on the buttons?
```
<ButtonGroup selectedKeys={['one']}>
  <ButtonGroup.Button key="one" />
  <ButtonGroup.Button key="two" />
  <ButtonGroup.Button key="three" />
</ButtonGroup>
```
vs
```
<ButtonGroup>
  <ButtonGroup.Button isSelected />
  <ButtonGroup.Button />
  <ButtonGroup.Button />
</ButtonGroup>
```

bjankord: This first one, using the selectedKeys prop, seems like it would be easier to do multiple selected keys for the multi-select button group from the perspective of the parent component which is rendering the controlled button group.

packages/terra-button-group/docs/README.md
```diff
@@ -19,12 +21,85 @@ import ButtonGroup from 'terra-button-group';
...
+          <ButtonGroup.Button text="Single-Select 1" key="1" />
+          <ButtonGroup.Button text="Single-Select 2" key="2" />
+          <ButtonGroup.Button text="Single-Select 3" key="3" />
+          <ButtonGroup.Button text="Single-Select 4" key="4" />
+        </ButtonGroup>
+      </div>
+    );
+  }
+}
+export default SingleSelectButtonGroup;
```
StephenEsser: This is just up to your preference, I find simplified examples easier to read and understand. Here is a potential example that has the bare minimum needed to work.
```
import React from 'react';
import ButtonGroup from 'terra-button-group';

class RadioButtonGroup extends React.Component {
  constructor(props) {
    super(props);
    this.state = { selectedKey: '1' };
    this.handleSelection = this.handleSelection.bind(this);
  }

  handleSelection(event, selectedKey) {
    this.setState({ selectedKey });
  }

  render() {
    return (
      <div>
        <ButtonGroup onChange={this.handleSelection} selectedKeys={[this.state.selectedKey]}>
          <ButtonGroup.Button text="Single-Select 1" key="1" />
          <ButtonGroup.Button text="Single-Select 2" key="2" />
          <ButtonGroup.Button text="Single-Select 3" key="3" />
          <ButtonGroup.Button text="Single-Select 4" key="4" />
        </ButtonGroup>
      </div>
    );
  }
}
export default RadioButtonGroup;
```
I personally think the added helper logic in the if block to conditionally set state adds more complexity and makes it harder to understand without increasing much value / performance. 

A consumer had an issue not too long ago with their logic inside their List onChange function never fired because they had copy pasted the [example blocks](https://github.com/cerner/terra-core/blob/master/packages/terra-list/docs/Selectable.md) without understanding the utils would ignore additional clicks to an already selected item.

packages/terra-button-group/docs/README.md
```diff
@@ -19,12 +21,85 @@ import ButtonGroup from 'terra-button-group';
...
 
-// Selectable (toggle style) button group with second button pre-selected
-  <ButtonGroup isSelectable>
-    <ButtonGroup.Button text="Button 1" key="button1" />
-    <ButtonGroup.Button isSelected text="Button 2" key="button2" />
-  </ButtonGroup>
+### Single-Select Button Group
+```jsx
+import React from 'react';
+import ButtonGroup from '../../src/ButtonGroup';
```
StephenEsser: This should probably import from terra-button-group
```
import ButtonGroup from 'terra-button-group';
```

packages/terra-button-group/docs/README.md
```diff
@@ -19,12 +21,85 @@ import ButtonGroup from 'terra-button-group';
...
+
+class SingleSelectButtonGroup extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { selectedKey: ['1'] };
```
StephenEsser: Because selectedKey is singular this should probably be a single key that just gets put into an array
```
this.state = { selectedKey: '1' };
```
```
<ButtonGroup onChange={this.handleSelection} selectedKeys={[this.state.selectedKey]}>
```

packages/terra-button-group/src/ButtonGroupUtils.jsx
```diff
@@ -0,0 +1,38 @@
...
+  const clonedSelectedKeys = currentSelectedKeys.slice();
+  const index = clonedSelectedKeys.indexOf(newSelectedKey);
+
+  if (index > -1) {
+    clonedSelectedKeys.splice(index, 1);
+  } else {
+    clonedSelectedKeys.push(newSelectedKey);
+  }
+
+  return clonedSelectedKeys;
```
StephenEsser: I would do a little more testing, but I think this whole function can written as follows:
```
const handleMultiSelectedKeys = (currentSelectedKeys, newSelectedKey) => (
  Array.from(new Set(currentSelectedKeys).add(newSelectedKey))
);
```

packages/terra-button-group/src/ButtonGroupButton.scss
StephenEsser: My vote is for `--terra-button-group-button`, it's explicit and would allow custom styles on `--terra-button-group` which might need them in the future

packages/terra-tabs/src/Tabs.jsx
benbcai: I was debugging Tabs and didn't realize the change got commited. Removed. https://github.com/cerner/terra-core/pull/1155/commits/81a22fc16181f57f506752dc4df5abb6c6eeea55

packages/terra-tabs/tests/nightwatch/Tabs/ControlledTabs.jsx
benbcai: I was debugging Tabs and didn't realize the change got committed. Reverted. https://github.com/cerner/terra-core/pull/1155/commits/81a22fc16181f57f506752dc4df5abb6c6eeea55

packages/terra-button-group/src/ButtonGroup.jsx
benbcai: Agree with Brett. The first option is also more consistent with Tabs, List, and Table. The second option is more modular but would requires a relatively more complicated structure on the parent component to map the selectedKey provided in onChange to setting the isSelected prop for each button. For these reasons, I prefer the first option.

packages/terra-button-group/src/ButtonGroupButton.jsx
```diff
@@ -7,55 +7,121 @@ import styles from './ButtonGroupButton.scss';
...
+  }
+
+  render() {
+    const {
+      icon,
+      isDisabled,
+      ...customProps
+    } = this.props;
+
+    const attributes = Object.assign({}, customProps);
```
yuderekyu: General question - what's our reasoning behind using `Object.assign()` versus the spread operator `...customProps`?

packages/terra-button-group/src/ButtonGroupButton.scss
benbcai: It makes sense to use `--terra-button-group-button` as the prefix for the Buttons in the group. Updated: https://github.com/cerner/terra-core/pull/1155/commits/c4e0000d6ceaf96b85053f0046154de1799e7072

packages/terra-button-group/docs/README.md
benbcai: I just copied it from the test. Examples are now slimmed down: https://github.com/cerner/terra-core/pull/1155/commits/c4e0000d6ceaf96b85053f0046154de1799e7072

packages/terra-button-group/docs/README.md
benbcai: Updated: https://github.com/cerner/terra-core/pull/1155/commits/c4e0000d6ceaf96b85053f0046154de1799e7072

packages/terra-button-group/docs/README.md
benbcai: Good catch: https://github.com/cerner/terra-core/pull/1155/commits/c4e0000d6ceaf96b85053f0046154de1799e7072

packages/terra-button-group/src/ButtonGroupButton.jsx
benbcai: customProps would not be null if there is really no additional props so I don't think the reason is to ensure that we don't have an undefined customProps. There are components that have a need to add additional attributes. [Here](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/Select.jsx#L251) is an example. Perhaps, one reason to copy the custom props to another constant is to avoid adding additional attributes to the original customProps. But there may be other better reasons that I'm not aware of.

This line was already here before this change but I went ahead and removed it and just use customProps. https://github.com/cerner/terra-core/pull/1155/commits/5867f82edd6f7d3559befec0429fe8ee11183fa0

packages/terra-button-group/tests/wdio/button-group-spec.js
benbcai: Each one of these wdio tests involves a series of clicks and take screenshots along the way. The parameter passed to the `Terra.should.matchScreenshot` helper is to give it a custom name for the screenshots. See this [comment](https://github.com/cerner/terra-toolkit/pull/37#issuecomment-355068062). I have changed the parameter to give it a more descriptive name: https://github.com/cerner/terra-core/pull/1155/commits/9d6dd46edcf8e4be2a28ff8717995749d8cc8dde

packages/terra-button-group/src/ButtonGroupButton.scss
```diff
@@ -1,14 +1,55 @@
 :local {
   .button-group-button {
+    background-color: var(--terra-button-group-button-background-color, #fff);
+    border-color: var(--terra-button-group-button-border-color, #c8cacb);
     border-radius: 0;
     border-style: solid;
     border-width: 1px;
+    box-shadow: var(--terra-button-group-button-box-shadow, none);
```
yuderekyu: We should define ltr/rtl values for the themeable box-shadows.

packages/terra-button-group/tests/nightwatch/ButtonGroupSingelSelect.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
```
jmsv6d: Typo in file name: SingleSelect

packages/terra-button-group/tests/nightwatch/ButtonGroupOnChange.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
```
jmsv6d: This file seems to be about the same as the SingleSelect, do we really need both of them? Testing the single and multi select implementations seems to be sufficient to test the onChange.

packages/terra-button-group/tests/wdio/button-group-spec.js
```diff
@@ -1,109 +1,153 @@
...
-      rules: { 'button-name': { enabled: false } },
+  describe('Single-Select', () => {
+    before(() => browser.url('/#/tests/button-group-tests/single-select-button-group'));
+    it('should select the first button', () => {
+      browser.keys('Tab');
+      browser.keys('Space');
+      expect(browser.getText('#selected-key')).to.equal('1');
+    });
+
+    Terra.should.themeEachCustomProperty({
```
jmsv6d: You only need to test these themeable variables once, it's not necessary to do it in each of the selectable test pages.

packages/terra-site/src/examples/button-group/ButtonGroupMultiSelect.jsx
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  render() {
+    return (
+      <ButtonGroup
+        id="button-group-multi-select"
+        onChange={this.handleSelection}
+        selectedKeys={this.state.selectedKeys}
+      >
+        <ButtonGroup.Button text="Mult-Select 1" key="multi-select1" />
```
jmsv6d: nit: Multi-Select

packages/terra-button-group/src/ButtonGroupButton.scss
benbcai: @yuderekyu Many other components do separate box-shadow into lrt and rtl so I think it makes sense to be consistent. However, I'm not quite sure when using box-shadow as the focus ring. 

@neilpfeiffer Do we anticipate that the offset-x and offset-y that cast the box-shadow for buttons in a button group (or just buttons in general) be something other than `0 0` when using box-shadow as the focus ring? Consumers can set what they want for these box-shadow themeable variables. 

Do we always break up box-shadow into two themeable variables for bi-di across the board or do we make that determination case by case?

packages/terra-button-group/tests/nightwatch/ButtonGroupMultiSelect.jsx
```diff
@@ -0,0 +1,35 @@
...
+  render() {
+    return (
+      <div>
+        <h3>Selected Button(s): <span id="selected-keys">{this.state.selectedKeys.join(', ')}</span></h3>
+        <ButtonGroup
+          id="button-group-multi-select"
+          onChange={this.handleSelection}
+          selectedKeys={this.state.selectedKeys}
+        >
+          <ButtonGroup.Button text="Mult-Select 1" key="1" />
```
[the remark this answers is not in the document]
benbcai: Updated: https://github.com/cerner/terra-core/pull/1155/commits/ed6ceb87b5bfd650e1259d62a29f077c6b4ef36f

packages/terra-button-group/tests/nightwatch/ButtonGroupOnChange.jsx
benbcai: Agree and removed: https://github.com/cerner/terra-core/pull/1155/commits/ed6ceb87b5bfd650e1259d62a29f077c6b4ef36f

packages/terra-button-group/tests/nightwatch/ButtonGroupSingelSelect.jsx
benbcai: Updated:https://github.com/cerner/terra-core/pull/1155/commits/ed6ceb87b5bfd650e1259d62a29f077c6b4ef36f

packages/terra-site/src/examples/button-group/ButtonGroupMultiSelect.jsx
benbcai: Updated: https://github.com/cerner/terra-core/pull/1155/commits/ed6ceb87b5bfd650e1259d62a29f077c6b4ef36f

packages/terra-button-group/package.json
```diff
@@ -27,13 +27,13 @@
     "react": "^15.4.2",
     "react-dom": "^15.4.2",
     "terra-base": "^2.10.0",
-    "terra-button": "^1.14.0"
+    "terra-button": "^x.0"
```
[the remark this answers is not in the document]
benbcai: I'm duplicating only the themeable styles but ButtonGroup still relies on the Button component because there is some logic in Button that I don't think we should duplicate.

packages/terra-button-group/tests/wdio/button-group-spec.js
benbcai: I'm doing this to give it a lighter background color so that the test doesn't complain about insufficient color contrast.

jmsv6d: That's pretty buggy that the theme tests will effect the accessibility tests. If any of default theme values fail accessibility then we need to get different values from @neilpfeiffer

benbcai: That makes sense.

@neilpfeiffer  According to the open-source theme, below are the styles for a selected button in a button group. The contrast for these colors are insufficient for accessibility. Should these be updated to have a larger contrast?
`background-color` :  `#b2b2b2`
`color` : `#434a4d`

For the Button component, its `background-color` and `color` for a neutral variant when selected are `rgba(0, 0, 0, 0.3)`  and `#434a4d` respectively and the contrast is also  insufficient. This may also need to up updated as well.

packages/terra-button-group/src/ButtonGroupButton.scss
```diff
@@ -1,14 +1,55 @@
...
     margin: 0;
     position: relative;
 
+    &.is-disabled,
+    &:disabled {
+      opacity: var(--terra-button-group-button-disabled-opacity, 0.3);
+    }
+
+    &.is-focused,
     &:focus {
```
jmsv6d: Remove the `&:focus` selector here. This is causing the focus styles to be applied on clicks, when they should only be applied on keyboard focus.

packages/terra-button-group/src/ButtonGroupButton.scss
```diff
@@ -1,14 +1,55 @@
...
+      background-color: var(--terra-button-group-button-focus-background-color, #fff);
+      border-color: var(--terra-button-group-button-focus-border-color, #c8cacb);
+      border-width: var(--terra-button-group-button-focus-border-width, 1px);
```
jmsv6d: I'm seeing some weird resizing/shifting happening on your mock theme. I'm pretty sure it's because of changing the border width on focus. I'm wondering if we should try to achieve this style with an inset box-shadow instead.

packages/terra-button-group/src/ButtonGroupButton.scss
benbcai: Agree. https://github.com/cerner/terra-core/pull/1155/commits/553693e81282805d5c9d4db1168757c4fcbca411

packages/terra-button-group/src/ButtonGroupButton.scss
benbcai: Good catch. I was trying to mimic the consumer theme by using a heavier border width for the focus style but this has the side effect of a jumpy screen. Using an inset box-shadow does work better. https://github.com/cerner/terra-core/pull/1155/commits/553693e81282805d5c9d4db1168757c4fcbca411

mmalaker:
One issue found during functional verification. the Text and Icon button group examples do not display the selected (dark gray) style. Depending on the browser, they display a very light gray color or no color instead. On Android, the selected style displays, but not consistently. Sometimes it will display dark gray after multiple clicks to the same button, but not always.

benbcai:
@mmalaker The default behavior for button group is that the buttons are not selectable. The "Text and Icon" example are not set as selectable. Although you can click on the buttons, they do not stay as selected after you release the mouse click. 

I'll take a look on Android to see what you're seeing. Basically, when it is not selected, the background color will be transparent. When you hover over the button, the background changes to light gray. When selected, the background color is dark gray.

benbcai:
@mmalaker This is the behavior I'm seeing on Android using the "Text and Icon" example, which is what I expect:
Background is transparent (white) initially. When I tap and hold on the button, the background color changes to dark gray because it is being selected. When I release the tap, the button is still in focus so the background color changes to light gray. If I then tap outside of the button, the background changes to white (because it is no longer selected or in focus). I also tried tapping on the same button and on different buttons repeatedly and the background color appear to display correctly. 

If you're seeing a different behavior, can you describe what steps you take to recreate it?

neilpfeiffer:
@benbcai Two more items found:
1. When a button has long text and wraps, it makes only that button taller within the group, rather than making all buttons taller within group as intended. Might be something we need to look for construction and feasibility.
2. Voice over currently not correctly telling of single-select or multi-select options so additional ARIA needs to be added

benbcai:
@neilpfeiffer 
1. I will take a look at the first item. 
2. For the second item, is [aria-multiselectable](https://www.w3.org/TR/wai-aria-1.1/#aria-multiselectable) the correct ARIA to use for multi-select? There doesn't seem to be one available for single-select. Since the button group is a controlled component, it is up to the consumer to manage the selections and determine whether it is a single-select or multi-select. For this reason, the button group doesn't know if it is single or multi select. This needs to be added by the consumer. What I can do is add aria-multiselectable in the multi-select examples.

mmalaker:
Functional verification completed. The button group examples display and function correctly based on the latest changes.

benbcai:
@neilpfeiffer The first item is now fixed. As discussed, I logged issue #1205 to address adding an ARIA for accessibility to indicate whether the button group is single or multi select once we have a better understanding of the intent of the component.

neilpfeiffer:
((Will add Enhancement considerations and comments here))
 
**Pass** |
-----------|
UX Review|

PR #1159, opened 2018-01-18
Status View: Support single child

noahbenham:
### Summary
This allows `terra-status-view` to display a single child element. Current behavior only allows multiple.

### Additional Details
~We might be able to leverage [React.Children.toArray()](https://reactjs.org/docs/react-api.html#reactchildrentoarray) as an alternative, if preferred.~

### DIFF
--- a/packages/terra-status-view/src/StatusView.jsx
+++ b/packages/terra-status-view/src/StatusView.jsx
@@ -183,7 +183,7 @@ class StatusView extends React.Component {
     }
 
     let actionSection;
-    if (children.length || !Array.isArray(children)) { // Single children are objects, not arrays
+    if (React.Children.count(children) > 0) {
       actionSection = (
         <div className={cx('actions')} ref={(element) => { this.actionsNode = element; }}>
           {children}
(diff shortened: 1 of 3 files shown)


packages/terra-status-view/src/StatusView.jsx
```diff
@@ -183,7 +183,7 @@ class StatusView extends React.Component {
     }
 
     let actionSection;
-    if (children.length) {
+    if (children.length || !Array.isArray(children)) { // Single children are objects, not arrays
```
emilyrohrbough: could we use `[React.Children.count](https://reactjs.org/docs/react-api.html#reactchildrencount) > 0` here instead?

noahbenham: Good call, that's cleaner.

noahbenham:
Is this something we actually want a CHANGELOG entry for?

bjankord:
@noahbenham We can skip the changelog on this since the component has not been released yet.

bjankord:
The build passed for this when I merged it down to master, but then it ran out of time and the build was cancelled.

PR #1162, opened 2018-01-19
Form field uplift theming to meet cerner-consumer-theme

amf9t2:
### Summary
Enable the selected form-field style for theming to meet cerner-consumer-theme requirements

### Additional Details
Adding theming variable for label, error text and help text

### DIFF
--- a/packages/terra-form-field/src/Field.scss
+++ b/packages/terra-form-field/src/Field.scss
@@ -13,10 +13,14 @@
   }
 
   .label-group {
-    font-size: var(--terra-form-field-label-font-size, 1rem);
+    color: var(--terra-form-field-label-font-color, #292d30);
+    font-family: var(--terra-form-field-label-font-family, 'Roboto');
+    font-size: var(--terra-form-field-label-font-size, 14px);
+    height: 16px;
     line-height: 1.29;
     margin-bottom: var(--terra-form-field-label-margin-bottom, 0.357em);
     text-align: left;
+    width: 267px;
   }
 
   .label-group-hidden {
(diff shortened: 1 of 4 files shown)


packages/terra-form-field/src/Field.scss
```diff
@@ -13,10 +13,13 @@
   }
 
   .label-group {
+    font-family: var(--terra-form-field-label-font-family, 'Roboto');
```
benbcai: The default theme should be the open-source theme and looks like the font-family for source-source is HelveticaNeue.

amf9t2: updated. Thanks.

packages/terra-form-field/src/Field.scss
```diff
@@ -13,10 +13,13 @@
   }
 
   .label-group {
+    font-family: var(--terra-form-field-label-font-family, 'Helvetica Neue');
```
emilyrohrbough: This does not need to be added. This style will be set by the Base component that wraps Field

packages/terra-form-field/src/Field.scss
```diff
@@ -13,10 +13,13 @@
...
     font-size: var(--terra-form-field-label-font-size, 1rem);
+    height: var(--terra-form-field-label-height, auto);
```
emilyrohrbough: Why are we adding the height and width variables? The height and width will be auto-sized depending on the field control supplied as the child.

packages/terra-form-field/src/Field.scss
amf9t2: [cerner-consumer-theme](https://github.cerner.com/terra/cerner-consumer-theme/pull/1/files#diff-967ea18b58462e0c403edefef7851abfR17) requires font-family 'Roboto' 

do you mean consumer should set the  required font family in base component and we should not worry about font-family at form-field level ?

emilyrohrbough: Yes the font will be inherited from the Base component. I thought the same thing when I created these variables :)

packages/terra-form-field/src/Field.scss
amf9t2: an attempt to meet this style for [cerner-consumer-theme](https://github.cerner.com/terra/cerner-consumer-theme/pull/1/files#diff-967ea18b58462e0c403edefef7851abfR21) 
image

I will double check with UX

amf9t2:
closing because the proposed changes are not required. Thanks

PR #1173, opened 2018-01-24
Refine and Update Dialog Component

saedar:
### Summary
This code review is for the basic functionality of the terra-dialog component. I have updated tests and examples.

Fixed some lint left over from removing terra-modal-manager and terra-app-delegate.

Commit log is a bit messy. Sorry about that. :(

@cerner/terra


yuderekyu:
Is this PR not also addressing an OCS uplift?

saedar:
> Is this PR not also addressing an OCS uplift?

@yuderekyu We are doing this particular component in a series of PRs. Felt like it would be more manageable to iterate rather than get everything out all at once.

packages/terra-dialog/tests/jest/Dialog.test.jsx
```diff
@@ -11,9 +11,9 @@ describe('Dialog', () => {
   });
 
   // Prop Tests
-  it('should use the default value when no value is given', () => {
+  it('should render with the provided props', () => {
```
yuderekyu: These first two tests are identical.

packages/terra-site/src/examples/dialog/DialogDefault.jsx
```diff
@@ -1,8 +1,17 @@
...
-import Dialog from 'terra-dialog/src/Dialog';
+import Dialog from 'terra-dialog';
+
+const header = <span>Header Stuff</span>;
+
+const footer = <span>Footer Stuff</span>;
 
 const DialogDefault = () => (
-  <Dialog />
+  <div style={{ height: '200px', width: '350px', border: 'dashed' }}>
```
yuderekyu: I know we're iterating on this component so this is a placeholder - in the end, we probably want to beef up this example. Currently, I think the test example is a better representation of the look/functionality.

saedar: @yuderekyu Can you elaborate on this a little? To me, the example already looks almost identical to the test. How would you recommend beefing it up?

packages/terra-dialog/tests/jest/Dialog.test.jsx
saedar: Updated some of these: https://github.com/cerner/terra-core/pull/1173/commits/a74b4cb2f1c6bf7a2fd274272909de8b940bf0ca

packages/terra-site/src/examples/dialog/DialogDefault.jsx
yuderekyu: Sorry, I should have been more specific - we could add long text to display the wrapping/ overflow scrolling, otherwise it looks fine

saedar: @yuderekyu Made an example with long content: https://github.com/cerner/terra-core/pull/1173/commits/dcaa6ec487b139b9d11769956f38a79df5147ea0

packages/terra-dialog/CHANGELOG.md
```diff
@@ -3,4 +3,7 @@ ChangeLog
 
 Unreleased
 -----------------
-* Initial stable release
+* New package scaffold.
```
bjankord: Typicaly we leave this text as `* Initial stable release` for the v1.0.0 release

packages/terra-dialog/docs/README.md
```diff
@@ -1,6 +1,6 @@
 # Terra Dialog
 
-{insert description}
+Wrapper for terra-content-container intended to be consumed from within a modal or popup.
```
bjankord: Let's move away from talking about the terra-content-container, it is an implementation detail and we may move away from using content-container in the future. Instead let's use this verbiage for the description:

```
Dialogs are temporary views that can be used in a myriad of ways. Dialogs have the ability to alert users to act on critical information. In doing so, Dialogs may allow users to avoid destructive decisions, and also extend user workflows without disorienting the user.
```

This is pulled from the Cerner Design Standard entry for dialogs.

packages/terra-dialog/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "terra-dialog",
   "main": "lib/Dialog.js",
   "version": "0.0.0",
-  "description": "{insert description}",
+  "description": "Wrapper for terra-content-container intended to be consumed from within a modal or popup.",
```
bjankord: Let's update this with the description update to packages/terra-dialog/docs/README.md

packages/terra-date-picker/package.json
```diff
@@ -24,8 +24,6 @@
     "react-intl": "^2.3.0",
     "react-redux": "^5.0.4",
     "redux": "^3.6.0",
-    "terra-app-delegate": "^1.9.0",
```
bjankord: Curious, why are these moving to be peerDependencies?

packages/terra-dialog/src/Dialog.scss
```diff
@@ -0,0 +1,12 @@
+:local {
+  .dialog {
+    height: 100%;
+    width: 100%;
+  }
+
+  .header,
```
bjankord: I'm thinking we should drop the padding the from the header, and possible the body and footer. With padding on the header, if we pass an action header in this prop, it will not be flush with the edges which is likely want UX intends.

bjankord: Or maybe we need a prop to toggle if padding is set on each container, I can see cases where it would be nice to have this default padding and cases where it will make it challenging to get what UX intends.

packages/terra-dialog/docs/README.md
saedar: https://github.com/cerner/terra-core/pull/1173/commits/0e9b0f65bb69c708fa9c6085f050ffd0bc137b6b

packages/terra-date-picker/package.json
saedar: No idea. I think I was moving stuff around when I was having problems with hot reloading. 

Updated: https://github.com/cerner/terra-core/pull/1173/commits/0e9b0f65bb69c708fa9c6085f050ffd0bc137b6b

PR #1180, opened 2018-01-29
Lock Down selenium-chrome Version

emilyrohrbough:
Two days ago docker-selenium [released a new version](https://github.com/SeleniumHQ/docker-selenium/commit/cfc1eeb5cdcea02395016685d7012688768ed342) of both selenium/node-chrome and selenium/standalone-chrome. In our tests we pull in the latest tag of both.

The previous version is `3.8.1-erbium` and the latest release is `3.8.1-francium`. 

This pulls the previous version to see if this resolves the issues we are seeing with the repeated
` Expected element ' ' to be present - element was not found`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ cache:
     - node_modules
 before_install:
   - docker pull selenium/standalone-chrome:3.8.1-erbium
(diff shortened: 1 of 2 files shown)


emilyrohrbough:
Interesting. The failures in the last build were trying to match screenshots with the [default] tag in test steps that override this tag with a custom name.

PR #1185, opened 2018-01-30
Updating SlideGroup visibility management to support multiple Slide additions/removals

tbiethman:
### Summary
The SlideGroup hides un-presented Slides to prevent tabbing/accessibility issues. When the SlideGroup is animated, it synchronizes the hiding and showing with the presentation animations to keep things looking nice.

The current implementation assumes that the animations will complete before other content is added/removed, but that may not always be the case. In some cases, Slides are not being revealed appropriately, causing the SlideGroup to appear blank with no recourse for correction.

I updated the visibility logic to show/hide Slides based on the currently presenting/removing component and support closely repeated updates


PaulSolDev:
Functional verification complete.  No new issues identified.

IE 11- backwards navigation is choppy but I confirmed that this is consistent with the previous code.

Issue #1199, opened 2018-02-02
Individual border properties in radio and checkbox are being merged into border-shorthand syntax during webpack build with production flag breaking PostCSS custom properties transform 


bjankord:
# Issue Description
When webpack is run in production mode, some of the checkbox and radio button theme styles are converted into a shorthand syntax which breaks the PostCSS plugin transformation of CSS custom properties to static values. This issue was originally discovered in the cerner-consumer-theme repo: https://github.cerner.com/terra/cerner-consumer-theme/issues/28

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Radio and checkbox styles can be custom themed

## Current Behavior
Individual border properties in radio and checkbox are being merged into border-shorthand syntax when webpack is run in production mode which causes issues when trying to compile static values via the PostCSS custom properties plugin.

## Extended description
Did some more debugging and came across and oddity between our styles in the local dev server and in the deployed environment.

With our local dev server running:

### Given this input

```scss
.native-input:checked + .outer-ring {
      background-color: var(--terra-form-radio-outer-ring-checked-background-color, transparent);
      border-color: var(--terra-form-radio-outer-ring-checked-border-color, transparent);
      border-radius: var(--terra-form-radio-outer-ring-checked-border-radius, 0);
      border-style: var(--terra-form-radio-outer-ring-checked-border-style, none);
      border-width: var(--terra-form-radio-outer-ring-checked-border-width, 0);
      box-sizing: border-box;
}
```

### We are getting the following output in our extracted CSS file. 
The border properties remain split out individually

```scss
[dir] .Radio__radio___1mKDF .Radio__native-input___2v9UZ:checked + .Radio__outer-ring___lGCNm {
    background-color: var(--terra-form-radio-outer-ring-checked-background-color, transparent);
    border-color: var(--terra-form-radio-outer-ring-checked-border-color, transparent);
    border-radius: var(--terra-form-radio-outer-ring-checked-border-radius, 0);
    border-style: var(--terra-form-radio-outer-ring-checked-border-style, none);
    border-width: var(--terra-form-radio-outer-ring-checked-border-width, 0);
    margin-top: var(--terra-form-radio-outer-ring-checked-margin-top, 0);
}
```

In our deployed environment for the PR deployment URLs, we run wepback with the production flag. It seems one of our style processors is merging the border rules into one property using the border shorthand.

### Given this input

```scss
[...]

PR #1202, opened 2018-02-02
Update radio styles

bjankord:
### Summary
This fixes the border issue for terra-form-radio described in #1199 
It also updates theming values to use valid initial browser default for various properties

### Deployment URL
https://terra-core-deployed-pr-1202.herokuapp.com/static/#/site/form-radio

@cerner/terra


bjankord:
Hmm, looks like there are some visual diffs I need to resolve.

bjankord:
I'm looking at this in Firefox and noticing with the default styles we set for the native input, the checkbox input doesn't seem to be rendering at the size we intended. They render much smaller.

## Firefox
screen shot 2018-02-06 at 9 13 13 pm

## Chrome
screen shot 2018-02-06 at 9 14 33 pm

Thinking on this some more. I believe the issue we are trying to find a solution for is from our need to have certain styles themable in the various OCS themes, which we don't need / we dont want to exist in the default theme. 

As an example, we want width and height set on the native input in the OCS theme, we use those values to [visually hide](https://github.com/h5bp/html5-boilerplate/blob/master/src/css/main.css#L130-L140) the checkbox on the OCS theme but keep it interactable in the DOM. However, we don't want any width or height set in the default theme. We leave it up to the browser default styles to set those styles. This originally worked in this code as we had width and height set as so in the terra-core default theme.

```css
.native-input {
  height: var(--terra-form-radio-native-input-height, none);
  width: var(--terra-form-radio-native-input-width, none);
}
```
width and height were never applied to the native input from terra-core styles because `none` is not a valid value for width or height, so the browser was free to apply its default width / height to the checkbox input.

With us wanting to provide a valid property for the fallback syntax, we looked at the default values Chrome uses for width and height on checkboxes and landed on the following.

```css
.native-input {
  height: var(--terra-form-radio-native-input-height, 0.929rem);
  width: var(--terra-form-radio-native-input-width, 0.857rem);
}
```

While these work in Chrome, they are causing oddities in Firefox and Safari, haven't looked at IE yet, but I assume checkboxes probably look odd there with these dimensions set.

Knowing we only want to set these values when were are applying a custom theme (not in the default theme), I'm interested to see how the following CSS comes out of the pipeline.

```css
.native-input {
  height: var(--terra-form-radio-native-input-height); /* no fallback value */
  width: var(--terra-form-radio-native-input-width); /* no fallback value */
}
```

packages/terra-form-radio/src/Radio.scss
```diff
@@ -45,90 +45,82 @@
...
-      border-style: var(--terra-form-radio-outer-ring-border-style, none);
-      border-width: var(--terra-form-radio-outer-ring-border-width, none);
+      background-color: var(--terra-form-radio-outer-ring-background-color, transparent);
+      border: var(--terra-form-radio-outer-ring-border, 0);
+      border-radius: var(--terra-form-radio-outer-ring-border-radius, 0);
       box-sizing: border-box;
       cursor: pointer;
       display: inline-block;
-      height: var(--terra-form-radio-outer-ring-height, none);
+      height: var(--terra-form-radio-outer-ring-height, 0);
```
JakeLaCombe: Do you want to change this height as well as the rest of the heights under native inputs to be blank instead of 0?

JakeLaCombe:
Also, do we need to update the changelog for terra-site?

packages/terra-form-radio/src/Radio.scss
bjankord: This is styling a span (.outer-ring) which is a sibling to the native input element. We could set this properties as CSS custom properties without fallback values like we did for the native input styles or keep as is. What do you think? I'm fine with either way.

bjankord:
@JakeLaCombe Added a changelog entry for terra-site updates.

packages/terra-form-radio/src/Radio.scss
bjankord: Talked with @JakeLaCombe offline. Gonna leave this as is.

PR #1203, opened 2018-02-02
Add Translations for time-input

JakeLaCombe:
### Summary
This adds translations for terra-time-input.

Since all locales do not support a 12 hour clock, I reverted to an empty string for the meridiem, and created a console warning so that the developer knows the locale only takes 24 hour time.

Deployment Link: https://terra-core-deployed-pr-1203.herokuapp.com/


packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -103,8 +103,15 @@ class TimeInput extends React.Component {
     let meridiem;
 
     if (props.variant === TimeUtil.FORMAT_12_HOUR) {
-      this.anteMeridiem = context.intl.formatMessage({ id: 'Terra.timeInput.am' });
-      this.postMeridiem = context.intl.formatMessage({ id: 'Terra.timeInput.pm' });
+      if (!context.intl.messages['Terra.timeInput.am'] || !context.intl.messages['Terra.timeInput.pm']) {
+        console.warn('This locale only uses 12 hour clock. The ante meridiem and post meridiem will not be displayed');
```
emilyrohrbough: Is a console.warn something we want to add?

JakeLaCombe: I think we need to give some kind of heads up to the developer, like we did up above for when an invalid input was provided to this package.

bjankord: If there is concern of this warning appearing in prod, we could wrap it in a check so it only shows the warning when webpack is run with non prod flag.

JakeLaCombe: Overlooked that we did that in the time-input with other warnings. Added it in the following commit.

https://github.com/cerner/terra-core/pull/1203/commits/2885e321bb44b28ee4ebc59c4aa3b4c7332132bf

JakeLaCombe:
That's the goal. I can add the label.

packages/terra-time-input/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Added
+* Translations for other locales
```
bjankord: Can we note the locales added here in the changelog.

bjankord:
With #1236 fixed, is this ready for re-verification @JakeLaCombe? cc/ @mmalaker

mmalaker:
Functional verification completed. The time input examples display and function correctly when switching locales.

PR #1208, opened 2018-02-05
Disallow capture outside of lines outside of bounds with touch events

### LINKED ISSUE #1191: Signature: Capture outside of drawable canvas with touch events
# Issue Description
You can capture line segments outside of the visible signature component using touch events. This does not occur with mouse events.

ezgif-4-e62032c65d

(Using modified version of terra-site, notice the negative values of the line segments output after drawing)

## Issue Type
- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Signature capture should stop when touches (touch events) go outside of the visible signature component. It should resume when touches re-enter the signature component.

## Current Behavior
When drawing on the signature pad component capture of touches (touch events) continue outside of the visible portion of the signature component.

## Steps to Reproduce
Using a touch-enabled device...
1. Navigate to http://engineering.cerner.com/terra-core/#/site/signature
2. Begin drawing on the signature component.
3. Without lifting, continue drawing while going outside of the signature component.
4. Lift when completed drawing.

## Environment
* Component Version: 0.9.0
* Browser Name and Version: Webkit, Safari, potentially others.
* Operating System and version (desktop or mobile): iOS 11 Mobile (Touch-enabled devices)

cyrusduong:
### Summary
Resolves #1191.

### Additional Details
I could not find ways to simulate touch testing in nightwatch. If anyone knows, point me in the right direction so I can include some. Otherwise I tested it with an iPad and hosting the test site I created in #1191 to verify it no longer contained negative values or values bigger than the right or bottom bounds.

Deployment Link: https://terra-core-deployed-pr-1208.herokuapp.com


packages/terra-signature/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "terra-signature",
   "main": "lib/Signature.js",
-  "version": "0.10.0",
+  "version": "0.11.0",
```
bjankord: No need to bump the version, this will be handled during release process.

JakeLaCombe:
For what you want, I don't see it possible to test with nightwatch. You'd want more of a jest test to ensure that when the event has specific attributes that it's not getting triggered. However, I'd have to think through that process a little more

The rest looked good to me though.

BEPPSY:
+1 - Safari Mobile on iOS.

mmalaker:
Functional verification completed. The signature example displays and functions correctly on touch devices.

Issue #1157, opened 2018-01-18
Create dialog component


bjankord:
# Issue Description
Create dialog component to use inside of modal and slide panel.

## Issue Type

- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Dialog component is development in accordance with design input provided by UX.

neilpfeiffer:
Replaces closed original issue #416

Issue #416, opened 2017-05-31
Modal Dialog Component


Matt-Butler:
# Issue Description
Create Modal Dialog component. 

## Issue Type

- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The Modal Dialog is a more opinionated, structured HOC of modal. It will include a header, content header, content, and footer subcomponent. 

## Current Behavior
N/A

Matt-Butler:
Work for the modal dialog component has been started on this branch: https://github.com/cerner/terra-core/tree/modal-dialog

tm4027:
we are dividing the elements up and have some requirements around Modals on the TRM side which will be a good starting point.

PaulSolDev:
Orion JIRA created

bjankord:
Closing until we get OCS design input

PR #1209, opened 2018-02-05
Dialog - Add Theme and ActionHeader

saedar:
Resolves: #1157

### Summary
Dialog has been updated to consume ActionHeader. I have also added theme variables to Dialog, Modal, and ModalOverlay.

### Deployment URL
https://terra-core-deployed-pr-1209.herokuapp.com/static/#/site/dialog

I also updated Dialog to use subcomponents for body and footer. Since I am having to pass the header/footer/body content to ContentContainer via props, I was unsure if how I have it currently working is the best approach. If not, I welcome tips on a better path.

EDIT: Subcomponent implementation has been removed for simplicity.

@cerner/terra


packages/terra-dialog/docs/README.md
```diff
@@ -13,9 +13,22 @@ Dialogs are temporary views that can be used in a myriad of ways. Dialogs have t
...
+  <p>This is my body content.</p>
+  <p>This is some more content.</p>
+</Dialog.Body>);
+
+<Dialog header={header} footer={footer} onClose={() => alert('Close')} >
+  {body}
+</Dialog>
 ```
 
+Note: This component has an i18n dependency that requires setup per [these steps](https://github.com/cerner/terra-core/#internationalization-i18n).
```
saedar: When trying to get ActionHeader working, I ran into issues because I didn't realize the need for Base as an ancestor. I wanted to include a link and brief description here to help future consumers avoid this confusion, too. Any tips on better wording appreciated.

packages/terra-dialog/tests/jest/Dialog.test.jsx
```diff
@@ -4,26 +4,12 @@ import Dialog from '../../src/Dialog';
...
   });
 
-  // Prop Tests
-  it('should render a header', () => {
-    const container = <Dialog header="Header Content">children required</Dialog>;
-    const wrapper = shallow(container);
-    expect(wrapper).toMatchSnapshot();
-  });
-
-  it('should render header & footer', () => {
```
saedar: Talked with Neal and Brett last week. Footer is required, so don't need the special test cases with and without.

packages/terra-dialog/docs/README.md
```diff
@@ -13,9 +13,22 @@ Dialogs are temporary views that can be used in a myriad of ways. Dialogs have t
 import React from 'react';
 import Dialog from 'terra-dialog';
 
-<Dialog header="Header Content" footer="Footer Content">some body content</Dialog>
+const header = 'Header Content';
+
```
JakeLaCombe: It'd be great if we can get this API consistent. Ex: Since header doesn't require a wrapper, do we need them for body and footer?

packages/terra-dialog/src/Dialog.jsx
```diff
@@ -3,39 +3,60 @@ import PropTypes from 'prop-types';
...
- */
+  /**
+   * The text string to be placed within the header area of the dialog.
+   */
+  header: PropTypes.string.isRequired,
+  /**
+   * The footer element to be placed within the footer area of the dialog.
+   * The footer content area will not display if this prop is not provided. UNDER EVALUATION
+   */
+  footer: PropTypes.node.isRequired,
```
JakeLaCombe: Remove the isRequired part, since it doesn't sound like it's required.

packages/terra-dialog/src/Dialog.jsx
```diff
@@ -3,39 +3,60 @@ import PropTypes from 'prop-types';
...
+  /**
+   * The children to be placed within the main content area of the dialog.
+   */
   children: PropTypes.node.isRequired,
```
JakeLaCombe: I'd make this optional as well to be honest.

packages/terra-dialog/src/Dialog.jsx
```diff
@@ -3,39 +3,60 @@ import PropTypes from 'prop-types';
...
+  /**
+   * Callback function for when the close button is clicked. The close button will not display if this is not set.
+   * On small viewports a back button will be displayed instead of a close button.
+   */
+  onClose: PropTypes.func,
 };
 
-const Dialog = ({ header, footer, children }) => {
+const defaultProps = {
+  header: null,
```
JakeLaCombe: Anything required doesn't need a defaultProp I believe.

packages/terra-dialog/src/Dialog.scss
```diff
@@ -3,4 +3,19 @@
     height: 100%;
     width: 100%;
   }
+
+  .header { // May not be needed if action header is already themeable.
```
JakeLaCombe: In my opinion, I would still theme it. Terra-time-input had to apply individual styles to all of it's buttons and inputs so that it would all be aligned with each other.

packages/terra-dialog/src/Dialog.scss
```diff
@@ -3,4 +3,19 @@
...
+    border-bottom: var(--terra-dialog-header-bottom-border, none);
+  }
+
+  .dialog-body {
+    background-color: var(--terra-dialog-body-background-color, transparent);
```
JakeLaCombe: How about the foreground? Typically, when you set a background, you also need to set a foreground. Or will that be handled by another component?

packages/terra-site/src/examples/dialog/DialogDefault.jsx
```diff
@@ -1,17 +1,23 @@
+/* eslint-disable no-alert */
 import React from 'react';
 import Dialog from 'terra-dialog';
 
-const header = <span>Header Stuff</span>;
+const header = 'Header Stuff';
```
JakeLaCombe: If it's possible, it would be better if we allowed this to take in react nodes as well. With react nodes, you can build components that easily let you opt in and out of overflowing text

packages/terra-site/src/examples/dialog/DialogDefault.jsx
```diff
@@ -1,17 +1,23 @@
...
-const footer = <span>Footer Stuff</span>;
+const footer = <Dialog.Footer>Footer Stuff</Dialog.Footer>;
+
+const body = (<Dialog.Body>
+  <p>This is my body content.</p>
+  <p>This is some more content.</p>
+</Dialog.Body>);
 
 const DialogDefault = () => (
   <div style={{ height: '200px', width: '350px', border: 'dashed' }}>
```
JakeLaCombe: Given how large this hash is, I think it's best to extract it to a CSS file. Also, should the border be themeable? @neilpfeiffer

JakeLaCombe: Disregard. This is an example page.

packages/terra-dialog/package.json
```diff
@@ -32,6 +32,7 @@
     "classnames": "^2.2.5",
     "prop-types": "^15.5.8",
     "terra-base": "^2.0.0",
+    "terra-clinical-action-header": "^1.6.0",
```
JakeLaCombe: Talked with @mjhenkes on this some, and we believe that it might be ideal to move action header to the core repository. terra-core was the building blocks for repos such as terra-clinical and terra-consumer, so it feels like a circular dependency when terra-core relies on stuff set inside one of the repos that extend off of this one.

packages/terra-dialog/src/Dialog.jsx
saedar: Actually, I forgot to update the doc. The footer is required now.

packages/terra-dialog/docs/README.md
saedar: I suppose it depends on what you mean by "need." I didn't create one for header because ActionHeader is already the wrapper.

The wrappers were suggested by Brett, so I will discuss this with him when he gets in.

packages/terra-dialog/src/Dialog.jsx
saedar: What is a dialog with no content? Seems...wrong.

packages/terra-site/src/examples/dialog/DialogDefault.jsx
saedar: Can't. This gets passed into ActionHeader, which requires String.

packages/terra-dialog/docs/README.md
JakeLaCombe: I'm thinking put the wrappers inside the Dialog, and the content will just populate the wrappers in the dialog.

packages/terra-dialog/src/Dialog.jsx
JakeLaCombe: It does and I'm sure it's an unlikely case, but it's still an edge case that I'd feel safer having than not. I know for table cells that api is a little confusing in that you can't have an empty table cell.

saedar: I don't know. That seems like a different situation, in that an empty table cell feels appropriate. That signifies "No data point" rather than "No content".

packages/terra-dialog/src/DialogBody.jsx
```diff
@@ -0,0 +1,35 @@
...
+   */
+  children: PropTypes.node.isRequired,
+};
+
+const defaultProps = {
+  children: null,
+};
+
+const DialogBody = ({ children }) => {
+  const DialogBodyClassNames = cx([
```
StephenEsser: This variable should probably be camelCase

packages/terra-dialog/package.json
saedar: Talked in Scrum. Current decision is to move forward with this review and add themes to the component in Clinical. Later, we are going to evaluate deprecating the clinical ActionHeader and including a new one in core.

packages/terra-dialog/src/Dialog.jsx
saedar: https://github.com/cerner/terra-core/pull/1209/commits/2fabef8d09a87699422ef4b44f00ce87220ae782

packages/terra-dialog/src/Dialog.jsx
saedar: https://github.com/cerner/terra-core/pull/1209/commits/2fabef8d09a87699422ef4b44f00ce87220ae782

packages/terra-dialog/src/Dialog.scss
saedar: https://github.com/cerner/terra-core/pull/1209/commits/2fabef8d09a87699422ef4b44f00ce87220ae782

packages/terra-dialog/src/DialogBody.jsx
saedar: https://github.com/cerner/terra-core/pull/1209/commits/2fabef8d09a87699422ef4b44f00ce87220ae782

packages/terra-dialog/docs/README.md
bjankord: Yeah, thinking on this some more that makes sense. For the dialog subcomponets, like`<Dialog.Body>`, these will switch to an internal implementation. Likely `<div className={dialogBodyClasses}>{dialogBody}</div>`

packages/terra-dialog/src/DialogBody.jsx
```diff
@@ -0,0 +1,35 @@
...
+    'dialog-body',
+  ]);
+
+  return (
+    <div className={DialogBodyClassNames}>
```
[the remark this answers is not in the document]
saedar: https://github.com/cerner/terra-core/pull/1209/commits/2fabef8d09a87699422ef4b44f00ce87220ae782

packages/terra-dialog/docs/README.md
saedar: https://github.com/cerner/terra-core/pull/1209/commits/bafa9ec603e96bfb692803bc6c334e0be2de9cea

packages/terra-dialog/src/Dialog.jsx
JakeLaCombe: But if a developer passes in null or no children, does it change the display logic of the component? If the dialog can't display if this is null or empty, I'm fine leaving it as required. However, it the dialog handles fine with null or empty children, I still vote to drop the isRequired off this prop. It doesn't change our implementation logic, but still allows for our API to handle that edge case.

saedar: This conversation is dumb and not important. I have removed the required flag because the impact in either direction is minimal and not worth the effort already invested in the discussion.

https://github.com/cerner/terra-core/pull/1209/commits/74c002d1475e1d182bbab294b424c4d971cae104

packages/terra-modal/src/Modal.scss
```diff
@@ -1,25 +1,32 @@
 :local {
   .modal {
-    background-color: #fff;
-    border-radius: 5px;
+    background-color: var(--terra-modal-background-color, #fff);
+    border: var(--terra-modal-border, none);
```
bjankord: If there is no property value for default theme, lets drop the fallback value for these theming properties.

e.g. switch `border: var(--terra-modal-border, none);` to `border: var(--terra-modal-border);`

Looks like this can be done for the following

* border
* box-shadow
* color
* height
* margin
* max-width
* width

saedar: https://github.com/cerner/terra-core/pull/1209/commits/5385d5a7b5cd88902bfd2a88ca54455ef554df0b

mmalaker:
Functional verification completed. The dialog examples display and function correctly with the action header changes applied.

packages/terra-dialog/src/Dialog.scss
```diff
@@ -3,4 +3,23 @@
     height: 100%;
     width: 100%;
   }
+
+  .dialog-header {
+    background-color: var(--terra-dialog-header-background-color, transparent);
+    border-bottom: var(--terra-dialog-header-bottom-border, none);
+    color: var(--terra-dialog-header-foreground-color, #000);
```
neilpfeiffer: Minor color updates:
```
.dialog-header {
   color: var(--terra-dialog-header-foreground-color, #1c1f21);
}
.dialog-body {
   color: var(--terra-dialog-body-foreground-color, #1c1f21);
}
.dialog-footer {
   border-top: var(--terra-dialog-footer-bottom-top, 1px solid #dedfe0);
   color: var(--terra-dialog-footer-foreground-color, #1c1f21);
}
```

neilpfeiffer:
Tested: PR #1209
Resolves: Issue #1157 #416

\
**Reviewed:**
- Follows pattern considerations documented in Cerner Design Standards: [Dialogs](https://wiki.ucern.com/x/8PTzXw)
- Follows UX defined design implementation for Terra-Dialog

\
**Future Enhancments:** 
- Action header will be uplifted to universal design in the future
- Action footer will be created with universal design in the future

\
Verified by @scottwilmarth and @neilpfeiffer, Ready for Release after color updates.

**Pass** |
-----------|
UX Review|

PR #1217, opened 2018-02-07
Update Checkbox Styles

JakeLaCombe:
### Summary
This fixes the border issue for terra-form-checkbox described in #1199 
It also updates theming values to use valid initial browser default for various properties

### Deployment URL
https://terra-core-deployed-pr-1217.herokuapp.com/static/#/site/form-checkbox

@cerner/terra


packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -131,39 +127,39 @@
...
       content: '';
-      height: var(--terra-form-checkbox-focus-ring-height, none);
-      left: var(--terra-form-checkbox-focus-ring-left, none);
+      height: var(--terra-form-checkbox-focus-ring-height);
+      left: var(--terra-form-checkbox-focus-ring-left);
       position: absolute;
-      top: var(--terra-form-checkbox-focus-ring-top, none);
-      width: var(--terra-form-checkbox-focus-ring-width, none);
+      top: var(--terra-form-checkbox-focus-ring-top);
+      width: var(--terra-form-checkbox-focus-ring-width);
```
StephenEsser: Do these just dissolve when they are left empty or do they get added as invalid and fall back to browser defaults?

bjankord: They get added as is, `width: var(--terra-form-checkbox-focus-ring-width);`, in the extracted CSS file that runs through all the loaders. In browsers that don't support CSS custom properties, it will skip over this style as the browser views it as an invalid property and continues to use the browser default style for width on checkbox inputs. 

In browsers that support CSS custom properties, we are relying on the fact that in the default theme, we don't set a value for `--terra-form-checkbox-focus-ring-width`, making it essentially undefined allowing the browser to continue to use the default style for width on checkbox inputs.

This is an approach we'll likely start using more often where components have styles that are needed for other themes, but not used in the default theme.

PR #1219, opened 2018-02-07
Fix wdio timeouts

### LINKED ISSUE #1216: Tabs webdriver tests timeout
# Issue Description
Tabs seems to be frequently timeout out during wdio tests. Need to look into this and figure out a fix for the timeout issue.

JakeLaCombe:
### Summary
Resolves #1216 WDIO tests are timing out. While looking through the tests, the tabs did a double for loop of rendering the same screenshots. You can see the original code here. 

https://github.com/cerner/terra-core/blob/master/packages/terra-tabs/tests/wdio/tabs-spec.js#L41

To fix it, we just take a screenshot of the current viewport the test is currently running at.

### DIFF
--- a/packages/terra-tabs/tests/wdio/tabs-spec.js
+++ b/packages/terra-tabs/tests/wdio/tabs-spec.js
@@ -16,7 +16,7 @@ describe('Tabs - Responsive', () => {
         browser.moveToObject('.tabContent');
       });
       Terra.should.matchScreenshot();
-      Terra.should.beAccessible({ viewports: [viewport] });
+      Terra.should.beAccessible();
     });
     describe('Extended', () => {
       beforeEach(() => {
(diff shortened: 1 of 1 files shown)


JakeLaCombe:
I also looked at the screenshots produced, and verified that it's the same number of screenshots as well as the same content as before.

packages/terra-tabs/tests/wdio/tabs-spec.js
```diff
@@ -15,26 +15,26 @@ describe('Tabs - Responsive', () => {
         browser.setViewportSize(viewport);
         browser.moveToObject('.tabContent');
       });
-      Terra.should.matchScreenshot({ viewports });
-      Terra.should.beAccessible({ viewports });
+      Terra.should.matchScreenshot();
+      Terra.should.beAccessible({ viewports: [viewport] });
```
emilyrohrbough: Actually this an be updated to be `Terra.should.beAccessible();`
By not passing a viewport, it uses the current viewport size (which is set in the loop above).

bjankord: Interesting, we should get this updated before we merge.

PR #1220, opened 2018-02-07
Fixed js error when handle search with a null searchText.

### LINKED ISSUE #1183: Cannot read property 'length' of undefined for initial terra-search
# Issue Description

[Line 135 of SearchField.jsx](https://github.com/cerner/terra-core/blob/master/packages/terra-search-field/src/SearchField.jsx#L135) isn't able to access the length property of undefined. 

length-error

If text is added, then removed, the search goes through fine. But if the search button is hit immediately, the error shows up. I believe it's because the default state is initialized from the prop defaultValue, which has a defaultProp of "undefined". I'm not super familiar with the code of SearchField so I'm not sure of the impact changing this default prop to empty string would be.

The error only seems to show up on this placeholder text example. The other examples on the page are fine.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
It should handle empty text, regardless if it was the initial run or subsequent runs

## Current Behavior
It handles empty text if and only if the search field had text in it any time before the handleSearch function is called.

## Steps to Reproduce

1. [Go to the Terra documentation for SearchField](http://engineering.cerner.com/terra-core/#/site/search-field)
2. Immediately click on the search button for the example "Search Field with placeholder text"
3. Open developer console and see error

## Environment

* Component Version: 1.19.0
* Browser Name and Version: Chrome Version 63.0.3239.132 (Official Build) (32-bit)
* Operating System and version (desktop or mobile): desktop

benbcai:
### Summary
Closes #1183. An console error occurs on handleSearch when a defaultValue is not provided.

### Deployment URL
https://terra-core-deployed-pr-1220.herokuapp.com/

@cerner/terra

### DIFF
--- a/packages/terra-search-field/CHANGELOG.md
+++ b/packages/terra-search-field/CHANGELOG.md
@@ -6,6 +6,9 @@ Unreleased
 ### Changed
 * Uplift to Button 2.0
 
+### Fixed
+* Fixed error when handle search with a null searchText. 
+
 1.20.0 - (February 1, 2018)
 ------------------
 ### Changed
--- a/packages/terra-search-field/src/SearchField.jsx
+++ b/packages/terra-search-field/src/SearchField.jsx
@@ -143,10 +143,12 @@ class SearchField extends React.Component {
   handleSearch() {
     this.clearSearchTimeout();
 
-    if (this.state.searchText.length >= this.props.minimumSearchTextLength && this.props.onSearch) {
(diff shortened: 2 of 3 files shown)


mmalaker:
Functional verification completed. The search example no longer throws a console error when clicking the search button without a search term in the field.

PR #1221, opened 2018-02-08
Fix to correct postcss issues with % being converted to numeric value

dkasper-was-taken:
### Summary
Testing fix for postcss issues using mediaqueries + calc + var.

@cerner/terra


dkasper-was-taken:
Looks like it works!

mjhenkes:
Should we add a wdio test for this?

packages/terra-slide-panel/src/SlidePanel.scss
```diff
@@ -70,7 +70,7 @@
     }
 
     [data-slide-panel-panel-size='large'] > .panel {
-      width: var(--terra-slide-panel-panel-width-large, 60%);
+      width: calc(100% * var(--terra-slide-panel-panel-width-large-scalar, 0.6));
```
benbcai: What is the reason for changing the themeable var from  `--terra-slide-panel-panel-width-large` to `--terra-slide-panel-panel-width-large-scalar`? Would this be non-passive? Should it also be updated in terra-site/themeable-variables.json.

bjankord: With this being a non-passive change, we can roll this out the major version bump of this project as it is uplifted to react to avoid double major version bumps (1 for react, 1 for theming variable change).

Issue #1231, opened 2018-02-09
Component:Terra Form Checkbox has default 28px margin


shriniketsarkar:
# Issue Description
The encasing DIV that gets added during rendering of the Checkbox component has an unexpected 28px right margin. 

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
The encasing DIV for the Checkbox component should only house the checkbox components and not have its own margin styles. 

## Current Behavior
It currently adds a 28px right margin which pushes other components further to the right. This is more evident when this component is used within a Table or a Terra TableCell as the column is extra wide despite of having its own padding.

## Steps to Reproduce

1. Add the component to a TableCell 
2. I used the following code : 
`<Table.Cell content={
       <Checkbox
          defaultChecked={this.props.readAccess.isReadChecked}
          id={idString + '-' + 'read-checkbox'}
          labelText='read'
          value={this.props.resourceName}
        />}
        />`
3. After its rendered you would be able to see the extra margin.
4. Output Screen Shots: 
Control: 
![checkbox_28_margin](https://user-images.githubusercontent.com/1454181/36049667-45e67810-0da9-11e8-9434-32d29c8b6b97.png)
Computed Values:
![checkbox_computedvalues](https://user-images.githubusercontent.com/1454181/36049666-45d293c2-0da9-11e8-8cd4-b1dc69625554.png)

## Environment
* Component Version: 1.2.0 :  http://engineering.cerner.com/terra-core/#/site/form-checkbox
* Browser Name and Version: Chrome Version 63.0.3239.132 (Official Build) (64-bit)
* Operating System and version (desktop or mobile): Mac 10.11.6 El Capitan

shriniketsarkar:
Note: I am using just the Terra Component by itself. Its being used without the Base.

yuderekyu:
The initial intention was to space out inline checkboxes according to the open source theme callouts - in hindsight it's best if we leave spacing up to consumers, seeing as there are unintended consequences depending on where the checkbox is placed within an element.

PR #1232, opened 2018-02-11
ISSUE-1231: Reduces erroneous unnecessary default 28px right margin

dylan-sftwr:
### Summary
Removes erroneous default `margin-right: 2rem` style.

Fixes: #1231.

### Deployment URL
https://terra-core-deployed-pr-1232.herokuapp.com/static/#/site/form-checkbox

@cerner/terra


packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -7,7 +7,7 @@
     font-family: inherit;
     margin-bottom: var(--terra-form-checkbox-container-margin-bottom, 0);
     margin-left: var(--terra-form-checkbox-container-margin-left, 0);
-    margin-right: var(--terra-form-checkbox-container-margin-right, 2rem);
+    margin-right: var(--terra-form-checkbox-container-margin-right, 0rem);
```
emilyrohrbough: This should be 0

dylan-sftwr: Good catch, that's what I get for changing things too quickly ha.

packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -7,7 +7,7 @@
     font-family: inherit;
     margin-bottom: var(--terra-form-checkbox-container-margin-bottom, 0);
     margin-left: var(--terra-form-checkbox-container-margin-left, 0);
-    margin-right: var(--terra-form-checkbox-container-margin-right, 2rem);
+    margin-right: var(--terra-form-checkbox-container-margin-right, 0);
```
StephenEsser: Removing this will result in inline checkboxes having no padding / margins between them by default. It would be worth discussing with UX how this should be handled. Whether that is by the user, by applying margins only to inline checkboxes, by a CheckboxGroup component, or some other way. 

image

yuderekyu:
Per offline discussion with @neilpfeiffer - lets remove the multiple checkbox examples (both inline and default checked) from site. While we can group these standalone control components together (as done in the examples), there's enough differences in grouping/spacing that merit new group components. Something like a `fieldset` or `checkbox group` and `radio group`.

packages/terra-form-checkbox/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ Unreleased
 * Updated to use React 16
 * Update theming values to initial browser default for various properties
 * Merged border-color, border-width, and border-style into short-hand syntax
+* Removed erroneous default margin-right value of 2rem
```
StephenEsser: You might update this description to say `reduced` instead of `removed` or something similar

dylan-sftwr: Good call! 23e96eab

neilpfeiffer:
Tested: PR #1232
Resolves: Issue #1231 

\
**Reviewed:**
- enhancement for smaller spacing

\
\
Verified @neilpfeiffer, Ready for Release.

**Pass** |
-----------|
UX Review|

Issue #111, opened 2017-03-10
Move id into component


Matt-Butler:
In all of our examples, we have the same two components.
```
    <Markdown id="readme" src={ReadMe} />
    <PropsTable id="props" src={IconSrc} />
```

We should move the id into the component itself.

Matt-Butler:
There is the possibility we have more than one markdown / props table on a page. Closing.

PR #1235, opened 2018-02-12
Implementation of terra-form-input component

### LINKED ISSUE #1118: Uplift Form Input Styles
# Issue Description
Uplift styles for the Form Input component.

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

### LINKED ISSUE #1207: Add terra-form-input component
# Technical Design: `terra-form-input`

* Originally inspired by [Ben Cai's preliminary tech-design](https://github.com/cerner/terra-core/issues/1118).
* As this is an uplift, it's our goal to allow for all original functionality allowable by terra-form's input element. As such, this tech design is not meant to be an all-inclusive design over this new component's implementation. Rather, a set of new/notable features not currently implemented by the terra-form input element that I believe are worth calling out.

## Summary:
Create a new package called `terra-form-input` to replace the Input component currently in `terra-form`.

## Accessibility
Keyboard focus will be available for this form input.

### Old
See existing component's props table [here](http://engineering.cerner.com/terra-core/#/site/form/input).
### New
| Props       | Type  | Default   | Description                                               |
| ----------- | ----- | --------- | --------------------------------------------------------- |
| `disabled`  | bool  | `false`   | Whether the input is disabled                             |
| `isInvalid` | bool  | `false`   | Whether the input is invalid                              |
| `onBlur`    | func  | undefined | Function callback to trigger when this input loses focus. |
| `onFocus`   | func  | undefined | Function callback to trigger when this input gains focus. |

## Themeable Styles:
* Background color
* Background image
* Border color
* Border radius
* Border style
* Border width
* Box Shadow
* Focus animation
	* Direction
	* Duration
* Font color
* Font size
* Font weight
* Opacity
* Padding

dylan-sftwr:
### Summary
Implementation and style uplift of the input component.

### Related
* Fixes #1118.
* Tech-Design: #1207 (Closes #1207)

### Deployment Link
* https://terra-core-deployed-pr-1235.herokuapp.com/static/#/site/form-input


packages/terra-form-input/package.json
```diff
@@ -0,0 +1,48 @@
+{
+  "name": "terra-form-input",
+  "main": "lib/FormInput.js",
+  "version": "0.0.0",
+  "description": "{insert description}",
```
JakeLaCombe: Insert description here.

packages/terra-form-input/package.json
```diff
@@ -0,0 +1,48 @@
...
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
+    "terra-props-table": "^1.10.1"
+  },
+  "peerDependencies": {
+    "react": "^15.4.2",
```
JakeLaCombe: Make this react 16.2.0, same with react-dom

packages/terra-form-input/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "react-dom": "^15.4.2",
+    "terra-base": "^2.0.0"
```
JakeLaCombe: we will need to update terra-base to the latest when it's released soon

packages/terra-form-input/src/FormInput.jsx
```diff
@@ -0,0 +1,118 @@
...
+      onChange,
+      onFocus,
+      name,
+      required,
+      value,
+      ...customProps
+    } = this.props;
+
+    const attributes = Object.assign({}, customProps);
+    const formInputClassNames = cx([
```
JakeLaCombe: After you set formInputClassNames, delete the className key from the attributes array. Otherwise, you risk passing in an invalid class to the input.

packages/terra-form-input/src/FormInput.jsx
```diff
@@ -0,0 +1,118 @@
...
+   */
+  onFocus: PropTypes.func,
+  /**
+   * Content to be displayed as the name.
+   */
+  name: PropTypes.string,
+  /**
+   * Whether the input is required.
+   */
+  required: PropTypes.bool,
```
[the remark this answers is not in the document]
dylan-sftwr: I've named it as `required` to keep consistent with other `terra-form-X` components (ex:
 [terra-form-textarea](https://github.com/cerner/terra-core/blob/master/packages/terra-form-textarea/src/Textarea.jsx#L58)). Doing so also is more consistent with the standard [html input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input).

packages/terra-form-input/src/FormInput.scss
```diff
@@ -0,0 +1,103 @@
+:local {
+  .form-input {
+    background-color: var(--terra-form-input-background-color, #fff);
+    background-image: var(--terra-form-input-focus-animation-background, none);
+    border-color: var(--terra-form-input-border-color, #4a4a4a);
```
JakeLaCombe: Is this an issue here? https://github.com/cerner/terra-core/issues/1199

packages/terra-form-input/package.json
dylan-sftwr: Done 6cb6425.

packages/terra-form-input/tests/jest/FormInput.test.jsx
```diff
@@ -0,0 +1,48 @@
...
+  // Snapshot Tests
+  it('should render a default component', () => {
+    const wrapper = shallow(defaultRender);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  // Prop Tests
+  it('should use the default value when no value is given', () => {
+    const wrapper = shallow(defaultRender);
+    expect(wrapper.find('.form-input').text()).toEqual('');
```
JakeLaCombe: Do we need this test? I feel that your snapshot tests are sufficient enough

packages/terra-form-input/package.json
dylan-sftwr: I'll add it to my todo list.

packages/terra-form-input/tests/jest/FormInput.test.jsx
```diff
@@ -0,0 +1,48 @@
...
+  it('should set the forminput to invalid when isInvalid is passed into the component', () => {
+    const forminput = <FormInput isInvalid />;
+    const wrapper = render(forminput);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  // Structure Tests
+  it('should have the class form-input', () => {
+    const wrapper = shallow(defaultRender);
+    expect(wrapper.prop('className')).toContain('form-input');
```
JakeLaCombe: Snapshot should have this test covered as well.

packages/terra-form-input/src/FormInput.jsx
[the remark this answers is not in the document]
samvaity: If I understand this right, this prop is to mark if this component is `mandatory`, right?

If so, can we change the comment description to `Whether or not the field is required.`

packages/terra-form-input/src/FormInput.jsx
dylan-sftwr: Doesn't [this](https://github.com/cerner/terra-core/blob/Form-Input-Element/packages/terra-form-input/src/FormInput.jsx#L109) ensure that the `className` prop always wins when prop collisions between any `className` that may be found in the `attributes` array that's spread on line 108?

packages/terra-form-input/tests/jest/FormInput.test.jsx
dylan-sftwr: I think this was one of the auto-generated tests, removed it. 297550fd

packages/terra-form-input/tests/jest/FormInput.test.jsx
dylan-sftwr: I think this was one of the auto-generated tests, removed it. 297550f

packages/terra-form-input/src/FormInput.jsx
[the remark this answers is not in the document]
dylan-sftwr: I'll change the word _input_ to _field_, but the _or not_ part of that statement is redundant. I'd originally written this to align with the [`terra-form-textarea`'s](https://github.com/cerner/terra-core/blob/master/packages/terra-form-textarea/src/Textarea.jsx#L56) comment, but given that this **is** the **input** component, I'd say _field_ is more clear 👍. 0603da15

packages/terra-form-input/src/FormInput.jsx
JakeLaCombe: I just now saw that. Given that, I think that risk isn't present here.

packages/terra-form-input/src/FormInput.scss
```diff
@@ -0,0 +1,103 @@
...
+    &[required] {
+      box-shadow: unset;
+    }
+
+    &[disabled] {
+      background-color: var(--terra-form-input-disabled-background-color, rgba(0, 0, 0, 0.05));
+      border-color: var(--terra-form-input-disabled-border-color, #4a4a4a);
+      border-style: var(--terra-form-input-disabled-border-style, solid);
+      border-width: var(--terra-form-input-disabled-border-width, 1px);
+      color: var(--terra-form-input-disabled-color, #000);
```
dm123455: needs to be updated.

packages/terra-form-input/src/FormInput.scss
```diff
@@ -0,0 +1,103 @@
...
+      border-style: var(--terra-form-input-error-hover-border-style, solid);
+      border-width: var(--terra-form-input-error-hover-border-width, 1px);
+    }
+
+    &[disabled] {
+      background-color: var(--terra-form-input-error-disabled-background-color, rgba(0, 0, 0, 0.05));
+      border-color: var(--terra-form-input-error-disabled-border-color, #ff0b0f);
+      border-style: var(--terra-form-input-error-disabled-border-style, solid);
+      border-width: var(--terra-form-input-error-disabled-border-width, 1px);
+      color: var(--terra-form-input-error-disabled-color, #000);
```
dm123455: needs to be updated.

packages/terra-form-input/src/FormInput.jsx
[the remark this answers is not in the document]
JakeLaCombe: Actually, I'll have to request to leave it as input in the description. We have another component terra-form-field, and I fear that naming this to field could cause confusion about what this prop sets.

packages/terra-form-input/src/FormInput.scss
```diff
@@ -0,0 +1,103 @@
...
+      }
+
+      [dir=rtl] & {
+        box-shadow: var(--terra-form-input-focus-keyboard-box-shadow-rtl, 0 0 8px #a6d9f4);
+      }
+    }
+  }
+
+  .form-error {
+    background-image: var(--terra-form-input-error-focus-animation-background, none);
```
dm123455: can we double check what the color of the text needs to be in this case as we dont have callouts for this scenario?Thanks

packages/terra-form-input/src/FormInput.jsx
dylan-sftwr: Cool :) it's always good to double check!

packages/terra-form-input/src/FormInput.jsx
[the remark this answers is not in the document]
dylan-sftwr: Switched back here: d5d1355c

packages/terra-site/src/examples/form-input/DefaultErrorExample.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+import FormInput from 'terra-form-input/lib/FormInput';
+
+class DefaultErrorExample extends React.Component {
```
JakeLaCombe: Is the point of the example to show what an inValid input looks like? If so, I'd just make it an uncontrolled input.

packages/terra-site/src/examples/form-input/DefaultExample.jsx
```diff
@@ -0,0 +1,32 @@
...
+  handleChange(event) {
+    this.setState({ input: event.target.value });
+  }
+
+  render() {
+    return(
+      <div>
+        <h4>Default</h4>
+        <FormInput
+          defaultValue="Default Input"
```
JakeLaCombe: I don't think this kind of input is correct react syntax actually. For an uncontrolled input, you just populate defaultValue while leaving onChange undefined. I'd make this one an uncontrolled input as well. You can read more about it [here](https://reactjs.org/docs/uncontrolled-components.html)

packages/terra-site/src/TestLinks.jsx
```diff
@@ -24,6 +24,7 @@ const TestLinks = () => (
     <li><Link to="/tests/form-checkbox-tests">Form Checkbox Tests</Link></li>
     <li><Link to="/tests/form-tests">Form Tests</Link></li>
     <li><Link to="/tests/form-field-tests">Form Field Tests</Link></li>
+      <li><Link to="/tests/form-input-tests">Form Input Tests</Link></li>
```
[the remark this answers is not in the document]
dylan-sftwr: Removed. 4ef88ad6

packages/terra-form-input/src/FormInput.scss
dylan-sftwr: per Chris:

> I don't think we want the text color to change.

packages/terra-form-input/docs/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+```jsx
+import React from 'react';
+import Input from 'terra-form-input';
+
+<Input
+  name="description"
+  value="I'm a software engineer working at Cerner"
+  required
+/>```
```
StephenEsser: Previewing this markdown file shows a couple syntax issues

packages/terra-form-input/src/FormInput.jsx
```diff
@@ -0,0 +1,118 @@
+import React from 'react';
```
StephenEsser: Nit: After reading the read me I would have expected this file to be named `Input.jsx`
```
import Input from 'terra-form-input';

<Input
  name="description"
  value="I'm a software engineer working at Cerner"
  required
/>
``` 
With the ability to alias it may not make too much of a difference, do you have a preference either way?

packages/terra-form-input/src/FormInput.jsx
```diff
@@ -0,0 +1,118 @@
...
+  defaultValue: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+  ]),
+  /**
+   * Whether the input is disabled.
+   */
+  disabled: PropTypes.bool,
+  /**
+   * Whether the input is invalid
```
StephenEsser: Nit: Missing punctuation.

packages/terra-form-input/src/FormInput.jsx
```diff
@@ -0,0 +1,118 @@
...
+  isInvalid: false,
+  onBlur: undefined,
+  onChange: undefined,
+  onFocus: undefined,
+  name: null,
+  required: false,
+  value: undefined,
+};
+
+class FormInput extends React.Component {
```
StephenEsser: This took me a little while to figure out. Declaring this as a class should have thrown the following [lint error](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prefer-stateless-function.md).

There is an exception to this rule when a class variable is declared. In this case the ref callback. 
```
ref={(input) => { this.textInput = input; }}
```

Because `this.textInput` is never used it should probably be removed which will result in the lint error mentioned above being flagged. 

Keeping this as a class allows users to access the component `ref`. Stateless functions do not allow the ref callback. It may be worth gathering opinions on whether the value of the ref callback is enough to disable this rule or if each stateless component should provide an alternative prop reference to the ref or neither.

packages/terra-form-input/src/FormInput.jsx
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    return (
+      <input
+        aria-label="input label"
+        name={name}
+        onBlur={onBlur}
+        onChange={onChange}
+        onFocus={onFocus}
+        ref={(input) => { this.textInput = input; }}
```
StephenEsser: This ref should probably be removed if it's not referenced anywhere.

packages/terra-form-input/src/FormInput.jsx
```diff
@@ -0,0 +1,118 @@
...
+        disabled={disabled}
+        required={required}
+        {...attributes}
```
StephenEsser: We don't appear to be consistent throughout all of the components but I believe the current standard for additional attributes is to spread them first so that explicitly declared prop values always win. 
```
      <input
        {...attributes}
        aria-label="input label"
        name={name}
        onBlur={onBlur}
        onChange={onChange}
        onFocus={onFocus}
        disabled={disabled}
        required={required}
        className={formInputClassNames}
      />
```

packages/terra-form-input/src/FormInput.jsx
```diff
@@ -0,0 +1,118 @@
...
+
+    if (value !== undefined) {
+      attributes.value = value;
+    } else {
+      attributes.defaultValue = defaultValue;
+    }
+
+    return (
+      <input
+        aria-label="input label"
```
StephenEsser: I'm not as familiar with screen reader accessibility but I don't think this can be a static string.

packages/terra-form-input/docs/README.md
bjankord: The closing tick marks (```) should be moved to a new line

<pre><code>
/>
```
</code></pre>

File: packages/terra-form-input/package.json [hunk 22]
```diff
@@ -0,0 +1,48 @@
...
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
+    "terra-props-table": "^1.10.1"
+  },
+  "peerDependencies": {
+    "react": "^16.2.0",
+    "react-dom": "^16.2.0",
+    "terra-base": "^2.0.0"
```
bjankord: Bump this to `^3.0.0`

packages/terra-form-input/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "UI"
+  ],
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
+    "terra-props-table": "^1.10.1"
```
bjankord: Bump this to `^2.0.0`

packages/terra-form-input/package.json
```diff
@@ -0,0 +1,48 @@
...
+  },
+  "peerDependencies": {
+    "react": "^16.2.0",
+    "react-dom": "^16.2.0",
+    "terra-base": "^2.0.0"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "prop-types": "^15.5.8",
+    "terra-base": "^2.0.0"
```
bjankord: Bump this to `^3.0.0`

packages/terra-form-input/src/FormInput.scss
bjankord: Yeah, we'll need to merge the themable border styles into 1 shorthand border property for this @dylan-sftwr

packages/terra-form-input/src/FormInput.jsx
bjankord: Let's leave this ref. While it may not be used in this file, I could see it being useful when this component is composed into other components like time input, date picker, etc. It also lines up with how we've coded the textarea component.

packages/terra-form-input/src/FormInput.jsx
bjankord: Let's rename this file to `Input.jsx` With all of the other new terra-form-component-name packages, the JSX file is named without the `Form` prefix. e.g. Radio.jsx, Select.jsx, Field.jsx

packages/terra-form-input/src/FormInput.jsx
```diff
@@ -0,0 +1,118 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import 'terra-base/lib/baseStyles';
+import styles from './FormInput.scss';
```
bjankord: With renaming the JSX file to Input.jsx, let's also rename the scss file to `Input.scss`

packages/terra-form-input/src/FormInput.jsx
bjankord: I'm fine making this a class based component to use the ref callback. While ref is not used directly in this file, I can see it being useful in more composed components we make using the input.

packages/terra-form-input/src/FormInput.scss
```diff
@@ -0,0 +1,103 @@
...
+      border-width: var(--terra-form-input-hover-border-width, 1px);
+    }
+
+    &[required] {
+      box-shadow: unset;
+    }
+
+    &[disabled] {
+      background-color: var(--terra-form-input-disabled-background-color, rgba(0, 0, 0, 0.05));
+      border-color: var(--terra-form-input-disabled-border-color, #dedfe0);
```
bjankord: we'll need to merge the themable border styles into 1 shorthand border property for this @dylan-sftwr. It works around this bug: https://github.com/cerner/terra-core/issues/1199

packages/terra-form-input/src/FormInput.scss
```diff
@@ -0,0 +1,103 @@
...
+      color: var(--terra-form-input-disabled-color, rgba(0, 0, 0, 0.25));
+      cursor: default;
+      opacity: var(--terra-form-input-disabled-opacity, 1);
+    }
+
+    &:focus {
+      background-color: var(--terra-form-input-focus-background-color, #fff);
+      border-color: var(--terra-form-input-focus-border-color, #4a4a4a);
+      border-radius: var(--terra-form-input-focus-border-radius, 0.2rem);
+      border-style: var(--terra-form-input-focus-border-style, solid);
```
bjankord: we'll need to merge the themable border styles into 1 shorthand border property for this @dylan-sftwr. It works around this bug: #1199
border-radius can be left stand-alone.

packages/terra-form-input/src/FormInput.scss
```diff
@@ -0,0 +1,103 @@
...
+        box-shadow: var(--terra-form-input-focus-keyboard-box-shadow-rtl, 0 0 8px #a6d9f4);
+      }
+    }
+  }
+
+  .form-error {
+    background-image: var(--terra-form-input-error-focus-animation-background, none);
+    border-color: var(--terra-form-input-error-border-color, #ff0b0f);
+    border-radius: var(--terra-form-input-error-border-radius, 0.2rem);
+    border-style: var(--terra-form-input-error-border-style, solid);
```
bjankord: we'll need to merge the themable border styles into 1 shorthand border property for this @dylan-sftwr. It works around this bug: #1199
border-radius can be left stand-alone.

packages/terra-form-input/src/FormInput.scss
```diff
@@ -0,0 +1,103 @@
...
+    border-width: var(--terra-form-input-error-border-width, 1px);
+    opacity: var(--terra-form-input-error-opacity, 1);
+
+    &:hover {
+      border-color: var(--terra-form-input-error-hover-border-color, #ff0b0f);
```
bjankord: we'll need to merge the themable border styles into 1 shorthand border property for this @dylan-sftwr. It works around this bug: #1199

packages/terra-form-input/src/FormInput.scss
```diff
@@ -0,0 +1,103 @@
...
+      border-style: var(--terra-form-input-error-hover-border-style, solid);
+      border-width: var(--terra-form-input-error-hover-border-width, 1px);
+    }
+
+    &[disabled] {
+      background-color: var(--terra-form-input-error-disabled-background-color, rgba(0, 0, 0, 0.05));
+      border-color: var(--terra-form-input-error-disabled-border-color, #ff0b0f);
```
bjankord: we'll need to merge the themable border styles into 1 shorthand border property for this @dylan-sftwr. It works around this bug: #1199

packages/terra-form-input/src/FormInput.scss
```diff
@@ -0,0 +1,103 @@
...
+      border-style: var(--terra-form-input-error-disabled-border-style, solid);
+      border-width: var(--terra-form-input-error-disabled-border-width, 1px);
+      color: var(--terra-form-input-error-disabled-color, rgba(0, 0, 0, 0.25));
+      opacity: var(--terra-form-input-error-disabled-opacity, 1);
+    }
+
+    &:focus {
+      background-color: var(--terra-form-input-error-focus-background-color, #fff);
+      border-color: var(--terra-form-input-error-focus-border-color, #ff0b0f);
```
bjankord: we'll need to merge the themable border styles into 1 shorthand border property for this @dylan-sftwr. It works around this bug: #1199
border-radius can be left stand-alone.

packages/terra-form-input/src/FormInput.scss
```diff
@@ -0,0 +1,103 @@
+:local {
```
bjankord: Left a comment on FormInput.jsx, lets rename this file to `Input.scss`

packages/terra-site/src/examples/form-input/DefaultExample.jsx
dylan-sftwr: I don't think the fact that this doesn't explicitly map the state back into the component via the `value` prop explicitly makes it invalid _React_ syntax. Admittedly, it is **not** a valid _controlled input_ example. As I want to keep the presentational part of this below (`<p>Input Provided: {this.state.input}</p>`), I'm going to make a clearer distinction between _controlled_ and _uncontrolled_ inputs by providing an example of both. 34572384

packages/terra-site/src/examples/form-input/DefaultErrorExample.jsx
dylan-sftwr: Addressed in 3457238

packages/terra-form-input/src/FormInput.jsx
StephenEsser: If we keep this component as a class we should be able to remove the hardcoded ref. After removing this line a consumer will be able to do the following.
```
<Input ref={this.mockFunction} />
```

packages/terra-form-input/package.json
dylan-sftwr: Done. c4abd525098d61a3b1e9238d8ca6a3382473a87b

packages/terra-form-input/src/FormInput.scss
dylan-sftwr: Done. 2b6d5b6031c0709535fe0bb2d78a16b7a335adc7

packages/terra-form-input/src/FormInput.scss
dylan-sftwr: Done. 2b6d5b6031c0709535fe0bb2d78a16b7a335adc7

packages/terra-form-input/docs/README.md
dylan-sftwr: Good catch. Fixed 629fc064

packages/terra-form-input/src/FormInput.jsx
dylan-sftwr: I left these as they're what the [generator-terra-module](https://github.com/cerner/generator-terra-module) output automatically. Updated them in 0fa82cf1

packages/terra-form-input/src/FormInput.jsx
dylan-sftwr: Updated bc4d5ad282200bb149837a71f72300a65584e35d

packages/terra-form-input/src/FormInput.jsx
dylan-sftwr: I'd also prefer to leave this as a class for the reasons Brett mentions above.

packages/terra-form-input/src/FormInput.jsx
dylan-sftwr: Removed. 05932d987a54c84ffd256b0c41d8925a168b67a6

packages/terra-form-input/src/FormInput.jsx
dylan-sftwr: I agree I had been following the example in `terra-form-textarea`. If this is the standard we want to follow, we should probably log an issue to update it [here](https://github.com/cerner/terra-core/blob/Form-Input-Element/packages/terra-form-textarea/src/Textarea.jsx#L164).

packages/terra-form-input/src/FormInput.jsx
dylan-sftwr: Dropped. 7ca869c070df993e6399cc6537b1682a114b19f7

packages/terra-form-input/package.json
dylan-sftwr: Done b7fb7ce2ada572d12750eb100d869c1e3cc3137b

packages/terra-form-input/package.json
dylan-sftwr: Done b7fb7ce2ada572d12750eb100d869c1e3cc3137b

packages/terra-form-input/package.json
dylan-sftwr: Done b7fb7ce2ada572d12750eb100d869c1e3cc3137b

packages/terra-form-input/src/FormInput.jsx
dylan-sftwr: Done 0fa82cf1ef4966c97404188b2b229c8f8385fb2b

packages/terra-form-input/src/FormInput.scss
dylan-sftwr: Done. 5e9eba487a5de8d996a8f96c29aa6381196df310

packages/terra-form-input/src/FormInput.scss
dylan-sftwr: Done. 5e9eba487a5de8d996a8f96c29aa6381196df310

packages/terra-form-input/src/FormInput.scss
dylan-sftwr: Done. 5e9eba487a5de8d996a8f96c29aa6381196df310

packages/terra-form-input/src/FormInput.scss
dylan-sftwr: Done. 5e9eba487a5de8d996a8f96c29aa6381196df310

packages/terra-form-input/src/FormInput.jsx
dylan-sftwr: Renamed 0fa82cf1ef4966c97404188b2b229c8f8385fb2b

packages/terra-form-input/src/FormInput.jsx
dylan-sftwr: Done. cd36335322ce48d3beb9b5e3edce6c8a899ae956

packages/terra-form-input/src/FormInput.scss
dylan-sftwr: Done 5e9eba487a5de8d996a8f96c29aa6381196df310

packages/terra-form-input/src/FormInput.scss
dylan-sftwr: Done 5e9eba487a5de8d996a8f96c29aa6381196df310

packages/terra-form-input/src/FormInput.scss
dylan-sftwr: Done 5e9eba487a5de8d996a8f96c29aa6381196df310

packages/terra-form-input/src/FormInput.scss
dylan-sftwr: Done 0fa82cf1ef4966c97404188b2b229c8f8385fb2b

packages/terra-form-input/package.json
dylan-sftwr: Done b7fb7ce2ada572d12750eb100d869c1e3cc3137b

dm123455:
+1

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,92 @@
+:local {
+  .form-input {
+    background-color: var(--terra-form-input-background-color, #fff);
+    background-image: var(--terra-form-input-focus-animation-background, none);
+    background-size: var(--terra-form-input-focus-background-start, 0 0);
```
saedar: This is lint. `background-size` should come after `background-repeat`.

dylan-sftwr: Fixed, good catch. ac2c322a

saedar:
I'm seeing four failing wdio tests.

dylan-sftwr:
Spoke offline to @saedar about failing wdio tests. I wasn't able to reproduce these failures, but if someone else is, please let me know.

dylan-sftwr:
@bjankord , can I get your +1 on this?

packages/terra-form-input/src/Input.jsx
```diff
@@ -0,0 +1,117 @@
...
+  isInvalid: false,
+  onBlur: undefined,
+  onChange: undefined,
+  onFocus: undefined,
+  name: null,
+  required: false,
+  value: undefined,
+};
+
+class FormInput extends React.Component {
```
bjankord: nit: Reword `FormInput` to `Input` to match how we are naming other terra form components

packages/terra-form-input/src/Input.jsx
```diff
@@ -0,0 +1,117 @@
...
+        onFocus={onFocus}
+        disabled={disabled}
+        required={required}
+        className={formInputClassNames}
+      />
+    );
+  }
+}
+
+FormInput.propTypes = propTypes;
```
bjankord: nit: Reword `FormInput` to `Input` to match how we are naming other terra form components

packages/terra-form-input/tests/jest/FormInput.test.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import FormInput from '../../src/Input';
```
bjankord: This file will need changes with updates in `packages/terra-form-input/src/Input.jsx` for exporting the component as `Input` instead of `FormInput`

packages/terra-form-input/tests/nightwatch/FormInputPropPermutations.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import FormInput from '../../lib/Input';
```
bjankord: This file will need changes with updates in packages/terra-form-input/src/Input.jsx for exporting the component as Input instead of FormInput

packages/terra-form-input/tests/wdio/form-input-spec.js
```diff
@@ -0,0 +1,132 @@
...
+      Terra.should.matchScreenshot();
+      Terra.should.themeEachCustomProperty({
+        '--terra-form-input-hover-border': '20rem dashed blue',
+      });
+    });
+
+    describe('Enabled Keyboard-Focused', () => {
+      beforeEach(() => {
+        browser.url('/#tests/form-input-tests/default');
+        browser.keys('Tab');
```
bjankord: Rather than tab to focus on the input, we should emulate a focus state by creating an `is-focused` class that applies the same styles as `:focus`. Then we need to create a test page for focus and set the `is-focused` class on the input. This is to get around the issue @saedar noticed where these tests have a potential to fail based on if the caret to indicate where the cursor is, is visible or not as it blinks.

This would allow us to capture focus styles without the blinking caret in the input.

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,92 @@
+:local {
+  .form-input {
+    background-color: var(--terra-form-input-background-color, #fff);
+    background-image: var(--terra-form-input-focus-animation-background, none);
```
bjankord: For styles which are needing for theming, but are not needed in the default theme, you can remove the fallback value. For example, you can update `background-image: var(--terra-form-input-focus-animation-background, none);` to `background-image: var(--terra-form-input-focus-animation-background);`.

It looks like the following properties can drop the fallback value:
* background-image
* background-size
* box-shadow
* color (possibly, need to test)
* opacity
* transition-duration (possibly, need to test)
* transition-timing-function (possibly, need to test)

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,92 @@
...
+    transition-property: background-size;
+    transition-timing-function: var(--terra-form-input-transition-timing-function, ease);
+    width: 100%;
+
+    &:hover {
+      border: var(--terra-form-input-hover-border, 1px solid #4a4a4a);
+    }
+
+    &[required] {
+      box-shadow: unset;
```
bjankord: For browser compatibility, it may be better to set this to `none`. Unset doesn't work in IE browsers or most of the Android browsers. https://caniuse.com/#search=unset

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    &[disabled] {
+      background-color: var(--terra-form-input-disabled-background-color, rgba(0, 0, 0, 0.05));
```
bjankord: Need to check with @neilpfeiffer if we need styling of background, border, and color in the disabled state or if we can just adjust the opacity.

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,92 @@
...
+    border: var(--terra-form-input-error-border, 1px solid #ff0b0f);
+    border-radius: var(--terra-form-input-error-border-radius, 0.2rem);
+    opacity: var(--terra-form-input-error-opacity, 1);
+
+    &:hover {
+      border: var(--terra-form-input-error-hover-border, 1px solid #ff0b0f);
+    }
+
+    &[disabled] {
+      background-color: var(--terra-form-input-error-disabled-background-color, rgba(0, 0, 0, 0.05));
```
bjankord: Need to check with @neilpfeiffer if we need styling of background, border, and color in the disabled state or if we can just adjust the opacity.

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,92 @@
...
+    background-repeat: no-repeat;
+    background-size: var(--terra-form-input-focus-background-start, 0 0);
+    border: var(--terra-form-input-border, 1px solid #4a4a4a);
+    border-radius: var(--terra-form-input-border-radius, 0.2rem);
+    box-shadow: var(--terra-form-input-box-shadow, none);
+    color: var(--terra-form-input-color, #000);
+    font-family: inherit;
+    font-size: var(--terra-form-input-font-size, 1.143rem);
+    line-height: var(--terra-form-input-line-height, 1.143);
+    opacity: var(--terra-form-input-opacity, 1);
```
bjankord: Does opacity of the input change in the consumer theme? Check with @neilpfeiffer on this, I'm thinking if there is an opacity change, it happens at background or color style with rgba instead of changing opacity.

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,92 @@
...
+      cursor: default;
+      opacity: var(--terra-form-input-disabled-opacity, 1);
+    }
+
+    &:focus {
+      background-color: var(--terra-form-input-focus-background-color, #fff);
+      background-size: var(--terra-form-input-focus-keyboard-animation-size, 0 0);
+      border: var(--terra-form-input-focus-border, 1px solid #4a4a4a);
+      border-radius: var(--terra-form-input-focus-border-radius, 0.2rem);
+      opacity: var(--terra-form-input-focus-opacity, 1);
```
bjankord: Check with @neilpfeiffer on this, I'm thinking if there is an opacity change, it happens at background or color style with rgba instead of changing opacity.

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,92 @@
...
+        box-shadow: var(--terra-form-input-focus-keyboard-box-shadow-rtl, 0 0 8px #a6d9f4);
+      }
+    }
+  }
+
+  .form-error {
+    background-image: var(--terra-form-input-error-focus-animation-background, none);
+    border: var(--terra-form-input-error-border, 1px solid #ff0b0f);
+    border-radius: var(--terra-form-input-error-border-radius, 0.2rem);
+    opacity: var(--terra-form-input-error-opacity, 1);
```
bjankord: Check with @neilpfeiffer on this, I'm thinking if there is an opacity change, it happens at background or color style with rgba instead of changing opacity.

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,92 @@
...
+      color: var(--terra-form-input-error-disabled-color, rgba(0, 0, 0, 0.25));
+      opacity: var(--terra-form-input-error-disabled-opacity, 1);
+    }
+
+    &:focus {
+      background-color: var(--terra-form-input-error-focus-background-color, #fff);
+      background-size: var(--terra-form-input-error-focus-keyboard-animation-size, 0 0);
+      border: var(--terra-form-input-error-focus-border, 1px solid #ff0b0f);
+      border-radius: var(--terra-form-input-error-focus-border-radius, 0.2rem);
+      opacity: var(--terra-form-input-error-focus-opacity, 1);
```
bjankord: Check with @neilpfeiffer on this, I'm thinking if there is an opacity change, it happens at background or color style with rgba instead of changing opacity.

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,92 @@
...
+      color: var(--terra-form-input-disabled-color, rgba(0, 0, 0, 0.25));
+      cursor: default;
+      opacity: var(--terra-form-input-disabled-opacity, 1);
+    }
+
+    &:focus {
+      background-color: var(--terra-form-input-focus-background-color, #fff);
+      background-size: var(--terra-form-input-focus-keyboard-animation-size, 0 0);
+      border: var(--terra-form-input-focus-border, 1px solid #4a4a4a);
+      border-radius: var(--terra-form-input-focus-border-radius, 0.2rem);
```
bjankord: Does the border-radius change when the input is focused? It seems like this may not be needed.

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,92 @@
...
+      [dir=rtl] & {
+        box-shadow: var(--terra-form-input-focus-keyboard-box-shadow-rtl, 0 0 8px #a6d9f4);
+      }
+    }
+  }
+
+  .form-error {
+    background-image: var(--terra-form-input-error-focus-animation-background, none);
+    border: var(--terra-form-input-error-border, 1px solid #ff0b0f);
+    border-radius: var(--terra-form-input-error-border-radius, 0.2rem);
```
bjankord: Does the border-radius change when the input is focused? It seems like this may not be needed.

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,92 @@
...
+      border: var(--terra-form-input-error-disabled-border, 1px solid #ff0b0f);
+      color: var(--terra-form-input-error-disabled-color, rgba(0, 0, 0, 0.25));
+      opacity: var(--terra-form-input-error-disabled-opacity, 1);
+    }
+
+    &:focus {
+      background-color: var(--terra-form-input-error-focus-background-color, #fff);
+      background-size: var(--terra-form-input-error-focus-keyboard-animation-size, 0 0);
+      border: var(--terra-form-input-error-focus-border, 1px solid #ff0b0f);
+      border-radius: var(--terra-form-input-error-focus-border-radius, 0.2rem);
```
bjankord: Does the border-radius change when the input is focused in error state? It seems like this may not be needed.

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,92 @@
+:local {
+  .form-input {
```
bjankord: The input appears much taller than before, which is ok, but is also appears taller compared to buttons. Buttons are 30px tall and the input is 35px tall. We'll want the input and button to have the same computed height for the default theme. This helps visually balance the input when its next to a button in an inline form. I can pair with you and figure out what all changes are needed, usually its a mix of line-height and padding.

packages/terra-site/src/examples/form-input/controlled_examples/DefaultErrorExample.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+import FormInput from 'terra-form-input/lib/Input';
```
bjankord: Switch to `import Input from` after renaming component in `packages/terra-form-input/src/Input.jsx`

packages/terra-site/src/examples/form-input/controlled_examples/DefaultExample.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+
+import FormInput from 'terra-form-input/lib/Input';
```
bjankord: Switch to `import Input from` after renaming component in `packages/terra-form-input/src/Input.jsx`

packages/terra-site/src/examples/form-input/controlled_examples/DisabledErrorExample.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+
+import FormInput from 'terra-form-input/lib/Input';
```
bjankord: Switch to `import Input from` after renaming component in `packages/terra-form-input/src/Input.jsx`

packages/terra-site/src/examples/form-input/controlled_examples/DisabledExample.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+import FormInput from 'terra-form-input/lib/Input';
```
bjankord: Switch to `import Input from` after renaming component in `packages/terra-form-input/src/Input.jsx`

packages/terra-site/src/examples/form-input/uncontrolled_examples/DefaultErrorExample.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+import FormInput from 'terra-form-input/lib/Input';
```
bjankord: Switch to `import Input from` after renaming component in `packages/terra-form-input/src/Input.jsx`

packages/terra-site/src/examples/form-input/uncontrolled_examples/DefaultExample.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+import FormInput from 'terra-form-input/lib/Input';
```
bjankord: Switch to `import Input from 'terra-form-input';` after renaming component in `packages/terra-form-input/src/Input.jsx`

packages/terra-site/src/examples/form-input/uncontrolled_examples/DisabledErrorExample.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+import FormInput from 'terra-form-input/lib/Input';
```
bjankord: Switch to `import Input from 'terra-form-input';` after renaming component in `packages/terra-form-input/src/Input.jsx`

packages/terra-site/src/examples/form-input/uncontrolled_examples/DisabledExample.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+import FormInput from 'terra-form-input/lib/Input';
```
bjankord: Switch to `import Input from 'terra-form-input';` after renaming component in `packages/terra-form-input/src/Input.jsx`

packages/terra-form-input/tests/wdio/form-input-spec.js
dylan-sftwr: Done 8918a8ddbdc438465dd5bab14b39ee7723b03652

packages/terra-form-input/tests/wdio/form-input-spec.js
```diff
@@ -0,0 +1,132 @@
...
+      Terra.should.matchScreenshot();
+      Terra.should.themeEachCustomProperty({
+        '--terra-form-input-hover-border': '20rem dashed red',
+      });
+    });
+
+    describe('Disabled Keyboard-Focused', () => {
+      beforeEach(() => {
+        browser.url('/#tests/form-input-tests/disabled');
+        browser.keys('Tab');
```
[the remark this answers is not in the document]
dylan-sftwr: Yes

packages/terra-form-input/src/Input.scss
dylan-sftwr: This came directly from [here](https://github.com/cerner/terra-core/blob/master/packages/terra-form-textarea/src/Textarea.scss#L35). I've made the change here 840073f2a7864a6bc4ee4d9c10501b5903ba460f

packages/terra-form-input/src/Input.scss
dylan-sftwr: I can change this if Neil wants, I just kept this consistent with the `form-textarea` component.

packages/terra-form-input/src/Input.scss
dylan-sftwr: I can change this if Neil wants, I just kept this consistent with the `form-textarea` component.

packages/terra-form-input/src/Input.scss
dylan-sftwr: I can change this if Neil wants, I just kept this consistent with the `form-textarea` component.

packages/terra-form-input/src/Input.scss
dylan-sftwr: I can change this if Neil wants, I just kept this consistent with the `form-textarea` component.

packages/terra-form-input/src/Input.scss
dylan-sftwr: I can change this if Neil wants, I just kept this consistent with the `form-textarea` component.

packages/terra-form-input/src/Input.scss
dylan-sftwr: I don't think so, I just went with 
> Theme all the things

ha, I've removed it
cd537ccc3b2bdedcafbec4a1f73eb25f76cb7330

packages/terra-form-input/src/Input.scss
dylan-sftwr: I don't think so, I just went with 
> Theme all the things

ha, I've removed it
cd537ccc3b2bdedcafbec4a1f73eb25f76cb7330

packages/terra-form-input/src/Input.scss
dylan-sftwr: I don't think so, I just went with 
> Theme all the things

ha, I've removed it
cd537ccc3b2bdedcafbec4a1f73eb25f76cb7330

packages/terra-form-input/src/Input.scss
dylan-sftwr: I set something up for Monday

packages/terra-form-input/src/Input.jsx
```diff
@@ -0,0 +1,117 @@
+/* eslint-disable react/prefer-stateless-function */
```
emilyrohrbough: What do we need this?
```
/* eslint-disable react/prefer-stateless-function */
```

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,89 @@
...
+    background-color: var(--terra-form-input-background-color, #fff);
```
emilyrohrbough: Should this value we white or transparent? @neilpfeiffer

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,89 @@
...
+    font-size: var(--terra-form-input-font-size, 1.143rem);
+    line-height: var(--terra-form-input-line-height, 1.143);
+    opacity: var(--terra-form-input-opacity, 1);
+    padding-bottom: var(--terra-form-input-padding-bottom, 0.571rem);
+    padding-left: var(--terra-form-input-padding-left, 0.357rem);
+    padding-right: var(--terra-form-input-padding-right, 0.357rem);
+    padding-top: var(--terra-form-input-padding-top, 0.571rem);
+    transition-duration: var(--terra-form-input-transition-duration, 0s);
+    transition-property: background-size;
+    transition-timing-function: var(--terra-form-input-transition-timing-function, ease);
```
emilyrohrbough: Do we need to define this if the transition duration is 0?

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,89 @@
...
+
+    &[disabled] {
+      background-color: var(--terra-form-input-error-disabled-background-color, rgba(0, 0, 0, 0.05));
+      border: var(--terra-form-input-error-disabled-border, 1px solid #ff0b0f);
+      color: var(--terra-form-input-error-disabled-color, rgba(0, 0, 0, 0.25));
+      opacity: var(--terra-form-input-error-disabled-opacity, 1);
+    }
+
+    &:focus {
+      background-color: var(--terra-form-input-error-focus-background-color, #fff);
```
emilyrohrbough: Again, should this be white or transparent?

packages/terra-form-input/tests/jest/Input.test.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import Input from '../../src/Input';
+
+describe('Input', () => {
+  const defaultRender = <Input />;
```
emilyrohrbough: Can we add snapshots for `required` as well.

packages/terra-form-input/src/Input.jsx
dylan-sftwr: Without it, we have linting errors. See [this](https://github.com/cerner/terra-core/pull/1235#discussion_r167760416) comment thread for more details.

packages/terra-form-input/tests/nightwatch/InputPropPermutations.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+import Input from '../../lib/Input';
+
+const DefaultInput = () => <Input aria-label="default form input" id="form-input-default" />;
+const InvalidInput = () => <Input aria-label="invalid form input" isInvalid id="form-input-invalid" />;
+const DisabledInput = () => <Input aria-label="disabled form input" disabled id="form-input-disabled" />;
```
emilyrohrbough: Add an example for `required`

packages/terra-form-input/tests/nightwatch/InputTestRoutes.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+// Test Cases
+import {
+  DefaultInput,
+  InvalidInput,
+  DisabledInput,
+} from './InputPropPermutations';
+
+const routes = (
+  <div>
```
emilyrohrbough: We need to add links to test for `onChange`, `onBlue`, `onFocus`, and `required` props

packages/terra-form-input/src/Input.scss
dylan-sftwr: My callouts had this listed as white. It's also listed as white [here](https://github.com/cerner/terra-core/blob/master/packages/terra-form-textarea/src/Textarea.scss#L3) which I was told was the example I should be following.

packages/terra-form-input/src/Input.scss
dylan-sftwr: My callouts had this listed as white. It's also listed as white [here](https://github.com/cerner/terra-core/blob/master/packages/terra-form-textarea/src/Textarea.scss#L3) which I was told was the example I should be following.

emilyrohrbough:
Given the nature of an input element, the screenshots on these appear as blank boxes, which makes sense.  However I am wondering if we need to add a screenshot or two which verifies input styles as well.

packages/terra-form-input/src/Input.scss
dylan-sftwr: I was advised to do this transition in the same way that the existing `form-textarea` component is [implemented](https://github.com/cerner/terra-core/blob/master/packages/terra-form-textarea/src/Textarea.scss#L25).

packages/terra-form-input/tests/wdio/form-input-spec.js
```diff
@@ -0,0 +1,129 @@
+/* global browser, Terra */
+describe('Form-Input', () => {
```
emilyrohrbough: These webdriver tests should be testing more than screenshot comparison and accessiblity. We also want to add test to ensure functionality behaves as expected. Adding functionality tests wound't necessarily require a screenshot afterwards, unless styles change for a behavior.

dylan-sftwr: @JakeLaCombe and I synced about these tests and determined that functional testing would be done via the jest tests.

packages/terra-form-input/src/Input.scss
emilyrohrbough: Sounds good. I wanted to double check because the button callouts also specified white, but it was intended to be transparent.

packages/terra-form-input/tests/wdio/form-input-spec.js
emilyrohrbough: 👍 Sounds good.

packages/terra-form-input/tests/jest/Input.test.jsx
dylan-sftwr: Done 159643ece2f150a8c2f2e00196c9f92a88254c2d

packages/terra-form-input/tests/nightwatch/InputPropPermutations.jsx
dylan-sftwr: Done c33141d942293cae899fc6e6efcd0587346cc7db

packages/terra-form-input/tests/nightwatch/InputTestRoutes.jsx
dylan-sftwr: Done c33141d942293cae899fc6e6efcd0587346cc7db

packages/terra-form-input/tests/wdio/form-input-spec.js
dylan-sftwr: Done c33141d942293cae899fc6e6efcd0587346cc7db

packages/terra-site/src/examples/form-input/controlled_examples/DefaultExample.jsx
```diff
@@ -0,0 +1,32 @@
...
+  constructor(props) {
+    super(props);
+
+    this.state = { input: 'Default Input' };
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  handleChange(event) {
+    this.setState({ input: event.target.value });
+  }
```
benbcai: nit: Looks like several of these examples are very similar except for the input value and name. If you like you could create a default example template and the other examples can just pass in the additional props to the default template. See [here](https://github.com/cerner/terra-core/blob/master/packages/terra-site/src/examples/search-field/SearchFieldExampleTemplate.jsx) as for an example.

packages/terra-form-input/src/Input.scss
neilpfeiffer: #fff - white. match this:
https://github.com/cerner/terra-core/blob/8855edcf3197bbd029e9d4a61da21c3cf2059308/packages/terra-form-textarea/src/Textarea.scss#L3

packages/terra-form-input/src/Input.scss
neilpfeiffer: #fff (white) for default theme

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,89 @@
...
+    &:hover {
+      border: var(--terra-form-input-hover-border, 1px solid #4a4a4a);
+    }
+
+    &[required] {
+      box-shadow: none;
+    }
+
+    &[disabled] {
+      background-color: var(--terra-form-input-disabled-background-color, rgba(0, 0, 0, 0.05));
```
neilpfeiffer: match colors for background-color, border, color, and opacity here:
https://github.com/cerner/terra-core/blob/8855edcf3197bbd029e9d4a61da21c3cf2059308/packages/terra-form-textarea/src/Textarea.scss#L38

packages/terra-form-input/src/Input.scss
neilpfeiffer: checked with @JakeLaCombe - do not add `opacity` here (keep in disabled style block). Too dangerous for someone to misuse that otherwise can already be achieved with rgba values for background, border, color, etc.

He will create an issue to remove it from textarea before it gets used an any other theme.

dylan-sftwr: Changed 44e3f71e18905ce93d017024abfc1047c92e6b98

packages/terra-form-input/src/Input.scss
dylan-sftwr: Changed 44e3f71e18905ce93d017024abfc1047c92e6b98

packages/terra-form-input/src/Input.scss
dylan-sftwr: Changed 44e3f71e18905ce93d017024abfc1047c92e6b98

packages/terra-form-input/src/Input.scss
dylan-sftwr: Changed 44e3f71e18905ce93d017024abfc1047c92e6b98

packages/terra-form-input/src/Input.scss
dylan-sftwr: Done dd73af05d4efa6dfec0077f918ef1b7efd9c0603

bjankord:
@dylan-sftwr Before we merge this PR, this will need to go through UX review via @neilpfeiffer and functional review via @mmalaker

bjankord:
@dylan-sftwr Looks like there are some lint errors that need cleaned up.

```
/home/travis/build/cerner/terra-core/packages/terra-form-input/tests/nightwatch/InputTestRoutes.jsx
  14:16  error  Missing trailing comma  comma-dangle
/home/travis/build/cerner/terra-core/packages/terra-form-input/tests/wdio/form-input-spec.js
  26:7  error  Duplicate key '--terra-form-input-line-height'  no-dupe-keys
```

mjhenkes:
You could add mock theme values here https://github.com/cerner/terra-core/blob/master/packages/terra-site/src/cerner-mock-theme.scss
to demonstrate the themeing capabilities on the doc site.

packages/terra-form-input/src/Input.jsx
emilyrohrbough: If that is the case, we should add an inline comment indicating why this rule was disabled

StephenEsser: If another component does this in the future it may be worth moving this into the root eslintrc, I would keep it here until a future discussion lands on this being a case by case or a standard.

packages/terra-site/src/examples/form-input/uncontrolled_examples/DisabledErrorExample.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
```
JakeLaCombe: Can you rename this file to DisableInvalidExample? It better aligns with the props

packages/terra-form-input/src/Input.jsx
dylan-sftwr: @emilyrohrbough Done. 096f1b6aa2c996a463a080f3483d3d1c282e2df2

dylan-sftwr:
@mjhenkes Done 9839bc6e53c712448641600f2deec4413ce02645

neilpfeiffer:
Edits to match Textarea and Terra-Legacy:

### Normal
1.  Update:  `border: var(--terra-form-input-border, 1px solid` #dedfe0 `);`
https://github.com/cerner/terra-core/blob/7876bf7cc04a97b56bdd83322706508dafed8c5e/packages/terra-form-input/src/Input.scss#L8 
#
2. Update:  `border: var(--terra-form-input-hover-border, 1px solid` #dedfe0 `);`
https://github.com/cerner/terra-core/blob/7876bf7cc04a97b56bdd83322706508dafed8c5e/packages/terra-form-input/src/Input.scss#L26
#
3. Update focus: `border: var(--terra-form-input-focus-border, 1px solid` #4cb2e9 `);`
https://github.com/cerner/terra-core/blob/7876bf7cc04a97b56bdd83322706508dafed8c5e/packages/terra-form-input/src/Input.scss#L44
---

### Error/Invalid
4. Update:  `border: var(--terra-form-input-error-border, 1px solid` #dedfe0 `);`
https://github.com/cerner/terra-core/blob/7876bf7cc04a97b56bdd83322706508dafed8c5e/packages/terra-form-input/src/Input.scss#L59
- other themes use red-border, default needs to match terra-legacy which uses field (error text) to notify input is in error
#
5. Update:  `border: var(--terra-form-input-error-hover-border, 1px solid` #dedfe0 `);`
https://github.com/cerner/terra-core/blob/7876bf7cc04a97b56bdd83322706508dafed8c5e/packages/terra-form-input/src/Input.scss#L62
#
6. Update focus:  `border: var(--terra-form-input-error-focus-border, 1px solid` #4cb2e9 `);`
https://github.com/cerner/terra-core/blob/7876bf7cc04a97b56bdd83322706508dafed8c5e/packages/terra-form-input/src/Input.scss#L75
---

### Padding & Line-Height
7. Update: `line-height: var(--terra-form-input-line-height,` 1.25 `);`
https://github.com/cerner/terra-core/blob/7876bf7cc04a97b56bdd83322706508dafed8c5e/packages/terra-form-input/src/Input.scss#L15
#
8. Update: 
    `padding-bottom: var(--terra-form-input-padding-bottom,` 0.286rem `);`
    `padding-left: var(--terra-form-input-padding-left,` 0.286rem `);`
    `padding-right: var(--terra-form-input-padding-right,` 0.286rem `);`
    `padding-top: var(--terra-form-input-padding-top,` 0.286rem `);`
https://github.com/cerner/terra-core/blob/7876bf7cc04a97b56bdd83322706508dafed8c5e/packages/terra-form-input/src/Input.scss#L16-L19
---

### Placeholder
8. Add for normal (can adjust as needed - something like [this](https://www.w3schools.com/howto/howto_css_placeholder.asp)): 
```
::placeholder { /* Chrome, Firefox, Opera, Safari 10.1+ */
    color: var(--terra-form-input-placeholder-color, #bcbfc0);
    opacity: 1; /* Firefox */
}
:-ms-input-placeholder { /* Internet Explorer 10-11 */
    color: var(--terra-form-input-placeholder-color, #bcbfc0);
}
::-ms-input-placeholder { /* Microsoft Edge */
    color: var(--terra-form-input-placeholder-color, #bcbfc0);
}
```
#
9. Add for disabled: 
```
::placeholder {
    color: var(--terra-form-input-disabled-placeholder-color, rgba(0, 0, 0, 0.1));
    opacity: 1;
}
:-ms-input-placeholder {
    color: var(--terra-form-input-disabled-placeholder-color, rgba(0, 0, 0, 0.1));
}
::-ms-input-placeholder {
    color: var(--terra-form-input-disabled-placeholder-color, rgba(0, 0, 0, 0.1));
}
```
- I will log an enhancement to add to textarea as well

packages/terra-form-input/src/Input.scss
dylan-sftwr: Done d9b680c536c0806fb22c71bc5f8ea3265244136b

dylan-sftwr:
@neilpfeiffer done
* d9b680c536c0806fb22c71bc5f8ea3265244136b
* 62e58cff7fdecea6f466ba236d1024329a80e831

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    &[disabled] {
+      background-color: var(--terra-form-input-disabled-background-color, rgba(111, 116, 119, 0.05));
+      border: var(--terra-form-input-disabled-border, 1px solid #dedfe0);
+      color: var(--terra-form-input-disabled-color, rgba(0, 0, 0, 0.25));
+      cursor: default;
+      opacity: var(--terra-form-input-disabled-opacity);
+
+      /* stylelint-disable, selector-no-vendor-prefix */
```
dylan-sftwr: Talked to @bjankord about an issue seen here with AutoPrefixer, this has his blessing.

packages/terra-form-input/src/Input.scss
```diff
@@ -0,0 +1,107 @@
...
+      [dir=ltr] & {
+        box-shadow: var(--terra-form-input-focus-keyboard-box-shadow-ltr);
+      }
+
+      [dir=rtl] & {
+        box-shadow: var(--terra-form-input-focus-keyboard-box-shadow-rtl);
+      }
+    }
+
+    /* stylelint-disable, selector-no-vendor-prefix */
```
dylan-sftwr: Talked to @bjankord about an issue seen here with AutoPrefixer, this has his blessing.

dm123455:
+1

neilpfeiffer:
Tested: PR #1235
Resolves: Issue #111

\
**Reviewed:**
- Follows pattern considerations documented in Cerner Design Standards: [Input](https://wiki.ucern.com/x/3EUHYw)
- Follows UX defined design implementation for Terra-Form-Input

\
**Fix with Enhancements:**
- themeable text color styling for placeholder text (will attach issue when created)

\
Verified by @scottwilmarth and @neilpfeiffer, Ready for Release.


**Pass** |
-----------|
UX Review|

dylan-sftwr:
@neilpfeiffer , we'll need to log the placeholder styling as a future enhancement.

mmalaker:
Functional verification completed. The form input examples display and function correctly after implementing the style changes.

PR #1240, opened 2018-02-13
Props table style update

### LINKED ISSUE #1201: Create Consistent PropsTable Layout
# Issue Description
The props-table displays inconsistently for each component; the column sizes vary depending on the prop names, prop types and descriptions. Additionally some components don't wrap their text and blow out beyond the screen size which makes it harder to read and navigation. 

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
We want to create a consistent layout/style for all props.

Blackop778:
### Summary
Adds styling to force long Props Table entries to line wrap. Most notably apparent on the Dynamic Grid 
page.
### Additional Details
Includes suggested change from and closes #1201 

@cerner/terra


emilyrohrbough:
In the tech design, we also outlined setting max-widths on specific columns to render consistent props table layouts when long text is present. Lets add these changes here as well.

packages/terra-props-table/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Added
+* Added styling to force long Props Table entries to line wrap
```
emilyrohrbough: lets be specific to indicate we are referring to wrapping the propsTypes: 
`Added styling to wrap long prop-type 'type' entries`

packages/terra-props-table/src/PropsTable.scss
```diff
@@ -0,0 +1,5 @@
+:local {
+    .props_table_pre {
```
emilyrohrbough: The style naming convention our team follows is using hyphenated class names. This should be updated to:
`.props-table-pre {`

Blackop778:
> In the tech design, we also outlined setting max-widths on specific columns to render consistent props table layouts when long text is present. Lets add these changes here as well.

The reason I suggested that was because I hadn't realized that the table would wrap lines in order to not require a scroll bar whenever possible. Another reason I removed it in 81bee80 was because I discovered my implementation actually had no effect on the table.

emilyrohrbough:
@Blackop778 & talked a bit offline. The designated width is still desirable for consistent displays. The use of `width` might be more appropriate. Going to look into it.

Blackop778:
We're once again unsure of whether or not we want to dictate `width` to the table headers. The problem is `max-width` gets ignored so we can only have a one size fits all `width` for each column.
Examples with the proposed 35% and 10% classes:
### Dynamic Grid Props Table
Dynamic Grid Props Table
### Embedded Content Consumer Props Table
Embedded Content Consumer Props Table

packages/terra-props-table/src/PropsTable.scss
```diff
@@ -0,0 +1,13 @@
+:local {
+    .props-table-pre {
+        white-space: pre-wrap;
+    }
+
+    .prop-table-name, .prop-table-required, .prop-table-default {
+        width: 10%;
```
bjankord: If you add `vertical-align: top` it may make these easier to read with the type is really long. 


It would align the text in the other columns to the top of their table cell.

packages/terra-props-table/src/PropsTable.scss
```diff
@@ -0,0 +1,13 @@
...
+    }
+
+    .prop-table-type, .prop-table-description {
+        width: 35%;
```
bjankord: If you add `vertical-align: top` it may make these easier to read with the type is really long. 


It would align the text in the other columns to the top of their table cell.

packages/terra-props-table/src/PropsTable.jsx
```diff
@@ -66,17 +70,19 @@ const PropsTable = ({ componentName, src, ...customProps }) => {
...
   const componentProps = componentMetaData.props;
 
+  const tableEntryClass = cx('prop-table-element');
+
   return (
     <div dir="ltr" className="markdown-body">
       <h2>{componentName} Props</h2>
       <table {...customProps}>
         <thead>
           <tr>
```
emilyrohrbough: You should be able to add the `prop-table-elemenet` class to this table row, instead of each table cell to achieve the vertical-top alignment

packages/terra-props-table/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Added
+* Added styling to wrap long prop-type 'type' entries
```
emilyrohrbough: lets update this to include added layout styling as well.

bjankord:
Talked with @mmalaker offline, we're gonna count this as tech-only.

PR #1245, opened 2018-02-13
Form checkbox label text alignment

praneethm381:
### Summary
Adding css themeable variables for label-text for mobile

### Additional Details
Added vertical-align css variable so that the label text is aligned inline to checkbox 

@cerner/terra

### Deployment URL
https://terra-core-deployed-pr-1245.herokuapp.com/


packages/terra-form-checkbox/src/Checkbox.scss
```diff
@@ -46,6 +46,12 @@
       max-width: 100%; //stops content overflow on IE 10/11
       user-select: none;
     }
+
+    .label-text.is-mobile {
+      display: inline-block;
+      vertical-align: var(--terra-form-checkbox-is-mobile-label-text-vertical-align, 0);
```
bjankord: You can remove the `, 0` part. e.g. `vertical-align: var(--terra-form-checkbox-is-mobile-label-text-vertical-align);`

yuderekyu:
Looks good. Can we set `--terra-form-checkbox-is-mobile-label-text-vertical-align: middle` on the mock theme? That way consumers viewing site can see the centered alignment when triggering a mobile resize.

bjankord:
Is the intent of the style to vertical align the label text optically in the middle next to the checkbox. If so, I don't believe this style is achieving that.

screen shot 2018-02-14 at 12 55 40 pm

praneethm381:
screen shot 2018-02-14 at 1 00 22 pm

seems like the changes work for me.

praneethm381:
Here is the deployment link with changes pointed to my forked terra-core:
https://pages.github.cerner.com/terra/cerner-consumer-theme/build/checkbox-update/#/site/components/site/form-checkbox/index

bjankord:
@praneethm381 That deployment link looks good to me. Looks like we won't get perfectly vertical centered with the height of checkbox, but the consumer font size for the label text on mobile makes it look close enough for me.

ChrisMageeJenks:
I think I'll let @neilpfeiffer make the official terra-core approval, but this works for me (oneup to @bjankord about still not getting perfect alignment, but it's better)

bjankord:
This looks like it is no longer needed.

PR #1263, opened 2018-02-19
Fix Issue with locale switching

### LINKED ISSUE #1236: Locale Switching in Terra Site is behaving inconsistently
# Issue Description
Locale switching in Terra site appears to be having issues. It seems to be rendering one locale back as opposed to rendering the selected locale, and sometimes the locale doesn't even display the messages.

## Issue Type
- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Locale shows the messages and formats correlated to the selected locale

## Current Behavior
Locale appears to be displaying messages and formats of the previously selected locale.

JakeLaCombe:
### Summary
Locale switching broke when I uplifted it in React 16. Sorry about that.

Resolves #1236

### DIFF
--- a/packages/terra-base/CHANGELOG.md
+++ b/packages/terra-base/CHANGELOG.md
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Fixed bug where new locales did not update correctly
 
 3.0.0 - (February 12, 2018)
 ------------------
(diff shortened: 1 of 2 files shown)


JakeLaCombe:
I tried to write a test for it, but couldn't find a good approach. It looks like our nightwatch tests only worked with customMessages, but the error appeared only when customMessages wasn't specified. For Jests tests, the intl only works with the render method, meaning I can't change the props on it.

PR #1275, opened 2018-02-21
Updating button types and variants syntax in Button Docs

Harwood:
### Summary
This isn't a code change, just an update to the documentation. I was adding variant styling and discovered the syntax had changed and had not been updated. This change corrects that.

### Additional Details
N/A

@cerner/terra


Harwood:
>Actually looking at this, updating the documentation to indicate One of Button.Opts.Types.BUTTON... is fine, however we haven't defined it yet to use in the Prop validation. This needs to be changed from

Should the same be reverted for variants?

emilyrohrbough:
@Harwood Yes. Looking at the code, you updated the proptypes to verify the prop against the Button.Opts.{thing}.{thing_type}. These Button.Opts are defined [here](https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L240) at the bottom.

PR #1277, opened 2018-02-22
Use toolkit 2.8.0

emilyrohrbough:
Collapseable menu triggers a portal with the menu popup, which needed an additional selector given the portal is rendered outside of the defaulted selector.


package.json
```diff
@@ -107,7 +107,7 @@
     "stylelint-order": "^0.7.0",
     "stylelint-scss": "^2.0.0",
     "stylelint-suitcss": "^1.0.0",
-    "terra-toolkit": "2.7.0",
+    "terra-toolkit": "2.8.0",
```
mjhenkes: Do you want to be optimistic again? `^`

bjankord:
Should this include an updated screenshot with the collapsible menu popup included in it now that the selector has been updated?

emilyrohrbough:
The selector was updated to grab the same screenshot as before

emilyrohrbough:
Lets wait to merge there. I am seeing some things in terra-framework I want to land on first.

PR #1280, opened 2018-02-23
Add themeing variable, Font-Size to Button component

### LINKED ISSUE #1246: Add theming for Button to meet cerner-consumer-theme requirements
# Issue Description
Uplift button theming to meet cerner-consumer-theme requirements.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

samvaity:
### Summary
Added theme-able variable, `font-size` to Button component to match cerner-consumer-theme- ing requirements.

Fixes #1246 

Deployment Link:
https://terra-core-deployed-pr-1280.herokuapp.com/static/#/site/button


packages/terra-button/src/Button.scss
```diff
@@ -30,7 +30,7 @@
     cursor: pointer;
     display: inline-block;
     font-family: inherit;
-    font-size: inherit;
+    font-size: var(--terra-button-font-size, 1rem);
```
dylan-sftwr: Do we want the fallback value here to really be `1 rem`, or `inherit`? This seems like a non-passive change to me for anyone else that's consuming button?

bjankord: Good question, looking back through commits, it looks like we had this set to 1rem prior to updating button to align with OCS in the v2 implementation.we probably should have originally coded the font-size to be 1rem, as using `inherit` allows outside styles to impact the size of the button. e.g A parent could set the font-size to 40px and now button's will have a 40px font-size.

The issue is users may currently be relying on this behavior which I'd consider a side-effect. 

We should keep the fallback as inherit for passivity. This does open a gap where people can change the buttons overall size by apply a large font size to a parent containing button

bjankord: Good question, [looking back through commits](https://github.com/cerner/terra-core/blob/f3061e7f6fb6e4b130bc5482742b4d6fbeb48b78/packages/terra-button/src/Button.scss#L36), it looks like we had this set to `1rem` prior to updating button to the v2 implementation. Looks like in v2, we set the font-size to `inherit`. I'd consider this a bug we missed in code review. Using `inherit` allows for outside modifications to alter the button appearance. e.g a parent could set the font-size to 40px and now button's will have a 40px font-size and appear much larger. I can also see how some might consider this a feature rather than a bug and users may currently be relying on this behavior that I'd consider a side-effect. 

Keeping the fallback as `inherit` keeps passivity with what we coded in v2, though setting the fallback to `1rem` gets us back to v1 behavior and more predictable font-size behavior in the button. 

I'd recommend we set the fallback to `1rem` to provide more predictable font-size and button appearance. Teams that need larger buttons can either theme them to be larger overall or request features to add the size prop back.

bjankord:
@samvaity I think we can skip UX review on this and count it as tech only.

PR #1288, opened 2018-02-26
Theming list component for focus border

praneethm381:
### Summary
Added themable variables for list component

### Additional Details
Added:

- Box shadow css variable for list-item focus
- css variable for default list-item color

Corresponding pull request in cerner-consumer-theme is here: https://github.cerner.com/terra/cerner-consumer-theme/pull/44

@cerner/terra


packages/terra-list/src/List.scss
```diff
@@ -50,14 +51,20 @@
...
-    &.list-item:hover,
-    &.list-item:focus {
+    &.list-item:hover {
       background-color: var(--terra-list-item-hover-background-color, lighten(#ebf6fd, 2%)) !important;
     }
 
+    &.list-item:focus {
+      background-color: var(--terra-list-item-focus-background-color, lighten(#ebf6fd, 2%)) !important;
+      outline: var(--terra-list-item-focus-outline);
+      box-shadow: var(--terra-list-item-focus-box-shadow);
```
samvaity: Should have a fallback value for this and others too?

bjankord: No fallback value is needed if the default theme does not need these styles.

packages/terra-list/src/List.scss
```diff
@@ -50,14 +51,20 @@
...
+    }
+
     &.selected:hover,
     &.selected:focus {
```
emilyrohrbough: What about selected:focus?

PR #1296, opened 2018-02-27
Updated Dialog Theme Variable Name

saedar:
### Summary
When I wrote up the Dialog component, I made a typo in the name of a theme variable.

### DIFF
--- a/packages/terra-dialog/src/Dialog.scss
+++ b/packages/terra-dialog/src/Dialog.scss
@@ -44,7 +44,7 @@
 
   .dialog-footer {
     background-color: var(--terra-dialog-footer-background-color, #f4f4f4);
-    border-top: var(--terra-dialog-footer-bottom-top, 1px solid #dedfe0);
+    border-top: var(--terra-dialog-footer-top-border, 1px solid #dedfe0);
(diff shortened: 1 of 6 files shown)


packages/terra-dialog/src/Dialog.scss
```diff
@@ -44,7 +44,7 @@
 
   .dialog-footer {
     background-color: var(--terra-dialog-footer-background-color, #f4f4f4);
-    border-top: var(--terra-dialog-footer-bottom-top, 1px solid #dedfe0);
+    border-top: var(--terra-dialog-footer-top-border, 1px solid #dedfe0);
```
mjhenkes: Should this be considered non-passive? (I know in this case we just released it and it's probably fine.) But if we made this change a year from now would it be a major version bump?

saedar: I was just thinking about that. I feel like it would be.

bjankord: Yeah, I think if we made this year a from now, or even a month from now, we should major version bump. I think due to this only being released for a day, we can count this as a bug and release a patch release.

PR #1298, opened 2018-02-27
Fix Terra Image Usage Documentation

ChaseSinclair:
The 'placeholder' property was incorrectly labeled 'placeholderSrc'.

### DIFF
--- a/packages/terra-image/docs/README.md
+++ b/packages/terra-image/docs/README.md
@@ -11,8 +11,8 @@ The terra-image component provides styling for visual imagery.
 ## Usage
 
 ```jsx
-import React from 'react;
-import Image from 'terra-image;
+import React from 'react';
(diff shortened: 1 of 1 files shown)


packages/terra-image/docs/README.md
```diff
@@ -18,11 +18,11 @@ import placeholderSrc from './mock/valid/path/placeholder.png';
 import invalidImageSrc from './mock/invalid/path/image.png';
```
StephenEsser: Would you mind adding the missing quote to the terra-image import above?

```
import Image from 'terra-image;
```

ChaseSinclair: Good catch. I fixed the react import as well.

bjankord:
@ChaseSinclair Once the build passes on this, you can merge it down.

Issue #1294, opened 2018-02-27
Enable Card Body to center align its contents


alex-bezek:
# Issue Description
I have found that in most of our use cases, the contents of a card end up having to utilize text-align center. While this works fine, it ends up with some random css scattered around to do this. 1 option we have to just make a reusable component for centering in a div and pass that to the card body. However, the pattern I have noticed is that cards almost always have their body centered.

My proposal is either Card or Card.Body enable a prop `isContentCentered` which adds this css. 

## Issue Type

- [x] Enhancement

## Expected Behavior
I would like to be able to center a card like this
```
<Card>
  <Card.Body isContentCentered>
     <MyContent />
  </Card.Body>
</Card>
```

## Current Behavior
Currently, a consumer can pass in their content wrapped in a component that applies centering

bjankord:
+1 on `isContentCentered` prop. We could use flexbox to center align or text-align: center. I think like flexbox may better align items in the center as `text-align: center` will only apply to `inline` and inline-ish displays like `inline-block`.

alex-bezek:
good point on the flex box. Who else would need to approve this idea before someone can make the PR? It should be a simple PR and I have people ready to make this centering component in the next few days. I would prefer they just contribute this to terra.

Thanks for the quick response!

dm123455:
I agree with Alex, there are lots of PR where people are adding custom css to align text to the center.
+1 on doing this. @bjankord  @mjhenkes are you fine with it?

mjhenkes:
+1 to `isContentCentered` prop

mananjain149:
I think we need maximum width also in Card Body for some cards like noDataCard.

alex-bezek:
@mananjain149 I'm not convinced the cards size should be a prop on the card itself. It could just be constrained by a container placed around it. You could take that 1 step further and handle center alignment of the card on the page and its now a layout component that can be used for things other than just card as well

![image](https://user-images.githubusercontent.com/8029578/36743052-e1ac299e-1bae-11e8-8f96-dbf7f9b73d4d.png)
![image](https://user-images.githubusercontent.com/8029578/36743062-e9c81dfe-1bae-11e8-9cd6-3b62fdd7f68b.png)

In these examples its not just a card for example

emilyrohrbough:
Resolved by #1300

PaulSolDev:
JIRA created

PR #1300, opened 2018-02-27
Center content in CardBody

vaibhavpaliwal:
### Summary
This PR resolves the issue https://github.com/cerner/terra-core/issues/1294 to center content inside CardBody.


dm123455:
add wdio test?

packages/terra-card/src/CardBody.scss
```diff
@@ -9,4 +9,10 @@
     padding-left: var(--terra-card-padding-left, 10px);
     padding-right: var(--terra-card-padding-right, 10px);
   }
+
+  .center-content {
+    align-items: center;
+    display: flex;
+    justify-content: center;
```
StephenEsser: It may be worth discussing if there should be a distinction between centering content vertically, horizontally, or both unless the only use-case is always centering content vertically and horizontally.

dm123455: I would think both vertically and horizontally for now. If we run into scenarios that need either, we can revisit?!

StephenEsser: I'm curious if vertically centering the content with `align-items: center;` has any effect if the Card.Body height is always auto.

packages/terra-site/src/examples/card/CardContentCentered.jsx
```diff
@@ -0,0 +1,14 @@
...
+import Card from 'terra-card';
+
+const CardContentCentered = () => (
+  <div>
+    <Card>
+      <Card.Body isContentCentered>
+         Content is centered
+      </Card.Body>
+    </Card>
+  </div>
```
dev-cprice: Does this need to be wrapped in a div?

packages/terra-card/src/CardBody.scss
StephenEsser: Because the Card.Body accepts multiple children it's worth mentioning `display: flex` has more side effects than just centering content. This may not be a huge issue if the intent is to always center a single child and it's opt in.

```
<Card>
  <Card.Body>
    <div>Block 1</div>
    <div>Block 2</div>
  </Card.Body>
</Card>;
```

Before:
image
With `isContentCentered`:
image

packages/terra-site/src/examples/card/CardContentCentered.jsx
vaibhavpaliwal: it doesn't have to be wrapped in a div I will update

packages/terra-card/src/CardBody.scss
alex-bezek: As @StephenEsser  noted, I don't think vertical-align makes much since because of how the height is calculated.

The blocks sitting inline like that seems like an issue though. It could be mitigated by conditionally wrapping the children in another div when the centering is used
https://codepen.io/anon/pen/MQZpaE
Multiple items is a definite use case, and forcing the consumer to put everything in a div seems bad

vaibhavpaliwal: Good point @StephenEsser I feel like there should be a distinction between horizontal and vertical alignment because of the side effect you mentioned but I don't have lot of experience to comment on if our use cases have been to center the contents horizontally and not vertically in which case I can just remove the vertical alignment and this component can be updated later when use case arises for vertical alignment.

If we decide to take in separate props I think we can create `isContentCenteredHorizontally` and `isContentCenteredVertically`. 

@bjankord @alex-bezek thoughts ?

saedar: I vote to split out the props.

StephenEsser: With the Card.Body always having a height of auto I don't think there is a concept of vertical centering. The content is always the full height of Card.Body ( minus padding ), so there are no adjustments to be made above or below the content. In order to center the content the Card.Body would need to be taller than the content.

bjankord: > With the Card.Body always having a height of auto I don't think there is a concept of vertical centering. 

Agreed, let's drop the `align-items: center;` style. I'm also open to renaming the prop to `isContentCenteredHorizontally` or just leaving it as `isContentCentered`.

> it's worth mentioning display: flex has more side effects than just centering content. This may not be a huge issue if the intent is to always center a single child and it's opt in.

bjankord: Theres a few different ways I can think of to center items

alex-bezek: I'm not really sure we want to get in the game of just providing react wrappers of different css styles do we? 
I feel like the problem we are trying to solve with this isn't "the ability to center things" but rather that I have found a common pattern in our callouts for centering the contents of a card.

We try to treat `any css specific to a page as a code smell` since it signifies something not buildable with our existing component library, but a concept that we need to deliver for UX. So if its a unique situation for our app, we just try to make it a re-usable component in our application. I'm almost against the center component just because it prevents us from easily finding those `css files that just consist of a single text-align: center class and that's it` and identifying that pattern.

For reference where this is coming from, I noticed that in our app, over the last couple months (and including the page @vaibhavpaliwal is working on) there have been
* 12 instances of `text-align: center;` added
* 4 are the only class/style in their components css
* 9 are in cards

We only have 1 area we added flex box. My understanding is that flex box applies to all the contents, but text-align center only affects non block level elements. I understand others may have other needs, but our stuff based off existing UX callouts hasn't encountered the issue of needing to center block elements.

I think the 3 options are
* Just use text-align. It solves the 90% use case, and consumer that actually have content with block items can make that center itself and put it in the card. https://codepen.io/anon/pen/QQYvBV
* Use the idea of throwing an extra div in there to fix flex box issue https://codepen.io/anon/pen/YeBVMQ
* Not allow cards to center their body, but just allow consumers to center their own contents. Provide a center component.

bjankord: I'm good with having the `isContentCentered` just set `text-align: center;`. I agree that for cases this style doesn't center stuff, the consumer can make that content / component center itself and put it in the card.

packages/terra-site/src/examples/card/CardContentCentered.jsx
vaibhavpaliwal: updated thanks!

packages/terra-card/src/CardBody.scss
vaibhavpaliwal: updated to use `text-align: center;`
Thanks!

vaibhavpaliwal:
@dm123455 Added wdio tests thanks!

Issue #1259, opened 2018-02-16
Remove Props-table Bin Executable


emilyrohrbough:
# Issue Description
Once terra-ui.com is uplifted to use terra-site v2.0.0 this removes the need to create add and create the props-table markdown files for each component. Thus, the prop-table markdown script can be removed from the package.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

emilyrohrbough:
Blocked by:
- [ ] https://github.com/cerner/terra-ui.com/issues/57

Blackop778:
In #1301 it was decided that the script will be kept in incase someone else was using it

PR #1301, opened 2018-02-28
Remove props table script

### LINKED ISSUE #1256: Remove PropsTable Scripts
# Issue Description
Once terra-ui.com is uplifted to use terra-site v2.0.0 this removes the need to create add and create the props-table markdown files for each component.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

Blackop778:
### Summary
Removes the `props-table` command used to generate markdown files for the Props Table and removes the Props Table as a dependency from all packages that required it for `props-table`.

### Additional Details
Closes #1256 and #1259 

@cerner/terra


bjankord:
I'm fine with removing the props-table dependency from all the packages, though removing the CLI functionality from props-table is non-passive. While we no longer need it, there may be teams relying on this. I'm curious what others think about this.

mjhenkes:
I would like to leave the cli functionality in. It doesn't hurt anything and if someone is using it they won't be broken.

package.json
```diff
@@ -57,6 +57,7 @@
     "lint:js": "eslint --ext .js,.jsx .",
     "lint:scss": "lerna run lint:scss",
     "nightwatch": "nightwatch",
+    "props-table": "lerna run props-table",
```
emilyrohrbough: This should be removed.

PR #1307, opened 2018-02-28
Terra Avatar

### LINKED ISSUE #1279: Avatar Component (Revised)
# Technical Design: `terra-avatar`

## Summary:
A discussion with UX has lead to a necessary redesign of the Avatar component. This revised tech design includes the following changes:

- only two avatar types (User/Facility)
- icon can no longer be provided as a prop (only the User and Facility icons are used)
- specifying that if an image or text is provided, it will display in place of the fallback icon. 

See the outdated [tech design](https://github.com/cerner/terra-core/issues/1226).

## Expected behavior:
Should be able to give the component an image or text which will display inside a circular wrapper. If neither of these are provided, then the default User or Facility icons display. Which icon is displayed is determined by the prop 'type'.

## Responsiveness:
The content within the avatar does not change, only the size. The size of the avatar is dictated by the component where it lives.

## I18N / Bi-directionality:
These changes should make no changes to the existing ltr/rtl functionality.

## Interaction States
Avatars will not have interaction states. The hover state shown on the callouts will be controlled by the parent component not the Avatar.

## Accessibility
We'll provide an aria-label to indicate that the component is an avatar.

## React Props:
| Props | Type | Default | Required? | Description |
| ----------- | ----- | --------- | ---- | --------------------------------------------------------- |
| `alt` | string | undefined | No | The text content that specifies the alternative text for the image. |
| `variant` | string | `user` | No | The type of avatar. Either `'user'` or `'facility'`. |
| `image` | string | undefined | No | The image to display. |
| `initials` | string | undefined | No | The initials to display. Limited to 2 or 3 characters. |

## Themeable Styles:
* Background
  * color
  * opacity
* Border
  * width
  * style
  * color
* Box shadow
* Font size
* Font color
* Size
* Icon color
* Line height

ChaseSinclair:
### Summary
The terra-avatar component displays an avatar which can be either an image or text. If neither an image or text is provided, a variant-based fallback icon is used. The avatar displays in a circular wrapper.

Resolves #1279. 

### Deployment Link

https://terra-core-deployed-pr-1307.herokuapp.com/static/#/site/avatar


packages/terra-avatar/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "terra-avatar",
+  "main": "lib/Avatar.js",
+  "version": "0.0.0",
+  "description": "{insert description}",
```
bjankord: Add description here `The terra-avatar component displays an avatar which can be either an image or text. If neither an image or text is provided, a variant-based fallback icon is used. The avatar displays in a circular wrapper.`

packages/terra-avatar/docs/README.md
```diff
@@ -0,0 +1,32 @@
...
+<Avatar
+  image="https://path/to/image.jpg"
+  alt="placeholder"
+  variant="user" 
+/>
+```
+
+## Component Features
+
+<!-- Uncomment supported features.
```
bjankord: You can uncomment any supported features this component has. This helps users better understand what the component offers.

packages/terra-avatar/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
+    "terra-props-table": "^1.10.1"
+  },
+  "peerDependencies": {
+    "react": "^15.4.2",
```
bjankord: Our generator is a bit out-dated. You'll want to bump react and react-dom to `^16.2.0`

packages/terra-avatar/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "UI"
+  ],
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
+    "terra-props-table": "^1.10.1"
```
bjankord: If https://github.com/cerner/terra-core/pull/1301 goes into master before this, you'll want to remove terra-props-table from the devDependencies as well as the props-table npm script: props-table

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,92 @@
...
+    let avatarContent = null;
+    if (image) {
+      avatarContent = <TerraImage src={image} placeholder={icon} alt={alt} />;
+    } else if (initials && (initials.length === 2 || initials.length === 3)) {
+      avatarContent = <text className={AvatarTextClassNames}>{initials.toUpperCase()}</text>;
+    } else {
+      avatarContent = icon;
+    }
+    return (
+      <circle aria-label="Avatar" {...attributes} className={AvatarClassNames} >
```
bjankord: `<circle />` should go inside of an SVG element.

packages/terra-avatar/src/Avatar.scss
```diff
@@ -0,0 +1,56 @@
+@import '~terra-mixins';
+
+:local {
+  .avatar {
+    align-items: center;
+    background-color: var(--terra-avatar-background-color, rgb(222, 223, 224));
+    border: var(--terra-avatar-border, 0 none);
```
bjankord: If border is not needed as part of the default theme, you can omit the fallback style.

packages/terra-avatar/src/Avatar.scss
```diff
@@ -0,0 +1,56 @@
...
+    border-radius: 50%;
+    box-shadow: var(--terra-avatar-box-shadow);
+    color: var(--terra-avatar-color, rgb(155, 159, 161));
+    display: inline-flex;
+    height: var(--terra-avatar-size, 3.429rem);
+    justify-content: center;
+    line-height: 0;
```
bjankord: Why is this needed?

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,92 @@
...
+   * The avatar variant. One of `AvatarVariants.FACILITY`, `AvatarVariants.USER`.
+   */
+  variant: PropTypes.oneOf([AvatarVariants.FACILITY, AvatarVariants.USER]),
+};
+
+const defaultProps = {
+  variant: AvatarVariants.USER,
+};
+
+export const propsErrorMsg = 'Only one of the props: [image, initials] should be supplied.';
```
StephenEsser: Why is this being exported?

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,92 @@
...
+
+const Avatar = ({
+  alt,
+  image,
+  initials,
+  variant,
+  ...customProps
+  }) => {
+  if (image && initials) {
+    throw new Error(propsErrorMsg);
```
StephenEsser: I think this should be a warning, I personally don't think presentational components should hard crash intentionally or should do so only in extremely special circumstances. Does anyone else have any opinions on this?

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,92 @@
...
+    <div className={cx('avatar-icon')}>
+      <svg className={cx(variant)} />
+    </div>
+  );
+
+  const AvatarContent = () => {
+    let avatarContent = null;
+    if (image) {
+      avatarContent = <TerraImage src={image} placeholder={icon} alt={alt} />;
+    } else if (initials && (initials.length === 2 || initials.length === 3)) {
```
StephenEsser: We should provide some documentation on the site and / or propTypes for the length requirements necessary to display initials.

packages/terra-avatar/src/Avatar.scss
```diff
@@ -0,0 +1,56 @@
...
+    &.avatar-text-large {
+      font-size: var(--terra-avatar-font-size-large, 1.571rem);
+    }
+    
+    &.avatar-text-small {
+      font-size: var(--terra-avatar-font-size-small, 1.143rem);
+    }
+  }
+  
+  circle img {
```
StephenEsser: You may want to scope these to a class. This will globally apply 100% height and width to any elements matching this selector.

packages/terra-avatar/src/Avatar.scss
```diff
@@ -0,0 +1,56 @@
...
+    height: 100%;
+    width: 100%;
+    
+    /* Ignoring so that single quotes can be used in svg string */
+    /* stylelint-disable string-quotes */
+    > .facility {
+      background: var(--terra-avatar-icon-facility, inline-svg("<svg data-name='Layer_1' xmlns='http://www.w3.org/2000/svg' fill='rgb(155, 159, 161)' viewBox='0 0 48 48'><path d='M33,20.2V12H15v8.2H9.8V48h12v-9h4.5v9h12V20.2H33z M17.2,42.8h-3V39h3V42.8z M17.2,33.8h-3V30h3V33.8z M25.5,33.8h-3V30h3 V33.8z M29.2,22.5h-3.8v3.8h-3v-3.8h-3.8v-3h3.8v-3.8h3v3.8h3.8V22.5z M33.8,42.8h-3V39h3V42.8z M33.8,33.8h-3V30h3V33.8z'/></svg>"));
+    }
+    
+    > .user {
```
StephenEsser: This is up to your preference, these classes can be flattened to remove descendent selectors.

```
.icon {
  // CSS for Icon
}

.facility {
  // CSS for facility
}

. user {
  // CSS for user
}
```

packages/terra-avatar/src/Avatar.scss
```diff
@@ -0,0 +1,56 @@
...
+    
+    > svg {
+      background-repeat: no-repeat;
+      background-size: contain;
+      height: 100%;
+      width: 100%;
+    }
+  }
+  
+  .avatar-text {
```
StephenEsser: If there are no styles applied to this class you might consider removing it and keeping `large-text` `small-text`

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  const attributes = Object.assign({}, customProps);
+
+  const AvatarClassNames = cx([
+    'avatar',
+    attributes.className,
+  ]);
+
+  const AvatarTextClassNames = cx([
```
StephenEsser: These variables should be lowercase

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,92 @@
...
+    { 'avatar-text-large': initials !== undefined && initials.length === 2 },
+  ]);
+
+  const icon = (
+    <div className={cx('avatar-icon')}>
+      <svg className={cx(variant)} />
+    </div>
+  );
+
+  const AvatarContent = () => {
```
StephenEsser: This appears to be a double nested functional component, is there a way to flatten this logic into a single component?

packages/terra-avatar/src/Avatar.scss
ChaseSinclair: Terra Base sets a line height that causes vertical spacing to be off centered.

jestconfig.json
```diff
@@ -7,7 +7,7 @@
     "./node_modules/enzyme-to-json/serializer"
   ],
   "moduleNameMapper": {
-    "\\.(svg)$": "<rootDir>/__mocks__/fileMock.js",
+    "\\.(svg|jpg)$": "<rootDir>/__mocks__/fileMock.js",
```
JakeLaCombe: Perhaps we should throw on png in here as well, as I do envision other teams will be using that probably sooner than later?

packages/terra-avatar/package.json
ChaseSinclair: Alright, I'll keep an eye on it. I think it'll probably make it in before this does.

packages/terra-avatar/src/Avatar.scss
saedar: If this ends up being needed, can we include a comment about why?

packages/terra-avatar/src/Avatar.jsx
bjankord: Yeah, we'll want to update this to throw a warning instead of an error as to not crash the app. Take a look at the code in terra-base. We log a console warning and only if the code is running in non-prod mode: https://github.com/cerner/terra-core/blob/master/packages/terra-base/src/baseStyles.js

packages/terra-avatar/src/Avatar.scss
bjankord: +1 for flattening these selectors.

packages/terra-avatar/src/Avatar.scss
bjankord: Yes, will want to style this with a class. e.g `.my-selector > circle > img` or `.img` and set the `.img` class directly on the `img` element.

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: I exported it so that I could use it in a test, which I forgot to write. I'll add the test.

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: So a console.warn() or is there a better way?

packages/terra-avatar/README.md
```diff
@@ -0,0 +1,25 @@
...
+- [Documentation](https://github.com/cerner/terra-core/tree/master/packages/terra-avatar/docs)
+- [LICENSE](#license)
+
+## Getting Started
+
+- Install from [npmjs](https://www.npmjs.com): `npm install terra-avatar`
+
+## LICENSE
+
+Copyright 2017 Cerner Innovation, Inc.
```
JakeLaCombe: I think we need a 2018 here? I also think the latest terra generator fixes this as well.

packages/terra-avatar/docs/README.md
```diff
@@ -0,0 +1,32 @@
+# Terra Avatar
+
+The terra-avatar component displays an avatar which can be either an image or text. If neither an image or text is provided, a variant-based fallback icon is used. The avatar displays in a circular wrapper.
```
JakeLaCombe: Nitpick.

Let's combine the first and third sentences. I think it will read a little better

    The terra-avatar component displays an avatar, which can be either an image or text, in a circular wrapper.

Also, I would rather say circular frame than circular wrapper. Wrapper feels like a code element while frame is a visual display.

packages/terra-avatar/src/Avatar.jsx
JakeLaCombe: In that case, let's just hard code it into the test as well. I don't think exporting an error message inside a package component is best practice.

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: Good idea, I'll add it to the props doc.

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,92 @@
...
+    let avatarContent = null;
+    if (image) {
+      avatarContent = <TerraImage src={image} placeholder={icon} alt={alt} />;
+    } else if (initials && (initials.length === 2 || initials.length === 3)) {
+      avatarContent = <text className={AvatarTextClassNames}>{initials.toUpperCase()}</text>;
```
JakeLaCombe: Where does the text node come from?

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: 👍

packages/terra-avatar/src/Avatar.scss
ChaseSinclair: Good idea. Anyone else with thoughts around this?

packages/terra-avatar/src/Avatar.jsx
bjankord: It's a SVG based element: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/text

packages/terra-avatar/package.json
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/855d5fc5c14ce5dc2a29cb07d7d5575af2320962).

packages/terra-avatar/src/Avatar.scss
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/855d5fc5c14ce5dc2a29cb07d7d5575af2320962).

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/855d5fc5c14ce5dc2a29cb07d7d5575af2320962).

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/855d5fc5c14ce5dc2a29cb07d7d5575af2320962).

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/855d5fc5c14ce5dc2a29cb07d7d5575af2320962).

packages/terra-avatar/src/Avatar.scss
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/855d5fc5c14ce5dc2a29cb07d7d5575af2320962).

packages/terra-avatar/src/Avatar.scss
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/855d5fc5c14ce5dc2a29cb07d7d5575af2320962).

packages/terra-avatar/src/Avatar.scss
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/855d5fc5c14ce5dc2a29cb07d7d5575af2320962).

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/855d5fc5c14ce5dc2a29cb07d7d5575af2320962).

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/855d5fc5c14ce5dc2a29cb07d7d5575af2320962).

jestconfig.json
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/855d5fc5c14ce5dc2a29cb07d7d5575af2320962).

packages/terra-avatar/README.md
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/855d5fc5c14ce5dc2a29cb07d7d5575af2320962).

packages/terra-avatar/docs/README.md
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/855d5fc5c14ce5dc2a29cb07d7d5575af2320962).

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: I don't actually need to use a circle here anyways. How about the following?
```jsx
<div {...attributes} aria-label={ariaLabel} className={avatarClassNames} >
  {avatarContent}
</div>
```

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: How about replacing this with a span so we don't have to wrap it with an svg?

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,92 @@
...
+
+const Avatar = ({
+  alt,
+  image,
+  initials,
+  variant,
+  ...customProps
+  }) => {
+  // Checks to run when not in production
+  if (process.env.NODE_ENV !== 'production') {
```
dm123455: trying to understand the reason for this?

ChaseSinclair: So we don't try to log to the console in prod.

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: Added required ariaLabel in this [commit](https://github.com/cerner/terra-core/commit/2d68f1a14c1c54e9d22de3287abe8049b03c8547).

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: I went ahead and replaced it with a span for now in this [commit](https://github.com/cerner/terra-core/commit/2d68f1a14c1c54e9d22de3287abe8049b03c8547).

packages/terra-avatar/src/Avatar.jsx
JakeLaCombe: I see that propsWarningMsg is used only once. For that case, I recommend not using a variable

packages/terra-avatar/src/Avatar.scss
```diff
@@ -0,0 +1,54 @@
...
+    line-height: 0;
+    overflow: hidden;
+    width: var(--terra-avatar-size, 3.429rem);
+  }
+  
+  .avatar-icon {
+    height: 100%;
+    width: 100%;
+    
+    > svg {
```
dylan-sftwr: **NOTE:** if we end up changing the tag like I mentioned above on this to a span, then these styles will no longer be applied, as we won't be using the `svg` element. I think it might be better to have a target class at the icon anyway, that the `facility` and `user` css classes are nested within, to ensure they only have the potential to target the element your inline-svg is being set as the background on, but i'm not super picky about this

packages/terra-avatar/tests/jest/Avatar.test.jsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
```
dylan-sftwr: Good coverage throughout this👍

packages/terra-avatar/tests/nightwatch/FacilityAvatar.jsx
```diff
@@ -0,0 +1,5 @@
+import React from 'react';
+
+import Avatar from '../../lib/Avatar';
+
+export default () => <Avatar id="facility-avatar" variant="facility" ariaLabel="facility" />;
```
dylan-sftwr: this is cheating, but I dig it ha

dylan-sftwr:
Just to confirm, are the [initials](https://github.com/cerner/terra-core/blob/318e88e21d45e6bdfee3101e0065d2de8b2b43ca/packages/terra-avatar/tests/wdio/__snapshots__/reference/chrome/Three_Initials%5B--terra-avatar-size%5D.1050x895.png) supposed to stay the same size, regardless of the avatar size?

dylan-sftwr:
Just to confirm, [this](https://github.com/cerner/terra-core/blob/318e88e21d45e6bdfee3101e0065d2de8b2b43ca/packages/terra-avatar/tests/wdio/__snapshots__/reference/chrome/User%5B--terra-avatar-icon-user%5D.1050x895.png) is supposed to appear as a solid colored circle?

ChaseSinclair:
> Just to confirm, are the initials supposed to stay the same size, regardless of the avatar size?

Yep, initials are sized as a separate themeable variable.

ChaseSinclair:
>Just to confirm, this is supposed to appear as a solid colored circle?

Yep, that's just how we're testing that the background (used for the svg's) is themed.

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+const propTypes = {
+  /**
+   * The text content that specifies the alternative text for the image.
+   */
+  alt: PropTypes.string,
+  /**
+   * The text content of the aria-label for accessibility.
+   */
+  ariaLabel: PropTypes.string.isRequired,
```
JakeLaCombe: On the original tech design, this wasn't a prop. Is there a way this component can populate the aria-label for you instead?

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,97 @@
...
+   */
+  initials: PropTypes.string,
+  /**
+   * The avatar variant. One of `AvatarVariants.FACILITY`, `AvatarVariants.USER`.
+   */
+  variant: PropTypes.oneOf([AvatarVariants.FACILITY, AvatarVariants.USER]),
+};
+
+const defaultProps = {
+  variant: AvatarVariants.USER,
```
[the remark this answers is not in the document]
StephenEsser: I would follow the standards set by Terra, but I'm personally not a fan of defaulting unnecessary props. It creates additional states and assumptions the props can be in. 

#### Defaulting empty arrays 
Leads to invalid logic statements:

| Syntax        | if (prop)           |
| ------------- |:-------------:|
| `<Component prop={['1', '2', '3']} />`     | true|
| `<Component />`      | true      |
| `<Component prop={null} />`  | false      |

#### Defaulting strings to empty
Developers may accidentally start making an assumption that prop can always have a function acted upon it. `prop.length`. But a common workflow for assigning props may involve a conditional statement that sets the value to null.
```
const title = shouldHaveTitle ? 'Title' : null;
<Component title={title} />

// Error
if (title.length > 0)
```

#### Defaulting values to undefined 
There are a few examples floating around, including one in this pull request, where the prop value is compared against undefined.

```
const initials = shouldHaveInitials ? 'SE' : null;

<Avatar initials={initials} />

// Error
{ 'avatar-text-small': initials !== undefined && initials.length === 3 }
```

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: We had originally considered doing this, but having the component populate the label didn't provide enough information to make the label useful from an accessibility standpoint. Using a required prop, the consumer could send in the user's/facility's name which provides much more value than "User Avatar" or "JS". I would say as is, this is what we believed to be the best approach. If you have a suggestion about how the component could provide a valuable label, particularly considering the complications that arise based on whether this is an _image_, _initials_, or _icon_ `Avatar` then i’d happily make the change.

packages/terra-avatar/src/Avatar.jsx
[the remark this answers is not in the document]
JakeLaCombe: I think for these props, I would keep the default value as undefined like we do here. https://github.com/cerner/terra-core/blob/master/packages/terra-demographics-banner/src/DemographicsBanner.jsx#L54. If the prop is something like an array or an iterator, possibly even a hash of objects, that's when I would set the default to an empty array or empty hash.

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,97 @@
...
+  if (image) {
+    avatarContent = <TerraImage className={cx('avatar-image')} src={image} placeholder={icon} alt={alt} />;
+  } else if (initials && (initials.length === 2 || initials.length === 3)) {
+    avatarContent = <span className={avatarTextClassNames}>{initials.toUpperCase()}</span>;
+  } else {
+    avatarContent = icon;
+  }
+
+  return (
+    <div {...attributes} aria-label={ariaLabel} className={avatarClassNames} >
```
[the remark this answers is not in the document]
JakeLaCombe: Strange that the linter didn't pick it up. I thought it did that for node elements.

packages/terra-avatar/tests/jest/Avatar.test.jsx
```diff
@@ -0,0 +1,71 @@
...
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render facility avatar when variant is facility', () => {
+    const avatar = <Avatar variant="facility" ariaLabel="facility" />;
+    const wrapper = render(avatar);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render two intials avatar when two initials are passed in', () => {
```
JakeLaCombe: *initials

packages/terra-avatar/tests/nightwatch/FacilityAvatar.jsx
JakeLaCombe: *effective coding* is a good synonym here!

packages/terra-avatar/src/Avatar.jsx
JakeLaCombe: That makes more sense. Perhaps for our documentation and test examples, can you change some of them to use names such as John Doe or a medical center?

packages/terra-avatar/src/Avatar.jsx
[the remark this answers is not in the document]
dm123455: I agree with @JakeLaCombe and @StephenEsser, if it is optional prop I would not set a default behavior to it until we hit some edge case we have not accounted for here.

packages/terra-avatar/src/Avatar.jsx
JakeLaCombe: Also, let's make this not required.

packages/terra-avatar/src/Avatar.jsx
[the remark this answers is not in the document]
JakeLaCombe: @StephenEsser I say we stick with the pattern inside the demographics banner, and then do a war room discussion on implementing standards for default props.

packages/terra-avatar/package.json
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/901237e7a6938c3655ffa65a772ace2bb6d7144e).

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/901237e7a6938c3655ffa65a772ace2bb6d7144e).

packages/terra-avatar/package.json
```diff
@@ -0,0 +1,51 @@
...
+  "peerDependencies": {
+    "react": "^16.2.0",
+    "react-dom": "^16.2.0",
+    "terra-base": "^2.0.0",
+    "terra-mixins": "^1.13.0"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "prop-types": "^15.5.8",
+    "terra-base": "^2.0.0",
```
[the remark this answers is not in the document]
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/901237e7a6938c3655ffa65a772ace2bb6d7144e).

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+  const attributes = Object.assign({}, customProps);
+
+  const avatarClassNames = cx([
+    'avatar',
+    attributes.className,
+  ]);
+
+  const avatarTextClassNames = cx([
+    { 'avatar-text-small': initials !== undefined && initials.length === 3 },
```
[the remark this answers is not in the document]
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/901237e7a6938c3655ffa65a772ace2bb6d7144e).

packages/terra-avatar/src/Avatar.jsx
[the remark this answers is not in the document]
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/901237e7a6938c3655ffa65a772ace2bb6d7144e).

packages/terra-avatar/src/Avatar.jsx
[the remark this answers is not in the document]
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/901237e7a6938c3655ffa65a772ace2bb6d7144e).

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * The text content of the aria-label for accessibility.
+   */
+  ariaLabel: PropTypes.string.isRequired,
+  /**
+   * The image to display.
+   */
+  image: PropTypes.string,
+  /**
+   * The initials to display. Can be two or three characters, otherwise a fallback icon displays.
```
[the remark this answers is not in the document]
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/901237e7a6938c3655ffa65a772ace2bb6d7144e).

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,97 @@
...
+   */
+  initials: PropTypes.string,
+  /**
+   * The avatar variant. One of `AvatarVariants.FACILITY`, `AvatarVariants.USER`.
+   */
+  variant: PropTypes.oneOf([AvatarVariants.FACILITY, AvatarVariants.USER]),
```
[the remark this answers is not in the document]
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/901237e7a6938c3655ffa65a772ace2bb6d7144e).

packages/terra-avatar/src/Avatar.jsx
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/901237e7a6938c3655ffa65a772ace2bb6d7144e).

packages/terra-avatar/tests/jest/Avatar.test.jsx
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/901237e7a6938c3655ffa65a772ace2bb6d7144e).

packages/terra-avatar/docs/README.md
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/63281dd2f99a1f829acc97efee1548a149368b78).

packages/terra-avatar/package.json
```diff
@@ -0,0 +1,47 @@
...
+  ],
+  "author": "Cerner Corporation",
+  "license": "Apache-2.0",
+  "bugs": {
+    "url": "https://github.com/cerner/terra-core/issues"
+  },
+  "homepage": "https://github.com/cerner/terra-core#readme",
+  "peerDependencies": {
+    "react": "^16.2.0",
+    "react-dom": "^16.2.0",
```
emilyrohrbough: remove terra-base and terra-mixins from the peer dependencies.

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,99 @@
...
+
+const propTypes = {
+  /**
+   * The text content that specifies the alternative text for the image.
+   */
+  alt: PropTypes.string,
+  /**
+   * The text content of the aria-label for accessibility.
+   */
+  ariaLabel: PropTypes.string,
```
emilyrohrbough: I am wondering if this should be required

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,99 @@
...
+  const icon = (
+    <div className={cx('avatar-icon')}>
+      <svg className={cx(variant)} />
+    </div>
+  );
+
+  let avatarContent;
+  if (image) {
+    avatarContent = <TerraImage className={cx('avatar-image')} src={image} placeholder={icon} alt={alt} />;
+  } else if (initials && (initials.length === 2 || initials.length === 3)) {
```
emilyrohrbough: This seems misleading, I propose we splice any initials provided to the first three letters, otherwise users may not understanding why their prop is not honored.

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,99 @@
...
+  /**
+   * The text content of the aria-label for accessibility.
+   */
+  ariaLabel: PropTypes.string,
+  /**
+   * The image to display.
+   */
+  image: PropTypes.string,
+  /**
+   * Two or three letters to display.
```
emilyrohrbough: Below there is a warning saying only image or only initials should be supplied. The docs should be updated to indicate this and mention image will be honors over initials.

packages/terra-avatar/src/Avatar.scss
```diff
@@ -0,0 +1,54 @@
...
+      background-repeat: no-repeat;
+      background-size: contain;
+      height: 100%;
+      width: 100%;
+    }
+  }
+  
+  .avatar-image {
```
emilyrohrbough: Look like this style is listed 2x

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,97 @@
...
+  ]);
+
+  const avatarTextClassNames = cx([
+    { 'avatar-text-small': initials !== undefined && initials.length === 3 },
+    { 'avatar-text-large': initials !== undefined && initials.length === 2 },
+  ]);
+
+  const icon = (
+    <div className={cx('avatar-icon')}>
+      <svg className={cx(variant)} />
```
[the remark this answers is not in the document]
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/18d816503ca37ccd585667ee835ecd2b04377d96).

packages/terra-avatar/src/Avatar.scss
ChaseSinclair: Updated in this [commit](https://github.com/cerner/terra-core/commit/18d816503ca37ccd585667ee835ecd2b04377d96).

packages/terra-avatar/src/Avatar.jsx
JakeLaCombe: Brett and I talked. We decided to leave it as an optional string.

ChaseSinclair: There's been some discussion around this already. @JakeLaCombe and @bjankord decided it was best to default it to undefined rather than make it required.

packages/terra-avatar/src/Avatar.scss
emilyrohrbough: maybe not. :)

packages/terra-avatar/src/Avatar.jsx
JakeLaCombe: Essentially, it's not always easy to provide a value for this attribute, and may not always have a desire to be accessible. Thus, to ensure we can still keep it as accessible when it needs to be, we can make this optional.

packages/terra-avatar/package.json
dylan-sftwr: Can you explain why these need to be removed? Looking quickly at the package.json's for [terra-arrange](https://github.com/cerner/terra-core/blob/master/packages/terra-arrange/package.json#L37), [terra-form-textarea](https://github.com/cerner/terra-core/blob/master/packages/terra-form-textarea/package.json#L29), and [terra-button](https://github.com/cerner/terra-core/blob/master/packages/terra-button/package.json#L37) all seem to show terra-base as a dependency?

packages/terra-avatar/src/Avatar.jsx
dylan-sftwr: @JakeLaCombe, if that's the case, is there a reason that we don't just let it get spread onto the `div` below via `customProps`? I know that react accepts the first-class prop `ariaLabel` in general for it's components, so I'm not sure if it's a matter of using `aria-label` vs`ariaLabel` being an issue, but if we aren't using Avatar's `ariaLabel` prop, is there a reason to explicitly call it out here?

packages/terra-avatar/src/Avatar.jsx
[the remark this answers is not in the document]
bjankord: Here is the output from the war room discussion on Friday for this topic:

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * The text content that specifies the alternative text for the image.
+   */
+  alt: PropTypes.string,
+  /**
+   * The text content of the aria-label for accessibility.
+   */
+  ariaLabel: PropTypes.string.isRequired,
+  /**
+   * The image to display.
```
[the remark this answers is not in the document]
bjankord: `imageSrc` could be a bit more descriptive prop name for this.

packages/terra-avatar/src/Avatar.jsx
bjankord: If we only take the first 3 letters of string, we should update the `initials` prop description from `Two or three letters to display.` to describe this behavior.

StephenEsser: For the case of initials I think splicing may be a little deceiving. Removing the last character(s) representing a last name or facility name could make the initials misleading and potentially un-mappable if there is another valid set of initials that are a subset of the spliced initials.

I can see the confusion of not showing any initials as well, so I guess the question becomes is an incorrect / incomplete set of initials more beneficial than the fallback icon? Is this a situation where we should manipulate user data rather than log a warning and fall back to the icon?

dylan-sftwr: I prefer the fallback approach. I don't like the idea of allowing components to be so opinionated that they modify data. The purpose of the fallback icon is to, IMO, is not only to provide a _default_ display for a user, facility, etc. (in the future), but to provide us a safe escape route for a consumer improperly using the component

dm123455: I agree with Dylan, it is better to fallback on the icon incase the required initials are not passed in. Lets go with that.

bjankord: Yeah, I agree with the consensus that we should keep the fallback behavior and not add the splicing behavior.

packages/terra-avatar/src/Avatar.jsx
bjankord: I agree with @JakeLaCombe on recommending the first classing prop, as it sticks out more to the consuming teams.

packages/terra-avatar/package.json
bjankord: Let's keep this as is, it aligns with how we handle this in other terra packages that @dylan-sftwr linked to.

neilpfeiffer:
Tested: PR #1307
Resolves: Issue #1279

\
**Reviewed:**
- N/A: Pattern considerations documented in Cerner Design Standards: Avatar (not yet posted)
- Follows UX defined design implementation for Terra-Avatar

\
_Also getting secondary verification by @scottwilmarth_

**Pass** |
-----------|
UX Review|

packages/terra-avatar/src/Avatar.jsx
emilyrohrbough: +1 to fallback. I think this fallback strategy should be outline in the props above to clarify the prop priority

juliezajic:
Functional verification complete. Avatar examples display and function as expected.

PR #1326, opened 2018-03-08
Export AvatarVariants constants from Avatar.

tbiethman:
### Summary

The constants defined in AvatarVariants are mentioned through the documentation; however, consumers have no access to them because they are not exported.

Avatar.jsx will now export that object.


packages/terra-avatar/src/Avatar.jsx
```diff
@@ -98,3 +98,4 @@ Avatar.propTypes = propTypes;
 Avatar.defaultProps = defaultProps;
```
emilyrohrbough: Should we be exporting these via the Opts object to remain consistent with other components? See [alert](https://github.com/cerner/terra-core/blob/master/packages/terra-alert/src/Alert.jsx#L205) for an example.

bjankord: Yeah, we should try and be consistent with our other components that expose variants.

bjankord: Discussed in war room. Plan is to start moving to separate exports for additional data or utils components may need.

PR #1329, opened 2018-03-08
Terra Action Footer

### LINKED ISSUE #1286: Action footer component
# Summary
An action footer is a footer bar that contains commonly used actions in the form
of pre-arranged buttons and hyperlinks. It can be used in multiple UI components, such as modal dialog, split views, popovers and cards.

# Expected behavior
Should be able to give the component buttons/hyperlinks and specify their layout.
If no action is passed in, render empty footer with 10px height

# Responsiveness
On large screen size action items are horizontally positioned, while on small size they are stacked

# React Props
| Prop | Type | Default | Description |
------|-------|--------|-------------
| start | node |  | actions to be displayed at the start side
| center | node |  | actions to be displayed in the center
| end | node | | actions to be displayed at the end side

# Themeable styles
background-color
border-top

# Example
```
<ActionFooter
  start={<Button text='cancel'/>}
  end={<div><Button text='save'/> <Button text='next'/></div>}
/>

<ActionFooter center={<a href='google.com'/>} />
```

dev-cprice:
### Summary
The terra-action-footer component is a footer bar that contains sockets to place items in the form of buttons and hyperlinks used to perform actions. The default variant contains a start and end socket, while the centered variant has only a center socket. If no actions are provided, the footer bar defaults to a 10px equivalent height and maintains the top border.

Created Action Footer component. Fixes #1286 

### Deployment Links
[Action Footer](https://terra-core-deployed-pr-1329.herokuapp.com/static/#/site/action-footer)
[Centered Action Footer](https://terra-core-deployed-pr-1329.herokuapp.com/static/#/site/action-footer/centered)


packages/terra-action-footer/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Action Footer
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-action-footer.svg)](https://www.npmjs.org/package/terra-action-footer)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+{insert description}
```
bjankord: Need to add a description here, I'd recommend combining the one from `packages/terra-action-footer/docs/README.md` and `packages/terra-action-footer/docs/CenteredActionFooter.md` into a shortened version.

dev-cprice: Ah.. missed this one. Good catch! 👍

packages/terra-action-footer/src/_ActionFooterContainer.jsx
```diff
@@ -0,0 +1,57 @@
...
+    customProps.className,
+  ]);
+
+  return (isEmpty ?
+    <div
+      {...customProps}
+      className={actionFooterContainerClassNames}
+    />
+    :
+    <Spacer
```
bjankord: Instead of using the `<Spacer />` component here, lets add padding styles to ActionFooterContainer.scss and make those themable with CSS custom properties.

bjankord:
@dev-cprice Looks like the snapshots for action-footer may need updated.

packages/terra-action-footer/docs/README.md
```diff
@@ -0,0 +1,45 @@
...
+```jsx
+import React from 'react';
+import ActionFooter from 'terra-action-footer';
+import Button from 'terra-button';
+
+<ActionFooter
+  start={<a href="/link/to/somewhere">A link</a>}
+  end={
+    <div>
+      <Button text="Left End Button"/>
```
dylan-sftwr: Can we use terms other than `Left` and `Right`?

Won't the order of the buttons, within the `end` socket reverse?

In the component's readme as well

packages/terra-action-footer/src/ActionFooterContainer.scss
```diff
@@ -0,0 +1,21 @@
...
+  .has-children {
+    min-height: 2.857142857142857rem;
+    padding-bottom: var(--terra-action-footer-padding-bottom, $default-padding);
+    padding-left: var(--terra-action-footer-padding-left, $default-padding);
+    padding-right: var(--terra-action-footer-padding-right, $default-padding);
+    padding-top: var(--terra-action-footer-padding-top, $default-padding);
+  }
+
+  .is-empty {
+    height: 0.714285714285714rem;
```
dylan-sftwr: nit: Can we change this to be `0.7142857143rem`. It's still 10px, it just better aligns with the called out naming convention equivalents [here](http://engineering.cerner.com/terra-core/#/site/spacer).

dylan-sftwr: same goes w/ the `default-padding`

packages/terra-action-footer/src/ActionFooterContainer.scss
```diff
@@ -0,0 +1,21 @@
...
+
+:local {
+  .action-footer-container {
+    background-color: var(--terra-action-footer-background-color, transparent);
+    border-top: var(--terra-action-footer-border-top, 0.071428571428571rem solid #cfd0d1);
+    display: flex;
+  }
+
+  .has-children {
+    min-height: 2.857142857142857rem;
```
dylan-sftwr: Rather than breaking up this and the `is-empty` class below, could you do something like
```css
  .action-footer-container {
    background-color: var(--terra-action-footer-background-color, transparent);
    border-top: var(--terra-action-footer-border-top, 0.071428571428571rem solid #cfd0d1);
    display: flex;
    height: 0.714285714285714rem;
    
    &.has-children {
      height: auto;
      min-height: 2.857142857142857rem;
      padding-bottom: var(--terra-action-footer-padding-bottom, $default-padding);
      padding-left: var(--terra-action-footer-padding-left, $default-padding);
      padding-right: var(--terra-action-footer-padding-right, $default-padding);
      padding-top: var(--terra-action-footer-padding-top, $default-padding);
    }
  }
```

That way you only have to worry about 1 boolean assignment of a classname and don't open yourself up to being in a situation where in the future, you have to try to decide where any future styles that could be added to `action-footer-container` belong? They are always owned at a top-level by the `action-footer-container` class, and its just a matter of whether or not it `has-children`?

packages/terra-action-footer/src/ActionFooter.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
dylan-sftwr: 👍 for breaking up the shared logic into a container!

packages/terra-action-footer/src/ActionFooter.jsx
```diff
@@ -0,0 +1,47 @@
...
+   */
+  start: PropTypes.node,
+  /**
+   * Actions to be displayed in the end socket
+   */
+  end: PropTypes.node,
+};
+
+const defaultProps = {
+  start: undefined,
```
dylan-sftwr: nit: probably don't **need** default props here, per [this](https://connect.ucern.com/docs/DOC-687470) i'd probably say that it's closest to a primitive? even though it's a node? I'm fine leaving it though

packages/terra-action-footer/src/_ActionFooterContainer.jsx
```diff
@@ -0,0 +1,48 @@
...
+import styles from './ActionFooterContainer.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  children: PropTypes.node,
+};
+
+const defaultProps = {
+  children: [],
```
dylan-sftwr: Because react handles `children` as an opaque data type, i think we'd be better not to have this defined in `defaultProps`

packages/terra-action-footer/src/_ActionFooterContainer.jsx
```diff
@@ -0,0 +1,48 @@
...
+
+  const actionFooterContainerClassNames = cx([
+    'action-footer-container',
+    { 'is-empty': isEmpty },
+    { 'has-children': !isEmpty },
+    customProps.className,
+  ]);
+
+  return (isEmpty ?
+    <div
```
dylan-sftwr: nit: these could probably be oneliners

packages/terra-action-footer/tests/nightwatch/action-footer/EmptyActionFooter.jsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+import ActionFooter from '../../../lib/ActionFooter';
+
+export default () => (
```
dylan-sftwr: cheating

packages/terra-site/src/examples/action-footer/ActionFooterComponentNavigation.jsx
```diff
@@ -0,0 +1,20 @@
...
+import List from 'terra-list';
+import styles from '../../site.scss';
+
+const propTypes = {
+  onClick: PropTypes.func,
+};
+
+const ActionFooterComponentNavigation = ({
+    onClick,
+  }) =>
```
dylan-sftwr: nit: can we please wrap this in parenthesis?

packages/terra-action-footer/docs/README.md
```diff
@@ -0,0 +1,45 @@
...
+The terra-action-footer component is a footer bar that contains two sockets, start and end, to place items in the form of buttons and hyperlinks used to perform actions. If no actions are provided, the footer bar defaults to a 10px equivalent height and maintains the top border.
+
+## Getting Started
+
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-action-footer`
+
+## Usage
+
+### Action Footer with a Start and End Actions
```
ChaseSinclair: Can you rephrase this to "Action Footer with a Start and End Action" or "Action Footer with Start and End Actions"?

packages/terra-action-footer/docs/CenteredActionFooter.md
```diff
@@ -0,0 +1,39 @@
...
+```jsx
+import React from 'react';
+import CenteredActionFooter from 'terra-action-footer/lib/CenteredActionFooter';
+
+<CenteredActionFooter
+  center={<a href="/link/to/somewhere">A link</a>}
+/>
+```
+
+### Centered Action Footer with mutliple Actions
```
ChaseSinclair: nit: Capitalize 'multiple'

packages/terra-site/src/examples/action-footer/ActionFooterComponentNavigation.jsx
dev-cprice: Interesting.. Not sure how that slipped through 🤷‍♂️

packages/terra-action-footer/src/ActionFooterContainer.scss
ChaseSinclair: Also, would we ever want the height/min-height to be themable?

packages/terra-action-footer/src/ActionFooterContainer.scss
dev-cprice: 👍consistency with spacer sounds good

packages/terra-action-footer/src/_ActionFooterContainer.jsx
dev-cprice: Seems like there's inconsistency on this so far..
Setting children defaultProp to `[]`:
screen shot 2018-03-09 at 2 11 54 pm

Setting children defaultProp to `null`:
screen shot 2018-03-09 at 2 12 46 pm

Setting children defaultProp to `undefined`:
screen shot 2018-03-09 at 2 14 00 pm

packages/terra-action-footer/src/_ActionFooterContainer.jsx
dev-cprice: Yeah.. originally with the spacer the `{...padding} className={actionFooterContainerClassNames}` pushed it past 80cols

The first one for sure can be one line, but the second div with children pushes it to 87 cols.. Could always just rename `actionFooterContainerClassNames` to something shorter.

packages/terra-action-footer/tests/nightwatch/action-footer/EmptyActionFooter.jsx
dev-cprice: 🙅‍♂️

packages/terra-action-footer/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Action Footer
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-action-footer.svg)](https://www.npmjs.org/package/terra-action-footer)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+The terra-action-footer component is a footer bar that contains sockets to place items in the form of buttons and hyperlinks used to perform actions. The default variant contains a start and end socket, while the centered variant has only a center socket. If no actions are provided, the footer bar defaults to a 10px equivalent height and maintains the top border.
```
ChaseSinclair: Would a consumer ever want the default height to _not_ be 10px?

packages/terra-action-footer/src/ActionFooterContainer.scss
dev-cprice: @dylan-sftwr good idea

@ChaseSinclair we would need to ask UX. The callouts didn't specify themeable height/min-height

packages/terra-action-footer/src/ActionFooter.jsx
dev-cprice: Hmm.. am I just missing where the actual resolution on this was stated in your link?

packages/terra-action-footer/src/_ActionFooterContainer.jsx
dylan-sftwr: I think of the above options, `undefined` would be the most appropriate, as I don't think the result from the warroom doc i linked you has had enough time to trickle out into all the components yet, though i know that's the intention going forward. Honestly, i really wouldn't be setting a `defaultProp` on `children`, personally. Because regardless, unless you start doing something with `React.Children.Only()` or `React.Children.Count()` to control the number of incoming children, you truly have no way of _knowing_ what the type `children` is going to be coming in. Hence why it's proptype always is `node`, the most generic (to my knowledge) react type. 

You have 2 situations i see as problematic.
1. You set `defaultProps = { children: [] }`, and say give a single child, then the **actual** type of `children` doesn't align with the default prop value you're supplying, therefore inconsistency.
2. You provide `defaultProps = { children: null }`, nothing **technically** blows up here, but why initialize grouping type items (objects and arrays) to `null`, when you can assign it to an empty version of either of the aforementioned items. 

If you really **want** to supply a `defaultProp` for `children`, i think `undefined` is the most logical. But again, i'd just recommend removing it as that's what it's going to be, if not supplied anyways. And if it is supplied, it's an opaque data type.

dev-cprice: Yeah, I had that kind of thought process when making the component.. went to look for a precedent and there wasn't a consistent one.

packages/terra-action-footer/src/ActionFooter.jsx
dylan-sftwr: Second bullet under 03/02/18 section

packages/terra-site/src/examples/action-footer/ActionFooterComponentNavigation.jsx
dylan-sftwr: hash-tag:who knows

packages/terra-action-footer/src/ActionFooter.jsx
bjankord: We discussed default-props for non-required props some more today and landed on  provide default-props for non-required props. However, there was no consensus on what values to use for default-props for non-required props (null/undefined vs empty array/object). For the non-required props for this, I'm cool with using `undefined` as the default prop values.

packages/terra-action-footer/src/_ActionFooterContainer.jsx
bjankord: Let's set this default-prop value for children to undefined.

packages/terra-action-footer/docs/README.md
mjhenkes: In item view we use the terms 'start' and 'end' for our accessories to avoid this problem.
http://engineering.cerner.com/terra-clinical/#/site/components/clinical-site/item-view/index

packages/terra-action-footer/src/CenteredActionFooter.jsx
```diff
@@ -0,0 +1,43 @@
...
+import ActionFooterContainer from './_ActionFooterContainer';
+import styles from './CenteredActionFooter.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+   * Actions to be displayed in the center socket
+   */
+  center: PropTypes.node,
```
mjhenkes: Should this just be children?

packages/terra-action-footer/tests/nightwatch/action-footer-container/ActionFooterContainerTestRoutes.jsx
```diff
@@ -0,0 +1,18 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
mjhenkes: lets move these examples out of the nightwatch folder and get rid of it all together.

mjhenkes: You could put the examples in:
`package
    examples
       test-examples
`
This will set us up to consume terra-dev-site in the future.
Here's an example: https://github.com/cerner/terra-framework/tree/master/packages/terra-aggregator/examples/test-examples

packages/terra-action-footer/src/CenteredActionFooter.jsx
dylan-sftwr: We had the discussion about it, I think that we decided to go with `center` as a means of keeping the API consistent with using `start` and `end` in the normal `ActionFooter` component, but I could be wrong

dev-cprice: There wasn't necessarily a consensus, but the discussion happened here: https://github.com/cerner/terra-core/issues/1286#issuecomment-370476119

packages/terra-action-footer/tests/nightwatch/action-footer-container/ActionFooterContainerTestRoutes.jsx
dev-cprice: Should I move the examples that I have under `packages/terra-site/src/examples/action-footer/` here under `examples/index-examples`?

mjhenkes: No we will move those later, once we deprecate site. Moving the test examples is an easy win and eliminates mention of nightwatch from the package

packages/terra-action-footer/README.md
dev-cprice: Spoke with UX, we're gonna go ahead and make it themeable, but default it to 10px.

packages/terra-action-footer/src/ActionFooterContainer.scss
dev-cprice: Spoke with UX, we're gonna go ahead and make height/min-height themeable, but default it to 10px/40px respectively.

dev-cprice:
@dylan-sftwr working on refactoring site examples to be better/easier to digest

packages/terra-action-footer/README.md
JakeLaCombe: Can you reword the first sentence? `In the form of` makes it a little hard to read and understand. I suggest doing `The terra-action-footer component is a footer bar that contains sockets for placing actionable items such as buttons and hyperlinks`.

packages/terra-action-footer/tests/jest/CenteredActionFooter.test.jsx
```diff
@@ -0,0 +1,81 @@
...
+              <button>Button 2</button>
+            </div>
+          }
+        />
+      );
+      expect(shallow(component)).toMatchSnapshot();
+    });
+  });
+
+  // Prop Tests
```
JakeLaCombe: Can snapshots be used for these prop tests instead? It looks like it's testing the exact same things the snapshot would test.

dev-cprice: For sure. Is that the best practice that we should follow moving forward?

packages/terra-action-footer/docs/CenteredActionFooter.md
```diff
@@ -0,0 +1,39 @@
+# Terra Centered Action Footer
+
+The terra-centered-action-footer component is a footer bar that contains a single, centered socket for placing actionable items such as buttons and hyperlinks. If no actions are provided, the footer bar collapses to a themeable height (default of 10px) and maintains the top border.
```
StephenEsser:
```
(default of 10px)
```

I would avoid calling out specific values in the read me documentation because it often times gets out of sync. I think the default value has already been converted into rems which will convert to a different px value depending on the themed font-size

packages/terra-action-footer/src/ActionFooter.jsx
```diff
@@ -0,0 +1,47 @@
...
+  return (isEmpty ?
+    <ActionFooterContainer {...customProps} className={cx(customProps.className)} />
+    :
+    <ActionFooterContainer {...customProps} className={cx(customProps.className)}>
+      { start && <div className={cx('socket')}>{start}</div> }
+      <div className={cx('empty-center')} />
+      { end && <div className={cx('socket')}>{end}</div> }
+    </ActionFooterContainer>
+  );
+};
```
StephenEsser: The formatting here is up to your preference. Possible other ways to format these could be:

```
const ActionFooter = ({
  start,
  end,
  ...customProps
}) => {
  if (!start && !end) {
    return <ActionFooterContainer {...customProps} className={cx(customProps.className)} />;
  }

  return (
    <ActionFooterContainer {...customProps} className={cx(customProps.className)}>
      { start && <div className={cx('socket')}>{start}</div> }
      <div className={cx('empty-center')} />
      { end && <div className={cx('socket')}>{end}</div> }
    </ActionFooterContainer>
  );
};
```
or
```
const ActionFooter = ({
  start,
  end,
  ...customProps
}) => (
  <ActionFooterContainer {...customProps} className={cx(customProps.className)}>
    { start && <div className={cx('socket')}>{start}</div> }
    { (start || end) && <div className={cx('empty-center')} />}
    { end && <div className={cx('socket')}>{end}</div> }
  </ActionFooterContainer>
);
```

dylan-sftwr: Single, implicit return, I like the second one 🙌

dev-cprice: I _do_ like the second option. I feel like this could get _real opinionated real quick_ though. Then again, I'm already using a ternary.. 🤔

packages/terra-action-footer/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Action Footer
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-action-footer.svg)](https://www.npmjs.org/package/terra-action-footer)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+The terra-action-footer component is a footer bar that contains sockets for placing actionable items such as buttons and hyperlinks. The default variant contains a start and end socket, while the centered variant has only a center socket. If no actions are provided, the footer bar collapses to a themeable height and maintains the top border.
```
dylan-sftwr: Can we use a word here other than `variant`? Makes me think there is a prop called `variant` :/

dev-cprice: Open to suggestions. `option`? `version`?

dylan-sftwr: [`Variation`?](http://engineering.cerner.com/terra-core/#/site/list)

image

packages/terra-action-footer/src/_ActionFooterContainer.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import 'terra-base/lib/baseStyles';
+import styles from './ActionFooterContainer.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  children: PropTypes.node,
```
dylan-sftwr: Can we mark this as `isRequired`? Doesn't make much sense to have a named `Container` without children, right??

dev-cprice: This is really a wrapper for common styling/behavior between both action footer variations. If there's no actions provided, I still want those styles to apply (empty height of 10px with a top border). No actions === no children

dylan-sftwr: 👍

packages/terra-action-footer/src/_ActionFooterContainer.jsx
```diff
@@ -18,7 +18,7 @@ const ActionFooterContainer = ({
   children,
   ...customProps
 }) => {
-  const isEmpty = !React.Children.count(children);
+  const isEmpty = !React.Children.toArray(children).length;
```
dev-cprice: Interestingly enough, `React.Children.count` _doesn't_ seem to filter out falsey values. Found this when using the terse syntax of the `ActionFooter`.. when no `start` or `end` was given, it was still rendering with the `.with-actions` class..

The fix: use `React.Children.toArray().length` because
>Returns the children opaque data structure as a flat array
See: [reference](https://reactjs.org/docs/react-api.html#reactchildrentoarray)

packages/terra-action-footer/src/_ActionFooterContainer.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+const defaultProps = {
+  children: undefined,
+};
+
+const ActionFooterContainer = ({
+  children,
+  ...customProps
+}) => {
+  const isEmpty = !React.Children.toArray(children).length;
```
dev-cprice: Interestingly enough, `React.Children.count` _doesn't_ seem to filter out falsey values. Found this when using the terse syntax of the `ActionFooter`.. when no `start` or `end` was given, it was still rendering with the `.with-actions` class..

The fix: use `React.Children.toArray().length` because
>Returns the children opaque data structure as a flat array

See: [reference](https://reactjs.org/docs/react-api.html#reactchildrentoarray)

packages/terra-site/src/examples/action-footer/examples/centered-action-footer/MultipleActions.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import CenteredActionFooter from 'terra-action-footer/lib/CenteredActionFooter';
+import Button from 'terra-button';
+import Spacer from 'terra-spacer';
+
+const MultipleActionsExample = () => (
+  <CenteredActionFooter
+    center={
+      <div>
```
neilpfeiffer: UX Edits (1 of 3): 
This layout of multiple actions inline would never be recommended to be used in actual practice, so lets change it to look closer to actual usage: 
- two default (neutral) style buttons, stack vertically with the `isBlock ` 
- text can read "First Action" and "Second Action" and make them
- add inline style to make the containing div have a max-width of 400px

packages/terra-site/src/examples/action-footer/examples/action-footer/MultipleStartEndActions.jsx
```diff
@@ -0,0 +1,20 @@
...
+import ActionFooter from 'terra-action-footer';
+import Button from 'terra-button';
+import Spacer from 'terra-spacer';
+
+const MultipleStartEndActionsExample = () => (
+  <ActionFooter
+    start={<a href="/#/site/action-footer">Start Action</a>}
+    end={
+      <div>
+        <Spacer isInlineBlock paddingRight="medium">
```
neilpfeiffer: UX Edits (2 of 3): 
To more closely follow common button placement, reverse the order so the first button is "Submit" and the second button is "Cancel"

neilpfeiffer:
UX Edits (3 of 3): 
Lets simplify the all the examples on the terra-site to just one set, and rather than doing the standalone example followed by the same example in a card, simpler to just wrap the standalone examples in a div and add dotted-line inline style (we can replace with a future dashed-line component):
```
import React from 'react';
import ActionFooter from 'terra-action-footer';

const BoxedExampleDivStyle = {
  border: "3px dotted #d3d4d5",
};

const EmptyExample = () => (
  <div style={BoxedExampleDivStyle}>
    <ActionFooter />
  </div>
);

export default EmptyExample;
```
So the examples would be:
### action-footer:
Empty.jsx
SingleStart.jsx
SingleEnd.jsx
MultipleActions.jsx (changes noted above in 2of3- switch order of submit and cancel)
### centered-action-footer:
Empty.jsx
SingleAction.jsx
MultipleActions.jsx (changes noted above in 1of3 - block buttons in "mobile-ish" sized container)

packages/terra-site/src/examples/action-footer/examples/centered-action-footer/MultipleActions.jsx
dev-cprice: Just to clarify, without using an inline style for the containing div for these vertically stacked buttons, is this fine?
screen shot 2018-03-13 at 4 55 34 pm

(will refactor to accomodate [your third edit below](https://github.com/cerner/terra-core/pull/1329#issuecomment-372768345)) Or are you wanting them to fill the width of the footer?

packages/terra-action-footer/tests/jest/CenteredActionFooter.test.jsx
JakeLaCombe: That's my recommendation. The snapshot is doing the same stuff here as checking prop values, and it takes care of all the demeter logic it takes to ensure props are the values they should be.

neilpfeiffer:
Tested: PR #1329
Resolves: Issue #1286

\
**Reviewed:**
- N/A: Pattern considerations documented in Cerner Design Standards: Action Footer (not yet posted) – related [Dialogs](https://wiki.ucern.com/x/8PTzXw)
- Follows UX defined design initial implementation for Terra-Section-Header
- N/A: Future enhancement to follow defined theme designs: terra-theme


**Fix with Enhancements:**
- will post future issue to update background-color and border-top-color to match terra default theme

\
Verified by @scottwilmarth and  @neilpfeiffer, Ready for Release after Functional Validation.


**Pass** |
-----------|
UX Review|

PR #1331, opened 2018-03-09
Revert "Update default prop snapshots"

bjankord:
Reverts cerner/terra-core#1330

More info: https://github.com/adriantoine/enzyme-to-json/pull/91#issuecomment-371563680


dev-cprice:
Okay, so I'm _not_ going crazy. Was wondering why snapshots were _still_ causing failed tests. 😮

Issue #1335, opened 2018-03-09
[terra-popup] Popup 'top' arrow is not themeable.


PseudoFaux:
# Issue Description
[PopupArrow.scss:33](https://github.com/cerner/terra-core/blob/master/packages/terra-popup/src/PopupArrow.scss#L33) is not themeable, while the others (left, right, bottom) are all themeable. Adding a variable check like the others would do.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior

![image](https://user-images.githubusercontent.com/7713715/37228594-cbb3e79e-23a6-11e8-9ebc-12c09109797b.png)
The arrow at the top of the popup should be themeable.

## Steps to Reproduce

1. Click on any component using a PopupArrow, the top arrow is not themeable.

## Environment

* Component Version: terra-popup@2.2.0
* Browser Name and Version: All
* Operating System and version (desktop or mobile): All

yuderekyu:
Addressed in #1336.

bjankord:
This has been released in terra-popup@2.3.0 cc/ @PseudoFaux

PR #1336, opened 2018-03-09
Allow 'Up' PopupArrow to be Themeable

PseudoFaux:
### Summary
These changes will allow the PopupArrow to accept the 'fallback syntax' and accept a variable for the 'top' PopupArrow.

Fixes Issue #1335

### DIFF
--- a/packages/terra-popup/CHANGELOG.md
+++ b/packages/terra-popup/CHANGELOG.md
@@ -3,6 +3,7 @@ Changelog
 
 Unreleased
 ----------
+* Allow Popup 'Up' Arrow to be Themeable
 
 2.2.0 - (March 6, 2018)
 ------------------
--- a/packages/terra-popup/CHANGELOG.md
+++ b/packages/terra-popup/CHANGELOG.md
@@ -3,6 +3,7 @@ Changelog
 
 Unreleased
 ----------
+### Changed
 * Allow Popup 'Up' Arrow to be Themeable
 
 2.2.0 - (March 6, 2018)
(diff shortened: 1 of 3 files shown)


ChaseSinclair:
I noticed there aren't any wdio tests set up for testing the theming nor is the mock theme set up for terra-popup. @bjankord Is this something that should be logged as an issue? Taken care of in this PR?

packages/terra-popup/src/PopupArrow.scss
```diff
@@ -30,8 +30,8 @@
       transform: translate3d(-100%, -100%, 0);
 
       &::after {
-        border-bottom-color: #fff;
         border-top-width: 0;
+        border-bottom-color: var(--terra-popup-arrow-inner-color, #fff);
```
ChaseSinclair: This needs to go before 'border-top-width' to avoid a linting error.

packages/terra-popup/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ Changelog
 
 Unreleased
 ----------
+* Allow Popup 'Up' Arrow to be Themeable
```
emilyrohrbough: Add `## Changed`

bjankord:
@ChaseSinclair Looks like we don't have any mock theme set up for popup, though wdio tests are [available here](https://github.com/cerner/terra-core/blob/master/packages/terra-popup/tests/wdio/popup-spec.js). I've logged #1341 to add tests for this addition and update the mock theme.

PR #1343, opened 2018-03-13
Button: refactor icon css selector

yuderekyu:
### Summary 
Creates a flattened css selector for the icon. This decreases specificity, which allows custom icon styles to not be overridden by the default styles.

@cerner/terra

### DIFF
--- a/packages/terra-button/src/Button.jsx
+++ b/packages/terra-button/src/Button.jsx
@@ -207,7 +207,8 @@ class Button extends React.Component {
 
     let buttonIcon = null;
     if (icon) {
-      const cloneIcon = React.cloneElement(icon, { className: cx('icon-svg') });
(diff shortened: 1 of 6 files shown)


packages/terra-button/src/Button.jsx
```diff
@@ -204,7 +204,12 @@ class Button extends React.Component {
     ]);
 
     const buttonText = !isIconOnly && variant !== 'utility' ? <span className={buttonTextClasses}>{text}</span> : null;
-    const buttonIcon = icon ? <span className={iconClasses}>{icon}</span> : null;
+
+    let buttonIcon = null;
+    if (icon) {
+      const cloneIcon = React.cloneElement(icon, { className: cx('icon-svg') });
```
StephenEsser: I would double check but I think this will overwrite any user provided classNames on the icon

bjankord: Yeah, we'll need to merge the classes here.

PR #1346, opened 2018-03-14
[terra-i18n] Export injectIntl for consumers of terra-i18n

tbiethman:
### Summary
The I18nProvider exposes its translated messages to its children through its React context. This in turn causes the child components to write boilerplate context interfacing code to actually get those values.

Something like this: https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-action-header/src/ActionHeader.jsx#L84

However, react-intl includes a higher-order component generator that does that interfacing for us. However, to access it, consumers would have to add a dependency to react-intl, which terra-18n is trying to mask. So I have updated the terra-i18n package to export injectIntl.

This might save us a little trouble when the old context API is deprecated. In the meantime, it will be a nice convenience.

### DIFF
--- a/packages/terra-i18n/docs/README.md
+++ b/packages/terra-i18n/docs/README.md
@@ -55,24 +55,6 @@ Note that the state of the object needs to contain keys as follows for the i18nL
 }
 ```
 
-Children of the `I18nProvider` can use the provided `injectIntl` higher-order component generator to interface with the `I18nProvider` context. The `intlShape` prop type is also available for use in the components' prop type specifications.
-
-```jsx
-import React from 'react';
-import { injectIntl, intlShape } from 'terra-i18n';
-
-const ChildComponent = ({ intl }) => (
-  <p>{intl.formatMessage({ id: 'my.string.id' })}</p>
-);
-
-ChildComponent.propTypes = {
-  intl: intlShape,
-};
-
-export default injectIntl(ChildComponent);
-```
-
-
 ## Component Features
 * [Cross-Browser Support](https://github.com/cerner/terra-core/wiki/Component-Features#cross-browser-support)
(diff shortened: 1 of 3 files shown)


tbiethman:
I'm going to update this PR to also export the `intlShape` from react-intl. We can use that for our propType declarations. It'll be great.

emilyrohrbough:
I am curious about how components like the `action-header` would utilize these exports? This this like it should also be exposed via the base component as well since `base` has `i18n` as a dependency and components like `action-header` have a dependency on `base`, but not on `i18n`.

Follow-up to my initial question:
1. If we **do not** expose this through Base, should components have `intl` or `i18n` as a direct dependency? This means base vs component dependency versions on `i18n` could get out of sync.
2. If we **do** expose this through Base, does this affect using Base without translations?

ryanthemanuel:
Right now to use injectIntl you have to have a dependency on react-intl.  https://github.com/cerner/terra-core/wiki/terra-i18n-Guide#pass-translated-message-as-props Is it better to have a dependency on react-intl or terra-i18n?

tbiethman:
Ah I didn't make the connection to terra-base. It should be exposed from terra-base as well. Simply exporting them won't impact the use of Base without translations. If there aren't any objections, I can update terra-base as well.

packages/terra-i18n/docs/README.md
```diff
@@ -55,6 +55,24 @@ Note that the state of the object needs to contain keys as follows for the i18nL
 }
 ```
 
+Children of the `I18nProvider` can use the provided `injectIntl` higher-order component generator to interface with the `I18nProvider` context. The `intlShape` prop type is also available for use in the components' prop type specifications.
+
+```jsx
```
bjankord: Let's drop this doc and add doc to Base on how to interact with this.

Issue #1351, opened 2018-03-15
Add aria-label for Close (X) icon in Dialog


vaibhavpaliwal:
# Issue Description
Html code sniffer with WCAG2 shows below error about accessibility
![image](https://user-images.githubusercontent.com/8019344/37494661-dad19a5c-2877-11e8-80ed-839c67dc6280.png)

## Issue Type
- [x] Enhancement

## Expected Behavior
Aria-label should be present so there is no accessibility error.

## Steps to Reproduce
- Go to terra site dialog example page and run the html code sniffer [extension](http://squizlabs.github.io/HTML_CodeSniffer/)

## Environment

* Component Version: Dialog Version: 1.2.0
* Browser Name and Version: Chrome Version 65.0.3325.146
* Operating System and version (desktop or mobile): desktop

zplata:
Related to the Close (X) button in Dialog, I'm seeing some errors when the Dialog renders due to the prop `text` being of value `undefined` on that Close `Button` component. I also noticed that issue in the source code of the Dialog component [here](https://github.com/cerner/terra-core/blob/master/packages/terra-dialog/src/Dialog.jsx#L43). 

<img width="1422" alt="screen shot 2018-03-28 at 4 23 11 pm" src="https://user-images.githubusercontent.com/5685058/38057412-fb23237a-32a4-11e8-85f8-67acb9f75290.png">

PR #1352, opened 2018-03-16
add aria-label for close icon in dialog

vaibhavpaliwal:
### Summary
This PR will resolve the minor accessibility issue in dialog reported here https://github.com/cerner/terra-core/issues/1351


packages/terra-dialog/src/Dialog.jsx
```diff
@@ -40,7 +40,11 @@ const Dialog = ({ header, footer, onClose, children, ...customProps }) => {
     customProps.className,
   ]);
 
-  const closeButton = onClose ? <div className={cx('dialog-header-close')}><Button variant="utility" onClick={onClose} isIconOnly icon={<span className={cx('close-icon')} />} /></div> : null;
+  const closeButton = onClose ? (
+    <div className={cx('dialog-header-close')}>
+      <Button variant="utility" onClick={onClose} isIconOnly icon={<span className={cx('close-icon')} aria-label="Close" />} />
```
emilyrohrbough: To add the aria-label close, we would need to add internationalization for this component and translations in each locale.

vaibhavpaliwal: closing this PR for now I thought it would be an easy change but looks like its more involved. Currently we do not have planned capacity for this work :( but I can look it into on side.

@emilyrohrbough @bjankord could you point me to existing components or examples/docs which I can refer to on how to add internationalization and translations in each locale.

emilyrohrbough: [Terra-alert](https://github.com/cerner/terra-core/tree/master/packages/terra-alert) is a good example of a component using translations. In the component, we ensure the [intl context is defined](https://github.com/cerner/terra-core/blob/master/packages/terra-alert/src/Alert.jsx#L98).  Then we use the [intl format message function on the component's message key)[https://github.com/cerner/terra-core/blob/master/packages/terra-alert/src/Alert.jsx#L121] to correctly pull in the translation. Once all translations are formated, we [export the intl context](https://github.com/cerner/terra-core/blob/master/packages/terra-alert/src/Alert.jsx#L204).

As for translation setup, we add translations to locale-named json files that contain key-value pairs where the component-name-spaced message name is the key and the translated message is the value. See terra-alert's [translations](https://github.com/cerner/terra-core/tree/master/packages/terra-alert/translations). Note: translations must be included for all of the [terra-supported locales](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/src/i18nSupportedLocales.js).

yuderekyu: Translations for 'close' have already been done for `terra-application-utility`. We can strip out them out [here](https://github.com/cerner/terra-framework/tree/master/packages/terra-application-utility/translations). fi-FI is currently missing, but will be added soon.

Issue #1349, opened 2018-03-15
[terra-form-fieldset] - There is no matching CSS for the isInline prop


StephenEsser:
# Issue Description

The Fieldset has a first-class prop for allowing it to be set as inline. The `fieldset-inline` class that is applied as a result of this prop has no corresponding CSS that it maps to.

The class is applied: 
https://github.com/cerner/terra-core/blob/master/packages/terra-form-fieldset/src/Fieldset.jsx#L47

But there is no CSS:
https://github.com/cerner/terra-core/blob/master/packages/terra-form-fieldset/src/Fieldset.scss

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior

The isInline prop should apply a class that sets the component to display inline.

## Current Behavior

The isInline prop applies a class that has no associated CSS

## Steps to Reproduce

1. Create a terra-form-fieldset component.
2. Set isInline to be true.
3. Observe the component does not display inline.

## Environment

* Component Version:
* Browser Name and Version:
* Operating System and version (desktop or mobile):

msaikaushik1992:
@StephenEsser Actually the prop allowing it to be set as inline should be deprecated as discussed with @bjankord. The intention of that prop is for terra-form-field component to allowed to be set inline. So we can close this and reopen a new bug to deprecate the inline prop.

bjankord:
The expected behavior is that the fieldset does not provide an inline variation at this time. Looks like we removed the styles but missed removing the prop. I'm fine leaving this issue open and repurposing it for removing the prop. This change will be passive as the `isInline` prop does nothing now.

bjankord:
This is available in terra-form-fieldset@1.1.0

PR #1354, opened 2018-03-19
Removed isInline prop for form-fieldset

msaikaushik1992:
### Summary
Resolves [this](https://github.com/cerner/terra-core/issues/1349).

### Additional Details
Removed isInline prop because field-set no longer needs it thinking of it as a wrapper to the form fields.

### Deployment links
https://terra-core-deployed-pr-1354.herokuapp.com/static/#/site/form-fieldset
https://terra-core-deployed-pr-1354.herokuapp.com/static/#/tests/form-fieldset-tests

### DIFF
--- a/packages/terra-form-fieldset/tests/jest/__snapshots__/FormFieldset.test.jsx.snap
+++ b/packages/terra-form-fieldset/tests/jest/__snapshots__/FormFieldset.test.jsx.snap
@@ -2,7 +2,8 @@
 
 exports[`should render a Fieldset when all the possible props are passed into it 1`] = `
 <fieldset
-  className="fieldset fieldset-inline fieldset-required fieldset-custom"
+  className="fieldset fieldset-required fieldset-custom"
+  isInline={true}
 >
   <legend
     className="legend healtheintent-legend"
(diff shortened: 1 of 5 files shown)


packages/terra-form-fieldset/tests/jest/__snapshots__/FormFieldset.test.jsx.snap
```diff
@@ -2,7 +2,8 @@
 
 exports[`should render a Fieldset when all the possible props are passed into it 1`] = `
 <fieldset
-  className="fieldset fieldset-inline fieldset-required fieldset-custom"
+  className="fieldset fieldset-required fieldset-custom"
+  isInline={true}
```
StephenEsser: Looks like the test for this one will need the prop removed

emilyrohrbough:
This doesn't affect the rendering of components, but does removing this make it non-passive as it will result in an unknown prop being spread to the feildset element?

JakeLaCombe:
I think I'd lean towards doing it as a non passive version. Like you said, jest tests would have to be rewritten most likely to use this change.

PR #1357, opened 2018-03-20
Add compile:prod task to root package.json

bjankord:
### Summary
In the [travis.yml file, it has a before_deploy hook that runs an npm script named compile:prod](https://github.com/cerner/terra-core/blob/master/.travis.yml#L14), but no script was defined.

@cerner/terra

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "clean:obsolete-snapshots": "npm test -- -u",
     "compile": "lerna run compile",
     "compile:heroku": "webpack --config webpack.prod.config --p",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -45,6 +45,7 @@
     "clean:obsolete-snapshots": "npm test -- -u",
     "compile": "lerna run compile",
     "compile:heroku": "webpack --config webpack.prod.config --p",
+    "compile:prod": "webpack --config webpack.prod.config --p",
```
mjhenkes: should we remove the compile heroku script since these are the same now?

PR #1358, opened 2018-03-20
Clean the repo without the confirmation.

mjhenkes:
### Summary
I always forget to confirm to clean. Lets change to have consent indicated by running the command.

@cerner/terra

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   },
   "scripts": {
     "bootstrap": "lerna bootstrap",
-    "clean:all": "lerna clean && node scripts/clean-modules/index.js",
+    "clean:all": "lerna clean --yes && node scripts/clean-modules/index.js",
     "clean:install": "npm run clean:all && npm install",
     "clean:obsolete-snapshots": "npm test -- -u",
     "compile": "lerna run compile",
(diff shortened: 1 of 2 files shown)


bjankord:
Looks like this hit the travis CI time limit :(

bjankord:
We had 1 last nightwatch test that were run in 3 breakpoints, I've switched this to only run in 1 breakpoint. After this, we'll need to find other ways to improve test time. Hopefully this is enough to get test times back under the limit until we can more properly address this issue.

mjhenkes:
Options:

1. Convert nightwatch tests (maybe faster)
2. Move packages to core (share the load)
3. Upgrade to webpack 4 (yay for terra-dev-site)
4. Pay for Travis-CI (unlikely 😢 )

emilyrohrbough:
Not sure wdio is any faster tbh, but at least with option four, to gain unlimited build time is pretty cheap! 

I feel option 3 could be helpful. 

Yesterday I was able to cut out 2 minutes of build time by changing the postInstall compile to use `compile:build` vs `compile`. This removed the lib cleanup for each package- initial install, we don't need this behavior. Was unsure if there would be repercussions for `npm install` any time after the repo setup or a `clean:install` though.

PR #1363, opened 2018-03-22
[wdio-config] Lock maxInstance to 1 for selenium/hub:3:11:0-anatomy

emilyrohrbough:
### Summary
Selenium released [3.11.0-anatomy](https://github.com/SeleniumHQ/docker-selenium/releases/tag/3.11.0-antimony) 03/11/18. 

Locally when using `selenium/hub:3.11.0-anatomy` the ChromeDriver instances were not being closed/cleaned as expected when using `wdio.maxInstances > 1`. We have not seen this error in Travis, because we were toggling `maxInstances` to be 1 with the `process.env.CI`varaible, where as locally `maxInstances`  is set to 5. This sets `maxInstances = 1` until this issue is resolved. 

This is being tracked in terra-toolkit. See https://github.com/cerner/terra-toolkit/issues/57

### DIFF
--- a/wdio.conf.js
+++ b/wdio.conf.js
@@ -19,8 +19,7 @@ const config = {
   baseUrl: `http://${localIP.address()}:${webpackPort}`,
   specs,
 
-  // Travis only has 1 browser instace, set maxInstances to 1 to prevent timeouts
-  maxInstances: process.env.CI ? 1 : wdioConf.config.maxInstances,
+  maxInstances: 1,
   seleniumDocker: {
     enabled: !process.env.TRAVIS,
   },
--- a/wdio.conf.js
+++ b/wdio.conf.js
@@ -19,7 +19,6 @@ const config = {
   baseUrl: `http://${localIP.address()}:${webpackPort}`,
   specs,
 
-  maxInstances: 1,
   seleniumDocker: {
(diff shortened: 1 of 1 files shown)


wdio.conf.js
```diff
@@ -19,8 +19,7 @@ const config = {
   baseUrl: `http://${localIP.address()}:${webpackPort}`,
   specs,
 
-  // Travis only has 1 browser instace, set maxInstances to 1 to prevent timeouts
-  maxInstances: process.env.CI ? 1 : wdioConf.config.maxInstances,
+  maxInstances: 1,
```
mjhenkes: can we put this change in toolkit?

emilyrohrbough: Yes. should i make the change there? & remove maxInstances from our config?

mjhenkes: Yeah, lets do that. If we want to handle max instances again after this bug is fixed, we should handle it from toolkit.

emilyrohrbough:
Per @mjhenkes request, made this change in toolkit https://github.com/cerner/terra-toolkit/pull/71 and will be remove `maxInstance` from the repo configs.

emilyrohrbough:
Note...We will need to wait for toolkit release to tests to pass in travis.

PR #1350, opened 2018-03-15
Enhancing Internationalization Support

emilyrohrbough:
### Summary
Resolves #1327 

This PR: 
* Implements locale fallback pattern in translations and locale loaders. Fallback pattern of try regional locale then base locale then en lastly throw error if none provided. Resolves #854
* Aggregate-translations pre-build tool to aggregate translations and build dynamic intlLoader and translationLoader modules that are configured for the specified locales. This tool is offered as a CLI script and as a setup function. Resolves #1247
* Open up the i18nLoader restrictions such that non-terra-supported locales will load.
* Only loads the specified aggregated-translation locales. Resolves #855

Thanks for contributing to Terra. 
@cerner/terra

PR #1367, opened 2018-03-23
Terra Action Header

### LINKED ISSUE #1315: ActionHeader Port from clinical
# Summary
There exists an action header in the terra-clinical repo already. We would like to push this into terra-core and deprecate the clinical package. To make upgrading easiest, we would like to try to maintain the api for passivity.

# Existing Props
| Prop | Type  |  Required |   Default |  Description 
|------|------|-------------|--------|-------------
| `title` | string | optional | `undefined` | Text to be displayed as the title in the header bar
|`level` | number | optional |  `1` |  Sets the heading level. One of `1`, `2`, `3`, `4`, `5`, `6`. (**ADDED**) 
| `onClose`  | func | optional | `undefined` | Callback function for when the close button is clicked. The back button will not display if this is not set. On small viewports a back button will be displayed instead of a close button when a separate onBack callback is not set.
| `onBack`  | func | optional | `undefined` | Callback function for when the back button is clicked. The back button will not display if this is not set.
| `onMaximize`   | func | optional | `undefined` | Callback function for when the expand button is clicked. The expand button will not display if this is not set or on small viewports. Only the expand button will be rendered if onMaximize and onMinimize are set.
| `onMinimize`   | func | optional | `undefined` | Callback function for when the minimize button is clicked. The minimize button will not display if this is not set or on small viewports. Only the expand button will be rendered if both onMaximize and onMinimize are set.
| `onNext`   | func | optional | `undefined` | Callback function for when the next button is clicked. The previous-next button group will not display if neither this or onPrevious are set.
| `onPrevious`   | func | optional | `undefined` | Callback function for when the previous button is clicked. The previous-next button group will not display if neither this or onNext are set.
| `keepCloseButton`  | bool | optional | `undefined` | A Boolean indicating if close button should be displayed on small viewports when separate onBack callback is not set.
| `children`   | element | optional | `undefined` | Child element to be displayed on the right end of the header. This is intended to be used with the CollapsibleMenuView.

# Changes
* Currently the back button, and the expand/collapse buttons appear to be mutually exclusive as they share the same spot and change based off screen size https://github.com/cerner/terra-clinical/blob/master/packages/terra-clinical-action-header/src/ActionHeader.jsx#L129 . They will no longer be mutually exclusive as the back button, and expand/collapse will now be in 2 different sockets.
* The existing action header uses functions specific to each special button (maximize, minimize, next, previous) to determine if they should be shown. These props will be kept for passivity, but it will be documented that they will be removed in the next major release, and instead the consumer should be passing them into their collapsible menu as children.
* Button styles have been changed to utility 
* @mjhenkes  , we discussed with @neilpfeiffer, and he thought we should put the buttons from legacy function props in their same existing position (between the back button and title) for now.

# How to deprecate the props
Seen here, https://discuss.reactjs.org/t/proposal-add-isdeprecated-to-react-proptypes/4037 something like this would work

```js
import React from 'react';

class Component extends React.Component {
  render() {
    const {oldProp, newProp} = this.props;
    if (process.env.NODE_ENV !== 'production' && typeof oldProp !== 'undefined') {
      console.warn('oldProp is deprecated, use newProp instead');
    }
    ...
  }
}
```
React chose not to add an isDeprecated function for prop-types https://github.com/facebook/react/pull/6357 . Dan Abramov did make a package for this https://github.com/Aweary/react-is-deprecated but its not worth pulling in

# Questions
* Do the buttons' icon need to be themable? I'm assuming atleast some of them do since I see a vertical ... and a horizontal version in different spots. 
* Digging in more now, I see ActionHeader has a dependency on the terra-clinical-header. Is this also a component we want ported over directly (I don't see anything about it in callouts)? I assume we don't want a dependency from core -> clinical. We could probably get most of the functionality we need utilizing arrange, heading, and truncating the text.

ChaseSinclair:
### Summary
The terra-action-header component is a header bar that contains sockets for placing actionable items such as buttons and hyperlinks.

[Deployment Link](https://terra-core-deployed-pr-1367.herokuapp.com/static/#/site/components/site/action-header/index)

Fixes #1315.


packages/terra-action-header/examples/test-examples/BackActionHeader.example.jsx
```diff
@@ -0,0 +1,14 @@
...
+import React from 'react';
+import { IntlProvider } from 'react-intl';
+import ActionHeader from '../../lib/ActionHeader';
+
+import messages from '../../translations/en-US.json';
+
+const locale = 'en-US';
+
+export default () => (
+  <IntlProvider locale={locale} messages={messages}>
```
emilyrohrbough: You should be able to remove this `IntlProvider` wrapper on these examples. Terra-dev-site now wraps each page in Base which provides this for us. Before these pages were standalone, which is why our current examples wrap with `IntlProvider` or `Base`

This also means you can drop the locale and messages form this example

packages/terra-action-header/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "compile:clean": "rimraf lib",
+    "compile:build": "babel src --out-dir lib --copy-files",
+    "lint": "npm run lint:js && npm run lint:scss",
+    "lint:js": "eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
+    "lint:scss": "stylelint src/**/*.scss",
+    "test": "npm run test:jest && npm run test:wdio",
+    "test:jest": "jest ./tests/jest/* --config ../../jestconfig.json",
+    "test:wdio": "wdio ../../wdio.conf.js"
+  },
+  "devDependencies": {
```
emilyrohrbough: We should be able to drop this dependency. See my comment above.

packages/terra-action-header/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Action Header
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-action-header.svg)](https://www.npmjs.org/package/terra-action-header)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+The terra-action-header component is a header bar that contains sockets for placing actionable items such as buttons and hyperlinks.
```
emilyrohrbough: > contains sockets 

This should probably updated. The props do not indicate there are sockets in this component.

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -0,0 +1,148 @@
...
+    backButtonSmall = backButton;
+  }
+
+  const leftButtonsDefault = (
+    <div className={cx('left-buttons')}>
+      {backButton}
+    </div>
+  );
+
+  const rightButtonsDefault = (
```
emilyrohrbough: Lets check if `closeButton` is null, otherwise we are adding an empty 'right-buttons' div

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -0,0 +1,148 @@
...
+      title={title}
+      endContent={rightButtonsDefault}
+      level={level}
+    >
+      {children}
+    </ActionHeaderContainer>
+  );
+
+  const smallActionHeader = (
+    <ActionHeaderContainer
```
emilyrohrbough: Is this an internal component expected to only be used by ActionHeader? If so, lets update the file  name to be `_ ActionHeaderContainer` the `_` indicates it is an internal component.

packages/terra-action-header/README.md
ChaseSinclair: Rephrased [here](https://github.com/cerner/terra-core/commit/b1e09f51f44a1123922961a673cc07c9e2925f33).

packages/terra-action-header/src/ActionHeader.jsx
ChaseSinclair: Updated to check for null on both close and back buttons [here](https://github.com/cerner/terra-core/commit/b1e09f51f44a1123922961a673cc07c9e2925f33).

packages/terra-action-header/src/ActionHeader.jsx
ChaseSinclair: It's internal only. Updated [here](https://github.com/cerner/terra-core/commit/b1e09f51f44a1123922961a673cc07c9e2925f33).

packages/terra-action-header/README.md
```diff
@@ -0,0 +1,25 @@
+# Terra Action Header
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-action-header.svg)](https://www.npmjs.org/package/terra-action-header)
+[![Build Status](https://travis-ci.org/cerner/terra-core.svg?branch=master)](https://travis-ci.org/cerner/terra-core)
+
+The terra-action-header component is a header bar containing a title and optional actionable items.
```
JakeLaCombe: I feel like for the description, we can add a small subset of examples it items

    actionable items, such as links and buttons.

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -0,0 +1,132 @@
...
+   */
+  keepCloseButton: PropTypes.bool,
+  /**
+   * Child element to be displayed on the right end of the header.
+   */
+  children: PropTypes.element,
+};
+
+const defaultProps = {
+  title: '',
```
JakeLaCombe: I think it's better to make the title undefined as opposed to an empty string to default, unless we noted in a tech design or decision document.

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -0,0 +1,132 @@
...
+  /**
+   * Text to be displayed as the title in the header bar.
+   */
+  title: PropTypes.string,
+  /**
+   * Optionally sets the heading level. One of `1`, `2`, `3`, `4`, `5`, `6`. Default `level=1`.
+   */
+  level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
+  /**
+   * Callback function for when the close button is clicked. The back button will not display if this is not set.
```
JakeLaCombe: We probably should clean up the documentation for the onClose, onBack and keepCloseButton features. Let me see what suggestion I can help with, but it feels like you have to do a scavenger hunt to determine if you will see a back button or not

packages/terra-action-header/package.json
ChaseSinclair: Removed [here](https://github.com/cerner/terra-core/commit/899ef844b8d2ed926f0df690f13e3b97417e8590).

packages/terra-action-header/examples/test-examples/BackActionHeader.example.jsx
ChaseSinclair: Removed [here](https://github.com/cerner/terra-core/commit/899ef844b8d2ed926f0df690f13e3b97417e8590).

packages/terra-action-header/README.md
ChaseSinclair: Updated [here](https://github.com/cerner/terra-core/commit/5b230102a59e8b12ddab3baf4971a109535b0884).

packages/terra-action-header/src/ActionHeader.jsx
ChaseSinclair: Updated [here](https://github.com/cerner/terra-core/commit/5b230102a59e8b12ddab3baf4971a109535b0884).

packages/terra-action-header/src/ActionHeader.jsx
ChaseSinclair: Glad you caught this. I removed the text "The back button will not display if this is not set." as it was inaccurate [here](https://github.com/cerner/terra-core/commit/5b230102a59e8b12ddab3baf4971a109535b0884).

packages/terra-action-header/src/ActionHeader.scss
```diff
@@ -0,0 +1,15 @@
+local {
```
msaikaushik1992: Shouldn't this be " :local { " to make the styles local to this file?

packages/terra-action-header/tests/jest/ActionHeader.test.jsx
```diff
@@ -0,0 +1,84 @@
...
+        <ActionHeader title="Action Header" level={3} />
+      </IntlProvider>,
+    );
+    expect(actionHeader).toMatchSnapshot();
+  });
+
+  it('should render an action header with custom button and title', () => {
+    const actionHeader = shallow(
+      <IntlProvider locale={locale} messages={messages}>
+        <ActionHeader title="Action Header" >
```
dylan-sftwr: nit: don't need the space before the closing arrow

packages/terra-action-header/tests/wdio/action-header-spec.js
```diff
@@ -0,0 +1,130 @@
...
+        '--terra-action-header-background-color': '#551A8B',
+        '--terra-action-header-border-bottom': '1px solid #00FF00',
+        '--terra-action-header-border-left': '1px solid #00FF00',
+        '--terra-action-header-border-right': '1px solid #00FF00',
+        '--terra-action-header-border-top': '1px solid #00FF00',
+        '--terra-action-header-box-shadow': '0 1px 0 0 #FF8C00',
+        '--terra-action-header-font-color': '#4682B4',
+        '--terra-action-header-font-family': 'Roboto',
+        '--terra-action-header-font-size': '0.5rem',
+        '--terra-action-header-padding': '0.714rem',
```
dylan-sftwr: how is this working? it looks like we're just setting the padding to the [default value](https://github.com/cerner/terra-core/blob/terra-action-header/packages/terra-action-header/src/ActionHeaderContainer.scss#L11)?

I thought we were supposed to be setting them to random values to see if they took, but I could be wrong

packages/terra-site/src/App.jsx
```diff
@@ -0,0 +1,246 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
dylan-sftwr: do we need this file, Index, or TestLinks?

packages/terra-action-header/src/ActionHeader.scss
ChaseSinclair: These styles weren't being used anyways. Removed [here](https://github.com/cerner/terra-core/commit/728d4f722bd7854c04324e910d74905c72737db1).

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -0,0 +1,115 @@
...
+const defaultProps = {
+  title: undefined,
+  level: 1,
+  onClose: undefined,
+  onBack: undefined,
+  keepCloseButton: false,
+  children: undefined,
+};
+
+const contextTypes = {
```
yuderekyu: Once `terra-base` has been updated to expose `injectIntl`, we should use that instead of context for translations.

packages/terra-action-header/tests/wdio/action-header-spec.js
ChaseSinclair: Good catch. Updated [here](https://github.com/cerner/terra-core/commit/50aed7a03dfe93b7dbe24a445f1395e38303e52a).

packages/terra-action-header/tests/jest/ActionHeader.test.jsx
ChaseSinclair: Fixed [here](https://github.com/cerner/terra-core/commit/50aed7a03dfe93b7dbe24a445f1395e38303e52a).

packages/terra-site/src/App.jsx
ChaseSinclair: These were added by mistake during a rebase. Removed [here](https://github.com/cerner/terra-core/commit/b6399f6704e36e5495678f8832b516bc88328f3e).

packages/terra-action-header/src/ActionHeaderContainer.scss
```diff
@@ -0,0 +1,61 @@
...
+
+  .title-container {
+    height: 100%;
+    left: 0;
+    position: relative;
+    top: 0;
+    width: 100%;
+  }
+
+  .title {
```
yuderekyu: Do we also need line height as a themeable variable here?

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -0,0 +1,115 @@
...
+  ...customProps }, {
+  intl,
+}) => {
+  const attributes = { ...customProps };
+  const backText = intl.formatMessage({ id: 'Terra.actionHeader.back' });
+  const closeText = intl.formatMessage({ id: 'Terra.actionHeader.close' });
+
+  const isCloseButtonAllowedOnSmallViewport = !(onClose && !onBack && !keepCloseButton);
+
+  const closeButton = onClose ? <Button variant="utility" isIconOnly icon={<IconClose />} text={closeText} onClick={onClose} /> : null;
```
yuderekyu: We should use the exported variants for the `variant` prop - `Button.Opts.Variants.UTILITY`

packages/terra-action-header/src/ActionHeaderContainer.scss
```diff
@@ -0,0 +1,62 @@
+:local {
+  .flex-header {
+    background-color: var(--terra-action-header-background-color, #fff);
+    border-bottom: var(--terra-action-header-border-bottom, 1px solid #1c1f21);
```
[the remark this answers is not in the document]
ChaseSinclair: Created a variable for this [here](https://github.com/cerner/terra-core/commit/9d712b84517806e8baa091e2dd1af61871936efc).

packages/terra-action-header/src/ActionHeaderContainer.scss
ChaseSinclair: Allowing the line height to be changed would allow for the consumer to break vertical centering on the title. Do we want to allow that?

packages/terra-action-header/src/ActionHeader.jsx
ChaseSinclair: Updated [here](https://github.com/cerner/terra-core/commit/6acfffda95af640e71096233a71f1ddac51f99f5).

packages/terra-action-header/src/ActionHeader.jsx
ChaseSinclair: Is that a part of #1350?

emilyrohrbough: No this is a PR coming soon by @tbiethman

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -0,0 +1,115 @@
...
+   * Text to be displayed as the title in the header bar.
+   */
+  title: PropTypes.string,
+  /**
+   * Optionally sets the heading level. One of `1`, `2`, `3`, `4`, `5`, `6`. Default `level=1`.
+   */
+  level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
+  /**
+   * Callback function for when the close button is clicked.
+   * On small viewports a back button will be displayed instead of a close button when a separate onBack callback is not set.
```
JakeLaCombe: I think this needs cleanup as well. Instead of saying a back button will be displayed, can we instead say "On small viewports, this will be triggered by a back button if onBack is not set."

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -0,0 +1,115 @@
...
+import IconClose from 'terra-icon/lib/icon/IconClose';
+import ResponsiveElement from 'terra-responsive-element';
+import 'terra-base/lib/baseStyles';
+import ActionHeaderContainer from './_ActionHeaderContainer';
+
+const propTypes = {
+  /**
+   * Text to be displayed as the title in the header bar.
+   */
+  title: PropTypes.string,
```
dm123455: not required?

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -0,0 +1,115 @@
...
+  /**
+   * Optionally sets the heading level. One of `1`, `2`, `3`, `4`, `5`, `6`. Default `level=1`.
+   */
+  level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
```
dm123455: I know we have used prop name as level for section header, but is it not better to name it as headingLevel?

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -0,0 +1,115 @@
...
+  keepCloseButton,
+  children,
+  ...customProps }, {
+  intl,
+}) => {
+  const attributes = { ...customProps };
+  const backText = intl.formatMessage({ id: 'Terra.actionHeader.back' });
+  const closeText = intl.formatMessage({ id: 'Terra.actionHeader.close' });
+
+  const isCloseButtonAllowedOnSmallViewport = !(onClose && !onBack && !keepCloseButton);
```
JakeLaCombe: I have a hardtime following this line. I think this would make it easier

    const isCloseButtonAllowedOnSmallViewport = keepCloseButton || onClose && !onBack

packages/terra-action-header/src/_ActionHeaderContainer.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+  /**
+   * Content to be displayed at the start of the header, placed before the title.
+   */
+  startContent: PropTypes.element,
+
+  /**
+   * Text to be displayed as the title in the header bar.
+   */
+  title: PropTypes.string,
```
dm123455: should be required.

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const closeButton = onClose ? <Button variant={Button.Opts.Variants.UTILITY} isIconOnly icon={<IconClose />} text={closeText} onClick={onClose} /> : null;
+  const backButton = onBack ? <Button variant={Button.Opts.Variants.UTILITY} isIconOnly icon={<IconLeft />} text={backText} onClick={onBack} /> : null;
+
+  const closeButtonSmall = isCloseButtonAllowedOnSmallViewport ? closeButton : null;
+  const backButtonSmall = isCloseButtonAllowedOnSmallViewport ? backButton : <Button variant={Button.Opts.Variants.UTILITY} isIconOnly icon={<IconLeft />} text={backText} onClick={onClose} />;
```
JakeLaCombe: what if backButton is not defined here?

packages/terra-action-header/src/ActionHeader.jsx
bjankord: I agree with @yuderekyu, though the work to update this code to use `injectIntl` can be completed in a subsequent PR.

packages/terra-action-header/src/ActionHeaderContainer.scss
```diff
@@ -0,0 +1,63 @@
...
+    height: 100%;
+    left: 0;
+    position: relative;
+    top: 0;
+    width: 100%;
+  }
+
+  .title {
+    color: var(--terra-action-header-font-color, #1c1f21);
+    font-family: var(--terra-action-header-font-family);
```
bjankord: `font-family` can be deleted. It will be inherited via the terra-base styles which have a themable font-family.

packages/terra-action-header/src/ActionHeaderContainer.scss
```diff
@@ -0,0 +1,63 @@
+$default-border: 1px solid #1c1f21;
+
+:local {
+  .flex-header {
+    background-color: var(--terra-action-header-background-color, #fff);
+    border-bottom: var(--terra-action-header-border-bottom, $default-border);
+    border-left: var(--terra-action-header-border-left, $default-border);
```
bjankord: Setting the left, right, and top border to `1px solid #1c1f21` seems odd to me as it doesn't match the design of the [action header in clinical](http://engineering.cerner.com/terra-ui/#/site/components/clinical/site/action-header/index) or the [header in the dialog](http://engineering.cerner.com/terra-ui/#/site/components/core/site/dialog/index). 

I think these styles should be updated to align more with the dialog header / clinical action header:

```css
background-color: var(--terra-action-header-background-color, #f4f4f4);
border-bottom: var(--terra-action-header-border-bottom, #dedfe0);
border-left: var(--terra-action-header-border-left);
border-right: var(--terra-action-header-border-right);
border-top: var(--terra-action-header-border-top);
```

cc/ @neilpfeiffer

packages/terra-action-header/src/ActionHeaderContainer.scss
bjankord: When the consumer changes the font-size, would they need to change the line-height to maintain vertical centering on the title?

packages/terra-action-header/tests/jest/ActionHeader.test.jsx
```diff
@@ -0,0 +1,84 @@
...
+import Button from 'terra-button';
+import ActionHeader from '../../src/ActionHeader';
+import messages from '../../translations/en-US.json';
+
+const locale = 'en-US';
+
+describe('ActionHeader', () => {
+  // Snapshot Tests
+  it('should render a default action header', () => {
+    const actionHeader = shallow(
```
bjankord: Switch shallow statements to use `render` instead. It provides more meaningful information of the DOM.

packages/terra-action-header/translations/en-US.json
```diff
@@ -0,0 +1,4 @@
+{
```
bjankord: You can pull translations for the other terra supported locales from the terra-clinical-action-header: https://github.com/cerner/terra-clinical/tree/master/packages/terra-clinical-action-header/translations

We should have a translation json file for the following:
* ar.json
* de.json
* en-GB.json
* en-US.json
* en.json
* es-ES.json
* es-US.json
* es.json
* fi-FI.json
* fr-FR.json
* fr.json
* pt-BR.json
* pt.json

packages/terra-action-header/src/ActionHeaderContainer.scss
ChaseSinclair: As currently implemented no, it would stay centered.

packages/terra-action-header/src/ActionHeaderContainer.scss
ChaseSinclair: I set these based off the "Standard Implementation" callouts in Zeplin.

packages/terra-action-header/src/ActionHeader.jsx
ChaseSinclair: It's not required in terra-clinical-action-header.

packages/terra-action-header/src/ActionHeaderContainer.scss
neilpfeiffer: Ahh, I understand the confusion -- @ChaseSinclair the "Standard Implementation" board is only for details of how the component is constructed and how it behaves. That board is not for Style or CSS. If you look at the `Styles: Themeable` & `Styles: Controlled` on that same board - it correctly lists only border-bottom.

What you should have used was details on the Terra-opensource-theme board (which currently is blank and wasn't provided in this case). But those details / colors are what already exist in the terra-clinical component, which @bjankord has included above.

packages/terra-action-header/src/_ActionHeaderContainer.jsx
neilpfeiffer: Title can be optional here.

packages/terra-action-header/src/ActionHeader.jsx
neilpfeiffer: Title can be optional here, since an ActionHeader can exist with just icons/actions in certain places, although it will be commonly used with one.

packages/terra-action-header/src/ActionHeaderContainer.scss
neilpfeiffer: I'll look into this further and see if needs to be added or how to handle.

packages/terra-action-header/src/ActionHeader.jsx
neilpfeiffer: The `level` prop so far has only been used for components that invoke a heading h1-h6 element, so it is currently consistent with all other components thus far as is. Only risk is if level is introduced elsewhere for another concept in the future - then we would want to adjust all others as well .

packages/terra-action-header/src/ActionHeader.jsx
tbiethman: FYI: #1379

packages/terra-action-header/src/ActionHeaderContainer.scss
ChaseSinclair: Updated styles to match terra-clinical-action-header [here](https://github.com/cerner/terra-core/commit/85b35f2f932e64ced0e37458cd33fde2f0633fe7).

packages/terra-action-header/src/ActionHeaderContainer.scss
ChaseSinclair: font-family removed [here](https://github.com/cerner/terra-core/commit/85b35f2f932e64ced0e37458cd33fde2f0633fe7).

packages/terra-action-header/src/ActionHeader.jsx
ChaseSinclair: Description updated [here](https://github.com/cerner/terra-core/commit/85b35f2f932e64ced0e37458cd33fde2f0633fe7#diff-adf8c495af771315ac46e269a49ed8bcR23).

packages/terra-action-header/tests/jest/ActionHeader.test.jsx
ChaseSinclair: Updated [here](https://github.com/cerner/terra-core/commit/85b48f23579d7aebe02de8f192ff8110e6b2727e).

packages/terra-action-header/src/ActionHeader.jsx
ChaseSinclair: Added backButton check [here](https://github.com/cerner/terra-core/commit/85b48f23579d7aebe02de8f192ff8110e6b2727e).

packages/terra-action-header/src/ActionHeader.jsx
ChaseSinclair: Updated [here](https://github.com/cerner/terra-core/commit/85b48f23579d7aebe02de8f192ff8110e6b2727e).

packages/terra-action-header/translations/en-US.json
ChaseSinclair: Added translations [here](https://github.com/cerner/terra-core/commit/da3b29962d2458b4be17f1b392be310150eec267).

dm123455:
Please resolve the conflicts before you merge.

packages/terra-action-header/src/ActionHeader.scss
```diff
@@ -0,0 +1,32 @@
...
+  
+  .header-close-button {
+    margin-left: var(--terra-action-header-margin-left, 1.429rem);
+  }
+  
+  .header-icon {
+    background-repeat: no-repeat;
+    background-size: contain;
+    display: inline-block;
+    margin-bottom: var(--terra-action-header-icon-margin-bottom);
```
hboege: Variables without defaults have been causing us (as consumers) console errors.  Please consider adding defaults.

packages/terra-action-header/src/ActionHeaderContainer.scss
```diff
@@ -0,0 +1,60 @@
...
+    display: flex;
+    flex: 0 0 auto;
+    height: 100%;
+    max-width: 40%;
+    position: relative;
+  }
+
+  .title-container {
+    height: 100%;
+    left: 0;
```
dkasper-was-taken: Left and Top values can be removed since it's no longer using absolute positioning.

packages/terra-action-header/src/ActionHeaderContainer.scss
```diff
@@ -0,0 +1,60 @@
...
+    flex: 1 1 auto;
+    height: 100%;
+    position: relative;
+    width: 100%;
+  }
+
+  .flex-end {
+    align-items: center;
+    display: flex;
+    flex: 0 0 auto;
```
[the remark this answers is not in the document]
ChaseSinclair: I think it was actually the .flex-collapse rule that needed updating. I also added brackets when using multiple classes with cx [here](https://github.com/cerner/terra-core/commit/5183093e2ead29ad95062eb5251e6dd714f5e6a9) since this could cause issues with older browsers.

packages/terra-action-header/src/_ActionHeaderContainer.jsx
```diff
@@ -0,0 +1,67 @@
...
+  endContent: undefined,
+};
+
+const ActionHeaderContainer = ({ children, title, startContent, endContent, level, ...customProps }) => {
+  const HeaderElement = `h${level}`;
+
+  return (
+    <header {...customProps} className={cx(['flex-header', customProps.className])}>
+      {startContent && <div className={cx('flex-end')}>{startContent}</div>}
+      <div className={cx('flex-fill')}>
```
emilyrohrbough: do we want to check title before creating the 'flex-fill' div?

packages/terra-action-header/src/_ActionHeaderContainer.jsx
```diff
@@ -0,0 +1,67 @@
...
+        {title &&
+          <div className={cx('title-container')}>
+            <HeaderElement className={cx('title')}>
+              {title}
+            </HeaderElement>
+          </div>
+        }
+      </div>
+      {children && React.cloneElement(children, { className: cx(['flex-collapse', children.props.className]) })}
```
emilyrohrbough: Does this work if multiple children are provided? Thinking we should map these with [React.Children](https://reactjs.org/docs/react-api.html#reactchildrenmap)

packages/terra-action-header/tests/jest/ActionHeader.test.jsx
```diff
@@ -0,0 +1,84 @@
...
+import ActionHeader from '../../src/ActionHeader';
+import messages from '../../translations/en-US.json';
+
+const locale = 'en-US';
+
+describe('ActionHeader', () => {
+  // Snapshot Tests
+  it('should render a default action header', () => {
+    const actionHeader = render(
+      <IntlProvider locale={locale} messages={messages}>
```
emilyrohrbough: We should drop the `IntlProvider` from these tests and test `ActionHeader` as itself. We can add intl context with [these setup configs](https://github.com/cerner/terra-core/blob/master/packages/terra-tabs/tests/jest/intl-context-setup.js), and we can provide the intl context [like this](https://github.com/cerner/terra-core/blob/master/packages/terra-tabs/tests/jest/Tabs.test.jsx#L8)

packages/terra-action-header/tests/jest/__snapshots__/ActionHeader.test.jsx.snap
```diff
@@ -0,0 +1,15 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`ActionHeader should render a default action header 1`] = `<div />`;
```
emilyrohrbough: These are not meaningful snapashots, likely because these are the IntlProvider divs

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -0,0 +1,117 @@
...
+  const backText = intl.formatMessage({ id: 'Terra.actionHeader.back' });
+  const closeText = intl.formatMessage({ id: 'Terra.actionHeader.close' });
+
+  const isCloseButtonAllowedOnSmallViewport = keepCloseButton || (onClose && !onBack);
+
+  const closeButton = onClose ? <Button className={cx('header-close-button')} isIconOnly icon={<span className={cx(['header-icon', 'close'])} />} text={closeText} onClick={onClose} /> : null;
+  const backButton = onBack ? <Button className={cx('header-back-button')} isIconOnly icon={<span className={cx(['header-icon', 'back'])} />} text={backText} onClick={onBack} /> : null;
+
+  const closeButtonSmall = isCloseButtonAllowedOnSmallViewport ? closeButton : null;
+  const backButtonSmall = isCloseButtonAllowedOnSmallViewport && backButton ? backButton : <Button className={cx(['header-back-button'])} isIconOnly icon={<span className={cx('header-icon', 'back')} />} text={backText} onClick={onClose} />;
```
JakeLaCombe: Can we leave a comment here detailing  that we need a default button if the user doesn't specify one? That'll make this code easier to maintain.

packages/terra-action-header/src/ActionHeaderContainer.scss
```diff
@@ -0,0 +1,58 @@
...
+
+  .title-container {
+    height: 100%;
+    position: relative;
+    width: 100%;
+  }
+
+  .title {
+    color: var(--terra-action-header-font-color, #1c1f21);
+    font-size: var(--terra-action-header-font-size, 1.286rem);
```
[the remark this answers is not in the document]
JakeLaCombe: Same for the background if they can theme the foreground. Otherwise, the foreground's themeability is limited.

packages/terra-action-header/tests/jest/ActionHeader.test.jsx
JakeLaCombe: Is that something we should consider moving to terra-toolkit? Seems like multiple tests can benefit from that.

JakeLaCombe: Another option potentially is to try shallow rendering. It looks like this component does it with proper messages. https://github.com/cerner/terra-core/blob/master/packages/terra-demographics-banner/tests/jest/DemographicsBanner.test.jsx

packages/terra-action-header/src/ActionHeader.scss
```diff
@@ -0,0 +1,30 @@
...
+    background-repeat: no-repeat;
+    background-size: contain;
+    display: inline-block;
+    position: relative;
+    top: var(--terra-action-header-icon-top, 0.1429rem);
+    
+    /* Disabling so that single quotes can be used in svg string */
+    /* stylelint-disable string-quotes */
+    &.back {
+      background: var(--terra-action-header-back-icon, inline-svg("<svg class='Icon__icon___XA18N Icon__is-bidi___2qF4d Button__icon-svg___D8gRQ' viewBox='0 0 48 48' xmlns='http://www.w3.org/2000/svg' data-name='Layer 1' aria-hidden='true' focusable='false'><path d='M48 21H10.6L27.3 3.9 23.5.1 0 24l23.5 23.9 3.8-3.8L10.6 27H48z'></path></svg>"));
```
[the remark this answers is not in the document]
bjankord: Let's remove those classes.

packages/terra-action-header/src/ActionHeaderContainer.scss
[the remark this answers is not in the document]
JakeLaCombe: I talked to UX about this. Let's remove the color style for now.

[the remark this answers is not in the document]
ChaseSinclair: Yes, font size should be themeable. The background is themeable on the div containing this heading.

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -0,0 +1,117 @@
...
+  title,
+  level,
+  onClose,
+  onBack,
+  keepCloseButton,
+  children,
+  ...customProps }, {
+  intl,
+}) => {
+  const attributes = { ...customProps };
```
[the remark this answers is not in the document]
ChaseSinclair: Updated [here](https://github.com/cerner/terra-core/commit/444fde0838557ce58e76c9faee9b854a6de8d62d).

packages/terra-action-header/src/ActionHeader.scss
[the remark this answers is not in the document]
ChaseSinclair: Updated and classes removed [here](https://github.com/cerner/terra-core/commit/444fde0838557ce58e76c9faee9b854a6de8d62d).

packages/terra-action-header/src/_ActionHeaderContainer.jsx
ChaseSinclair: Added a check for title before the 'flex-fill' div [here](https://github.com/cerner/terra-core/commit/444fde0838557ce58e76c9faee9b854a6de8d62d).

packages/terra-action-header/src/_ActionHeaderContainer.jsx
ChaseSinclair: Updated to use React.Children.map() [here](https://github.com/cerner/terra-core/commit/444fde0838557ce58e76c9faee9b854a6de8d62d).

packages/terra-action-header/tests/jest/ActionHeader.test.jsx
```diff
@@ -0,0 +1,84 @@
...
+  it('should render an action header with level three header element and title', () => {
+    const actionHeader = render(
+      <IntlProvider locale={locale} messages={messages}>
+        <ActionHeader title="Action Header" level={3} />
+      </IntlProvider>,
+    );
+    expect(actionHeader).toMatchSnapshot();
+  });
+
+  it('should render an action header with custom button and title', () => {
```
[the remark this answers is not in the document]
ChaseSinclair: Added example with multiple custom buttons as children [here](https://github.com/cerner/terra-core/commit/444fde0838557ce58e76c9faee9b854a6de8d62d).

packages/terra-action-header/tests/jest/ActionHeader.test.jsx
ChaseSinclair: Updated [here](https://github.com/cerner/terra-core/commit/444fde0838557ce58e76c9faee9b854a6de8d62d).

packages/terra-action-header/tests/jest/__snapshots__/ActionHeader.test.jsx.snap
ChaseSinclair: The snapshots were updated in this [commit](https://github.com/cerner/terra-core/commit/444fde0838557ce58e76c9faee9b854a6de8d62d).

packages/terra-action-header/src/ActionHeaderContainer.scss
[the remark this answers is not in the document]
ChaseSinclair: Removed color style [here](https://github.com/cerner/terra-core/commit/84f6e551794dc374b8f45fe1a2c9ab729ebd940f).

packages/terra-action-header/tests/wdio/action-header-spec.js
```diff
@@ -0,0 +1,130 @@
...
+    Terra.should.themeEachCustomProperty(
+      {
+        '--terra-action-header-background-color': '#551A8B',
+        '--terra-action-header-border-bottom': '1px solid #00FF00',
+        '--terra-action-header-border-left': '1px solid #00FF00',
+        '--terra-action-header-border-right': '1px solid #00FF00',
+        '--terra-action-header-border-top': '1px solid #00FF00',
+        '--terra-action-header-box-shadow': '0 1px 0 0 #FF8C00',
+        '--terra-action-header-font-color': '#4682B4',
+        '--terra-action-header-font-family': 'Roboto',
```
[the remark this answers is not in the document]
ChaseSinclair: Removed [here](https://github.com/cerner/terra-core/commit/84f6e551794dc374b8f45fe1a2c9ab729ebd940f).

packages/terra-action-header/tests/wdio/action-header-spec.js
```diff
@@ -0,0 +1,130 @@
...
+      },
+    );
+  });
+
+  describe('Back', () => {
+    beforeEach(() => browser.url('/#/raw/tests/action-header/back-action-header'));
+
+    Terra.should.beAccessible({ viewports });
+    Terra.should.matchScreenshot({ viewports });
+    Terra.should.themeEachCustomProperty(
```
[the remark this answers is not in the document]
ChaseSinclair: Updated [here](https://github.com/cerner/terra-core/commit/84f6e551794dc374b8f45fe1a2c9ab729ebd940f).

packages/terra-action-header/tests/wdio/action-header-spec.js
```diff
@@ -114,16 +79,11 @@ describe('ActionHeader', () => {
...
-        '--terra-action-header-border-bottom': '1px solid #00FF00',
-        '--terra-action-header-border-left': '1px solid #00FF00',
-        '--terra-action-header-border-right': '1px solid #00FF00',
-        '--terra-action-header-border-top': '1px solid #00FF00',
-        '--terra-action-header-box-shadow': '0 1px 0 0 #FF8C00',
-        '--terra-action-header-font-color': '#4682B4',
-        '--terra-action-header-font-family': 'Roboto',
-        '--terra-action-header-font-size': '0.5rem',
-        '--terra-action-header-padding': '1rem',
+        '--terra-action-header-margin-left': '2rem',
```
emilyrohrbough: These should be removed. They were tested above.

packages/terra-action-header/tests/wdio/action-header-spec.js
```diff
@@ -72,16 +58,9 @@ describe('ActionHeader', () => {
...
```
emilyrohrbough: These can be removed. These were tested above.

packages/terra-action-header/tests/wdio/action-header-spec.js
```diff
@@ -51,16 +42,11 @@ describe('ActionHeader', () => {
...
+        '--terra-action-header-margin-right': '2rem',
+        '--terra-action-header-icon-top': '0.2rem',
```
emilyrohrbough: Remove ` '--terra-action-header-icon-top': '0.2rem',` and ` '--terra-action-header-back-icon': '#551A8B',` these were tested in the `Back` test step.

packages/terra-action-header/tests/wdio/action-header-spec.js
```diff
@@ -14,9 +14,7 @@ describe('ActionHeader', () => {
         '--terra-action-header-border-left': '1px solid #00FF00',
         '--terra-action-header-border-right': '1px solid #00FF00',
         '--terra-action-header-border-top': '1px solid #00FF00',
-        '--terra-action-header-box-shadow': '0 1px 0 0 #FF8C00',
-        '--terra-action-header-font-color': '#4682B4',
-        '--terra-action-header-font-family': 'Roboto',
+        '--terra-action-header-box-shadow': '5px 5px 5px #551A8B',
```
emilyrohrbough: This screenshot should have been updated but I doesn't look like it took

themeable-variables.json
```diff
@@ -20,8 +20,8 @@
     "--terra-action-header-background-color": "#f4f4f4",
     "--terra-action-header-border-bottom": "1px solid #dedfe0",
     "--terra-action-header-padding": "0.714rem",
-    "--terra-action-header-font-color": "#1c1f21",
-    "--terra-action-header-font-size": "1.286rem"
+    "--terra-action-header-font-size": "1.286rem",
+    "--terra-action-header-box-shadow": "5px 5px 5px #551a8b"
```
emilyrohrbough: @bjankord @JakeLaCombe @mjhenkes @ryanthemanuel Looks like we have a bug in the theming-plugin. This new value `"--terra-action-header-box-shadow": "5px 5px 5px #551a8b"` was set in the wdio test and is not the default value.

packages/terra-action-header/tests/wdio/action-header-spec.js
ChaseSinclair: Removed [here](https://github.com/cerner/terra-core/commit/98ed32090a91a4cc4aeb54e121d94d0aa060c2c4).

packages/terra-action-header/tests/wdio/action-header-spec.js
ChaseSinclair: Removed [here](https://github.com/cerner/terra-core/commit/98ed32090a91a4cc4aeb54e121d94d0aa060c2c4).

packages/terra-action-header/tests/wdio/action-header-spec.js
ChaseSinclair: Removed [here](https://github.com/cerner/terra-core/commit/98ed32090a91a4cc4aeb54e121d94d0aa060c2c4).

packages/terra-action-header/tests/wdio/action-header-spec.js
ChaseSinclair: Discussed this with @bjankord, this might be a bug. I think we are going to move forward without the fix for now.

packages/terra-action-header/tests/jest/ActionHeader.test.jsx
```diff
@@ -0,0 +1,84 @@
...
+    );
+    expect(actionHeader).toMatchSnapshot();
+  });
+});
+
+it('throws error on missing locale prop in Base', () => {
+  try {
+    render(<ActionHeader />);
+  } catch (e) {
+    expect(e.message).toContain('add locale prop to Base component');
```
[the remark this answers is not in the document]
bjankord: I'm fine leaving this as is as it matches how we test this in other components. Though, I also agree we should update this type of test across all places where are already doing this. Logged https://github.com/cerner/terra-core/issues/1393 to follow up on this.

packages/terra-action-header/tests/wdio/action-header-spec.js
emilyrohrbough: Out of curiosity, are you able to apply a box-shadow and see the changes locally? If the box shadow is being applied to the outside of the header component itself, its likely that the screenshot test is not capturing it because the box shadow is outside the element itself. One thing to try would be applying an inset-box shadow.

mjhenkes:
I propose that we add back the following functions props for passivity:
onMaximize
onMinimize
onNext
onNext
I know there has been discussion of where they would go. I would recommend we keep them in the same position the were in in the old action header for passivity (and maybe mark that they may be depreciated)

Secondly, I'd like to remove the keepCloseButton prop and have that behavior by default. No switching between the back arrow and the close button.

packages/terra-action-header/tests/jest/ActionHeader.test.jsx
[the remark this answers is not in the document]
JakeLaCombe: Personally, it feels weird that we are testing a debug message inside our code. I'm fine with testing that an exception was raised, but I think testing that console.warn was called feels like a code smell.

bjankord:
@mmalaker @neilpfeiffer This is ready for functional and UX validation.

mmalaker:
Functional verification completed. The action header examples display and function correctly.

neilpfeiffer:
Reviewing UX now

packages/terra-action-header/src/ActionHeaderContainer.scss
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  .title {
+    font-size: var(--terra-action-header-font-size, 1.286rem);
+    font-weight: normal;
+    height: 100%;
+    margin-bottom: 0;
+    margin-top: 0;
+    width: 100%;
+    word-break: normal;
```
neilpfeiffer: Found an issue when a really long word with no spaces/breaks is used for the title, it will overflow the layout. Possible with translation when action-header is used in a small container and also includes a collapsible-menu. 

Recommend changing to `word-break: break-all;` and also adding:
```
-webkit-hyphens: auto;
    -ms-hyphens: auto;
        hyphens: auto;
```

packages/terra-action-header/src/ActionHeaderContainer.scss
```diff
@@ -0,0 +1,57 @@
...
+    border-bottom: var(--terra-action-header-border-bottom, 1px solid #dedfe0);
+    border-left: var(--terra-action-header-border-left);
+    border-right: var(--terra-action-header-border-right);
+    border-top: var(--terra-action-header-border-top);
+    box-shadow: var(--terra-action-header-box-shadow);
+    display: flex;
+    justify-content: space-between;
+    padding: var(--terra-action-header-padding, 0.714rem);
+    position: relative;
+    width: 100%;
```
neilpfeiffer: Found that the action-header height is not consistent without buttons or menu (also when completely empty).  

- add themeable miniumum height on **.flex-header**: 
`min-height: var(--terra-action-header-min-height, 3.6428571429rem);`

- add themeable padding-top or margin-top to **.title**: 
`margin-top: var(--terra-action-header-title-margin-top, 0.1428571429rem);`

neilpfeiffer:
Tested #1367
Resolves #1315

\
**Reviewed:**
- verified port of Terra-Clinical-Action-Header from [terra-clinical](https://github.com/cerner/terra-clinical) into [terra-core](https://github.com/cerner/terra-core)
- follows defined theme designs: terra-theme

\
**Left for Future Enhancements:**
- uplifting to Cerner universal design with utility style buttons and new menu
- replacing internal “layout” with new reusable component for all common headers

\
\
Verified by @neilpfeiffer, Ready for Release after updating to include changes in previous 2 comments.


**Pass** |
-----------|
UX Review|

PR #1369, opened 2018-03-23
Terra Tag examples fix

### LINKED ISSUE #1368: Terra-Tag example update
# Issue Description
Update Terra-tag examples documentation page to separate interactive vs. non-interactive versions of tags when no onClick or href is provided. 

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Should look something like this:
image

neilpfeiffer:
Changed Tag examples and added new section for Fallback Styles with
alert message for non-interactive tags. 

Resolves #1368


packages/terra-site/src/examples/tag/TagFallbacks.jsx
```diff
@@ -0,0 +1,26 @@
...
+import IconTag from 'terra-icon/lib/icon/IconTag';
+import Tag from 'terra-tag';
+import Alert from 'terra-alert';
+
+const alertTitle = 'Not Reccomended Usage.';
+const alertText = 'These styles are provided for when this compoment is missing recomended elements for best practice use.';
+
+const TagHeadingMarginStyle = {
+  marginTop: '200px',
+};
```
StephenEsser: Go big or go home. This does feel like a lot of space though.

yuderekyu: I agree, that's a ton of whitespace. What was the reasoning behind adding this?

PR #1379, opened 2018-03-28
Exporting terra-i18n helpers from terra-base

tbiethman:
### Summary
With #1346, terra-i18n was updated to export injectIntl and intlShape for consumer convenience. Since most consumption of terra-i18n happens through terra-base, we need to export injectIntl/intlShape from Base so that consumers of Base can access them.


packages/terra-base/docs/README.md
```diff
@@ -20,6 +20,41 @@ import Base from 'terra-base';
 </Base>
 ```
 
+## I18n
+
+The `locale` prop provided to `Base` will be used to loaded the appropriate translation messages. `Base` additionally wraps its children with an `I18nProvider` component (as provided by `terra-i18n`) that exposes the loaded messages to its children through its context.
```
yuderekyu: *load the appropriate translation messages.

bjankord: Updated: https://github.com/cerner/terra-core/pull/1379/commits/68616339486b3447a8a0f9e297f2f6cde08eebad

PR #1386, opened 2018-03-29
Clean up terra-mixin dependency

### LINKED ISSUE #1361: Remove cerner-consumer-theme from Base
# Issue Description
This behavior should be supplied by the consumer theme, not by the base component. 
See https://github.com/cerner/terra-core/blob/master/packages/terra-base/src/_themes.scss

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

bjankord:
### Summary
Removed unused mixin package and theme mixins from various modules.
Also resolves #1361


packages/terra-avatar/src/Avatar.scss
```diff
@@ -1,3 +1,5 @@
+@import '~terra-mixins';
```
emilyrohrbough: Why did we re-add this?

packages/terra-badge/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Removed
```
JakeLaCombe: Is this change version correct?

packages/terra-avatar/src/Avatar.scss
bjankord: Avatar.scss uses a sass function named `inline-svg` that is provided in terra-mixins.

packages/terra-badge/CHANGELOG.md
bjankord: I believe so. The terra-mixins package has been removed from terra-badge as it is not used.

Issue #1244, opened 2018-02-13
[Table] - HeaderCell onClick prop is overwritten


StephenEsser:
# Issue Description

Consumers wanting to implement sorting need to be able to pass an onClick to a Table's HeaderCell. Currently the HeaderCell's onClick is always overwritten by the click event of the Table.Header

https://github.com/cerner/terra-core/blob/master/packages/terra-table/src/TableHeader.jsx#L33

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior

A Table.HeaderCell should honor an onClick directly passed.

## Current Behavior

A Table.HeaderCell's onClick is always overwritten by the onClick of the Table.Header.
```
 <Table.Header>
    <Table.HeaderCell content={'Name'} minWidth="small" sort="desc" onClick={() => console.log('This onClick will never be invoked')} />
    <Table.HeaderCell content={'Address'} key={'ADDRESS'} minWidth={'medium'} />
    <Table.HeaderCell content={'Phone Number'} key={'PHONE_NUMBER'} minWidth={'large'} />
 </Table.Header>
```

## Steps to Reproduce

1. Create a Table
2. Add an onClick to a Table.HeaderCell
3. Observe the onClick is always overwritten,

## Environment

* All environments

dylan-sftwr:
@StephenEsser , looking at the line you posted, and tracking it up to [here](https://github.com/cerner/terra-core/blob/master/packages/terra-table/src/TableHeader.jsx#L22-L24) where `newProps` is declared, would you expect the behavior to allow for **both** `onClick`s to be fired (both the `Table.Header`'s and the `Table.HeaderCell`'s, or **only** the `Table.HeaderCell`'s, when provided?

I ask because in the former case, we could do something like [this](https://github.com/kentcdodds/react-toggled/blob/master/src/index.js#L4) (Kent C. Dodds provides an explanation of doing this in terms of a `prop getter` being used with the `render prop` pattern. But, in the latter, we could modify the line you linked above to be 
```jsx
return React.cloneElement(
  child,
  [...newProps, child.props.onClick]
);
```
(I believe, this would need to be tested) and essentially ensure that the `Table.HeaderCell`'s `onClick` is the one that will always _win_ when the props get hashed as they're passed into the new element.

StephenEsser:
[...]

PR #1387, opened 2018-03-29
Add table with multiple onClick handlers in header

bjankord:
### Summary
Add functional example for #1244


packages/terra-site/src/examples/table/Index.jsx
```diff
@@ -53,6 +54,9 @@ const TableExamples = () => (
     <h2>Table without padding</h2>
     <NoPaddingTable />
     <br />
+    <h2>Table with multiple onClick handlers</h2>
```
mjhenkes: Table with multiple *header* onClick handlers?

bjankord: Updated: https://github.com/cerner/terra-core/pull/1387/commits/65f361141b4fd72a306954165f0251cd15d0c469

mmalaker:
Functional verification completed. The table example with multiple onClick handlers displays and functions as expected.

Issue #1302, opened 2018-02-28
Replace Nightwatch Tests with Webdriver.io Tests


emilyrohrbough:
# Issue Description
Remove the existing nightwatch tests and replace them with webdriver.io tests. This has been done for some components but these need to happen for the remaining components:

Checked off means done or claimed

- [x] terra-content-container
- [x] terra-demographics-banner
- [x] terra-divider
- [x] terra-dynamic-grid
- [x] terra-form-checkbox
- [x] terra-form-field
- [x] terra-form-textarea
- [x] terra-form-select
- [x] terra-heading
- ~terra-hookshot~ Migrated to terra-framework https://github.com/cerner/terra-framework/pull/128
- [x] terra-icon
- [x] terra-image
- [x] terra-profile-image
- [x] terra-progress-bar
- [x] terra-props-table
- ~terra-slide-panel~ Migrated to terra-framework https://github.com/cerner/terra-framework/pull/131
- [x] terra-spacer
- [x] terra-status
- [x] terra-text
- [x] terra-toggle
- [x] terra-grid
- [x] terra-list
- [x] terra-responsive-element
- [x] terra-signature
- [x] terra-slide-group
- [x] terra-table
- [x] terra-toggle-button
- [x] terra-abstract-modal
- [x] terra-date-picker
- [x] terra-menu
- [x] terra-search-field

## Remaining Components
- [x] terra-data-time-picker
- [x] terra-embedded-content-consumer

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

emilyrohrbough:
Be sure to specify terra-defined viewports in the test setup: see https://github.com/cerner/terra-core/issues/1444

msaikaushik1992:
picking up terra-status for replacing nightwatch with wdio tests.

bjankord:
picking up content container tests

bjankord:
picking up the demographics banner tests

bjankord:
picking up the dynamic grid

bjankord:
picking up the form field

Blackop778:
I'll grab content container

bjankord:
@Blackop778 There is a PR out for content container: https://github.com/cerner/terra-core/pull/1468
terra-heading or terra-text would be another good one to pick up instead.

Blackop778:
I'll grab props table. terra-heading and terra-text don't actually test anything with nightwatch.

[...]

PR #1392, opened 2018-03-30
Replaced nightwatch with wdio for divider component

msaikaushik1992:
### Summary
Resolves one of [this](https://github.com/cerner/terra-core/issues/1302).


packages/terra-divider/tests/wdio/divider-spec.js
```diff
@@ -0,0 +1,16 @@
+/* global browser, Terra */
+const viewports = Terra.viewports('tiny', 'medium');
```
bjankord: Let's only take a screenshot at the `medium` size. This component doesn't change at different screen widths.

PR #1396, opened 2018-03-31
Initial webpack4 update

mjhenkes:
### Summary
Testing webpack4 update.

* Moved site.config.js from root to dev-site-config/site.confg.js
* Using webpack-serv now instead of webpack-dev-server

@cerner/terra


package.json
```diff
@@ -44,14 +44,13 @@
...
     "compile": "lerna run compile",
-    "compile:heroku": "webpack --config webpack.prod.config --p",
     "compile:prod": "webpack --config webpack.prod.config --p",
     "danger": "danger",
     "deploy": "webpack --config webpack.prod.config -p && gh-pages -d build",
     "dependency-markdown": "node scripts/dependency-markdown-generator/DependencyMarkdownGenerator.js",
-    "generate-config": "generate-config -s 'packages/terra-site/src/examples/**/+(Icon!(*Navigation|*Routes|*Tabs)|*Index).jsx' -s 'packages/terra-*/tests/{nightwatch,nightwatch/*}/!(*-spec|*Template|common|MockComponent|SignatureConstants)+(.jsx|.js)' -s 'packages/terra-table/tests/nightwatch/*/components/!(*-spec|*Template|common|MockComponent|SignatureConstants)+(.jsx|.js)'",
+    "generate-config": "generate-config -o 'dev-site-config' -s 'packages/terra-site/src/examples/**/+(Icon!(*Navigation|*Routes|*Tabs)|*Index).jsx' -s 'packages/terra-*/tests/{nightwatch,nightwatch/*}/!(*-spec|*Template|common|MockComponent|SignatureConstants)+(.jsx|.js)' -s 'packages/terra-table/tests/nightwatch/*/components/!(*-spec|*Template|common|MockComponent|SignatureConstants)+(.jsx|.js)'",
```
emilyrohrbough: We should update the generate-config scrip  to provide 'dev-site-config' as the default with webpack 4 changes

mjhenkes: I'd rather add a setting to the site.config to auto-generate component config prior to running the server and remove this script entirely.

webpack.config.js
```diff
@@ -1,17 +1,9 @@
 const path = require('path');
-const webpackConfig = require('terra-dev-site/src/config/webpack.config');
-const aggregateTranslations = require('./packages/terra-i18n/scripts/aggregate-translations/aggregate-translations');
-
-aggregateTranslations({ baseDirectory: __dirname });
+const webpackConfig = require('terra-dev-site/src/webpack/webpack.config')();
```
JakeLaCombe: From looking how we import this and other sub components like out of terra-form and the tabs, would it be possible to get terra-dev-site building out to a lib folder instead of an src one?

mjhenkes: I'd like to leave sorting that out to another PR. There isn't a reason to transpile the webpack config because it's executed by node. But we do have some code in terra-dev-site intended for the client side and I think we should transpile that.

JakeLaCombe: I do like that approach.

mjhenkes: I'll make that change in terra-dev-site and terra-toolkit

package.json
```diff
@@ -44,31 +44,30 @@
...
     "heroku-prebuild": "npm install rimraf && npm install -g lerna@2.1.2 && lerna init",
-    "heroku-postbuild": "npm install --only=dev && npm run compile:heroku",
+    "heroku-postbuild": "npm install --only=dev && npm run compile:prod",
     "jest": "jest",
     "jest:coverage": "jest --coverage",
     "link-parent-bin": "link-parent-bin",
     "lint": "npm run lint:js && npm run lint:scss",
     "lint:js": "eslint --ext .js,.jsx .",
-    "lint:scss": "stylelint 'packages/**/src/**/*.scss' --syntax scss",
```
bjankord: Looks like a bad merge here, we want to keep the `stylelint 'packages/**/src/**/*.scss' --syntax scss` command, not the `lerna run lint:scss` command.

PR #1401, opened 2018-04-04
Transpile default loaders

### LINKED ISSUE #1400: I18n default loaders are not transpiled
# Issue Description
The default I18n loaders are added directly to the node modules folder and escape transpilation. This causes issues in IE 10 and 11.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Passive behaviour.

## Steps to Reproduce
1. Run webpack using the old plugin but the newest version of terra i18n
2. Run the app in IE 10 or 11
3. Error

mjhenkes:
### Summary
Move default loader and translation loaders to src and reference the lib file.
Fixes #1400

@cerner/terra


mjhenkes:
I know not why the date picker spec is failing, but it's also failing in master.

bjankord:
The date picker spec if failing on the `Triggers onClickOutside when clicking outside of the picker to dismiss it` test. It also just started failing in the action header PR. I'm guessing it is an issue in one of the dependencies.

mjhenkes:
I logged a new bug for the date time picker issue. https://github.com/cerner/terra-core/issues/1402

packages/terra-i18n/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Transpile defaultIntLoaders and defaultTranslationsLoaders.
```
JakeLaCombe: *Intl

packages/terra-date-picker/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Lock in version of popper to prevent date picker from pointing to an incorrect position when invoked. https://github.com/cerner/terra-core/issues/1402
```
JakeLaCombe: Leave the link out of the changelog. I don't think it's necessary, and we can easily trace back to this issue through git history.

bjankord: I'd prefer to keep the link, I think it makes it easier for consumers to find out about the change

PR #1403, opened 2018-04-04
Props table line wraps all code sections

### LINKED ISSUE #1391: [Props Table] Description column doesn't line wrap <pre> Sections
# Issue Description
Preformatted sections within the Description column added with markdown aren't automatically line wrapped, unlike preformatted sections in the type column

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Preformatted sections in the props table should line wrap

## Current Behavior
Preformatted sections in the Description column don't line wrap 

## Steps to Reproduce

1. Add \```Really long lines of text\``` to the description column of a Props Table

## Environment

* Component Version: 2.2.0
* Browser Name and Version: Chrome 65
* Operating System and version (desktop or mobile): OS X Sierra

Blackop778:
### Summary
Fixed `PropsTable` not line wrapping code sections in the description column

### Additional Details
Had to change the css to affect all code elements within the props table because single \`\` don't include a pre tag

[Effect can be seen here](https://terra-core-deployed-pr-1403.herokuapp.com/static/#/raw/tests/props-table/mock-props-table)

Closes #1391 

@cerner/terra


packages/terra-props-table/src/PropsTable.jsx
```diff
@@ -75,7 +75,7 @@ const PropsTable = ({ componentName, src, ...customProps }) => {
   return (
     <div dir="ltr" className="markdown-body">
       <h2>{componentName} Props</h2>
-      <table {...customProps}>
+      <table className={cx('propsTable')} {...customProps}>
```
JakeLaCombe: Keep the same structure as the rest of the classes. In this case, props-table.

packages/terra-props-table/src/PropsTable.jsx
```diff
@@ -75,7 +75,7 @@ const PropsTable = ({ componentName, src, ...customProps }) => {
   return (
     <div dir="ltr" className="markdown-body">
       <h2>{componentName} Props</h2>
-      <table {...customProps}>
+      <table className={cx('props-table')} {...customProps}>
```
StephenEsser: Classnames and customProps should be flipped, if custom props includes a classname it will always overwrite any className declared before it when it spreads.

```
<table {...customProps} className={cx('props-table')}>
```

We should probably also be merging the customProps classNames into our classNames. Here is an example: https://github.com/cerner/terra-core/blob/ada5669999f1f56efd8420e3e296716a681d7643/packages/terra-card/src/Card.jsx#L21-L26

PR #1404, opened 2018-04-05
Make Demographics Banner responsive to parent instead of window

### LINKED ISSUE #1399: Demographics Banner - Wrapping Behavior
# Issue Description
screen shot 2018-04-03 at 4 24 33 pm

DOB year is wrapping to second line, even with plenty of room. 

## Issue Type
- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The DOB shouldn't wrap if there is space to accommodate it. If there's an instance where data on the right is pushing data on the left to wrap, the entire line should wrap down to the next line to avoid this behavior: 

screen shot 2018-04-03 at 4 26 38 pm

## Current Behavior
With this amount of space on the right side, the year shouldn't wrap. 

## Steps to Reproduce
https://kaiju.cerner.com/projects/divine-amano-shiratori-GtgdCA/workspaces/twilight-king-kong-LWU15w/preview

## Environment
* Component Version: Whatever is being used in Kaiju
* Browser Name and Version: Chrome Version 64.0.3282.186 (Official Build) (64-bit)
* Operating System and version (desktop or mobile): Mac OS Desktop

JakeLaCombe:
### Summary
Resolves #1399 

Terra Demographics banner should be responsive to it's parent instead of the window object. This allows the demographics banner to be used in small regions such as side panels while also having the same look and feel as it would on a mobile device.

### Additional Details
I wanted to add a wdio test for this case, but wdio doesn't exist for this component. Also, nightwatch wouldn't have a benefit here, as we essentially need a screenshot of the banner inside a small region. When wdio is added to this package, we need to be sure that we add a test for this scenario.

@cerner/terra

Verification Link - https://terra-core-deployed-pr-1404.herokuapp.com/static/#/site/components/site/demographics-banner/index


packages/terra-site/src/examples/demographics-banner/DemographicsBannerSmallContainer.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import DemographicsBanner from 'terra-demographics-banner';
+import Image from 'terra-image';
+
+import demographicsImage from '../../assets/demographicsCat.jpg';
+
+const DemographicsBannerSmallContainer = () => (
+  <div style={{ width: '525px' }}>
```
bjankord: nit: let's change this to `max-width`.

mmalaker:
Functional verification completed. The demographics examples display and function correctly and are responsive based on parent container.

neilpfeiffer:
Tested #1404
Resolves #1399
 
\
**Reviewed:**
- Resolves the layout issues noted in the issue – demo banner now responds to container instead of window, and moved breakpoint down to tiny (544px).
  

\
 
 
**Pass** |
-----------|
UX Review|

PR #1406, opened 2018-04-05
Move aggregator script down to terra toolkit

mjhenkes:
### Summary
With this change I have moved the aggregate-translations script down to toolkit and have aliased it in terra-i18n for passivity.
I don't really like that terra-i18n has a dependency on terra-toolkit, but in the long run I don't think it harms anything and we should remove it during the next major version bump.

@cerner/terra


mjhenkes:
I'm having second thoughts about adding terra-toolkit as a dependency to terra-i18n. Specifically Terra toolkit is going to start having dependencies like webpack4 and other things be added to it and I don't want to put us into the position of causing strange behaviour in a project using terra-i18n and an older version of terra-toolkit.

At this point I think it's just best to leave the duplication and the warnings recommending that our users switch to the newer commands.

Issue #1398, opened 2018-04-03
Update terra-modal css to correct the max-width and allow for additional z-index


dkasper-was-taken:
# Issue Description
Remove/Update the max-width set on the modal content in order to enabled modal managers expanded styling.
Add a secondary z-index style in order to enable 2 tiers of modal.

## Issue Type
- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Adjust the max-width css style to no longer be limited to 48em.
Add another z-index tier for modal display.

Issue #1323, opened 2018-03-07
Update modal documentation to reference modal manager.


dkasper-was-taken:
# Issue Description
Enhance modal documentation to point to modal-manager. We've received many requests for the disclosure and sizing offered by modal manager, with users unaware that modal manager exists. As well as modal manager is our general recommendation rather than consuming modal directly, as a it handles popups and datepicker.

## Issue Type
- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

PR #1409, opened 2018-04-06
Update modal styling for max width and z-index

dkasper-was-taken:
### Summary
Correct the maximum width for fixed size modal and add ability to set z-index additional layer.

After offline discussion with @neilpfeiffer we are looking to make these changes as well:
Removing responsive break-point behavior at this level, so each consuming implementation can handle responsiveness differently. The first two examples we have are within modal-manager and the alert/prompt.  Depending on the modal manager height, in some cases we don't want to go full screen in small form factor. Within the prompt we never want to go to full screen on small form factors.

Removing the theme-able variables for height, width, and margin. They are not being consumed and we do not want to forcibly set heights of all modal content across implementations.

Resolves: #1398, Resolves: #1323 

@cerner/terra


packages/terra-modal/src/Modal.jsx
```diff
@@ -4,6 +4,8 @@ import { Portal } from 'react-portal';
 import 'terra-base/lib/baseStyles';
 import ModalContent from './ModalContent';
 
+const zIndexes = ['6000', '8000'];
```
yuderekyu: Just curious, how did we decide on these two z-index values?

dkasper-was-taken: 6000 was the previous value, the popup is at 7000, so i went an additional layer to 8000.

mjhenkes: I vote 9001, for no reason...

packages/terra-abstract-modal/src/AbstractModal.scss
```diff
@@ -0,0 +1,26 @@
...
+    max-height: calc(100% - 2 * var(--terra-abstract-modal-vertical-inset, 10px));
+    max-width: calc(100% - 2 * var(--terra-abstract-modal-horizontal-inset, 10px));
+    outline: none;
+    overflow: hidden;
+    position: fixed;
+    top: 50%;
+    transform: translate(-50%, -50%);
+
+    &.layer-6000 {
+      z-index: 6000 + 1;
```
tbiethman: Any reason not to just make this 6001?

dkasper-was-taken: Not especially, it was the pattern used in modal before, and kind of made a weird kind of sense to have 'layer-6000' be '6000'(+1). I'm fine updating it if people have an opinion either way.

packages/terra-abstract-modal/src/AbstractModal.jsx
```diff
@@ -55,6 +51,10 @@ const propTypes = {
    * Role attribute on the modal dialog.
    */
   role: PropTypes.string,
+  /**
+   * Z-Index layer to apply to the ModalContent and ModalOverlay.
```
mjhenkes: document the available options here so they end up in the props table?

PR #1413, opened 2018-04-11
Text line height fix

### LINKED ISSUE #1064: Line-height incorrectly inherited in text component
# Issue Description
The line-height in the text component does not correctly apply when a parent element has a larger font-size.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Text component's line height is applied by its font-size, not by parent element font-size

## Current Behavior
Text component's line height is applied by parent if parent sets a different font-size than the one in the text component font-size

jmsv6d:
### Summary
Fixes #1064 
Line height on the text component was getting calculated based on the parent's font size. This is because the line-height is only applies to block level elements. Updated display to be inline-block and went ahead and switched all the tests to wdio since they were simple. 

@cerner/terra


packages/terra-text/tests/nightwatch/TextLineHeight.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import Text from '../../lib/Text';
+
+const TextLineHeight = () => (
+  <div>
+    <div style={{ 'font-size': '100px' }}>
```
JakeLaCombe: Can you do fontSize here?

PR #1416, opened 2018-04-12
Update .travis.yml

bjankord:
### Summary
Update chromedriver version in travis.yml to use same version as terra-toolkit.


bjankord:
I'm gonna switch the 2 components that now have failing nightwatch tests to use webdriver and then hopefully this is ready to go.

PR #1423, opened 2018-04-13
Update package.json to use terra-dev-site@0.4.1

bjankord:
### Summary
Update to use latest terra-dev-site release

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "stylelint-order": "^0.7.0",
     "stylelint-scss": "^2.0.0",
     "stylelint-suitcss": "^1.0.0",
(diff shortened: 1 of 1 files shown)


mjhenkes:
I'm making changes around this in the next version of terra-dev-site. I'm not sure why it's complaining now though.
screen shot 2018-04-13 at 8 46 24 am

bjankord:
@mjhenkes Yeah looks like this is going to need a fix in terra-dev-site before we can consume it.

mjhenkes:
With terra-dev-site 0.4.2 i think the initial draw of the site is decently faster.

PR #1433, opened 2018-04-18
Update root stylelint check

bjankord:
### Summary
Replaces using lerna to run individual commands for stylelint to 1 global command for stylelint. This is similar to how we run eslint from the root package.

### DIFF
--- a/packages/terra-site/CHANGELOG.md
+++ b/packages/terra-site/CHANGELOG.md
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Fixed stylelint issues
 
 2.6.0 - (April 15, 2018)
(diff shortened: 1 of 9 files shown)


packages/terra-site/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Fixed stylelint issues
 
 2.6.0 - (April 15, 2018)
```
emilyrohrbough: I thought we weren't releasing site anymore?

bjankord: Haven't added it to the ignore config in lerna.json yet so it is still being published. I can update that in a separate PR.

PR #1435, opened 2018-04-18
Lerna hoist package lock

bjankord:
Testing install / build time with hoist and package.lock in root package


bjankord:
Build time on this seems to be pretty consistently around 29 minutes

PR #1451, opened 2018-04-27
Nightwatch -> Wdio 

emilyrohrbough:
### Summary
This was a quick change. #1302


packages/terra-toggle/tests/wdio/toggle-spec.js
```diff
@@ -0,0 +1,47 @@
...
+    beforeEach(() => browser.url('/#/raw/tests/toggle/open-toggle'));
+
+    Terra.should.beAccessible();
+    Terra.should.matchScreenshot();
+  });
+
+  describe('Animated', () => {
+    beforeEach(() => browser.url('/#/raw/tests/toggle/animated-toggle'));
+
+    it('disables focusable elements when closed', () => {
```
JakeLaCombe: Instead of doing attribute and nightwatch style testing, can we use screenshots here instead?

Also, if we need to ensure that specific attributes are set on an element, I think that would be better done with jest that wdio.

packages/terra-toggle/tests/jest/Toggle.test.jsx
```diff
@@ -6,6 +6,7 @@ describe('Toggle', () => {
   it('should render a default toggle', () => {
     const toggle = render(<Toggle>Test</Toggle>);
     expect(toggle).toMatchSnapshot();
+    expect(toggle[0].attribs['aria-hidden']).toBe('true');
```
JakeLaCombe: Doesn’t the jest test have this in the snapshot?

Issue #1044, opened 2017-12-01
Create enumeration for PropTypes.oneOf


benbcai:
# Issue Description
For [any package](https://github.com/cerner/terra-core/search?utf8=%E2%9C%93&q=oneOf%28&type=) that has a prop with a type of PropTypes.oneOf, an enumerations should be made available for that prop to make it easier for consumers to know the acceptable values. This also minimizes consumers from making typos when setting these props. 

An example of of a prop using enumeration is [alert type](https://github.com/cerner/terra-core/blob/master/packages/terra-alert/src/Alert.jsx#L16-L24) in the Alert component.

## Issue Type
- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [X] Other

benbcai:
Packages that need updated:

- [ ] terra-spacer
- [ ] terra-heading
- [ ] terra-slide-panel
- [ ] terra-button
- [ ] terra-arrange
- [ ] terra-badge
- [ ] terra-table
- [ ] terra-text
- [ ] terra-button-group
- [ ] terra-progress-bar
- [ ] terra-image
- [ ] terra-hookshot
- [ ] terra-overlay
- [ ] terra-popup
- [ ] terra-form
- [ ] terra-responsive-element
- [ ] terra-collapsible-menu-view
- [ ] terra-collapsible-menu-view
- [ ] terra-modal-manager
- [ ] terra-menu
- [ ] terra-signature
- [ ] terra-form-textarea

noahbenham:
When declaring the PropType, it might make more sense to use `Object.values()` instead of listing the options to improve maintainability.
For example, `type: PropTypes.oneOf(Object.values(AlertTypes))`.

noahbenham:
Question for the team: Would it make more sense to export options directly instead of as part of `Opts`?

For example,
#### Internally
```jsx
const AlertTypes = {
  ALERT: 'alert',
  ...
};
```
would become
```jsx
export const AlertTypes = {
  ALERT: 'alert',
  ...
};
```


#### Consumers
```jsx
import Alert from 'terra-alert';

<Alert type={Alert.Opts.Types.WARNING}>
  This is a warning!
</Alert>
```
could become
```jsx
import Alert, { AlertTypes } from 'terra-alert';

<Alert type={AlertTypes.WARNING}>
  This is a warning!
</Alert>
```
This also allows consumers flexibility while still achieving our goal of typed props. For example,
```jsx
[...]

PR #1458, opened 2018-04-30
Add a new prop to card in order to flex a class for a new style

### LINKED ISSUE #1432: Enhance Card On Card for better Box Shadow
# Issue Description
Enhance Card to allow for an alternative box shadow when on a white background (like when in another card)

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The card should be flexible enough to allow for a different box shadow when nested inside another card.

## Current Behavior
Terra core has a css variable for box-shadow set to none, but themes may override this to have a box shadow. The ask is for the ability to also flex it to be different when placed inside another card. This is so that a different box shadow can be applied for a better visual affect.

##### A new Prop on Card
One possible way to do this is to just add a type/variant on the card that allows it to change some visuals based off that type. In this case it might be a `stacked` card.

##### A new Card Component
In order to prevent the existing card from having its props balloon out with different specific scenarios, a simple new `StackedCard` could be created that utilizes the existing card and overrides the box shadow.

##### Smart Card Body Approach
If the rule is that "A card inside a card has a different box shadow" then 1 possibility is to make the card put a class on it that scopes a css variable inside it to change the card box shadow. This would not work in ie though, so its probably a no go. Another way of doing something similar is via a descendent selector to look for another Card inside Card and to change the box shadow. This would work in IE but comes with alot of assumptions that would inevitably cause an edge case. If target only direct children, it wouldn't work if anyone puts a div or spacer around their inner card. If you target all descendents, someone might put a container inside their card with a none-white background, and then another card in that expecting the normal box shadow.

##### Separate the box shadow concern from the Card itself
Create a wrapper component or HOC that can be used to add a box shadow on the outer most element. This would be useful to not couple specific box shadows to card, and to allow them to be used with other components. Would only apply to the outer element of a component however. After speaking with UX a bit more about this, their guidance is that the 6 levels of box shadow wouldn't need to be applied randomly to other components, but is already encapsulated by existing components.

## Cards on Light Backgrounds
So what if a card is on the background, and a client themes their background to white. After discussing with UX, this scenario shouldn't be handled by the Card trying to be smart and flex. In this case, the client would have to theme the card box shadow appropriately.

# My Vote
I think the new prop on card makes the most sense for now. There are already no props on Card (other than children) so we aren't making it super custom. The new component seems a bit overkill and would require more refactoring of the internals to share code

### New React Prop:
| Prop     | Type | Default | Description  |
|----------|------|---------|--------------|
| TBD | boolean | false | Sets a class that overrides the default box shadow css property with a new value and a new css custom property|

To keep it passive, the default would be false. Additionally, since terra currently doesn't have a box shadow for the card at all, I'm assuming even when true, the default value for the css custom property would still be no box shadow. Would need UX input

### Proposed Names - We don't want
* stacked - isn't right, since there isn't a card stack 
* nested - Doesn't always have to be nested
* isInsideCard - ins't right, since it is for any card used on any white background,
* isRaised - Doesn't seem right since we aren't truly changing the elevation, just changing the way the light hits it and makes a shadow
* hasBoxShadow - Tied directly to the specific style it adds, not its usage. Also normal cards can already have box shadows

### Proposed Names - That we are open to using
* isOnLightBg
* enbeddedCard
* cardStyle/variant - A bit vague and non-prescriptive on how to use it, but not tied to any particular use case and open for themes to use it differently. Default would be `baseLine`, new one would be 1raised1, and in the future it could expand for other scenarios.

alex-bezek:
### Summary
New prop called variant used to flex to different card styles for new use cases. Falls back to default. Currently only adds the option for raised which adds a new css custom attribute for the box shadow.

### Additional Details
Resolves #1432
Deployed https://terra-core-deployed-pr-1458.herokuapp.com/#/site/components/site/card/index

@cerner/terra


packages/terra-card/src/Card.scss
```diff
@@ -3,8 +3,15 @@
...
     color: var(--terra-card-color, inherit);
     overflow: hidden;
   }
+
+  .default {
+    box-shadow: var(--terra-card-box-shadow, none);
+  }
+
+  .raised {
+    box-shadow: var(--terra-card-raised-box-shadow, none);
```
alex-bezek: @neilpfeiffer let me know if you have a style you want applied by default for this class in terra-core. Otherwise, it's just a new themeable option we can utilize

packages/terra-card/src/Card.jsx
```diff
@@ -12,20 +12,31 @@ const propTypes = {
    * Child Nodes
    */
   children: PropTypes.node,
+  /**
+   * An option used in order to change to a different style of card for different use cases
+   */
+  variant: PropTypes.oneOf(['default', 'raised']),
```
alex-bezek: I couldn't find a pattern on what the prefered way to do this is. I see
* just an array inline like this
* a const array up above and then used here
* an object where the keys and values are all caps and then doing object.values here
* adding `opts` to the component, exporting, and using those.

Let me know if there is a preference!

dylan-sftwr: I'm currently working on #1044. I volunteered to go through all the components and do the uplift to allow for exported enumeration of variants, I'd prefer it be done following [this](https://github.com/cerner/terra-core/commit/8bae7969950a3868334cf871b72da372dea373ee) pattern, as it's how I plan on implementing all component enumerations.

packages/terra-card/src/Card.jsx
```diff
@@ -12,20 +12,31 @@ const propTypes = {
    * Child Nodes
    */
   children: PropTypes.node,
+  /**
+   * An option used in order to change to a different style of card for different use cases
```
emilyrohrbough: Let list the options in the description so consumers know what variants there are.

packages/terra-card/tests/jest/Card.test.jsx
```diff
@@ -14,6 +14,13 @@ describe('Card', () => {
   it('should use the default value when no value is given', () => {
     const wrapper = shallow(defaultRender);
     expect(wrapper.find('.card'));
+    expect(wrapper.find('.default'));
+  });
+
+  it('should use the raised variant when prop is provided', () => {
```
emilyrohrbough: Lets add a jest test using the `default` variant as well

packages/terra-card/tests/jest/Card.test.jsx
```diff
@@ -14,6 +14,13 @@ describe('Card', () => {
...
+    const wrapper = shallow(<Card variant="raised" />);
+    expect(wrapper.find('.raised'));
```
JakeLaCombe: Do we need the wrapper.find tests? Your snapshot had a card with the raised class on it. To me, it feels like duplicate testing when we do test jest like this.

packages/terra-card/src/Card.jsx
dev-cprice: I like the use of enums, but I think there needs to be a conversation on the consumption of them. Because it's easier and more terse to do something like `<Card variant="raised" ... />` than `<Card variant={Card.Opts.Variants.RAISED} ... />` as a consumer. I get why the precedence is to put it behind another `Opts` object, but it just adds to verbosity and feels like writing Java all over again. 

`Card.Variants.RAISED` >>> `Card.Opts.Variants.RAISED`, in my opinion

You could also name export `Variants` or `CardVariants` from here, but `CardVariants` is only one key stroke less than `Card.Variants` and `Variants` will need to be renamed by the consumer if they use multiple terra components in the same file (all too common), e.g. `import { Variants as CardVariants } from 'terra-card';

The above can also be done with destructuring from the Card like `const { Variants } = Card` or renaming it like `const { Variants: CardVariants } = Card;`

I don't putting the burden of renaming an enum on the consumer and I think if getting to the enum takes too long a consumer is just going to do `<Card variant="raised" ... />`

packages/terra-card/src/Card.jsx
alex-bezek: Good call, these are what generate the doc site props table, aren't they?

packages/terra-card/src/Card.jsx
alex-bezek: Updated

packages/terra-card/tests/jest/Card.test.jsx
alex-bezek: My personal opinion is that snapshot tests are not very helpful and often times ignored. I typically prefer explicit jest tests. 
That being said, i can remove the test if you want

packages/terra-card/src/Card.jsx
```diff
@@ -6,26 +6,47 @@ import styles from './Card.scss';
...
 };
 
+const opts = {
+  Variants: CardVariants,
+};
+
 Card.propTypes = propTypes;
+Card.defaultProps = defaultProps;
 Card.Body = CardBody;
+Card.Opts = opts;
```
mjhenkes: For the record, I'm against exporting these props as an enum. Using a string is more concise and lowers the complexity of maintaining and creating these components. With prop type warning you will be warned if you've had a spelling mistake or pass in a non supported type.

alex-bezek: For the record, I agree with you. The first is easier
`<Card variant="raised" />`
vs
`<Card variant={Card.Opts.Variants.RAISED} />`

I also ran my wdio tests for 15 min only for them to fail because I used
`<Card variant={Card.Opts.RAISED} />` the first time on accident

Also it makes typing the wrong thing way worse. With the string it will just throw a prop type warning in dev and continue along its way. But with the object of object of object thing, mis typing like
`<Card id="raisedCard" variant={Card.Opt.Variants.RAISED}>`
throws a run time exception and crashes the page

image

or `<Card id="raisedCard" variant={Card.Opts.Variant.RAISED}>`
image

But, I don't really have a dog in this. Let me know what the consensus is to go with, and it sounds like there is an open issue to reconcile them all to be consistent anyways.

dylan-sftwr:
@mjhenkes if @alex-bezek just wants to remove the enumeration from this PR and do it with strings, I can just pick it up in #1044 if we decide to go forward with it.

mjhenkes:
@dylan-sftwr Sounds good. Lets not block this PR.

packages/terra-card/src/Card.scss
neilpfeiffer: @alex-bezek - Yes, we will want to have a visual version for default theme, otherwise it would look like a dead prop (not recommended). I'll grab a box-shadow style to use.

neilpfeiffer: @alex-bezek Values to used for default terra-opensource-theme:
```
background: white;
border: 1px solid #c8cacb; 
box-shadow: 0 3px 8px -2px rgba(0, 0, 0, 0.18), 0 3px 10px 0 rgba(0, 0, 0, 0.12), 0 3px 3px -3px rgba(0, 0, 0, 0.3);
```

alex-bezek:
Updated to use an object inside the component, but not export it out. 
Also updated the default from none to the box shadow the @neilpfeiffer provided.
Deployment link here: https://terra-core-deployed-pr-1458.herokuapp.com/#/site/components/site/card/index

Let me know if there are any other issues

neilpfeiffer:
Tested: PR #1458
Resolves: Issue #1432

**Reviewed:**
- Reviewed Enhancement to add an additional card variant style: `raised`
- Follows pattern considerations documented in Cerner Design Standards: [Cards](https://wiki.ucern.com/x/JG4VYQ)
- Follows UX defined design implementation for Terra-Card 
- Follows all defined theme designs: terra-theme

\
Verified by @neilpfeiffer and @scottwilmarth, Ready for Release after Functional Validation.


**Pass** |
-----------|
UX Review|

PR #1470, opened 2018-05-01
Bump dependency on terra-popup.

dkasper-was-taken:
### Summary
Bump terra-menu's dependency on terra-popup to use the 3.0 version.

@cerner/terra


bjankord:
@dkasper-was-taken Should terra-form-select also get a bumped popup version?

dkasper-was-taken:
@bjankord Good catch, updating now. #

PR #1472, opened 2018-05-02
Converted nightwatch to wdio tests for terra-status

msaikaushik1992:
### Summary
Resolves [this](https://github.com/cerner/terra-core/issues/1302).


packages/terra-status/tests/nightwatch/StatusIcon.jsx
```diff
@@ -1,9 +1,8 @@
 import React from 'react';
```
emilyrohrbough: Lets move all these test examples to the new dir structure
```
/component
    /examples
        /test-examples
            ...examples
```

packages/terra-status/tests/wdio/status-spec.js
```diff
@@ -0,0 +1,39 @@
+/* global browser, Terra */
+const viewports = Terra.viewports('medium');
+
+describe('Status', () => {
+  describe('Default', () => {
```
emilyrohrbough: Lets add
```before(() => browser.setViewportSize(Terra.viewports('medium')[0]));```
and drop each specification of `veiwports` in the test steps
so,
```diff
+    Terra.should.beAccessible();
-    Terra.should.beAccessible({ viewports });
```

packages/terra-status/tests/wdio/status-spec.js
```diff
@@ -0,0 +1,39 @@
...
+    beforeEach(() => browser.url('/#/raw/tests/status/status-default'));
```
emilyrohrbough: nit: each of these beforeEach can be changed to before since we do not need to refresh the page between test steps

PR #1481, opened 2018-05-04
Add Error Prop to Fieldset

### LINKED ISSUE #1480: Add error prop to Fieldset
# Issue Description
The fieldset doesn't have an error prop like it's previous component contained. We need to add one back in order to display an error message for a group of forms.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Fieldset has an error prop on it.

## Current Behavior
Fieldset does not have an error prop.

JakeLaCombe:
### Summary
Resolves #1480 

This adds the error prop back to the fieldset

One thing I'm debating is if we want to use an isInvalid prop on this fieldset as well. Let me know what you guys think.


emilyrohrbough:
+1 to adding `isInvalid`. This would make field set consistent with field then.

packages/terra-form-fieldset/src/Fieldset.scss
```diff
@@ -25,6 +25,15 @@
     width: auto;
   }
 
+  .error-text {
+    color: var(--terra-form-fieldset-error-text-color, #e50000);
+    font-size: var(--terra-form-fieldset-error-text-font-size, 0.857rem);
+    font-weight: var(--terra-form-fieldset-error-text-font-weight, normal);
+    line-height: 1.25;
```
yuderekyu: Should we be making `line-height` themeable for error-text and help-text? @neilpfeiffer

neilpfeiffer: Thanks @yuderekyu - currently discussion the options of what fieldset should have/do with @JakeLaCombe

bjankord:
Talked with UX and found out that adding `error` and `isInvalid` props to the fieldset component is not desired. Instead, we'll be building a new component to handle rendering checkbox or radio buttons inside of a fieldset along with displaying associated help text / error text with the field. @JakeLaCombe is working on a tech design for the new component.

PR #1497, opened 2018-05-08
Update changelogs and dependencies in prep for release

bjankord:
### Summary
Updating changelogs and bumping dependencies with major version releases from frameworks.

With this change, terra-collapsible-menu-view, terra-form-select, terra-menu, and terra-tabs are slated to get major version bumped in this release.


cg.js
```diff
@@ -0,0 +1,24 @@
+const replace = require('replace-in-file');
```
emilyrohrbough: do we want to check this in?

bjankord: Did not mean to include that, its a temporary script I use to automate some of changelog updates. Removed in this commit: https://github.com/cerner/terra-core/pull/1497/commits/84f713d15c2efb6f775dbcde6ee469825a35aafb

Issue #1362, opened 2018-03-21
[Select] - Support multiple selections


StephenEsser:
# Issue Description

This is a placeholder issue for adding multiple selection support into the Select component. A tech design will be added here in the near future.

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior

The Select component should support multi-select.

## Current Behavior

The Select component does not support multi-select.

emilyrohrbough:
Adding numbered list for easy reference:

1. Should it be _noResult**s**Content_  instead of _noResultContent_.

2. I think `onFilter` might better articulate that the proposed `optionFilter` is a callback function for option filtering. 
3. Why are we dropping the `requestFocus` and `releaseFocus` props? Are we dropping the terra-popup dependency? Or rolling our own request & release focus methods?

4.  
>The event object should be abstracted as it no longer contains true change information.

Would it make sense to pass the event information down to the consumer, even if its the the 'true' event? 

5. Is `Opt Group` a  new component within select?
6. Are the props listed under `Option` the props for the current `SelectOption` component props? I notice `isSelected` is not included. Or is `Option` A new component as well?

emilyrohrbough:
You had mentioned the implementation behind select would be similar for each variant type, will the prop combinations (outside of default) be similar as well?

How do you expect to handling removing a user-defined option from the list of options? I did not see any props or description of this behavior.  For example, what if I meant to add `component` but accidentally added `compon`

StephenEsser:
1. Because the search can return a single result or multiple results I'm not strongly opinionated either way. Does anyone else have an opinion between `noResultsContent` or `noResultContent`? 

For additional information Vue uses the singular approach: https://vue-multiselect.js.org/#sub-slots
Ant avoids the plural / singular issue by using `notFoundContent`: https://ant.design/components/select/#Select-props

2. In general the `onHandler` event pattern is used to react to a single change that has already happened and is bubbling up through the ancestors. It feels a little awkward for the return value of the handler to conditionally filter options. Do you have any examples where the `onHandler` callback return value is used in this way? 

3. The popup is being removed as a dependency and the focus requests are no longer necessary.

[...]

Issue #1495, opened 2018-05-08
terra-form-select ignores value change


ChaseSinclair:
# Issue Description
When the value is changed externally, the internal state is not updated. this.props.value is not used meaningfully anywhere [here](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/Select.jsx#L133). On the other hand, DatePicker actually changes the state [here](https://github.com/cerner/terra-core/blob/master/packages/terra-date-picker/src/DatePicker.jsx#L133) which is consistent with other terra form input components.

Before:
![image](https://user-images.githubusercontent.com/5703475/39779712-ae3674a4-52cf-11e8-8cc2-ebe5d30cb663.png)

OnChange:
![image](https://user-images.githubusercontent.com/5703475/39779738-ba26ce6c-52cf-11e8-9b0e-90f9abd0bd25.png)

OnReset:
![image](https://user-images.githubusercontent.com/5703475/39779758-c7261b04-52cf-11e8-98d2-63d3d51ad4f3.png)

## Issue Type
- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior
When the value prop is changed, the option matching the value should appear selected.

## Environment
* Component Version: React 16
* Browser Name and Version: Chrome Version 66.0.3359.139
* Operating System and version (desktop or mobile): MacOS Sierra 10.12.6

PR #1501, opened 2018-05-11
[terra-select] Added OptGroup and Tag, Combobox, Multiple, and Search variants

StephenEsser:
@cerner/terra


packages/terra-form-select/package.json
```diff
@@ -44,8 +44,7 @@
     "lint": "npm run lint:js && npm run lint:scss",
     "lint:js": "eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
     "lint:scss": "stylelint src/**/*.scss",
-    "test": "npm run test:jest && npm run test:nightwatch",
-    "test:jest": "jest ./tests/jest/* --config ../../jestconfig.json",
-    "test:nightwatch": "nightwatch -c ../../nightwatch.conf.js"
+    "test": "npm run test:jest",
+    "test:jest": "jest ./tests/jest/* --config ../../jestconfig.json"
```
StephenEsser: More tests are in progress

packages/terra-form-select/src/_Frame.scss
```diff
@@ -0,0 +1,252 @@
...
+    font-size: var(--terra-select-font-size, 1.143rem);
+    line-height: var(--terra-select-line-height, 1.75);
+  }
+
+  .default {
+    background-size: var(--terra-select-background-size, 0 0);
+    border: var(--terra-select-border, 1px solid #dedfe0);
+    border-radius: var(--terra-select-border-radius, 3px);
+    cursor: pointer;
+    padding: var(--terra-select-padding, 0 0 0 0.42857rem);
```
bjankord: Let's split this out into individual styles (padding-top, padding-right, padding-left, padding-bottom). The RTL postcss plugin doesn't seem to process this as a bidi based style as it is now.

packages/terra-form-select/tests/jest/FrameUtil.test.jsx
```diff
@@ -0,0 +1,200 @@
+import FrameUtil from '../../src/_FrameUtil';
+import { Variants } from '../../src/_constants';
+
+describe('FrameUtil', () => {
```
bjankord: Nice test coverage on these utils.

packages/terra-form-select/CHANGELOG.md
```diff
@@ -3,6 +3,27 @@ ChangeLog
 
 Unreleased
 ----------
+### Breaking Changes
+* Removed boundingRef
+* Removed isPlaceholderHidden - The placeholder option is no longer added by default.
+* Removed name - Terra recommends handling validations with [react-final-form](https://github.com/final-form/react-final-form).
+* Removed required - Terra recommends handling validations with [react-final-form](https://github.com/final-form/react-final-form).
```
bjankord: I'm on the fence removing this, I think it may still be valuable just for setting aria-required on the element with the `role="listbox"` or equivalent element in the variations.

packages/terra-form-select/src/_FrameUtil.js
```diff
@@ -0,0 +1,114 @@
...
+   */
+  static includes(props, query) {
+    if (query.trim().length === 0) {
+      return false;
+    }
+    return (props.value || []).indexOf(query) > -1;
+  }
+
+  /**
+   * Determines whether the dropdown should be respositioned.
```
JakeLaCombe: *repositioned*

Also, if that's the intent of this function, could it be renamed to shouldDropdownReposition?

packages/terra-form-select/src/_FrameUtil.js
```diff
@@ -0,0 +1,114 @@
...
+   * @param {ReactNode} dropdown - The component dropdown.
+   * @return {boolean} - True if the dropdown should be respositioned.
+   */
+  static shouldPositionDropdown(previousState, currentState, dropdown) {
+    if (!currentState.isOpen) {
+      return false;
+    }
+
+    const { bottom } = dropdown.getBoundingClientRect();
+    if (previousState.isOpen === false || bottom > window.innerHeight) {
```
JakeLaCombe: I think you can reduce this if statement to this?

    return previousState.isOpen === false || bottom > window.innerHeight

packages/terra-form-select/src/_Option.jsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import 'terra-base/lib/baseStyles';
+import styles from './_Option.scss';
+
+const cx = classNames.bind(styles);
+
+/* Disabling validation for controlled props. These props will always be overwritten. */
```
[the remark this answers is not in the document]
StephenEsser: This prevents them from showing up in the props table, I wasn't able to find an existing strategy to hide these from the public API. The parent of these will always override them during cloning, so it will never honor a custom users value if one is provided so I didn't want to expose them in the readme

packages/terra-form-select/src/_Tag.jsx
```diff
@@ -0,0 +1,37 @@
...
+   * Callback function triggered when the tag is deselected.
+   */
+  onDeselect: PropTypes.func.isRequired,
+  /**
+   * The value of the tag.
+   */
+  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
+};
+
+/* eslint-disable jsx-a11y/no-static-element-interactions */
```
JakeLaCombe: We should still add an aria label to the span element that's interactionable.

packages/terra-form-select/src/_FrameUtil.js
StephenEsser: I fixed the typo and changed the grammar to read as `should be positioned` to preserve the name and avoid some confusion

packages/terra-form-select/src/_FrameUtil.js
StephenEsser: Good call, updated to shorted the function

packages/terra-form-select/tests/jest/MenuUtil.test.jsx
```diff
@@ -0,0 +1,467 @@
...
+    it('should return false if the string is undefined', () => {
+      let string;
+      expect(MenuUtil.contains(string, 'query')).toBeFalsy();
+    });
+
+    it('should return false if the query is not conatained within the string', () => {
+      expect(MenuUtil.contains('string', 'query')).toBeFalsy();
+    });
+
+    it('should return false if the query is not conatained within the string', () => {
```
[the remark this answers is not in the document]
StephenEsser: Removed

packages/terra-form-select/src/_MenuUtil.js
```diff
@@ -0,0 +1,222 @@
...
+    }
+    return a.toLowerCase() === b.toLowerCase();
+  }
+
+  /**
+   * Determines whether the variant allows multiple selections.
+   * @param {string} variant - The variant.
+   * @return {boolean} - True if the variant allows multiple selections.
+   */
+  static isMultiple(variant) {
```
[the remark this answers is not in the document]
StephenEsser: This was originally `allowsMultipleSelections` but I changed it to isMultiple for fewer characters in the function name. I've changed it back to `allowsMultipleSelections` to help avoid any confusion that might come from the multiple variant

packages/terra-form-select/CHANGELOG.md
JakeLaCombe: I'm curious if that's something we can do. Technically, input elements are the only ones that should have that attached, and I'm not sure if we have any elements attached here that could be used.

packages/terra-form-select/src/_Frame.jsx
```diff
@@ -0,0 +1,392 @@
...
+  /**
+   * Callback function triggered when an option is selected.
+   */
+  onSelect: PropTypes.func,
+  /**
+   * Callback function for option filtering. function(searchValue, option)
+   */
+  optionFilter: PropTypes.func,
+  /**
+   * Placholder text.
```
JakeLaCombe: *placeholder

StephenEsser: Fixed

packages/terra-form-select/CHANGELOG.md
bjankord: Yeah, that's a good point, I'm fine with it as is.

packages/terra-form-select/src/_Menu.jsx
```diff
@@ -0,0 +1,291 @@
...
+/* eslint-disable react/no-unused-prop-types */
+const propTypes = {
+  /**
+   * The content of the menu.
+   */
+  children: PropTypes.node,
+  /**
+   * Content to display when no results are found.
+   */
+  noResultContent: PropTypes.node,
```
StephenEsser: There was a comment on the tech design about this being plural or singular. Does anyone have an opinion on this? 

Potential Options include, but not limited to. ( With or without the Content suffix )
`noResultContent`
`noResultsContent`
`notFoundContent`

packages/terra-form-select/src/_Tag.jsx
JakeLaCombe: I'm fine with that. I found a defect on it's repo that says it won't go away unless you update.

mmalaker:
The popup container of the select floats above the top nav on the page (see screenshot). It is doing this on all browsers and mobile platforms, but most pronounced on mobile.
screen shot 2018-05-18 at 1 39 30 pm

StephenEsser:
@mmalaker We discussed this in the war room and we believe this is expected behavior, it behaves in a similar way to the native select in regards to the menu is always displayed over content. 

This also prevents the clipping / hiding of the menu if it was to display under the nav.

Do you have any additional concerns about this behavior?

mmalaker:
@StephenEsser I'm OK with it as long as UX agrees. @neilpfeiffer what do you think? Looks very odd especially when scrolling on mobile, but probably acceptable since native behavior is the same.

packages/terra-form-select/examples/test-examples/ControlledDefault.example.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import Base from 'terra-base';
```
emilyrohrbough: All example are wrapped in base by terra-dev-site so we can drop wrapping these in Base

StephenEsser: Removed

neilpfeiffer:
Tested: PR #1501
Resolves: Issue #1362 #1495

**Reviewed:**
- Follows pattern considerations documented in Cerner Design Standards: [Select]
- Follows UX defined design implementation for Terra-Form-Select
- Follows all defined theme designs: terra-theme
- Adds functionality for new searchable select, combobox, and multiselect

\
\
**Resolve with Future Enhancements:**
- Apply additional outstanding enhancements needed for Cerner Universal Design for combobox and multiselect variants.

\
\
Verified by @neilpfeiffer, Ready for Initial Release.

**Pass** |
-----------|
UX Review|

PR #1504, opened 2018-05-13
Enable parallel testing during builds

ryanthemanuel:
### Summary
In order to speed up builds and avoid the travis time limit, this pull request intends to enable parallel testing during the build process.  To accomplish this, the build will now use docker to run npm install during a build phase.  Then during the test phase, it will run different wdio test suites in parallel.  If all is successful (and we're on the master branch), we will then run the deploy phase.

The goal here is also to be able to run tests locally the exact same way as previously.


.travis.yml
```diff
@@ -1,21 +1,43 @@
...
-deploy:
-  provider: pages
-  skip_cleanup: true
-  github_token: $GITHUB_TOKEN # Set in travis-ci.org dashboard
-  local_dir: /build
-  on:
-    branch: master
+script:
+  - 'docker-compose pull'
+  - "docker-compose run -e DANGER_GITHUB_API_TOKEN=$DANGER_GITHUB_API_TOKEN -e TRAVIS=$TRAVIS -e CI=$CI -e HAS_JOSH_K_SEAL_OF_APPROVAL=$HAS_JOSH_K_SEAL_OF_APPROVAL -e TRAVIS_PULL_REQUEST=$TRAVIS_PULL_REQUEST -e TRAVIS_REPO_SLUG=$TRAVIS_REPO_SLUG test npm run $TEST_SUITE"
```
ryanthemanuel: These environment variables are needed for running danger.

.travis.yml
```diff
@@ -1,21 +1,43 @@
...
+  - TEST_SUITE="wdio -- --suite suite1"
+  - TEST_SUITE="wdio -- --suite suite2"
+  - TEST_SUITE="wdio -- --suite suite3"
+  - TEST_SUITE="wdio -- --suite suite4"
+  - TEST_SUITE=danger
+jobs:
+  include:
+    - stage: build
+      script:
+        - echo "$DOCKER_PASSWORD" | docker login -u "ryanthemanuel" --password-stdin
```
ryanthemanuel: Eventually, I'll want this to be the cerner username and password so that it gets published to the cerner docker user.

Dockerfile
```diff
@@ -0,0 +1,19 @@
+# use node as base image
```
ryanthemanuel: This file will be the same across all repos.  It's not that complicated but we could DRY it up by having a docker image that takes care of lines 1-16.  Thoughts?

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
+version: '3'
```
ryanthemanuel: We could DRY this up by moving it to terra-toolkit and curl'ing in the travis.yml file before running the docker-compose commands.  We could have the image name be an environment variable.  We also need to have a tag.  I'm thinking it could be the travis build id or travis build number if the branch is not master and "latest" otherwise.  Thoughts?

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+
+services:
+  base:
+    build: ./
+    image: "ryanthemanuel/terra-core"
+  standalone-chrome:
+    image: 'selenium/standalone-chrome:3.11.0-californium'
```
JakeLaCombe: Is toolkit locking this down? I know we talked about toolkit doing that? If so, is there a way to ensure we always reference that same version?

nightwatch.conf.js
```diff
@@ -12,6 +14,92 @@ if (isRepoTest) {
...
+  page_objects_path: '',
+  persist_globals: true,
+  test_workers: false,
+  detailed_output: !process.env.TRAVIS && !process.env.CI,
+  test_settings: {
+    default: {
+      launch_url: `http://${seleniumHost}:${port}`,
+      persist_globals: true,
+      selenium_port: seleniumPort,
+      selenium_host: process.env.TRAVIS ? 'standalone-chrome' : seleniumHost,
```
ryanthemanuel: This is the only line that is different from terra-toolkit.  I'm ok with moving everything here as nightwatch is going away eventually.  Thoughts?

docker-compose.yml
JakeLaCombe: However, should the image name be an environment variable? Didn't we want to lock that version down?

wdio.conf.js
```diff
@@ -1,17 +1,32 @@
...
     enabled: !process.env.TRAVIS,
   },
 
+  host: 'standalone-chrome',
+
+  // Define wdio test suites
+  suites: {},
 };
 
+const numberOfSuites = 4;
```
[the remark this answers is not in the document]
emilyrohrbough: Does this break package level testing?

wdio.conf.js
```diff
@@ -1,17 +1,32 @@
...
+const PackageUtilities = require('lerna/lib/PackageUtilities');
+const Repository = require('lerna/lib/Repository');
+
+// eslint-disable-next-line no-underscore-dangle
+const packageLocations = PackageUtilities.getPackages(new Repository(__dirname)).map(pkg => pkg._location);
 
 const config = {
   ...wdioConf.config,
 
-  webpackConfig,
```
emilyrohrbough: Why did drop our webpack config? Will this work locally?

wdio.conf.js
```diff
@@ -1,17 +1,32 @@
 /* eslint-disable import/no-extraneous-dependencies */
```
mjhenkes: We were so close to being able to eliminate wdio config from our repos, any way this could be moved to terra-toolkit?

Dockerfile
emilyrohrbough: I think it would be cleaner to have one build image in toolkit

.travis.yml
bjankord: Who do we need to work with to get this info?

.travis.yml
```diff
@@ -1,21 +1,49 @@
...
-  provider: pages
-  skip_cleanup: true
-  github_token: $GITHUB_TOKEN # Set in travis-ci.org dashboard
-  local_dir: /build
-  on:
-    branch: master
+script:
+  - 'wget -O docker-compose.yml https://raw.github.com/cerner/terra-toolkit/test_parallel/config/wdio/ci-docker-compose.yml'
+  - 'docker-compose pull'
+  - "docker-compose run -e DANGER_GITHUB_API_TOKEN=$DANGER_GITHUB_API_TOKEN -e TRAVIS=$TRAVIS -e CI=$CI -e HAS_JOSH_K_SEAL_OF_APPROVAL=$HAS_JOSH_K_SEAL_OF_APPROVAL -e TRAVIS_PULL_REQUEST=$TRAVIS_PULL_REQUEST -e TRAVIS_REPO_SLUG=$TRAVIS_REPO_SLUG test npm run $TEST_SUITE"
```
bjankord: HAS_JOSH_K_SEAL_OF_APPROVAL ? are we using that for anything?

package.json
```diff
@@ -92,6 +92,7 @@
     "gh-pages": "^0.12.0",
     "glob": "^7.1.1",
     "identity-obj-proxy": "^3.0.0",
+    "ip": "^1.1.5",
```
yuderekyu: Are we using `ip` anywhere?

ryanthemanuel: Not anymore.  I can remove this.

nightwatch.conf.js
ryanthemanuel: Just for posterity, I moved this code back to terra-toolkit.

.travis.yml
ryanthemanuel: @bjankord it is necessary for danger integration.

.travis.yml
ryanthemanuel: We should be good to go now.

PR #1506, opened 2018-05-14
Only Highlight JSX to make the package size smaller.

mjhenkes:
### Summary
Only Highlight JSX to make the package size smaller.
Based on instructions here:
https://github.com/conorhastings/react-syntax-highlighter#light-build

@cerner/terra


mjhenkes:
This saved .2 Mb from the site size.

PR #1513, opened 2018-05-15
Prepare for toolkit

ryanthemanuel:
### Summary
This PR is preparing the project for a change that is going to happen in terra-toolkit to support parallelization.  This change will help prevent a major version bump in terra-toolkit.


wdio.conf.js
```diff
@@ -11,7 +11,8 @@ const config = {
   seleniumDocker: {
     enabled: !process.env.TRAVIS,
   },
-
+  
+  host: '0.0.0.0',
```
mjhenkes: why?

ryanthemanuel: This is temporary until I can get all of the parallelization changes propagated.  It is the default host used in webdriver.

PR #1524, opened 2018-05-17
Update date time picker to consume abstract-modal.

dkasper-was-taken:
### Summary
Remove the DateTimePicker's dependency on the deprecated terra-modal. Now implements a custom abstract modal that is analogous to the terra-dialog-modal. Now handles focus state when presented from modal manager.

@cerner/terra

### DIFF
--- a/packages/terra-date-time-picker/src/DateTimePicker.jsx
+++ b/packages/terra-date-time-picker/src/DateTimePicker.jsx
@@ -136,6 +136,8 @@ class DateTimePicker extends React.Component {
     this.handleOnDateInputFocus = this.handleOnDateInputFocus.bind(this);
     this.handleOnTimeInputFocus = this.handleOnTimeInputFocus.bind(this);
     this.handleOnCalendarButtonClick = this.handleOnCalendarButtonClick.bind(this);
+    this.handleOffsetButtonClick = this.handleOffsetButtonClick.bind(this);
+    this.handleOnRequestClose = this.handleOnRequestClose.bind(this);
   }
 
   componentWillMount() {
@@ -347,6 +349,14 @@ class DateTimePicker extends React.Component {
     }
   }
 
+  handleOffsetButtonClick(event) {
+    this.setState({ isTimeClarificationOpen: !this.state.isTimeClarificationOpen });
+  }
+
+  handleOnRequestClose(event) {
+    this.setState({ isTimeClarificationOpen: false });
+  }
+
   renderTimeClarification() {
     return (
       <TimeClarification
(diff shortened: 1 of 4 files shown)


packages/terra-date-time-picker/src/_TimeClarification.jsx
```diff
@@ -25,13 +25,33 @@ const propTypes = {
...
+   */
+  requestFocus: PropTypes.func,
 };
 
 const defaultProps = {
   isOpen: false,
   isOffsetButtonHidden: false,
   onDaylightSavingButtonClick: undefined,
   onStandardTimeButtonClick: undefined,
+  onOffsetButtonClick: undefined,
```
JakeLaCombe: I didn't see any Jest tests updated for these new props, which I find a little surprising. Is there a reason they are not needed?

dkasper-was-taken: There wasn't previously a jest test for it. I'll add one.

packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -347,13 +349,25 @@ class DateTimePicker extends React.Component {
     }
   }
 
+  handleOffsetButtonClick() {
+    this.setState({ isTimeClarificationOpen: !this.state.isTimeClarificationOpen });
```
bjankord: React may batch multiple `setState()` calls into a single update for performance.

Because `this.state` may be updated asynchronously, you should not rely on its values for calculating the next state. You can fix this by using the functional form of `setState` which accepts a function as an argument.

More info in [docs here](https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous):

```
this.setState(prevState => ({
  isTimeClarificationOpen: !prevState.isTimeClarificationOpen
}));
```

packages/terra-date-time-picker/src/_TimeClarification.scss
```diff
@@ -1,5 +1,14 @@
 /* stylelint-disable selector-max-compound-selectors, max-nesting-depth */
 :local {
+  .time-clarification {
+    background-color: var(--terra-date-time-clarification-modal-background-color, #fff);
+    border: var(--terra-date-time-clarification-modal-border);
+    border-radius: var(--terra-date-time-clarification-modal-border-radius, 5px);
+    box-shadow: var(--terra-date-time-clarification-modal-box-shadow, 0 3px 7px rgba(0, 0, 0, 0.3));
+    color: var(--terra-date-time-clarification-modal-foreground-color);
+    width: 640px;
```
bjankord: Should this be a max-width? I'm curious how this works on smaller viewports.

dkasper-was-taken: Max width is already handled by the base abstract modal, with the 100% - 20px etc. The text just wraps inside the dialog.

neilpfeiffer:
Only issue I saw was new TimeClarification modal on actual iOS browser doesn't fully resize to width with screen, stays a little wider (400-440px or so) and then requires side-scroll. Not critical and can be fixed with future enhancement.

Appears fine when tested in desktop browser at the same mobile viewport size.

mmalaker:
Functional verification completed. The date time picker example in the modal displays and functions as expected using the new abstract modal implementation.

PR #1525, opened 2018-05-17
[terra-doc-template] Added some styling to the example template

StephenEsser:
### Summary
Added some styling to the doc template and wanted to get some feedback. If anyone has suggestions or ideas let me know.

# Before
image

# After
image

@cerner/terra


saedar:
Maybe not needed, but I saw that we have the Doc Template listed in our component set. Should this also be included or nah?

emilyrohrbough:
This looks so good!!

yuderekyu:
Dangggg. Looking sharp

packages/terra-doc-template/src/ExampleTemplate.scss
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  .footer {
+    display: flex;
+    flex-direction: column;
+    overflow: hidden;
+  }
+
+  .code {
+    height: 0;
```
yuderekyu: I don't have exact transition values, but adding back an animation when toggling the src code would be cool.
```
transition: height <duration> <timing function>
```

packages/terra-doc-template/src/ExampleTemplate.scss
```diff
@@ -0,0 +1,73 @@
...
+    border: 0;
+    cursor: pointer;
+    display: flex;
+    margin-left: auto;
+    outline: none;
+    padding: 0.625rem;
+  }
+
+  .chevron-left {
+    background: url('data:image/svg+xml;base64,PHN2ZyBkYXRhLW5hbWU9IkxheWVyIDEiIHhtbG5zPSJodHRwOi8vd3d3LnczLm9yZy8yMDAwL3N2%0D%0AZyIgdmlld0JveD0iMCAwIDQ4IDQ4IiBjbGFzcz0iaXMtYmlkaSI+PHBhdGggZD0iTTEwLjMgMjRM%0D%0AMzMuOCAwbDMuOSAzLjhMMTggMjRsMTkuNyAyMC4yLTMuOSAzLjh6Ii8+PC9zdmc+');
```
yuderekyu: Should we use chevrons from `terra-icon`? Also, we can use `inline-svg` from our mixins here.

packages/terra-doc-template/src/ExampleTemplate.scss
StephenEsser: The animations in the original terra-toggle-button are done by react-animate-height. The only other way to animate the height is to set a generic max-height on the component. 

https://github.com/cerner/terra-core/blob/master/packages/terra-toggle/src/Toggle.jsx#L37

Because this component is becoming a hard dependency in every package I think we should remove as many dependencies as possible. If we want to experiment with a max-height solution there are a few caveats that come along with that.

packages/terra-doc-template/src/ExampleTemplate.scss
StephenEsser: This is using the icons from terra-icon, but in CSS to avoid adding an additional dependency to the package. Because this component is becoming a hard dependency in every package I think we should avoid adding any extra dependencies if there is an alternative solution. 

The inline-svg from the mixins converts a raw SVG into the base64 encode. The SVG here is already encoded. Unless there are additional mutations the mixin performs I think this should behave the exact same way, and doesn't require adding the terra-mixin dependency. 

@bjankord Do you see any issues not using the inline-svg mixin if the SVG is already encoded?

packages/terra-doc-template/src/ExampleTemplate.scss
yuderekyu: Ah gotcha, makes sense

packages/terra-doc-template/src/ExampleTemplate.scss
bjankord: No concern, the mixin is a nicety so people don't need to convert to dataURI, if its already converted, the mixin isn't needed.

PR #1526, opened 2018-05-18
added wdio tests for checkbox

msaikaushik1992:
### Summary
Resolved [this](https://github.com/cerner/terra-core/issues/1302).

### Additional Details
Converted nightwatch tests to wdio tests for just checkbox(checkbox-field alsready has wdio tests).


packages/terra-form-checkbox/tests/wdio/form-checkbox-spec.js
```diff
@@ -0,0 +1,161 @@
...
+    '--terra-form-checkbox-disabled-background-color': '#e7e8e8',
+    '--terra-form-checkbox-disabled-background-image': 'none',
+    '--terra-form-checkbox-disabled-border-color': '#cfd0d1',
+    '--terra-form-checkbox-disabled-hover-border-color': '#cfd0d1',
+    '--terra-form-checkbox-disabled-hover-border-width': '1px',
+    '--terra-form-checkbox-focus-border-color': '#77b1fd',
+    '--terra-form-checkbox-focus-border-width': '2px',
+    '--terra-form-checkbox-transition': 'all 0.3s cubic-bezier(0.78, 0.14, 0.15, 0.86)',
+    '--terra-form-checkbox-disabled-font-opacity': '0.5',
+    '--terra-form-checkbox-focus-ring-background-color': 'orange',
```
yuderekyu: We should split some of these themeable variables out to specific tests. For example, the check box has both a regular focus and a focus ring. So we need themeable tests for both, similar to radio button ([here](https://github.com/cerner/terra-core/blob/master/packages/terra-form-radio/tests/wdio/radio-spec.js#L188-L189)) and [here](https://github.com/cerner/terra-core/blob/master/packages/terra-form-radio/tests/wdio/radio-spec.js#L225-L229)).

bjankord: Updated: https://github.com/cerner/terra-core/pull/1526/commits/16a43813088c6c3dea6c4841fbea3eba7ce83bda

yuderekyu:
Hmm, looks like the focus ring screen shot isn't rendering correctly

PR #1529, opened 2018-05-20
Theme tabs

alex-bezek:
### Summary
As an engineer, in order to theme the tabs component, I want a couple more css variables available

### Additional Details
The specific new piece added is a before content in order to control the blue stripe at the top. Typically we do this with the background-image and set the size. However this doesn't work in our specific theme case because we need to be able to set the background image to a linear gradient thats transparent. Since we can't use it for both, the before content is an optional way to do it.

I left the existing way in there for passivity.

@cerner/terra

### DIFF
--- a/packages/terra-tabs/examples/test-examples/Tabs/TabsTemplate.jsx
+++ b/packages/terra-tabs/examples/test-examples/Tabs/TabsTemplate.jsx
@@ -19,7 +19,7 @@ const defaultProps = {
 
 const TabsTemplate = props => (
   <Base locale={locale}>
-    <div style={{ backgroundColor: 'lightgrey', height: props.containerHeight }}>
+    <div style={{ height: props.containerHeight }}>
       <Tabs
         tabFill={props.tabFill}
         fill={props.fill}
--- a/packages/terra-tabs/src/StructuralTabs.scss
+++ b/packages/terra-tabs/src/StructuralTabs.scss
@@ -9,6 +9,7 @@
       background-color: var(--terra-tabs-structural-content-background-color, #fff);
       border-radius: var(--terra-tabs-structural-content-border-radius, 0);
       box-shadow: var(--terra-tabs-structural-content-box-shadow, 0);
+      //Used to make the content sit above the tab to block the box shadow that can be applied
       z-index: var(--terra-tabs-structural-content-z-index, 0);
     }
 
--- a/packages/terra-tabs/src/StructuralTabs.scss
+++ b/packages/terra-tabs/src/StructuralTabs.scss
@@ -8,6 +8,8 @@
     > :nth-child(2) {
       background-color: var(--terra-tabs-structural-content-background-color, #fff);
       border-radius: var(--terra-tabs-structural-content-border-radius, 0);
+      box-shadow: var(--terra-tabs-structural-content-box-shadow, 0);
+      z-index: var(--terra-tabs-structural-content-z-index, 0)
     }
 
     .collapsed-tabs-container {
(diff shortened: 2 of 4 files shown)


packages/terra-tabs/examples/test-examples/Tabs/TabsTemplate.jsx
```diff
@@ -19,7 +19,7 @@ const defaultProps = {
 
 const TabsTemplate = props => (
   <Base locale={locale}>
-    <div style={{ backgroundColor: 'lightgrey', height: props.containerHeight }}>
+    <div style={{ height: props.containerHeight }}>
```
alex-bezek: this was causing issues with the themed examples (with transparent tabs) and doesn't seem to be needed for core when comparing my branch to master. Let me know if i'm missing something
http://engineering.cerner.com/terra-core/#/site/components/site/tabs/index
https://terra-core-deployed-pr-1529.herokuapp.com/#/site/components/site/tabs/index

alex-bezek:
will add some wdio tests for new theme variables if this approach is fine

alex-bezek:
@bjankord I added some test cases. I also added in some more selectors and css to target hover for active tabs specifically. With the way terra-tabs css currently work the hover affect is applied to both active and non active

packages/terra-tabs/src/StructuralTabs.scss
```diff
@@ -21,18 +24,37 @@
...
 
     .collapsible-tabs-container {
       border-bottom-color: var(--terra-tabs-structural-tab-bar-border-color, #dddedf);
       border-bottom-style: solid;
       border-bottom-width: var(--terra-tabs-structural-tab-bar-border-width, 1px);
+
+      // Only target the tabs when in collapsible-tabs-container. When it is collapsed, the menu shouldn't have the stripe
+      .tab,
+      .tab-menu {
+        position: var(--terra-tabs-structural-position, static);
```
StephenEsser: This variable is missing the `-tab-` substring, or the ones below have an extra `-tab-`

packages/terra-tabs/src/StructuralTabs.scss
```diff
@@ -8,6 +8,9 @@
     > :nth-child(2) {
       background-color: var(--terra-tabs-structural-content-background-color, #fff);
       border-radius: var(--terra-tabs-structural-content-border-radius, 0);
+      box-shadow: var(--terra-tabs-structural-content-box-shadow, 0);
+      //Used to make the content sit above the tab to block the box shadow that can be applied
```
StephenEsser: Nit: Missing a space on the comment

packages/terra-tabs/src/StructuralTabs.scss
alex-bezek: hmm, the existing ones even further below are
`--terra-tabs-structural-background-image`
so I'm gonna go with I have an extra `-tab` for my before content variables. So i'll leave this one the same, and update the ones below to 
`--terra-tabs-structural-before-content-background`
`--terra-tabs-structural-before-content-height`

sound good?

packages/terra-tabs/src/StructuralTabs.scss
```diff
@@ -63,7 +85,25 @@
       &.is-active {
         background-color: var(--terra-tabs-structural-active-background-color, #fff);
         background-size: var(--terra-tabs-structural-active-background-size, 100% 2px);
+        box-shadow: var(--terra-tabs-structural-active-box-shadow, none);
         color: var(--terra-tabs-structural-active-color, #000);
+        position: relative;
+
+        &::before {
+          transition-duration: var(--terra-tabs-structural-transition-duration, 0);
+          transition-property: width;
```
yuderekyu: should we open up `transition-property` to be themeable, or constrain it to width?

packages/terra-tabs/src/StructuralTabs.scss
StephenEsser: Sounds good to me

PR #1532, opened 2018-05-22
Terra time input wdio

JakeLaCombe:
### Summary
Part of #1302 

Converting the time-input tests to wdio.

**Note:** I purposely left the accessibility tests out of this module, and there are several accessibility issues with the time-input that need to be address. I'll run the accessibility tests on my local device, and make an issue on this repo designed to get rid of those issues, as this will take UX effort as well in addressing the concerns.


packages/terra-time-input/tests/wdio/time-input-twelve-hour-mobile-spec.js
```diff
@@ -0,0 +1,81 @@
...
+        '--terra-time-input-meridiem-padding-bottom': '0',
+        '--terra-time-input-meridiem-padding-left': '0',
+        '--terra-time-input-meridiem-padding-right': '0',
+        '--terra-time-input-meridiem-padding-top': '0',
+        '--terra-time-input-meridiem-button-background-color': '#1c1f4d',
+        '--terra-time-input-meridiem-button-border-color': '#ab1c4f',
+        '--terra-time-input-meridiem-button-color': '#ffffff',
+        '--terra-time-input-meridiem-button-padding-bottom': '0.714em',
+        '--terra-time-input-meridiem-button-padding-top': '1.157em',
+        '--terra-time-input-meridiem-button-hover-background-color': '#bcbfff',
```
emilyrohrbough: Some of these will need to be broken out to correctly capture the themability. For example, focus styles will need to be separate from hover styles.

packages/terra-time-input/tests/wdio/time-input-spec.js
```diff
@@ -0,0 +1,268 @@
...
+
+  describe('Time Input does not prepends 0 on hour input less than 6', () => {
+    beforeEach(() => {
+      browser.url('/#/raw/tests/time-input/time-input/on-change');
+
+      browser.click('#timeInput input[name="terra-time-minute-time-input"]');
+      browser.keys('5');
+    });
+
+    Terra.should.matchScreenshot({ viewports });
```
emilyrohrbough: I think these screenshots need checked out. Check out this corresponding screenshot: https://github.com/cerner/terra-core/blob/d5f60148c62709480244d36f679c960553207db8/packages/terra-time-input/tests/wdio/__snapshots__/reference/chrome/Time_Input_does_not_prepends_0_on_hour_input_less_than_6%5Bdefault%5D.838x768.png

JakeLaCombe: The test was renamed incorrectly. It should be referencing the minute, not the hour. I've updated and fixed it.

PR #1536, opened 2018-05-24
[terra-form-select] Added SelectField

StephenEsser:
### Summary
Added a SelectField component that wraps a Select within a Field.

Deployment Link:
https://terra-core-deployed-pr-1536.herokuapp.com/#/site/components/site/form-select/select-field-index

@cerner/terra


packages/terra-form-select/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Added
+* A SelectField component.
```
JakeLaCombe: Nit: The log looks a little weird with just the sentence. I recommend changing A to Added

packages/terra-form-select/docs/SelectField.md
```diff
@@ -0,0 +1,30 @@
+# Terra Form Select
```
JakeLaCombe: Terra Form SelectField*

packages/terra-form-select/CHANGELOG.md
StephenEsser: It does feel a little strange, but it's also strange to double the word Added under a section called Added. Any ideas to make the sentence read cleaner without doubling the same section word?

StephenEsser: Updated to read as: 
`* SelectField - A convenience component that wraps a select within a field.`

packages/terra-form-select/src/SelectField.jsx
```diff
@@ -0,0 +1,163 @@
...
+  isInvalid: false,
+  isLabelHidden: false,
+  labelAttrs: {},
+  onChange: undefined,
+  placeholder: undefined,
+  required: false,
+  selectAttrs: {},
+  showOptional: false,
+  value: undefined,
+  variant: Variants.DEFAULT,
```
emilyrohrbough: should this be 'default'?

packages/terra-site/src/examples/form-select/examples/SelectField.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import SelectField from 'terra-form-select/lib/SelectField';
+
+const SelectFieldExample = () => (
+  <SelectField label="T-shirt size" placeholder="Select a size" selectId="select-field" required style={{ width: '300px' }}>
```
bjankord: Need to update this to use `maxWidth`. @mmalaker noted it gets cut off on mobile devices.

packages/terra-form-select/docs/SelectField.md
```diff
@@ -0,0 +1,30 @@
+# Terra Form Select Field
+
+A convenience wrapper assembling a [terra-form-select](https://github.com/cerner/terra-core/tree/master/packages/terra-form-select) within a [terra-form-field](https://github.com/cerner/terra-core/tree/master/packages/terra-form-field).
+
+## Getting Started
+
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-form-select/lib/SelectField`
```
emilyrohrbough: Does this work? Or should this just be `npm install terra-form-select`? then use via `import SelectField from 'terra-form-select/lib/SelectField`?

packages/terra-form-select/tests/wdio/select-field-spec.js
```diff
@@ -0,0 +1,47 @@
+/* global before, browser, Terra */
+const viewports = Terra.viewports('tiny');
+
+describe('Select Field', () => {
+  describe('should select an option by click', () => {
```
emilyrohrbough: Since are only testing one viewport you could simplify this to be be

```
describe('Select Field', () => {
    before(() => browser.setViewportSize(Terra.viewports('tiny')[0]));
     describe('should select an option by click', () => {
         ... everything here can drop { viewports }
     });
});
```

packages/terra-form-select/docs/SelectField.md
StephenEsser: Whoops, no this would not work. This has been updated

packages/terra-form-select/tests/wdio/select-field-spec.js
StephenEsser: I think I'll keep this as it is, I like how it reads and keeps a consistent spot to update the viewports if we want to test this at different sizes

mmalaker:
Functional verification completed. The select field examples display and function correctly.

PR #1538, opened 2018-05-24
Is it Example or Examples?

emilyrohrbough:
### Summary
On pages where there is only one example, the header still says 'Examples'.

### DIFF
--- a/packages/terra-doc-template/src/DocTemplate.jsx
+++ b/packages/terra-doc-template/src/DocTemplate.jsx
@@ -86,6 +86,12 @@ const DocTemplate = ({ packageName, readme, srcPath, examples, propsTables, ...c
     customProps.className,
   ]);
 
+  let exampleHeader;
+  if (localExamples.length > 0) {
+    const exampleHeaderText = localExamples.length > 1 ? 'Examples' : 'Example';
+    exampleHeader = <h1 className={cx('.examples-header')} >{exampleHeaderText}</h1>;
+  }
+  
   return (
     <div {...customProps} className={docTemplateClassNames}>
       {packageName && <a href={`https://www.npmjs.org/package/${packageName}`}>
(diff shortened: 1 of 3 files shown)


JakeLaCombe:
I like the pull request title! A title with a question....

packages/terra-doc-template/src/DocTemplate.jsx
```diff
@@ -86,6 +86,12 @@ const DocTemplate = ({ packageName, readme, srcPath, examples, propsTables, ...c
     customProps.className,
   ]);
 
+  let exampleHeader;
+  if (localExamples.length > 0) {
+    const exampleHeaderText = localExamples.length > 1 ? 'Examples' : 'Example';
+    exampleHeader = <h1 className={cx('.examples-header')} >{exampleHeaderText}</h1>;
```
StephenEsser: nit: There's a space leftover from the original syntax on the opening h1 tag. 

className={cx('.examples-header')}` `>

This is just preference so keep what you have if you prefer it. Another way to write this could be something like:
```
const exampleHeader = examples.length === 0 ? undefined : (
  <h1 className={cx('.examples-header')}>{localExamples.length > 1 ? 'Examples' : 'Example'}</h1>
);
```

emilyrohrbough: Updated to remove the comment.

In regards how we could write it, I'm going to leave as is for now. Its how we've been writing conditional logic so I want to stay consistent.

packages/terra-doc-template/src/DocTemplate.jsx
```diff
@@ -86,6 +86,12 @@ const DocTemplate = ({ packageName, readme, srcPath, examples, propsTables, ...c
     customProps.className,
   ]);
 
+  let exampleHeader;
+  if (localExamples.length > 0) {
+    const exampleHeaderText = localExamples.length > 1 ? 'Examples' : 'Example';
+    exampleHeader = <h1 className={cx('.examples-header')}>{exampleHeaderText}</h1>;
+  }
+
```
bjankord: This is whitespace on this line that is causing a lint error. Depending on which code editor you are using, you can set it up to trim whitespace on file save.

emilyrohrbough:
@bjankord whoops. Thats what I get for updating from the web.

PR #1547, opened 2018-05-30
[terra-responsive-element] Removed the wrapping div when rendered into the dom

StephenEsser:
### Summary
- Removed the wrapping div of the ResponsiveElement
- Removed nightwatch tests
- Added wdio tests

### Additional Details
This works by rendering a div into the dom on the initial render and then rendering the expected content when an element has been determined on successive renders. 

Resolves https://github.com/cerner/terra-core/issues/1571.

### Deployment Link:
https://terra-core-deployed-pr-1547.herokuapp.com/#/components/terra-responsive-element/responsive-element

@cerner/terra


packages/terra-responsive-element/src/terra-dev-site/doc/ResponsiveElement.doc.jsx
```diff
@@ -1,72 +1,30 @@
+/* eslint-disable import/no-extraneous-dependencies, import/no-webpack-loader-syntax, import/first, import/no-unresolved, import/extensions, import/no-duplicates */
```
mjhenkes: I prefer the ignores to just wrap the responsive table example and example src files. It doesn't look as pretty but is more targeted to which items actually need to be ignored.

packages/terra-responsive-element/src/ResponsiveElement.jsx
```diff
@@ -99,13 +101,11 @@ class ResponsiveElement extends React.Component {
...
+    if (this.state.element) {
+      return this.props[this.state.element] || null;
+    }
 
-    return (
-      <div {...customProps} ref={this.setContainer}>
-        {this.props[this.state.element]}
-      </div>
-    );
+    return <div ref={this.setContainer} />;
```
emilyrohrbough: Is this case for the initial render?

StephenEsser: Yes, on the initial render this will load into the dom, bind the appropriate resize observers, determine the appropriate element to display based on the available width and then rerender. It is not needed after the initial render.

mmalaker:
Functional verification completed. The responsive element example displays and functions correctly with the above changes.

Issue #1359, opened 2018-03-20
Add refCallback function prop and data attribute to ContentContainer


dkasper-was-taken:
# Issue Description
ContentContainer is a functional component and consumers would benefit from being able to get it's ref, so we should add a refCallback similar to what we've done in terra-button and terra-list.

In conjunction with this, we've received several requests for access to the inner normalizer div that provides scrolling, so I recommend adding a data attribute to that div. Something along the lines of 'data-content-container-scroll'. That way scroll listeners, adjustments to scrollTop, or selectively disabling scrolling could be done more efficiently.

## Issue Type
- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
refCallback's node value could be mapped to a variable, and a simple selector could provide the scroll:
```
const scrollNode = this.containerNode.querySelector('[data-content-container-scroll]');
```

StephenEsser:
As we uncover components that warrant a ref callback I think there would be value in making the component a real React Class to inherit the natural and more consistent `ref`. We currently have a mixed and matched ref callback API. True components can utilize `ref` and stateless functional components have added a `refCallback` into the API. 

Making the components true classes may require disabling the lint rule, similar to what was done here: https://github.com/cerner/terra-core/blob/master/packages/terra-form-input/src/Input.jsx#L2

Does anyone else have any thoughts on uplifting components to be React Classes vs adding in a `refCallback` as needed into functional components?

emilyrohrbough:
+1 on uplifting

dkasper-was-taken:
Before we uplift components, I think we should look at the performance. I've noticed there's a fairly significant performance impact between the two when working with the infinite scroller. The additional life cycle events and checks add up. I would think we would only want to convert components to pure react components if we were able to implement shouldComponentUpdate. As many of these components would have a too expensive of a comparison, I'm not sure it's advisable to convert them without that check to improve performance.

StephenEsser:
I agree with running some performance testing before a decision is reached. I'm hesitant to vote for generically implementing shouldComponentUpdate unless we are certain there are no edge cases during prop comparisons.

The Button was uplifted to be a React class during the last version bump.

PR #1549, opened 2018-05-30
Create terra-scroll-view

dkasper-was-taken:
### Summary
Implement a component that provides consistent scroll css styling for consumers ( e.g. momentum scrolling and IE prefixes). This component can be used internally as well, our best example being terra-content-container. Could also be used for terra-slide-panel and terra-application-menu-layout once released.

Resolves: #1359 

Feel free to pitch any names you feel would be better for this component.

@cerner/terra


StephenEsser:
Starting a petition for terra-scroll-view

packages/terra-scrollable-container/docs/README.md
```diff
@@ -0,0 +1,15 @@
+# Terra Scrollable Container
+
+The terra-scrollable-container component scrollable appropriate styling for standard scrolling css and a refCallback.
```
JakeLaCombe: Reword this description. Instead of saying that it applies styling, say that it's a content view that hides data accessible with scrolling.

JakeLaCombe:
Can we add an example of horizontal overflow scroll, perhaps by putting a table inside a container? I imagine that being the biggest use case of overflow scroll at that level.

packages/terra-scrollable-container/src/ScrollableContainer.jsx
```diff
@@ -0,0 +1,32 @@
...
+
+const propTypes = {
+  /**
+   * The child content to present in the scrollable region.
+   */
+  children: PropTypes.node,
+  /**
+   * Function callback for the ref of the scrollable container.
+   */
+  refCallback: PropTypes.func,
```
JakeLaCombe: Instead of first classing this prop, how about we just attach ref to customProps and allow it to go through that way?

dkasper-was-taken: React does not allow 'ref' for be passed through as a prop. It functions similar to 'key' in that it is a reserved value. React will throw warning at you saying the value is inaccessible.  Also as a functional component ref is not mapped in the standard way, it can only be directly attached to a primitive.

dkasper-was-taken:
Added horizontal example.

packages/terra-content-container/src/ContentContainer.scss
```diff
@@ -11,30 +25,14 @@
...
       height: 100%;
       min-height: 50px;
       overflow: hidden;
       position: relative;
       width: 100%;
 
       > .normalizer {
-        height: 100%;
-        overflow: auto;
         position: absolute;
```
tbiethman: Just curious, have we tested things without the normalizer in a while? Wasn't it needed to fix an issue with Safari? I'd be curious to see if versions released in the last year and a half corrected the issue or not, though we might still run into issues with the UIWebView regardless until we're 100% on WKWebViews.

StephenEsser:
The `-view` suffix is a little generic, but to me it feels natural to describe what the view is. This might be a carry over opinion from Android or Apple ScrollViews.

https://developer.apple.com/documentation/uikit/uiscrollview
https://developer.android.com/reference/android/widget/ScrollView

An unrelated React-native ScrollView lost in time: 
https://facebook.github.io/react-native/docs/using-a-scrollview.html

If we decide to rename this component I think we should also avoid the `Container` or `Wrapper` suffixes. I know we have a few of these already. By definition every React component is a container / wrapper of content. (At least until React 16 which allows siblings). The concept of `Container` also has a very opinionated meaning in Redux.

tbiethman:
Everything in iOS/Android has view in the name because they extend from UIView/View, respectively.  It'd be like us naming every React component we have X-Component. Even if I can reconcile using view as an alternative to component, I don't know why we are including view in the name here (and in a small fraction of our components overall).

It's just weird to me to conflate the two concepts. I'll concede that Container/Wrapper aren't much better, and I don't want to cross multiple issue streams here. I think ContentContainer would be better named ContentLayout at this point. What I don't want to do is continue the use of 'view' in our naming schemes without really having a definition for 'view' within Terra or React at large. If we're going to do this, we should define it clearly.

Even something like `terra-scrollable` would better match existing naming styles in Terra/the React community at large.

StephenEsser:
# terra-scrolltron

megatron_bay

Or I would vote for one of `terra-scroll`, `terra-scroller`, `terra-scrollable`. Preference in the same order listed.

bjankord:
I vote `terra-elder-scroll`

packages/terra-scroll-view/docs/README.md
```diff
@@ -0,0 +1,18 @@
+# Terra Scroll View
+
+The terra-scroll-view is a content view that hides data accessible with scrolling and provides a refCallback. The expectation is that any consumer needing to provide a scrollable area within their implementation should be using this component, or one of our other terra components that implements ScrollView.
+
+ScrollView provides a menthod to access the node managing scrolling, so values such as scrollTop can be easily set, as well as the correct prefixes to consistently sytle scrolling across supported browsers.
```
yuderekyu: *method to access
*style scrolling across

Issue #815, opened 2017-09-14
Running nightwatch tests from a nested package fails in regards to file path issues on Windows


bjankord:
# Issue Description
On Windows 10, running nightwatch tests from a nested package fails in regards to file path issues. 

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Contributors can run nightwatch tests for individual packages on Windows

## Current Behavior
Contributors cannot run nightwatch tests for individual packages Windows

## Steps to Reproduce
1. Clone repo
2. Run `npm install`
3. Change directories to a component directory. e.g `packages/terra-alert, packages/terra-button`
4. Run `npm run test:nightwatch`

## Environment
* Operating System and version (desktop or mobile): Windows 10

Issue #813, opened 2017-09-14
Running jest tests from a nested package fails in regards to file path issues on Windows


bjankord:
# Issue Description
On Windows 10, running jest tests from a nested package fails in regards to file path issues. 

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Contributors can run jest tests for individual packages on Windows

## Current Behavior
Contributors cannot run jest tests for individual packages Windows

## Steps to Reproduce
1. Clone repo
2. Run `npm install`
3. Change directories to a component directory. e.g `packages/terra-alert, packages/terra-button`
4. Run `npm run test:jest`

## Environment
* Operating System and version (desktop or mobile): Windows 10

Singleton06:
Any time that I run this on windows from a package's directory, it usually just fails because it is unable to find any tests that match.
```
C:\installs\workspaces\dustin-terra\terra-core\packages\terra-search-field>npm run test:jest

> terra-search-field@1.12.1 test:jest C:\installs\workspaces\dustin-terra\terra-core\packages\terra-search-field
> jest ./tests/jest/* --config ../../jestconfig.json

No tests found
In C:\installs\workspaces\dustin-terra\terra-core
  1987 files checked.
  testMatch: **/__tests__/**/*.js?(x),**/?(*.)(spec|test).js?(x) - 88 matches
  testPathIgnorePatterns: \\node_modules\\ - 1987 matches
Pattern: ./tests/jest/* - 0 matches
```

However, if I run this from the command line it works (after installing jest globally):
```
jest ..\\terra-search-field\\tests\\jest --config ../../jestconfig.json
```

I then tried updating the package.json to the following and was able to get it to work as well:
```
"test:jest": "jest ..\\\\terra-search-field\\\\tests\\\\jest --config ../../jestconfig.json",
```

I realize this isn't the solution to this but it gives us an idea of how the matching is off.

emilyrohrbough:
Try updating the script to `"test:jest": "jest node tests/jest/* --config ../../jestconfig.json",`

PR #1554, opened 2018-05-31
Fixing the Windows nested package issues with running Nightwatch and Jest

BenBoersma:
### Summary
Resolves https://github.com/cerner/terra-core/issues/815, this fixes the pathing issues trying to run nightwatch, but nightwatch itself still does not run.
Resolves https://github.com/cerner/terra-core/issues/813

@cerner/terra


packages/terra-props-table/package.json
```diff
@@ -44,7 +44,6 @@
     "lint:js": "eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
     "lint:scss": "stylelint src/**/*.scss",
     "test": "npm run test:jest && npm run test:wdio",
-    "test:jest": "jest ./tests/jest/* --config ../../jestconfig.json",
-    "test:wdio": "../../node_modules/.bin/wdio ../../node_modules/terra-dev-site/config/wdio/wdio.conf.js"
-  }
+    "test:jest": "node ../../scripts/run-jest/index.js",
+    "test:wdio": "../../node_modules/.bin/wdio ../../node_modules/terra-dev-site/config/wdio/wdio.conf.js"  }
```
mjhenkes: move curly brace to next line

scripts/run-jest/index.js
```diff
@@ -0,0 +1,20 @@
+const spawn = require('child_process').spawn;
```
mjhenkes: What is this file (add comments)? And if it's common it should live somewhere common so we don't have to duplicate it across repos.

BenBoersma: Its a nodejs module, https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options

BenBoersma:
@mjhenkes 
I just checked and indeed it is having that issue.
Specifically went to "packages/terra-card" and ran "test:wdio: and got 

> '..' is not recognized as an internal or external command,

scripts/run-jest/index.js
```diff
@@ -0,0 +1,20 @@
...
+const os = require('os');
```
emilyrohrbough: I notice we didn't add this as a dependency in our root level package.json

scripts/run-jest/index.js
```diff
@@ -0,0 +1,20 @@
...
+  case 'Linux':
+  case 'Darwin':
+    regex = `${process.env.npm_package_name}/tests`;
+    jestcmd = 'jest';
+    break;
+  case 'Windows_NT':
+    regex = `${process.env.npm_package_name}\\\\tests`;
+    jestcmd = 'jest.cmd';
+    break;
+  default:
```
emilyrohrbough: I am worried this check is too specific? What are the possible type returns? could we check for windows and if not windows use `jestcmd= 'jest' by default?

emilyrohrbough: Is the `.cmd` a by product of all bin scripts on windows machines? If so we might look into opening this up to accept a cmd paramater to toggle `jest || jest.cmd` or for example `wdio || wdio.cmd`

emilyrohrbough: Does this command correctly pass command line arguments ? For example, running from terra-button, does this run?

```bash
npm run tt:wdio -- -u
```

BenBoersma:
Changed the way I fixed the issue with Jest. Now instead of creating a custom script to run jest, I have switch jestconfig.json to jestconfig.js and updated its testMatch regex to dynamically match the directory it is called from.

jestconfig.js
```diff
@@ -0,0 +1,16 @@
+module.exports = {
```
mjhenkes: Fun Fact: We have two jest configs floating around in our repo. This one is used when running jest in the individual packages, one in package.json is run when running jest on the lerna repo overall?
Two questions,
1. Does running jest at the main repo have an issue on PC?
2. Could you consolidate these configs (I think you could just delete the config from package.json)

rockfact

BenBoersma: 1. Running Jest on the main repo has no issue on PC.

2. Looking into it now

PR #1557, opened 2018-06-01
Convert table tests to wdio

JakeLaCombe:
### Summary
Table Tests wdio conversion


packages/terra-table/package.json
```diff
@@ -42,6 +42,6 @@
     "lint:scss": "stylelint src/**/*.scss",
     "test": "npm run test:jest && npm run test:nightwatch",
```
Blackop778: This should run the web driver tests instead of nightwatch

Blackop778:
The nightwatch tests can be removed now that we have web driver tests.

Issue #1544, opened 2018-05-29
Signature: Theming is not working for the strokes in canvas


mayurikini:
# Issue Description
For signature component, in canvas the strokes style is made themeable [here](https://github.com/cerner/terra-core/blob/master/packages/terra-signature/src/Signature.scss#L5). But it is not working as expected. Even if we try to change the [context.strokesStyle](https://github.com/cerner/terra-core/blob/master/packages/terra-signature/src/Signature.jsx#L78) to a different color, it will still result in strokesStyle color to be the default #000.  

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior
StokesStyle should be themeable.

## Current Behavior
Theming does not work for StokesStyle in signature component. 

## Steps to Reproduce

1. Try to pass a different color in Signature.scss file, instead of theming variable.

bjankord:
cc/ @vikonava @e3ujwh

PR #1558, opened 2018-06-01
Convert Signature Tests to wdio

JakeLaCombe:
### Summary
Convert Signature Tests to wdio

**Note:** Theming has a defect in it at https://github.com/cerner/terra-core/issues/1544, so the line code isn't being updated.


packages/terra-signature/package.json
```diff
@@ -41,6 +41,6 @@
     "lint:scss": "stylelint src/**/*.scss",
     "test": "npm run test:jest && npm run test:nightwatch",
```
emilyrohrbough: Update test script

packages/terra-signature/tests/wdio/signature-spec.js
```diff
@@ -0,0 +1,64 @@
+/* global browser, Terra */
+
+const viewports = Terra.viewports('medium');
+
+describe('Signature', () => {
+  describe('Default', () => {
```
emilyrohrbough: Looks like you checked in a 'large' or 'huge' viewport sized screenshot for 'Default' but is not longer defined in the test file

JakeLaCombe: It's removed here. https://github.com/cerner/terra-core/pull/1558/commits/56ac2511e13e96325b4308e03156c921bf564617

PR #1576, opened 2018-06-05
[terra-list] Nightwatch -> Webdriverio tests

tomleewu:
### Summary
Converts `terra-list` nightwatch tests into webdriverio tests
Part of #1302


yuderekyu:
Lets add tests that make sure the themeable properties (like [this](https://github.com/cerner/terra-core/blob/master/packages/terra-list/src/List.scss#L14) variable) work as expected. We have convenience methods `themeEachCustomProperty ` and `themeCombinationOfCustomProperties` that apply these properties either one at a time, or as a group, for a screenshot test. Here's an example [test]( https://github.com/cerner/terra-core/blob/master/packages/terra-table/tests/wdio/table-spec.js#L11) doing that

emilyrohrbough:
Side-note: list & table would be prime components for [page-objects]( http://webdriver.io/guide/testrunner/pageobjects.html)

packages/terra-list/src/terra-dev-site/test/list/list-item/DefaultListItem.test.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 
 import ListItem from '../../../../ListItem';
+import List from '../../../../List';
 
-export default () => <ListItem content={<p>test content</p>} style={{ listStyle: 'none' }} />;
```
emilyrohrbough: The point of initially reacting these `ListItem` examples as standalone without being wrapped in the `List `component was to ensure the `ListItem`s would render as expected without influence of the `List`. Given the ListItem should never be rendered outside of the `List` given the implementation expectations and accessibility issues that would occur, it almost seems as if these tests are not gaining us anything.

I would be curious on what others thoughts if removed ListItem visual-regression tests and example and only tested the ListItem via jest. My thought that this is okay is that each List example is composed of the ListItems, which provide they render as expected.

@mjhenkes @ryanthemanuel @bjankord @JakeLaCombe

packages/terra-list/tests/wdio/selectable-list-spec.js
```diff
@@ -0,0 +1,139 @@
...
+      browser.keys(['Tab', 'Space']);
+    });
+    Terra.should.matchScreenshot('first item selected');
+
+    it('selects the second item', () => {
+      browser.keys(['Tab', 'Enter']);
+    });
+    Terra.should.matchScreenshot('second item selected');
+
+    it('attemps to select the third item', () => {
```
emilyrohrbough: This example should be able to select the third item just fine: https://terra-core-deployed-pr-1576.herokuapp.com/#/tests/terra-list/list/selectable-list/on-change-selectable-list

attemps to selects && then update the click to click on the third item in the list.

packages/terra-list/tests/wdio/selectable-list-spec.js
```diff
@@ -0,0 +1,139 @@
...
+      browser.click('ul li:nth-child(1)');
+    });
+    Terra.should.matchScreenshot('first item selected');
```
emilyrohrbough: Update to be `third item selected`

packages/terra-list/tests/wdio/selectable-list-spec.js
```diff
@@ -0,0 +1,139 @@
...
+
+  describe('Selectable List Non Selectable Item', () => {
+    before(() => browser.url('/#/raw/tests/terra-list/list/selectable-list/non-selectable-item-selectable-list'));
+
+    Terra.should.matchScreenshot();
+    it('attempts to select non-selectable item', () => {
+      browser.click('ul li:nth-child(3)');
+    });
+
+    Terra.should.matchScreenshot();
```
emilyrohrbough: lets take add a screenshot name to both of these `Terra.should.matchScreenshot();` so we can differentiate between the before and after click interaction.

packages/terra-list/tests/wdio/multi-select-list-spec.js
```diff
@@ -0,0 +1,207 @@
...
+  });
+
+  describe('Multi Select List No Deselect Item', () => {
+    before(() => browser.url('/#/raw/tests/terra-list/list/multi-select-list/multi-select-list-with-no-deselect-item'));
+    Terra.should.matchScreenshot();
+
+    it('should not deselect item one', () => {
+      browser.click('ul li:nth-child(1)');
+    });
+    Terra.should.matchScreenshot();
```
emilyrohrbough: lets take add a screenshot name to both of these Terra.should.matchScreenshot(); so we can differentiate between the before and after click interaction.

packages/terra-list/tests/wdio/multi-select-list-spec.js
```diff
@@ -0,0 +1,207 @@
...
+    });
+    Terra.should.matchScreenshot('selected item(s) 0,1');
+
+    it('should not highlight and not display the third item', () => {
+      browser.keys(['Tab', 'Enter']);
+    });
+    Terra.should.matchScreenshot('selected item(s) 0,1');
+
+    it('should unhighlight and deselect the second item', () => {
+      browser.addValue('ul li:nth-child(2)', 'Space');
```
emilyrohrbough: Is this sending the Space key to the second item? I haven't seen this combination before.

packages/terra-list/tests/wdio/multi-select-list-spec.js
```diff
@@ -0,0 +1,207 @@
...
+    });
+    Terra.should.matchScreenshot('selected item(s) 0 second item unhighlighted');
```
emilyrohrbough: nit: update to deselect second item?

packages/terra-list/tests/wdio/multi-select-list-spec.js
```diff
@@ -0,0 +1,207 @@
...
+    Terra.should.matchScreenshot('first and third item highlighted');
+
+    Terra.should.beAccessible();
+  });
+
+  describe('Multi Select List Focus', () => {
+    before(() => {
+      browser.url('/#/raw/tests/terra-list/list/multi-select-list/default-multi-select-list');
+      browser.waitForExist('ul');
+      browser.keys(['Tab', 'Enter', 'Tab', 'Enter', 'Shift', 'Tab', 'Shift']);
```
emilyrohrbough: This is sending quite a few keys here, why `Shift`, `Tab`, `Shift`? it might be beneficial to add an inline comment here to  describe the outcome

packages/terra-list/tests/wdio/multi-select-list-spec.js
```diff
@@ -0,0 +1,207 @@
...
+        browser.keys(['Shift', 'Tab', 'Space', 'Shift']);
+      });
+      Terra.should.matchScreenshot('first and third item highlighted');
+
+      it('should highlight the selected items upon space keydown', () => {
+        browser.keys(['Tab', 'Space']);
+      });
+      Terra.should.matchScreenshot('first item highlighted third item unhighlighted');
+    });
+    Terra.should.beAccessible();
```
emilyrohrbough: Looks like this guy is by itself? I think this is for the initial render. It might be more readable to push this guy up next to the initial should.matchScreenshot call

packages/terra-list/tests/wdio/selectable-list-spec.js
tomleewu: Oops, I will rename that test case. The original Nightwatch test presses the first item again.
```javascript
'Triggers onChange for a selectable list': (browser) => {
  browser.url(`${browser.launchUrl}/#/raw/tests/terra-list/list/selectable-list/on-change-selectable-list`);
  browser.sendKeys('ul li:nth-child(1)', browser.Keys.SPACE);
  browser.assert.containsText('#current-index', '0');

  browser.sendKeys('ul li:nth-child(2)', browser.Keys.ENTER);
  browser.assert.containsText('#current-index', '1');

  browser.click('ul li:nth-child(1)');
  browser.assert.containsText('#current-index', '0');
},
```

packages/terra-list/tests/wdio/selectable-list-spec.js
tomleewu: See comment above.

packages/terra-list/tests/wdio/multi-select-list-spec.js
tomleewu: Yeah, the screenshots verify that it works. For some reason, I couldn't get 
```javascript
browser.element('ul li:nth-child(2)').keys('Space');
``` 
to work.

emilyrohrbough: Interesting! Good thing to know 😄 I wasn't sure if we could send keys to a specific element in wdio till now. Thanks!!

packages/terra-list/src/terra-dev-site/test/list/list-item/DefaultListItem.test.jsx
mjhenkes: @emilyrohrbough your argument makes sense to me. Remove list item visual regression, test list item through jest and list item in a list with wdio.

packages/terra-list/src/List.jsx
```diff
@@ -39,8 +39,13 @@ const List = ({
     customProps.className,
   ]);
 
+  let roleAttr;
+  if (children.length !== 0) {
```
bjankord: This logic conflicts with how we are setting a [role in the menu component](https://github.com/cerner/terra-core/blob/8c34cdca8c56eaaeb16b26a2b780e97934f662aa/packages/terra-menu/src/_MenuContent.jsx#L312). If a custom role is set like we do in menu, we should use that first, then apply this logic.

packages/terra-list/src/ListItem.jsx
```diff
@@ -68,7 +68,7 @@ const ListItem = ({
   }
 
   return (
-    <li {...customProps} className={listItemClassNames} ref={refCallback}>
+    <li {...customProps} role="option" aria-selected={isSelected} className={listItemClassNames} ref={refCallback}>
```
bjankord: Not every list item will be a role of `option`. For the base list item, I'd keep the role as its default HTML semantics, e.g. no role, and then implement the role="option" in the selectable list items: 
* https://github.com/cerner/terra-core/blob/8c34cdca8c56eaaeb16b26a2b780e97934f662aa/packages/terra-list/src/MultiSelectList.jsx#L71
* https://github.com/cerner/terra-core/blob/8c34cdca8c56eaaeb16b26a2b780e97934f662aa/packages/terra-list/src/SelectableList.jsx#L70
* https://github.com/cerner/terra-core/blob/8c34cdca8c56eaaeb16b26a2b780e97934f662aa/packages/terra-list/src/SingleSelectList.jsx#L69

We'll want logic that defaults to "option" in the selectable list but takes a custom role for when we use the selectable lists in terra-menu.

packages/terra-list/tests/wdio/multi-select-list-spec.js
```diff
@@ -0,0 +1,209 @@
...
+      it('should unhighlight the third item upon enter', () => {
+        browser.keys(['Tab', 'Enter']);
+      });
+      Terra.should.matchScreenshot('first item highlighted third item unhighlighted');
+    });
+
+    describe('Space Bar', () => {
+      before(() => browser.url('/#/raw/tests/terra-list/list/multi-select-list/default-multi-select-list'));
+
+      it('should highlight the selected items upon space keydown', () => {
```
emilyrohrbough: "highlight" seems like the wrong word here reading this interaction. Is this selecting the first item on space or just indication the item has focus without selection?

PR #1577, opened 2018-06-05
[terra-form-textarea] Add setRef prop

noahbenham:
### Summary
This exposes a ref to the textarea field for advanced functionality such as managing focus, selection, or animations. CareAware Connect requires it for focusing & resizing the textarea.

### Additional Details
I explored a [polyfill](https://www.npmjs.com/package/create-react-ref) for [React 16.3](https://reactjs.org/docs/forwarding-refs.html)'s `createRef` & `forwardRef`. While this could be cleaner, the syntax was problematic. If anyone has a good solution, I'd be happy to modify this PR accordingly.

@cerner/terra


packages/terra-form-textarea/tests/jest/Textarea.test.jsx
```diff
@@ -12,7 +12,6 @@ it('should render a TextArea when all the possible props are passed into it', ()
   const textarea = (
     <Textarea
       name="description"
-      cols={15}
```
emilyrohrbough: why the removal?

emilyrohrbough:
@noahbenham Is there an issue logged for this?

noahbenham:
I don't believe this has an associated issue, @emilyrohrbough. Does it need one?

packages/terra-form-textarea/tests/jest/Textarea.test.jsx
noahbenham: It's not a supported prop type, I didn't see much value in testing it.. besides specifying columns for a textarea doesn't make much sense.

packages/terra-form-textarea/src/Textarea.jsx
```diff
@@ -70,6 +70,11 @@ const propTypes = {
    * The value of the textarea. Use this to create a controlled textarea.
    */
   value: PropTypes.string,
+  /**
+   * Callback to set ref of the textarea. Useful for advanced functionality
+   * such as managing focus, selection, or animations.
+   */
+  setRef: PropTypes.func,
```
StephenEsser: There's somewhat of a standard for this callback name: `refCallback`.

Here are a few examples:

https://github.com/cerner/terra-core/blob/master/packages/terra-scroll/src/Scroll.jsx#L17
https://github.com/cerner/terra-core/blob/master/packages/terra-list/src/List.jsx#L22
https://github.com/cerner/terra-core/blob/master/packages/terra-tabs/src/_TabMenu.jsx#L25

packages/terra-form-textarea/src/Textarea.jsx
```diff
@@ -188,7 +195,10 @@ class Textarea extends React.Component {
 
     return (
       <textarea
-        ref={(textarea) => { this.textarea = textarea; }}
+        ref={(textarea) => {
+          this.textarea = textarea;
```
StephenEsser: This is probably frowned upon, but I wanted to mention that you already have access to this textarea ref from the component ref. This is a part of the public API.

```
<Textarea ref={(ref) => this.componentRef = ref} />
```
```
console.log(this.componentRef) // Class reference
```

```
console.log(this.componentRef.textarea); // Dom node
```

noahbenham: I learned something new today. Is this PR still desired? I'd be happy to close it, looks like this solves our needs.

StephenEsser: I don't know if it's appropriate to access variables this way, but it does work. 

It might be worth doing a little investigation and discuss if this is considered an implementation leak or a valid way to access underlying references. 

It gets a little blurry when accessing an exposed / public internal variable.

noahbenham: I suspect this is probably an anti-pattern, unless we expect the private `textarea` variable name to persist. Although it's somewhat "cleaner" in terms of implementation, it feels wrong to expect a private variable to be maintained as a public API contract.

tbiethman: As far as I know, we've only ever considered passivity within the context of propTypes and module exports. The second we expose the entire instance itself as an API, we put ourselves in a really tricky position long-term.

noahbenham:
test this please

noahbenham:
@emilyrohrbough I hesitate to add `setRef` to components where it's not actively needed by a consumer _right now_, since the OCS transition will change prop names and what's supported. Also, once we upgrade to React 16.3+, it might be better to change these props from `setRef` to simply `ref`.

If you think an issue would still be relevant given this, I'd be happy to log one.

packages/terra-form-textarea/tests/jest/Textarea.test.jsx
JakeLaCombe: It's actually an html attribute on the dom textarea. https://www.w3schools.com/tags/att_textarea_cols.asp. Does it produce errors if we keep it?

PR #1581, opened 2018-06-08
[terra-avatar] Fix image flicker (when possible)

### LINKED ISSUE #1334: [terra-avatar] Image state is reset on prop updates
# Issue Description
I noticed that an image I provided to the Avatar component was flickering when receiving new props. 

The flickering is occurring because the Avatar provides a new element instance to the Image as the placeholder with each render (https://github.com/cerner/terra-core/blob/master/packages/terra-avatar/src/Avatar.jsx#L79), which causes the Image's internal state to reset and display the Image placeholder for a short time. You can see the equality check in the Image that's causing the problem here: (https://github.com/cerner/terra-core/blob/master/packages/terra-image/src/Image.jsx#L67).

We can fix this by defining single instances of each placeholder type and passing a consistent reference to the child Image component when the placeholder has not changed.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior
The Avatar should provide its child Image with a consistent placeholder prop when the placeholder has not changed. This will keep the Image's internal state from updating unnecessarily.

## Current Behavior
The Avatar provides its child Image with a new placeholder prop, even the placeholder has not changed. This causes the unnecessary and visible updates to the Image.

Blackop778:
### Summary
Fixes the image briefly flickering when updating Avatar props by storing the image in the state so a new image doesn't have to be generated upon rerendering if `image`, `alt`, and `variant` props haven't been updated.

I looked into hooking `shouldComponentUpdate` in terra-image but it had no effect on the issue.

There is a **[test page](https://terra-core-deployed-pr-1581.herokuapp.com/#/tests/terra-avatar/avatar/update-avatar)** added but the effect will have to manually be observed because I could not get an automated test to see the flicker.

### Additional Details
Closes #1334 

@cerner/terra


packages/terra-avatar/src/Avatar.jsx
```diff
@@ -43,56 +44,92 @@ const defaultProps = {
...
+      'avatar',
+      attributes.className,
+    ]);
+
+    const avatarTextClassNames = cx([
+      { 'avatar-text-small': initials && initials.length === 3 },
+      { 'avatar-text-large': initials && initials.length === 2 },
+    ]);
+
+    const icon = <span className={avatarIconClassNames} />;
```
emilyrohrbough: This needs removed, along with `avatarIconClassNames` as you moved where this is used.

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -43,56 +44,89 @@ const defaultProps = {
...
-    <div {...attributes} aria-label={ariaLabel} className={avatarClassNames}>
-      {avatarContent}
-    </div>
-  );
-};
+  componentWillReceiveProps(newProps) {
+    // If we have an image to display (they take precedence) and one of its attributes have changed
+    if (newProps.image && (newProps.image !== this.props.image ||
+      newProps.alt !== this.props.alt || newProps.variant !== this.props.variant)) {
+      const { alt, image, variant } = newProps;
```
yuderekyu: Nit: for reusability, would it be better this into a static function
```javascript
//not sure on the function name
static <function> (props) {
  const { alt, image, variant } = props;
  const imageComponent = Avatar.generateImage(image, variant, alt);
  this.setState({ imageComponent });
}
//call it in both places
Avatar.function(newprops);
```
This way we won't repeat the logic twice.

Blackop778: Calling `this.setState({ imageComponent });` from the constructor is a runtime error and reassigning state after construction is bad practice, so the most a static function for this could do is extract the individual variables from props. If you still think I should go ahead and make a static method for it, I'll do so.

yuderekyu: ahhhh missed that, good catch 👍

packages/terra-avatar/src/Avatar.jsx
```diff
@@ -43,56 +44,89 @@ const defaultProps = {
...
+
+      const imageComponent = Avatar.generateImage(image, variant, alt);
+      this.setState({ imageComponent });
+    }
+  }
+
+  render() {
+    const { alt,
```
yuderekyu: Nit: format is a lil off. We can put the props and `} = this.props` on new lines

PaulSolDev:
Functional review complete.  Some image resizing was noted on Edge and IE10/11 (and later Chrome on further investigation by @Blackop778 ) when changing between the initials v. image versions of the Avatar.  This variance was noted to exist in Master and is unrelated to the work completed in this PR.  A new issue will be logged to address the sizing issue.  In terms of the PR in question, no new issues identified were identified.

PR #1588, opened 2018-06-11
[terra-abstract-modal] Nightwatch -> Wdio

tomleewu:
### Summary
Converting `terra-abstract-modal` nightwatch tests to webdriverio.

Part of #1302.

@cerner/terra


packages/terra-abstract-modal/tests/wdio/abstract-modal-spec.js
```diff
@@ -0,0 +1,138 @@
+/* global before, browser, Terra */
+
+describe('Abstract Modal', () => {
+  before(() => browser.setViewportSize(Terra.viewports('medium')[0]));
+
+  describe('Append Class', () => {
+    before(() => {
+      browser.url('/#/raw/tests/terra-abstract-modal/abstract-modal/abstract-modal-append-class');
+      browser.setViewportSize(Terra.viewports('medium')[0]);
```
emilyrohrbough: Since setViewport is defined in the Abstract Modal before hook, all following instances can of setting the viewport can be removed.

tomleewu: I'll remove setViewport from `Abstract Modal` since I want to customize the viewport for `Content Overflow` tests. 

Otherwise the viewport I'm setting in `Content Overflow` will be propagated to all the other tests.

emilyrohrbough: Would it make sense for `Content Overflow` steps should set and then readjust the viewport back?

packages/terra-abstract-modal/tests/wdio/abstract-modal-spec.js
```diff
@@ -0,0 +1,115 @@
...
+      selector: 'div[role=document]',
+      viewports: [overflowModal],
+    });
+  });
+
+  describe('Custom Props', () => {
+    before(() => browser.url('/#/raw/tests/terra-abstract-modal/abstract-modal/abstract-modal-custom-props'));
+    const viewport = Terra.viewports('medium')[0];
+
+    Terra.should.matchScreenshot({ viewports: [viewport], selector: 'div[role=document]' });
```
bjankord: Rather than taking a screenshot, this test should check for the custom props, looks like the ID attribute is the only custom prop set on this. This test could also be replaced with a jest test.

PR #1615, opened 2018-06-18
Paginator update props

### LINKED ISSUE #1609: Terra-Paginator needs to sync new selectedPage value
# Issue Description
Currently the Paginator does not re-sync the selectedPage props when that prop is updated. We have a search box which rebinds the information and replaces it in the table.

screen shot 2018-06-15 at 2 33 56 pm

## Issue Type
- [ ] Bug

## Expected Behavior

Terra should implement a `componentWillReceiveProps ` or the new `getDerivedStateFromProps` to sync the selectedPage to the correct value

## Steps to Reproduce
1. Create a function that will change how many items are in the paginator
2. Go to the next page
3. Change the function to return less or more items, essentially filtering them
4. Component is in an incorrect state

## Environment
* Component Version: 1.4.0
* Browser Name and Version: All
* Operating System and version (desktop or mobile): All

JakeLaCombe:
### Summary
Fixes #1609


packages/terra-paginator/src/Paginator.jsx
```diff
@@ -33,6 +33,19 @@ const propTypes = {
 };
 
 class Paginator extends React.Component {
+  static getDerivedStateFromProps(props, state) {
```
bjankord: With using `getDerivedSateFromProps`, we need to polyfill it for react 16.2 since that's our lower bound version of react. To do this, you can [import a polyfill from react-lifecycles-compat](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/_Menu.jsx#L4) and [wrap the paginator component like we do in the select component code](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/_Menu.jsx#L285-L289).

Then you'll want to add the react-lifecycles-compat modules as a dependency to this package [similar to terra-select](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/package.json#L35).

CoryMcDonald:
I noticed that the paginator component doesn't seem to have internationalization? Specifically the Next/Previous links
https://github.com/cerner/terra-core/blob/master/packages/terra-paginator/src/Paginator.jsx#L172

Is this intended?

CoryMcDonald:
Oh! No, I didn't realize that there was an issue log for that 😄 Thanks! I'll review this PR

JakeLaCombe:
Going with a new approach in https://github.com/cerner/terra-core/pull/1621. Closing this pull request.

Issue #1370, opened 2018-03-23
Terra.should.beAccessible() failures


cwalten:
# Issue Description
Our project recently moved to WebdriverI/O following the terra-toolkit upgrade. With that, we are wanting to use the beAccessible helper service, but it is failing on some Terra Core components.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

## Expected Behavior
The Terra components should be able to pass the beAccessible service in terra-toolkit

## Current Behavior
The following 3 failures I believe are linked back to implementation details on the Terra side:

- No aria-label on Search Field's input element. `search-input` was suggested in #1619 but will need translations.
- No aria-label or alt attribute on the img tag for Demographics Banner
- No aria-label for the ... button in Collapsible Menu Button

BenBoersma:
I haven't had much luck recreating the described issue for Demographics Banner. Can you give an example of how you're using it?

bjankord:
@BenBoersma I tested the axe browser extension on https://engineering.cerner.com/terra-core/#/raw/tests/terra-demographics-banner/demographics-banner/populated-demographics-banner and I too wasn't able to reproduce the issue with the demographics banner. I believe this has been resolved in a release of terra-image.

bjankord:
The collapsible menu issue has been resolved in https://github.com/cerner/terra-core/pull/1456

bjankord:
There was some prior discussion on what to do with the aria-label for for the input and button [here](https://github.com/cerner/terra-core/pull/1619#discussion_r197267778) but we never landed on anything.

I'd recommend we use the string, Search, for the aria-label on the input and the string, Submit search, on the aria-label for the button. We should have translations for the string, Search. We'll need to get translations for the string, Submit search.

BenBoersma:
There was an offline discussion about the search-field aria-label, where the consensus was to add a new prop to terra-search-field called inputAttributes that would pass down custom props to the Input component to use. This way the aria-label wouldn't be constrained to a generic label and the search-field could be labeled as something like "Personnel search".

bjankord:
This has now been resolved.

PR #1619, opened 2018-06-19
[search-field] Nightwatch => Wdio

Blackop778:
### Summary
Converted Nightwatch tests to WebdriverIO, removed a duplicate test page, and added clickable text to a couple of test pages.

Accessibility issues logged in #1370

### Additional Details
Part of #1302

@cerner/terra


packages/terra-search-field/src/terra-dev-site/test/search-field/DefaultSearchField.test.jsx
```diff
@@ -1,10 +1,9 @@
...
-import Base from 'terra-base';
 import SearchField from '../../../SearchField';
 
-const locale = document.getElementsByTagName('html')[0].getAttribute('lang');
-
-export default () =>
-  <Base locale={locale}>
+export default () => (
+  <div>
+    <h4>Extraneous Clickable Text</h4>
```
emilyrohrbough: why are we adding `<h4>Extraneous Clickable Text</h4>` to the test examples?

Blackop778: Clicking on it removes the blinking mouse cursor from the `<input>` element which can tend to ruin tests.

emilyrohrbough: Lets add something like [this](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/tests/wdio/select-spec.js#L175) to set the cursor to be transparent. We will want to capture the focus states of the elements during screenshots and adding this extra text is not desirable.
```
        browser.execute(() => {
          // Removes the blinking cursor to prevent screenshot mismatches.
          document.querySelector('input').style.caretColor = 'transparent';
        });
```

packages/terra-search-field/tests/wdio/search-field-spec.js
```diff
@@ -0,0 +1,252 @@
...
+  describe('With Default Value', () => {
+    before(() => browser.url('/#/raw/tests/terra-search-field/search-field/search-field-default-value'));
+
+    Terra.should.matchScreenshot('default value');
+    Terra.should.beAccessible({ rules: ignoredA11y });
+
+    it('should enter a new search term', () => {
+      browser.click('input');
+      for (let i = 0; i < 7; i += 1) {
+        browser.keys('Backspace');
```
JakeLaCombe: I would expect the backspace to be in an array like ['Backspace'] Otherwise, I think the test is writing all those values to the input, and getting cleared when addValue is called.

Blackop778: `addValue` appends to the input but `setValue` would overwrite the input. I decided to use `browser.keys('Backspace');` to test to be sure that the default value could actually be cleared.

packages/terra-search-field/tests/wdio/search-field-spec.js
```diff
@@ -0,0 +1,252 @@
...
+    });
+
+    Terra.should.matchScreenshot('with text');
+
+    it('should scroll text that is too long', () => {
+      browser.addValue('input', ' ipsum dolor sit amet');
+    });
+
+    Terra.should.matchScreenshot('scrolled text');
+    Terra.should.beAccessible({ rules: ignoredA11y });
```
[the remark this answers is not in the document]
Blackop778: The one accessibility rule that is being ignored has an issue for it at #1370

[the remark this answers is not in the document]
emilyrohrbough: @Blackop778 what is the fix for the issue seen here? Can we resolve it now? If not, can we add a commented about the ignoreA11y const to indicate an issue has been logged.

packages/terra-search-field/tests/wdio/search-field-spec.js
emilyrohrbough: We should still be able to use an array like Jake suggested if we update to use 

```
 browser.keys(['Backspace', 'Backspace', 'Backspace',...]);
```
Though I have no preference on which approach is cleaner. 

In regard to check if the default value was actually cleared, there is no assertion or screenshot taken to ensure the value is actually empty before we use `addValue('input', 'Lorem')`

packages/terra-search-field/tests/wdio/search-field-spec.js
```diff
@@ -0,0 +1,252 @@
...
+      browser.execute(() => {
+        // Removes the blinking cursor to prevent screenshot mismatches.
+        document.querySelector('input').style.caretColor = 'transparent';
+      });
+    });
+
+    Terra.should.matchScreenshot('with too short text');
+
+    it('should enter a long enough search term', () => {
+      browser.addValue('input', 'm');
```
emilyrohrbough: nit:  can we use `setValue` with an input that is obviously long enough here to improve clarity on what is happening in this test?

packages/terra-search-field/tests/wdio/search-field-spec.js
```diff
@@ -0,0 +1,252 @@
...
+
+    it('should search using enter', () => {
+      browser.addValue('input', ' ipsum');
+      browser.keys('Enter');
+    });
+
+    Terra.should.matchScreenshot('extended search');
+  });
+
+  describe('Delayed', () => {
```
emilyrohrbough: Nit: update to AutoSearch Delayed

Then, we should add non-screenshot assertion here every 250 ms like nightwatch did to ensure the search was not yet triggered. A screenshot test overly appropriate here because we are testing the callback. It might make sense to move this to jest.

packages/terra-search-field/tests/wdio/search-field-spec.js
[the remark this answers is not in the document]
Blackop778: It appears to be that the Search Button is missing an `aria-label`. I'll go ahead and resolve that.

Blackop778:
While no test is explicitly is named as such, Callback, Auto Search Delayed, and Minimum Length all test Auto Search being enabled.

packages/terra-search-field/tests/wdio/search-field-spec.js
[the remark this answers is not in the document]
emilyrohrbough: Great! Thank you

packages/terra-search-field/tests/wdio/search-field-spec.js
Blackop778: Ended up moving this to jest, but left the test page.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -198,6 +198,7 @@ class SearchField extends React.Component {
           onChange={this.handleTextChange}
           disabled={isDisabled}
           aria-disabled={isDisabled}
+          aria-label="Text to search for"
```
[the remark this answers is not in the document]
emilyrohrbough: If new translations are required for this a11y fix, we should push this to a new PR.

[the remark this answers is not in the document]
bjankord: Yeah I'd go with the phrase "Search" for the aria-label. Agreed, we should tackle that work in another PR once we get translations.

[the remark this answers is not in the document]
Blackop778: "Search" is already translated for SearchField because that's what is used for the `aria-label` for the button @bjankord.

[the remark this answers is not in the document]
emilyrohrbough: Lets update the issue you reference with these details that the button's aria-label already uses the term search. Then we know translations of some sort need requested. IMO it seems the aria-label we would want would be `search-input` or the equiv. 

Will remove my block. Looks good 😄 Nice Job Nathan

PR #1639, opened 2018-06-23
Fix extra space displaying below image avatar

### LINKED ISSUE #1626: [terra-avatar] Image avatars take up more space than facility or user avatars
# Issue Description
When loading a page with an image avatar in IE 10/11, Edge, or after resizing a chrome window the image avatar will take an additional 5-6 pixels (from testing) of space below the image.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Avatar takes up as much space as it's image

## Current Behavior
Image avatars in IE 10/11, Edge, Chrome after resizing take an additional 5-6 pixels (from testing) of space below the image

## Steps to Reproduce
Create a border around an Avatar component that is displaying an image then open the page in IE or Edge and observe the extra space taken at the bottom, or open page in Chrome and resize the window.

bjankord:
### Summary
* Resolves #1626
* Setting `vertical-align: middle` on the container with `display: inline-flex` seems to have resolved this in the noted browsers.
* Added test case for wdio

screen shot 2018-06-23 at 9 25 42 am
screen shot 2018-06-23 at 9 25 18 am
screen shot 2018-06-23 at 9 21 29 am
screen shot 2018-06-23 at 9 20 27 am

More info on this issue: https://dev.to/christiankaindl/the-strange-img-gap-in-html


bjankord:
Need to add wdio screenshot for new test case

packages/terra-avatar/src/Avatar.scss
```diff
@@ -47,4 +48,8 @@
   .avatar-text-small {
     font-size: var(--terra-avatar-font-size-small, 1.143rem);
   }
+
+  .extra-white-space-fix {
```
JakeLaCombe: Where is this class used? I'd expect a markup update in this PR

bjankord: Ah that isn't needed, removed in this commit: https://github.com/cerner/terra-core/pull/1639/commits/f536b389d54312dfc16760e13d1cacb20093e6d9

PR #1647, opened 2018-06-25
Fix error ResizeObserver loop limit exceeded

ap056120:
### Summary
Resolves the ResizeObserver loop limit exceeded error that is thrown when Responsive elements are loaded.

#### Additional Details
When a ResponsiveElement is mounted for the first time, the ResizeObserver receives an event with the size of the ResponsiveElement’s parent, which the ResponsiveElement uses to render the correct component in place of the div it used to initially get the ref.  In cases where the ResponsiveElement’s parent is wrapped to its contents, the act of rendering the actual responsive component will cause another resize event to be triggered, which the ResizeObserver will detect as a potential infinite loop.  The error thrown by the ResizeObserver is intended as a warning, doesn’t actually prevent infinite loops, and the ResponsiveElement already has logic to prevent accidental infinite loops, so the error just gets in the way.  In our case, however, other components we are consuming don’t realize this and overreact to the error, so we need to suppress it.  We can fix this by delaying the second rendering (the first one where an actual child component is rendered) for one frame after the first rendering with the ref div.

The 'ResizeObserver loop limit exceeded' error is not thrown in the JavaScript console of a browser, but it can be caught using a simple [extension](https://chrome.google.com/webstore/detail/javascript-errors-notifie/jafmfknfnkoekkdocjiaipcnmkklaajd?hl=en) that is available for Chrome/Firefox: 

The error can currently be demonstrated in [Kaiju](https://kaiju.cerner.com/projects/dark-gyaos-TAZ-MA/workspaces/muddy-shadow-_9LZPw/preview) as well. This [JS Fiddle]( https://jsfiddle.net/que_etc/ba1ad26e/) demonstrates that the loop limit is 0, and any size modification within the resize handler triggers the warning.

[Documentation for ResizeObserver](https://developers.google.com/web/updates/2016/10/resizeobserver)

This [GitHub issue](https://github.com/WICG/ResizeObserver/issues/38) proposes an alternate solution to the loop limit problem, but the current implementation of the ResizeObserver doesn’t allow that to solution to work.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -65,6 +65,7 @@ Cerner Corporation
 - Tom Wu [@tomleewu]
 - Cory McDonald [@corymcdonald]
 - Anthony Ross [@AnthonyRoss]
+- Ajay Philip Sabu [@ap056120]
 
 [@ryanthemanuel]: https://github.com/ryanthemanuel
 [@Matt-Butler]: https://github.com/Matt-Butler
@@ -132,3 +133,4 @@ Cerner Corporation
 [@tomleewu]: https:/github.com/tomleewu
 [@CoryMcDonald]: https://github.com/CoryMcDonald
 [@AnthonyRoss]: https://github.com/AnthonyRoss
+[@ap056120]: https://github.com/ap056120
--- a/packages/terra-responsive-element/CHANGELOG.md
+++ b/packages/terra-responsive-element/CHANGELOG.md
@@ -4,6 +4,11 @@ Changelog
 Unreleased
 ----------
 
+3.2.0 - (June 25, 2018)
+------------------
+### Fixed
+* Fixed ResizeObserver loop limit exceeded error for Responsive Elements
+
 3.1.0 - (June 22, 2018)
 ------------------
 ### Changed
--- a/packages/terra-responsive-element/src/ResponsiveElement.jsx
+++ b/packages/terra-responsive-element/src/ResponsiveElement.jsx
@@ -52,8 +52,11 @@ class ResponsiveElement extends React.Component {
 
   componentDidMount() {
     if (this.container) {
-      this.handleResize(this.container.innerWidth);
-      this.resizeObserver = new ResizeObserver((entries) => { this.handleResize(entries[0].contentRect.width); });
+      this.resizeObserver = new ResizeObserver((entries) => {
+        window.requestAnimationFrame(() => {
+          this.handleResize(entries[0].contentRect.width);
+        });
+      });
       this.resizeObserver.observe(this.container);
     } else {
       this.handleResize(window.innerWidth);
--- a/packages/terra-responsive-element/src/ResponsiveElement.jsx
+++ b/packages/terra-responsive-element/src/ResponsiveElement.jsx
@@ -92,7 +92,9 @@ class ResponsiveElement extends React.Component {
     }
 
     if (this.state.element !== element) {
-      this.setState({ element });
+      window.requestAnimationFrame(function() {
+        this.setState({ element });
+      });
     }
   }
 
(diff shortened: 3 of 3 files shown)


emilyrohrbough:
Is there a terra issue logged for this?

packages/terra-responsive-element/tests/jest/__snapshots__/ResponsiveElement.test.jsx.snap
```diff
@@ -1,7 +1,3 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`should render a default component 1`] = `
-<div>
-  Default
-</div>
-`;
+exports[`should render a default component 1`] = `<div />`;
```
StephenEsser: I don't think any tests should have been affected by this change, was this file auto generated or manually updated?

ap056120: Auto Generated.

StephenEsser: This snapshot dropped the default text. I would try compiling and running the tests again. 

If this change is a result of the changes in ResponsiveElement.jsx it is non-passive and is not rendering the element. 

It looks like this test renders the initial div but never renders the defaultElement.

```
it('should render a default component', () => {
  const responsiveElement = shallow(<ResponsiveElement defaultElement={<div>Default</div>} />);
  expect(responsiveElement).toMatchSnapshot();
});
```

shinepd: It doesn't render the default text in the same frame, likely due to the call to requestAnimationFrame. Are we expecting the default text to be rendered at the same time as the initial div?

StephenEsser: The call to freeze the repaint breaks this test. My concern is that it will also cascade and break a significant amount of consumer tests down the chain.

StephenEsser: It also renders snapshot testing responsive elements useless, as the expected element will never be shallow rendered.

shinepd: An alternative to the requestAnimationFrame would be to call handleResize before the resizeObserver is attached with the current size of the of the container. Would this be acceptable?

```
  componentDidMount() {
    if (this.container) {
      this.handleResize(this.container.innerWidth);
      this.resizeObserver = new ResizeObserver((entries) => { this.handleResize(entries[0].contentRect.width); });
      this.resizeObserver.observe(this.container);
    } else {
      this.handleResize(window.innerWidth);
      window.addEventListener('resize', this.handleWindowResize);
    }
  }
```

shinepd:
@emilyrohrbough we didn't log an issue for this change. I'm not aware of an open issue in terra-core pertaining to this change.

packages/terra-responsive-element/tests/jest/__snapshots__/ResponsiveElement.test.jsx.snap
StephenEsser: Calling handResize before observe might cause a triple render

mjhenkes:
@ap056120 @shinepd, We've talked internally and we've come up with another fix. We're going to close this pull request and open up one of our own by end of day tomorrow at the latest. Thanks for the debugging time you guys put into this.

MadhavaNaresh:
@mjhenkes Wanted to know if this issue was fixed. I'm getting the ResizeObserver loop error when integrating a react component to a mpages view through workflow_server

mjhenkes:
@MadhavaNaresh if you are seeing this issue again, it would be best to log a new issue.

PR #1648, opened 2018-06-26
Centered action footer is block

### LINKED ISSUE #1640: isBlock Prop on CenteredActionFooter
# a New isBlock property on the CenteredActionFooter component

## Description
the isBlock Property will stretch the width of the children to take up the width of the parent div inside the CenteredActionFooter, right now in order to get the children to stretch to the width of the footer the parent div needs to have a style set to width:'100%', the consumer shouldn't have to set this prop in order to get the children to take up the whole width

Here is a screenshot of what it looks like right now without setting the width of the parent div of the children to 100%:
screen shot 2018-06-22 at 10 38 17 am

this is what it should look like with an "isBlock" property:
screen shot 2018-06-22 at 5 08 20 pm

dev-cprice:
### Summary

* resolves #1640 
* add `isBlock` prop to `<CenteredActionFooter />` that forces the center socket node to fill the entirety of the footer
* add test cases
* update docs

@cerner/terra


packages/terra-action-footer/src/CenteredActionFooter.jsx
```diff
@@ -12,27 +12,37 @@ const propTypes = {
...
+    { 'is-block': isBlock },
+  ]);
+
   return (
     <ActionFooterContainer
       {...customProps}
       className={centeredActionFooterClassNames}
     >
-      {center}
+      {center && cloneElement(center, { className: isBlockClassName })}
```
yuderekyu: Wouldn't this override custom classes applied to center?

tbiethman: It seems like we almost want a new variant of ActionFooter ('BlockButtonFooter' or something less awful) that simply takes an array of Buttons. That footer could own and style container as it needs to without bleeding styles into consumer-provided components.

dev-cprice: I totally missed @bjankord's suggestion in Teams. Thoughts on this instead of what I currently have?
screen shot 2018-06-26 at 9 38 20 am

tbiethman: What styles would the `isBlock` prop be changing on the CenteredActionFooter? 

I could see having a `layoutDirection` prop (with `horizontal` or `vertical` values) that would change the flex-direction accordingly. With it set to vertical, an array of block Buttons could be rendered without the need of a wrapper.

We could also play with the align-items value and change to to `strech` when vertical to prevent the need for `isBlock` on the Buttons, but I'm not sure if that behavior is always going to be desired.

dev-cprice: Here's @bjankord's suggested change: 6395924e497ca9978903e9ff390454fa140dc183

deployed [here](https://terra-core-deployed-pr-1648.herokuapp.com/#/components/terra-action-footer/action-footer/centered)

bjankord: > What styles would the isBlock prop be changing on the CenteredActionFooter?

The display style of the action footer container would be changed from `flex` to `block`.

bjankord: Ah I see the issue with that. I like the idea of having a `layoutDirection` prop to switch `flex-direction`. I do think the stacked button use case is the primary reason the CenteredActionFooter exists. I wouldn't be opposed to setting `align-items: stretch` if `layoutDirection` is set to `vertical`. If it becomes a concern those are coupled, we could split the ability to set `align-items: stretch` to a separate prop.

dev-cprice: @bjankord so this? One side effect of this is that buttons automatically fill out the center socket. This of course could be alleviated if we decouple the `align-items: stretch;` from `layoutDirection`.
image

packages/terra-action-footer/tests/jest/CenteredActionFooter.test.jsx
```diff
@@ -73,5 +96,13 @@ describe('CenteredActionFooter', () => {
         expect(centeredActionFooter.find('ActionFooterContainer').length).toBe(1)
       ));
     });
+
+    describe('with a block center action', () => {
```
Blackop778: Do we want to repeat the test name?

emilyrohrbough: No each test step should be unique. Good catch.

packages/terra-action-footer/src/CenteredActionFooter.jsx
bjankord: Discussed in a meeting last Friday a path forward for this. Plan is to expose [_ActionFooterContainer.jsx](https://github.com/cerner/terra-core/blob/master/packages/terra-action-footer/src/_ActionFooterContainer.jsx) as a new component named BlockActionFooter.

To update, we can rename `_ActionFooterContainer.jsx` to `BlockActionFooter.jsx`. Then in the other components, update the import of `_ActionFooterContainer.jsx` to `BlockActionFooter.jsx`.

bjankord:
Talked with @dev-cprice about this PR. Plan is to close this out as its grown stale compared to master and make a new PR to expose `_ActionFooterContainer.jsx` as a new component named `BlockActionFooter`.

To update, we can rename `_ActionFooterContainer.jsx` to `BlockActionFooter.jsx`. Then in the other components, update the import of `_ActionFooterContainer.jsx` to `BlockActionFooter.jsx`.

PR #1655, opened 2018-06-27
Embedded content consumer wdio

saedar:
### Summary
Migrate `terra-embedded-content-consumer` from Nightwatch to WDIO.

### Additional Details
I just wanted to get this up early. There is still some weird behavior with wdio throwing an exception every 3-6 runs that I'm trying to figure out.


saedar:
I think the exception issue has been dealt with. There's either something wrong with how `waitUntil()` works or I was using it wrong. Probably the second.

### Summary of Updates
* Removed the snapshot testing for the `data-status` example. They render inconsistently and cannot be snapshot tested at this time.
* Added iframe testing to `data-status`.
* Updated wait timers in other tests, as I misunderstood how it worked.

PR #1658, opened 2018-06-27
Implement request animation frame to correct loop limit error.

dkasper-was-taken:
### Summary
Update life cycle of renders for tabs and collapsible menu to handle consistent render.
Updated tabs, collapsible menu, and responsive element to use animation frames.

@cerner/terra


packages/terra-tabs/src/_CollapsibleTabs.jsx
```diff
@@ -92,6 +98,13 @@ class CollapsibleTabs extends React.Component {
     this.menuRef = node;
   }
 
+  resetCache() {
```
emilyrohrbough: out of curiosity, why did we move these values out of state?

dkasper-was-taken: The primary reason was to have the values be synchronus, even if the actual render was async, that way even if multiple renders occur, it's only uses the most up to date data. Since the animation frame is introducing a delay between 1-16ms, we don't reliably know if it will coincide with react's own batch updates. This initially lead to problems with missing refs, or incorrect sizes being used in calculations.

Also, previously the onResize was applying two back to back setStates.

PR #1664, opened 2018-06-28
[terra-form-select] Null check the active option before calling findByValue

### LINKED ISSUE #1663: [terra-form-select] - Null check active state for dynamic options
# Bug Report

## Description

Loading dynamic options into the Select can cause an error. The stored state value of the active option gets reset to null and a new search is performed that tries to use null as a search key. 

## Steps to Reproduce

1. Create a search variant Select.
2. Implement an onSearch callback that will update the available options.
3. Type or backspace the search.
4. Observe an error.

## Expected Behavior

Dynamic options should be loaded without using a null active value to perform a search.

## Possible Solution

Check for null before doing the findByValue search.

## Environment

* Component Name and Version:
* Browser Name and Version:
* Webpack Version:

StephenEsser:
### Summary
When options are changed dynamically the active option will be reset to null. 

This change checks to make sure the active option is valid before performing a search.

Fixes #1663

@cerner/terra


packages/terra-form-select/src/_MenuUtil.js
```diff
@@ -221,7 +221,7 @@ class MenuUtil {
       return selected === undefined ? options[0].props.value : selected.props.value;
     } else if (searchValue !== state.searchValue) {
       return options[0].props.value;
-    } else if (MenuUtil.findByValue(options, active)) {
+    } else if (active !== null && MenuUtil.findByValue(options, active)) {
```
yuderekyu: Curious, is it possible for active to ever be undefined?

StephenEsser: No, it shouldn't be possible. We also can't do `active &&` or a variation of `!active` because an empty string is a valid value and evaluates as false.

PR #1665, opened 2018-06-28
Correct terra-content-container IE10 'initial' style bug.

dkasper-was-taken:
### Summary
Change static normalizer styles to use explicit values, rather than 'initial' as it's not supported.

@cerner/terra


Blackop778:
Should we fix the few other `initial` values? (https://github.com/cerner/terra-core/search?l=SCSS&q=initial)

PR #1669, opened 2018-07-02
Input forward ref

### LINKED ISSUE #1653: Add the ability to forward refs for terra-form-input
# Feature Request
terra-form-input is the replacement for terra-form. One thing that the original implementation did was set the ref property on the html input element, so that the components input can be accessed through element.textInput. However, this is not present in terra-form-input, and it preventing packages such as terra-time-input from using this package.

## Description
Add functionality that allows us to either forward the ref, or have access to the dom input for the terra-form-input package. Either approach is acceptable, but I think we would have to uplift our code to support React 16.3 in order to use forwarded ref. In addtion, the forwarded ref may introduce a non passive change. You can view more here. https://reactjs.org/docs/forwarding-refs.html

JakeLaCombe:
### Summary
Resolves #1653 

After the war room discussion, we determined the best case for our apps to pass refs is through a refCallback prop attached to the input element. We have a need to do this for inputs, as terra-time-input and terra-date-time-picker needs this.


packages/terra-form-input/src/terra-dev-site/test/form-input/InputField.test.jsx
```diff
@@ -11,6 +11,11 @@ class InputFieldExample extends React.Component {
     };
 
     this.toggleInvalid = this.toggleInvalid.bind(this);
+    this.refCallback = this.refCallback.bind(this);
+  }
+
+  refCallback(input) {
+    this.input = input;
```
Blackop778: Should we do something with this callback to demonstrate a use case?

JakeLaCombe: I'll do the autofocus like I did with the other tests

packages/terra-form-input/tests/wdio/form-input-spec.js
```diff
@@ -145,6 +145,12 @@ describe('Form-Input', () => {
       });
     });
 
+    describe('RefCallback', () => {
+      beforeEach(() => browser.url('/#/raw/tests/terra-form-input/form-input/input-ref-callback'));
+
+      Terra.should.matchScreenshot({ viewports });
```
emilyrohrbough: This screenshot test doesn't seem to be proving the refcallback is triggered or valid. Can we add some sort on-screen text change or visual different that show the callback was called?

StephenEsser: I think the visual test should be removed in favor of a mocked jest test. Something similar to:
```
  it('should call refCallback', () => {
    const mockRefCallback = jest.fn();
    const wrapper = shallow(<Input refCallback={mockRefCallback} />);

    expect(mockRefCallback).toBeCalled();
  });
```

packages/terra-form-input/src/Input.jsx
```diff
@@ -100,6 +106,9 @@ class Input extends React.Component {
     return (
       <input
         {...attributes}
+        ref={(inputRef) => {
```
dkasper-was-taken: This can be simplified to just 
```
ref={refCallback}
```

packages/terra-form-input/tests/wdio/form-input-spec.js
JakeLaCombe: The blue border is the visual indication. When the page loads, it auto focuses to the given input, causing the blue border to appear.

I'd also be more favored as well towards the jest test.

emilyrohrbough: To someone unfamiliar with the test, a blue border is not a clear indicator of what is happening here. I likely would have assumed focus was applied. I also agree testing via jest seems appropriate here.

packages/terra-form-input/src/terra-dev-site/test/form-input/InputField.test.jsx
```diff
@@ -30,7 +30,7 @@ class InputFieldExample extends React.Component {
 
   render() {
     return (
-      <div>
+      <div style={{ caretColor: 'transparent' }}>
```
emilyrohrbough: We should not add this to our test examples directly. These examples are still functional verification candidate on PR reviews.

Issue #1670, opened 2018-07-02
Time Input Fails Accessibility at Desktop Level


JakeLaCombe:
# Bug Report

## Description
Time Input needs to be accessible.

## Steps to Reproduce
1. Go to https://engineering.cerner.com/terra-core/#/raw/tests/terra-time-input/time-input/twelve-hour/default
2. Do a lighthouse audit in desktop mode

## Additional Context / Screenshots
* Form Elements do not have associated labels. Minutes and Hours should have their own labels poentially.
* Color contrast is invalid on the mobile inputs. That needs designer feedback to correct.

## Expected Behavior
Accessible Time Picker

bjankord:
@neilpfeiffer Can we get some color updates on this for the default theme to make it match contrast ratios.

neilpfeiffer:
@JakeLaCombe What are the items that are currently failing with contrast ratio checks? Is it specifically the help-text for mobile inputs or is it the placeholder color for both desktop+mobile?

JakeLaCombe:
I'm not getting any accessibility failures on the placeholder elements. I think since it's the placeholder value that no audits are being fully performed on that element. The only failures flagging are color contrast on the help texts.

As for other failures, for the 12 hour clocks, an aria-label attribute needs to be added to both the meridiem display, and the select box that contains the values of the meridiems. We'd have to work through how we want those attributes to be applied to those elements.

bjankord:
Related: https://github.com/cerner/terra-core/issues/1840

PR #1672, opened 2018-07-03
Fix Time Input wrong duplication

### LINKED ISSUE #1671: Default Hour and Minute Attributes on Time Input are changed on the defaultProps object.
# Bug Report

## Description
terra-time-input has hourAttributes and minuteAttributes props with default values set as a hash. From looking at the implementation, these values are not new objects every time a new time input is initialized. This is a problem as the time input attaches an id to the object, and that id gets shared across all time inputs that do not have hourAttributes or minuteAttributes defined. Thus, several time inputs with the same input id can be initialized at once unexpectedly.

## Steps to Reproduce
Create a terra component composed of several different time inputs without setting hourAttributes or minuteAttributes.

https://engineering.cerner.com/terra-core/#/tests/terra-time-input/time-input/twelve-hour/mobile/#/tests/terra-time-input/time-input/twelve-hour/mobile

## Expected Behavior
The hourAttributes and minuteAttributes are not modified every time a time input is initialized.

## Possible Solution
Duplicate the hash inside the render method, or switch the default value to hourAttributes to undefined. The feasibility of the latter solution would have to be investigated.

JakeLaCombe:
### Summary
Fixes #1671 

As an added bonus, fixed some issues with accessibility.

### Additional Details
To test the duplicate id's, I did two things. Added a jest test, and added the accessibility test to wdio which tests for duplicate id's on a page.


packages/terra-time-input/tests/wdio/time-input-twelve-hour-mobile-spec.js
```diff
@@ -1,13 +1,18 @@
 /* global browser, Terra, before */
 const viewports = Terra.viewports('medium');
 
+const ignoredA11y = {
```
emilyrohrbough: Is there an issue logged regarding the inaccessible color contrast? If so, can we include the issue link in a commend so we know why we disabled this rule?

packages/terra-time-input/tests/jest/TimeInput.test.jsx
```diff
@@ -96,6 +96,23 @@ it('should render a 12 hour timepicker meridiem with buttons when viewed on a mo
   delete window.ontouchstart;
 });
 
+it('should not have duplicate ids on the page when multiple date pickers are initialized', () => {
```
emilyrohrbough: Since the consumer must supply a unique `name` to receive a unique `id` is this test needed?

packages/terra-time-input/tests/wdio/time-input-twelve-hour-mobile-spec.js
JakeLaCombe: Right here. I'll add the comment to this test. https://github.com/cerner/terra-core/issues/1670

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -516,6 +516,9 @@ class TimeInput extends React.Component {
       ...customProps
     } = this.props;
 
+    const instanceHoursAttrs = Object.assign({}, hourAttributes);
```
emilyrohrbough: I'm curious, does creating a new hours and minutes attributes object change how the id are attached to the rendered component?

packages/terra-time-input/tests/jest/TimeInput.test.jsx
JakeLaCombe: Yeah. If I left the names as time-input, time-input-2, time-input-3, it would generate three time-inputs all with the same id's of terra-time-hour-time-input and terra-time-minute-time-input. I want to ensure this test generates terra-time-hour-time-input-1,  terra-time-hour-time-input-2, and  terra-time-hour-time-input-3.

packages/terra-time-input/tests/wdio/time-input-twelve-hour-mobile-spec.js
emilyrohrbough: This comment seems to be mentioning accessibly in regards to incorrect labels, not color contrast. Do these go hand in hand?

packages/terra-time-input/src/TimeInput.jsx
JakeLaCombe: Yeah. What I found is that all instances of the time input share the same default prop object as opposed to having their own instance. If I rely on the default hourAttributes prop for a time input and it get's modified (like setting the id), that same id carries over to the next time input.

packages/terra-time-input/tests/wdio/time-input-twelve-hour-mobile-spec.js
JakeLaCombe: I'll update the issue. We should resolve all the accessibility issues in that github issue that couldn't be resolved in this commit.

packages/terra-time-input/tests/wdio/time-input-twelve-hour-mobile-spec.js
```diff
@@ -1,13 +1,19 @@
 /* global browser, Terra, before */
 const viewports = Terra.viewports('medium');
 
+// Color contrast will be resolved in https://github.com/cerner/terra-core/pull/1672
```
Blackop778: I don't believe this PR is resolving the color contrast issue

packages/terra-time-input/src/TimeInput.jsx
emilyrohrbough: oh, that is interesting. who knew 🤷‍♀️

packages/terra-time-input/tests/wdio/time-input-twelve-hour-mobile-spec.js
JakeLaCombe: One thing you'll learn. Even years of experience can't prevent copy paste errors like this, lol

packages/terra-time-input/src/TimeInput.jsx
```diff
@@ -529,12 +532,12 @@ class TimeInput extends React.Component {
       timeValue = 'T'.concat(hour, ':', this.state.minute);
     }
 
-    if (hourAttributes && !hourAttributes.id) {
-      hourAttributes.id = 'terra-time-hour-'.concat(name);
+    if (instanceHoursAttrs && !instanceHoursAttrs.id) {
```
tbiethman: Nitpicking here but you should be able to remove the presence check for instanceHoursAttrs (and instanceMinuteAttrs below).

PR #1677, opened 2018-07-03
[terra-form-select] - Added a customizable maxHeight

### LINKED ISSUE #1579: Terra Select - make size of list dropdown configurable
# Issue Description
the dropdown list height of Terra Select spills over the size of the modal

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
1. When user clicks on terra select field, the list remains within the modal.  
Have the list display below the dropdown field, with the dropdown list height filling any available space in the modal below the dropdown and 10px above the footer.

## Current Behavior
1. Click on terra select field to display list.  Note that list spills over the size of the modal.

## Steps to Reproduce
1. Click on terra select field to display list.  Note that list spills over the size of the modal.
image

[Consumer select component](https://pages.github.cerner.com/terra/cerner-consumer-theme/build/master/#/site/components/core/site/form-select/index)

StephenEsser:
### Summary
Added a first class prop for customizing the max height of the Select dropdown.

### Additional Details
There has been some concern and confusion from consumers when setting the max height of the dropdown using the custom dropdownAttrs. To alleviate some of this concern we have moved the maxHeight into a first class prop and provided an example on the documentation page for the recommended way to set it.

The previous way to set the maxHeight:
```
<Select dropdownAttrs={{ style: { maxHeight: '300px' } }}>
```
After this change:
```
<Select maxHeight={300}>
```

Resolves #1579 

@cerner/terra


packages/terra-form-select/src/_FrameUtil.js
```diff
@@ -15,24 +15,24 @@ class FrameUtil {
...
-    const maxHeight = style.maxHeight ? parseInt(style.maxHeight, 10) : Infinity;
-    const canFitBelow = maxHeight < spaceBelow || height < spaceBelow || spaceBelow > top;
+    const maximumHeight = parseInt(style.maxHeight || maxHeight, 10) || Infinity;
+    const canFitBelow = maximumHeight < spaceBelow || height < spaceBelow || spaceBelow > top;
     const availableSpace = canFitBelow ? spaceBelow : top;
 
     return {
       width,
-      maxHeight: Math.min(maxHeight, availableSpace - 10),
+      maxHeight: Math.min(maximumHeight, availableSpace - 10),
```
jmsv6d: Not really related to these changes, but what is the `- 10` for?

StephenEsser: It's an arbitrary padding between the dropdown menu and the window. 

When the dropdown is opened we calculate the maximum available space between the Select frame and the window to determine the maximum height the dropdown can be. The drop down will never open and expand larger than the available window space. 

image


The 10 pixels is the distance between the bottom of the dropdown and the window in this picture.

packages/terra-form-select/src/_FrameUtil.js
```diff
@@ -15,24 +15,24 @@ class FrameUtil {
...
-  static dropdownPosition(props, state, target, dropdown) {
+  static dropdownPosition(props, target, dropdown, maxHeight) {
     const style = Object.assign({}, props).style || {};
     const { height } = dropdown.getBoundingClientRect();
     const { bottom, width, top } = target.getBoundingClientRect();
 
     const spaceBelow = window.innerHeight - bottom;
-    const maxHeight = style.maxHeight ? parseInt(style.maxHeight, 10) : Infinity;
-    const canFitBelow = maxHeight < spaceBelow || height < spaceBelow || spaceBelow > top;
+    const maximumHeight = parseInt(style.maxHeight || maxHeight, 10) || Infinity;
```
JakeLaCombe: If we are first classing the maxHeight, I think it would be better to put precedence on that first, and then check the styles object.

StephenEsser: I lean towards inline styles always having precedence. It follows similar expectations from CSS. Realistically it probably doesn't matter. This scenario could only be observed if a user sets a different value in both places which would be ambiguous either way. 

If there are strong preferences one way or another I'll update the precedence.

PR #1680, opened 2018-07-05
Remove CSS 'initial' values

### LINKED ISSUE #1667: Update components that are using css 'initial'
# Bug Report

## Description
IE 10&11 doesn't support 'initial' with css.
https://github.com/cerner/terra-core/search?l=SCSS&q=initial

## Possible Solution
Either set the specific browser default value or remove the entry.

## Environment
* Component Name and Version:
* Browser Name and Version: IE 10 & !!

Blackop778:
### Summary
Changes `initial` value to equivalent where possible. Adds black as a fallback color because there is not one defined initial color value.

### Additional Details
Closes #1667

@cerner/terra


packages/terra-tabs/src/StructuralTabs.scss
```diff
@@ -140,6 +140,7 @@
     .tab-menu:focus.is-active {
       background-color: var(--terra-tabs-structural-active-focus-background-color, #fff);
       background-image: var(--terra-tabs-structural-background-image, linear-gradient(to bottom, #007ee0, #007ee0));
+      color: var(--terra-tabs-structural-active-focus-color, #000);
       color: var(--terra-tabs-structural-active-focus-color, initial);
```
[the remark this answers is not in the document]
Blackop778: If this line is removed then the color would still be set to initial if `--terra-tabs-structural-active-focus-color` isn't specified. On IE the initial will be ignored so the color gets set to black. On any other browser the initial will overwrite the black.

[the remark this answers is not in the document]
emilyrohrbough: You have color defined twice, so by adding
```
 color: var(--terra-tabs-structural-active-focus-color, #000);
```
but leaving 
```
color: var(--terra-tabs-structural-active-focus-color, initial);
```
your addition will never be honors. Are you saying you should not have added ?
```
 color: var(--terra-tabs-structural-active-focus-color, #000);
```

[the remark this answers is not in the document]
Blackop778: In Internet Explorer the black color setting takes effect
screen shot 2018-07-06 at 9 46 40 am
While in Chrome the black gets overwritten
screen shot 2018-07-06 at 9 47 55 am

I also tested and this change doesn't break theming the theme-able color.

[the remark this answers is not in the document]
yuderekyu: Would it make more sense to remove the duplicate then have no value for the theme var?
`color: var(--terra-tabs-structural-active-focus-color);`

packages/terra-tabs/src/StructuralTabs.scss
```diff
@@ -155,6 +156,7 @@
 
     .tab-menu:focus {
       background-color: var(--terra-tabs-structural-focus-background-color, #fff);
+      color: #000;
       color: var(--terra-tabs-structural-focus-color, initial);
```
dkasper-was-taken: We might be able to use 'unset' for this.

Blackop778: Unfortunately that isn't supported by Internet Explorer either

dkasper-was-taken: BOOO!!! IE! BOOOOOOOOO!

dkasper-was-taken: [](http://i.imgur.com/8m9mzPf.gif)

emilyrohrbough: Could you explain why you added the following? Is this what the original value is? I assume you grabbed this from terra-base?
```
color: #000;
```

Blackop778: I believe it is the default color for most browsers.

Interestingly, it is used just as much in terra-core as the color set by terra-base, `#1c1f21`.

packages/terra-paginator/src/Paginator.module.scss
```diff
@@ -51,8 +51,9 @@
     }
 
     .is-selected {
+      color: #000;
```
bjankord: Let's set this to `color: var(--terra-paginator-nav-link-selected-color, inherit);`. This will inherit the parent elements color which is likely inherited from [base](https://github.com/cerner/terra-core/blob/master/packages/terra-base/src/Base.scss#L14). Then we should be good to just have 1 instance of color in this selector.

packages/terra-tabs/src/StructuralTabs.scss
```diff
@@ -140,6 +140,7 @@
     .tab-menu:focus.is-active {
       background-color: var(--terra-tabs-structural-active-focus-background-color, #fff);
       background-image: var(--terra-tabs-structural-background-image, linear-gradient(to bottom, #007ee0, #007ee0));
+      color: #000;
```
bjankord: Let's set this to color: var(--terra-tabs-structural-active-focus-color, inherit);. This will inherit the parent elements color which is likely inherited from base. Then we should be good to just have 1 instance of color in this selector.

packages/terra-tabs/src/StructuralTabs.scss
```diff
@@ -155,6 +156,7 @@
 
     .tab-menu:focus {
       background-color: var(--terra-tabs-structural-focus-background-color, #fff);
+      color: #000;
```
bjankord: Let's set this to color: `var(--terra-tabs-structural-focus-color, inherit);`. This will inherit the parent elements color which is likely inherited from base. Then we should be good to just have 1 instance of color in this selector.

Issue #1590, opened 2018-06-11
Create Filter View Component


yuderekyu:
# Issue Description
Create Filter View (name pending) component.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

tomleewu:
# Tech Design

## Summary
Create a `Filter View` component that is a search bar and a suggested search drop-down.

Implementation is similar to the [single search select](https://engineering.cerner.com/terra-core/#/tests/terra-form-select/form-select/controlled-search).
In addition to the dropdown style, there will also be a variant with a persistent result view that updates during search.

Example of search with [persistent result view and multiple search qualifiers.](https://albertlucianto.github.io/vuex-search/)

## Expected behavior
When inputting text to the `Filter View`, either
* a dropdown menu will be shown with the options matching the current value in the input.
* a persistent result view will populate with relevant results.

## Responsiveness
Similar behavior to `terra-form-select`. 
* Options will wrap when width is overflowed.
* Input value will be horizontally scrollable when input width is overflowed. 
* Results list will scroll when number of items overflows height.

## I18N / Bidirectionality
Alignment will be mirrored in rtl orientation.

## Accessibility
* User should be able to horizontally scroll through input value with arrow keys.
* User should be able to navigate the options with arrow keys and select using space or enter.
* User should be able to tab to focus the field.

## React Props:

### FilterView
Mostly borrowed from `terra-search-field`.

| Props					| Type          |isRequired 	| Default  		| Description                                                                                           |
| ------------- 		        |:-------------:|:-------------:| --------		|------------                                                                                           |
| **children**      	                | node 		|   required    | **undefined**	        |The dropdown/result menu options.                                                                      |
| **defaultValue** 		        | string      	|   optional 	| **undefined**	        |The default value displayed in input                                                                   |
| **disableAutoSearch**      	        | bool 		|   optional    | **false**	        |When true, will disable the auto-search.                                                               |
| **disabled**      	                | bool      	|   optional 	| **false**		|Whether the search field is disabled.                                                                  |
| **dropdownAttrs**                     | object      	|   optional 	| **undefined**	        |Additional attributes to spread onto the dropdown. (Style, classNames, etc.)                           |
| **isBlock** 			        | bool      	|   optional    | **false**	        |Whether the field should display as a block.                                                           |
[...]

PR #1685, opened 2018-07-09
[DO NOT MERGE] Filter View Verification

tomleewu:
### Summary
This is a preliminary PR for the new Filter View component.

#### In Progress
* Adding styles
    * Clear search button is currently tacked on, need a more elegant implementation
* General functionality
    * Keeping search value within input when dropdown is toggled
* Tests
    * Jest tests
    * WDIO tests
* Remove/Rename borrowed code from `terra-form-select`.
    * Rename helper/util files to avoid collisions/confusion from `terra-form-select`
    * Find other unnecessary code.

Part of #1590.


ryanthemanuel:
Can we have an option to options displayed up front if the container for the options is displayed up front?

tomleewu:
`showResultsInitially` prop added in https://github.com/cerner/terra-core/pull/1685/commits/48e3d910747cfa0fc8e85788c3bd335666290951

packages/terra-filter-view/docs/README.md
```diff
@@ -0,0 +1,32 @@
...
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-filter-view`
+
+## Usage
+
+```jsx
+import React from 'react';
+import FilterView from 'terra-filter-view';
+
+<FilterView>
```
emilyrohrbough: I feel we should drop the usage section. Now that we can view example code in the site, this section is not as valuable.

packages/terra-filter-view/src/FilterView.jsx
```diff
@@ -0,0 +1,130 @@
...
+  return (
+    <Select
+      id="search"
+      defaultValue={defaultValue}
+      onChange={onChange}
+      onSelect={onSelect}
+      placeholder={placeholder}
+      searchDelay={searchDelay}
+      variant={variant}
+      {...otherProps}
```
yuderekyu: Let's spread `otherProps` first so there isn't a chance to override the default props.
```jsx
<Select
  {...otherProps}
  id="search"
  ...
</Select>
```

packages/terra-filter-view/src/_Frame.module.scss
```diff
@@ -0,0 +1,183 @@
...
+  }
+  
+  .dropdown,
+  .list {
+    > .display {
+      background-size: var(--terra-filter-background-size, 0 0);
+      border: var(--terra-filter-border, 1px solid #dedfe0);
+      border-bottom-left-radius: var(--terra-filter-border-radius, 3px);
+      border-bottom-right-radius: 0;
+      border-right: 0;
```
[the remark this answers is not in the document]
tomleewu: Yeah, I mocked the current [search-field](https://engineering.cerner.com/terra-core/#/components/terra-search-field/search-field/search-field) style. I'd be interested in knowing which style is needed here.

packages/terra-filter-view/src/FilterView.jsx
```diff
@@ -0,0 +1,129 @@
...
+    placeholder,
+    variant,
+    searchDelay,
+    ...otherProps
+  } = props;
+
+  return (
+    <Select
+      {...otherProps}
+      id="search"
```
JakeLaCombe: We need to remove this id. We shouldn't set it automatically on a component.

packages/terra-filter-view/src/_Frame.module.scss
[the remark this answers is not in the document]
yuderekyu: Talked offline with @neilpfeiffer - let's add the border back in.
Logged https://github.com/cerner/terra-core/issues/1730 for search field.

packages/terra-filter-view/src/_Frame.module.scss
```diff
@@ -0,0 +1,181 @@
...
+
+    &:active {
+      background-color: #6f7477;
+      border-color: #52585b;
+      color: #fff;
+    }
+  }
+
+  .clear-button {
+    color: var(terra-filter-clear-button-color, #b9bbbc);
```
bjankord: The variable name needs a double dash in front of it, e.g. `--terra-filter-clear-button-color`

packages/terra-filter-view/src/_Frame.module.scss
```diff
@@ -0,0 +1,181 @@
...
+  .placeholder {
+    color: var(--terra-filter-search-placeholder-color, #85898b);
+    font-size: var(--terra-filter-search-placeholder-font-size, 1.143rem);
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+
+  .box {
+    border: var(terra-filter-list-border, 1px solid #dedfe0);
```
bjankord: The variable name needs a double dash in front of it

packages/terra-filter-view/src/_Frame.module.scss
```diff
@@ -0,0 +1,181 @@
...
+    flex: 1;
+    margin-top: var(terra-filter-list-margin-top, 0.5rem);
```
bjankord: The variable name needs a double dash in front of it

packages/terra-filter-view/src/_Frame.module.scss
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    .button {
+      background-color: var(--terra-filter-disabled-button-background-color, #e2e3e4);
+      border: var(--terra-filter-disabled-border, 1px solid #85898b);
+      opacity: var(--terra-filter-disabled-button-opacity, 0.5);
+    }
+
+    .box {
+      background-color: var(--terra-filter-disabled-background-color, rgba(111, 116, 119, 0.2));
```
bjankord: I'd rename this to `--terra-filter-disabled-list-background-color` to match the variable names on line 94 and 96

packages/terra-filter-view/src/_Frame.module.scss
```diff
@@ -0,0 +1,181 @@
...
+      border: var(--terra-filter-disabled-border, 1px solid #85898b);
```
bjankord: I'd rename this to `--terra-filter-disabled-list-border` to match the variable name on line 94

packages/terra-filter-view/package.json
```diff
@@ -0,0 +1,52 @@
...
+  "peerDependencies": {
+    "react": "^16.2.0",
+    "react-dom": "^16.2.0",
+    "terra-base": "^3.1.0"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "prop-types": "^15.5.8",
+    "react-lifecycles-compat": "^3.0.4",
+    "react-portal": "^4.1.5",
```
StephenEsser: Is react portal used anywhere?

packages/terra-filter-view/src/FilterView.jsx
```diff
@@ -0,0 +1,128 @@
...
+  minimumSearchTextLength: 2,
+  noResultContent: undefined,
+  onChange: undefined,
+  onInvalidSearch: undefined,
+  onSearch: undefined,
+  onSelect: undefined,
+  optionFilter: undefined,
+  placeholder: undefined,
+  searchDelay: 250,
+  showResultsInitially: true,
```
StephenEsser: Booleans should default to false otherwise a user is required to override a lint error to satisfy `showResultsInitially={false}`

This could be changed to `hideResultsInitially` to satisfy the requirements of showing the results initially but a true flag would hide them.

packages/terra-filter-view/src/_Dropdown.jsx
```diff
@@ -0,0 +1,92 @@
...
+   */
+  onResize: PropTypes.func.isRequired,
+  /**
+   * Callback function triggered for setting a ref to the dropdown.
+   */
+  refCallback: PropTypes.func.isRequired,
+  /**
+   * The attachment target.
+   */
+  target: PropTypes.shape({}).isRequired,
```
StephenEsser: Sneaky... You might be able to use `PropTypes.instanceOf(Element)` to avoid the lint error

packages/terra-filter-view/src/_Dropdown.module.scss
```diff
@@ -0,0 +1,23 @@
+:local {
+  .dropdown {
+    background-color: var(--terra-select-dropdown-background-color, #fff);
```
StephenEsser: All of these variable names will need to be scoped to the filter view

packages/terra-filter-view/src/_Frame.jsx
```diff
@@ -0,0 +1,540 @@
...
+
+const propTypes = {
+  /**
+   * Whether the select is disabled.
+   */
+  disabled: PropTypes.bool,
+  /**
+   * When true, will disable the auto-search.
+   */
+  disableAutoSearch: PropTypes.bool,
```
StephenEsser: nit: This is a preference, but I think documentation should read consistently for PropTypes, in this case specifically booleans. I would choose between `Whether the filter...` and `When true, will...` for each boolean to consistently document descriptions.

packages/terra-filter-view/src/_Frame.jsx
```diff
@@ -0,0 +1,540 @@
...
+import Util from './_FrameUtil';
+import styles from './_Frame.module.scss';
+
+const cx = classNames.bind(styles);
+
+const Icon = <IconSearch />;
+
+const propTypes = {
+  /**
+   * Whether the select is disabled.
```
StephenEsser: There's a few occurrences of `select` floating around that should be updated

packages/terra-filter-view/src/_Frame.jsx
```diff
@@ -0,0 +1,540 @@
...
+
+  /**
+   * Clears the search-bar content.
+   */
+  clearSearch() {
+    if (this.state.searchValue !== '') {
+      this.setState({ searchValue: '', excludeBlur: false });
+    }
+    // Refocus input after clearing
+    this.input.focus();
```
StephenEsser: Blur and focus is a dangerous game of catch. To prevent the blur entirely you may be able to prevent default on the event. Examples of this can be found throughout the Select component and this component. 

https://github.com/cerner/terra-core/blob/9bade521c11c07ca67263b439936e8fed4a742ab/packages/terra-form-select/src/_Dropdown.jsx#L58

packages/terra-filter-view/src/_Frame.jsx
```diff
@@ -0,0 +1,540 @@
...
+      this.props.onBlur(event);
+    }
+  }
+  /**
+   * Handles the focus event.
+   */
+  handleFocus(event) {
+    // Show results once input is focused
+    if (!this.state.showResults) {
+      this.setState({ showResults: true, isOpen: true });
```
StephenEsser: This might be challenging or unavoidable but this function can call setState twice and do a double render. Once here and once again if focus needs to be set. If possible setState should only be called once in a function

packages/terra-filter-view/src/_Frame.jsx
```diff
@@ -0,0 +1,540 @@
...
+    }
+
+    return (
+      <div
+        className={cx(['select-parent', { 'is-disabled': disabled }])}
+        onMouseDown={this.handleMouseDown}
+        onBlur={this.handleBlur}
+      >
+        <div
+          {...customProps}
```
StephenEsser: customProps should be spread on the outer most div

packages/terra-filter-view/src/_Frame.jsx
```diff
@@ -0,0 +1,540 @@
...
+          {this.state.searchValue && this.state.isFocused &&
+            <IconIncomplete
+              className={cx('clear-button')}
+              onClick={this.clearSearch}
+              onMouseEnter={this.handleMouseEnter}
+              onMouseLeave={this.handleMouseLeave}
+            />
+          }
+          <Button
+            className={cx(['button'])}
```
StephenEsser: nit: array brackets are unneeded. 

Brackets are actually unneeded for cx entirely. Cx will accept any list of arguments and couple them together. `cx('one', 'two', 'three')` is equivalent to `cx(['one', 'two', 'three'])`

packages/terra-filter-view/src/_Frame.module.scss
```diff
@@ -0,0 +1,181 @@
...
+    }
+  }
+
+  .clear-button {
+    color: var(--terra-filter-clear-button-color, #b9bbbc);
+    margin-left: 0.3571428571rem;
+    margin-right: 0.3571428571rem;
+    position: absolute;
+    right: 30px;
+    top: 7px;
```
StephenEsser: If the filter view is themed to be larger or smaller I think these static values may position the button inaccurately.

Flexbox positioning the clear button and the input together may position content a little better.

packages/terra-filter-view/src/_Frame.jsx
```diff
@@ -0,0 +1,540 @@
...
+   */
+  disabled: PropTypes.bool,
+  /**
+   * When true, will disable the auto-search.
+   */
+  disableAutoSearch: PropTypes.bool,
+  /**
+   * The select display.
+   */
+  display: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.array]),
```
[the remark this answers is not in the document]
tomleewu: No, not in the current implementation. Should the display() function in `Select.jsx` and the display prop from `_Frame.jsx` removed?

packages/terra-filter-view/src/_Frame.module.scss
```diff
@@ -0,0 +1,181 @@
...
+    opacity: 1;
+    outline: none;
+    text-overflow: ellipsis;
+    width: auto;
+
+  }
+
+  .right-padding {
+    // Add padding to avoid text from hiding under clear-search button
+    padding: 0 26px 0 0;
```
StephenEsser: Similar to the comment above, a flex solution might solve a few of these static positioning issues.

packages/terra-filter-view/src/_Menu.module.scss
```diff
@@ -0,0 +1,8 @@
+:local {
+  .menu {
+    list-style: none;
+    margin: var(--terra-select-menu-margin, 0);
+    overflow: hidden;
+    padding: var(--terra-select-menu-padding, 0 0 0.142857rem);
```
StephenEsser: These should be name spaced

packages/terra-filter-view/src/_OptGroup.module.scss
```diff
@@ -0,0 +1,32 @@
...
+
+    &:not(:first-child) > .label {
+      border-top: var(--terra-select-optgroup-label-border-top, 0.1428571429rem solid #dedfe0);
+    }
+  }
+
+  .label {
+    background-color: var(--terra-select-optgroup-label-background-color, #f4f4f4);
+    color: var(--terra-select-optgroup-label-color, #64696c);
+    font-size: var(--terra-select-optgroup-font-size, 0.9286rem);
```
StephenEsser: When these are namespaced this one should also be updated to include the `label` prefix

packages/terra-filter-view/src/_Frame.jsx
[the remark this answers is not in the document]
StephenEsser: It doesn't hurt anything to keep the function so removing it is up to your preference. Those file names should be changed to be scoped to the component. `FilterView.jsx`

.gitignore
```diff
@@ -60,3 +60,4 @@ props-table
 **/__snapshots__/diff
 errorScreenshots
 .DS_Store
+tags
```
tomleewu: Adding this to .gitignore to ignore ctags

packages/terra-filter-view/src/_FilterFrameUtil.js
```diff
@@ -0,0 +1,70 @@
...
+  /**
+   * Determines the dropdown position.
+   * @param {Object} props - The component props.
+   * @param {Object} state - The component state.
+   * @param {ReactNode} target - The select wrapper.
+   * @param {ReactNode} dropdown - The dropdown.
+   * @return {Object} - The calculated dropdown attributes.
+   */
+  static dropdownPosition(props, state, target, dropdown) {
+    const style = Object.assign({}, props).style || {};
```
Blackop778: What is the purpose of doing `Object.assign({}, props)`? I don't think I've seen that done before.

packages/terra-filter-view/src/_FilterMenuUtil.js
```diff
@@ -0,0 +1,230 @@
...
+      return false;
+    }
+
+    return array.find(option => MenuUtil.isEqual(option, query)) !== undefined;
+  }
+
+  /**
+   * Flattens the available options. Ignores disabled options.
+   * @param {ReactNode} object - The node being flattened.
+   * @param {boolean} - Whether disabled props should be filtered out.
```
Blackop778: This parameter is not named

packages/terra-filter-view/src/terra-dev-site/doc/example/OptGroup.jsx
```diff
@@ -0,0 +1,16 @@
...
+  <FilterView variant="list" >
+    <FilterView.OptGroup label="Colors">
+      <FilterView.Option value="red" display="Red" key="red" />
+      <FilterView.Option value="blue" display="Blue" key="blue" />
+      <FilterView.Option value="green" display="Green" key="green" />
+    </FilterView.OptGroup>
+  </FilterView>
+);
+
+export default <OptGroup />;
```
Blackop778: Is there a reason we're exporting this as a element instead of function?
(Why not use `export default OptGroup;`?)

packages/terra-filter-view/src/terra-dev-site/doc/example/DefaultFilterView.jsx
```diff
@@ -0,0 +1,14 @@
...
+
+const FilterViewDefault = () => (
+  <FilterView>
+    <FilterView.Option value="red" display="Red" key="red" />
+    <FilterView.Option value="blue" display="Blue" key="blue" />
+    <FilterView.Option value="green" display="Green" key="green" />
+  </FilterView>
+);
+
+export default () => FilterViewDefault();
```
Blackop778: Is there a reason we're exporting a function to call the `FilterViewDefault` function?

packages/terra-filter-view/src/_FilterFrameUtil.js
emilyrohrbough: if props is undefined and we try to access the style key, it results in the error `cannot access style key of undefined` (or something very similar). An alternative option to this could be:
```
const style = (props || {}).style || {};
```

packages/terra-filter-view/src/terra-dev-site/doc/example/OptGroup.jsx
tomleewu: The [DocTemplate expects a React Element](https://github.com/cerner/terra-core/blob/master/packages/terra-doc-template/src/ExampleTemplate.jsx#L17) instead of a function

packages/terra-filter-view/src/terra-dev-site/doc/example/DefaultFilterView.jsx
tomleewu: I'll change this to JSX instead to stay consistent.

packages/terra-filter-view/src/terra-dev-site/doc/example/OptGroup.jsx
Blackop778: We typically create the react element in the doc template page and export the function.

tomleewu: Ah, I see, I can change that.

Blackop778: There are other instances of odd exports in your doc/test pages, I didn't review them all incase you had a reason you needed to export like wrote it.

tomleewu: will take a look and fix those

packages/terra-filter-view/src/_FilterDropdown.jsx
```diff
@@ -0,0 +1,92 @@
...
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * Whether the dropdown is presented above the target.
+   */
+  isAbove: PropTypes.bool,
+  /**
+   * Whether the dropdown is visible.
+   */
+  isEnabled: PropTypes.bool,
```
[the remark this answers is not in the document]
tomleewu: It looks like isEnabled isn't being manipulated in `_FilterDropdown` and is directly passing it to Hookshot. [Hookshot has defined a default for isEnabled](https://github.com/cerner/terra-framework/blob/master/packages/terra-hookshot/src/Hookshot.jsx#L108).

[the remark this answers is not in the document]
ryanthemanuel: It'd be worth setting a default explicitly so that consumers are aware without having to dig into how it's being used

packages/terra-filter-view/src/_FilterFrame.jsx
```diff
@@ -0,0 +1,523 @@
...
+import IconSearch from 'terra-icon/lib/icon/IconSearch';
+import IconIncomplete from 'terra-icon/lib/icon/IconIncomplete';
+import { KeyCodes, Variants } from './_FilterConstants';
+import Dropdown from './_FilterDropdown';
+import Util from './_FilterFrameUtil';
+import styles from './_FilterFrame.module.scss';
+
+const cx = classNames.bind(styles);
+
+const Icon = <IconSearch />;
```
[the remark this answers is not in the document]
tomleewu: @neilpfeiffer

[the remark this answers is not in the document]
neilpfeiffer: Yes - Icon we should provide by inline svg with themeable variable.

packages/terra-filter-view/src/_FilterOption.module.scss
```diff
@@ -0,0 +1,84 @@
+@import '~terra-mixins';
```
StephenEsser: These variables will need to be scoped under terra-filter

packages/terra-filter-view/src/_FilterNoResults.module.scss
```diff
@@ -0,0 +1,10 @@
+:local {
+  .no-results {
+    font-size: 1.142rem;
+    line-height: 1.125;
+    overflow: hidden;
+    padding: var(--terra-select-no-results-padding, 0.28571rem 0.714286rem);
```
StephenEsser: These variables will need to be scoped under terra-filter

packages/terra-filter-view/src/_FilterSelect.jsx
```diff
@@ -0,0 +1,167 @@
+import React from 'react';
```
StephenEsser: When the package name is finalized we should update the file name to match.

`Filter.jsx` / `FilterView.jsx` / `FilterList.jsx` / `SortMaster5000.jsx`

StephenEsser: This will also influence the scss variable names

packages/terra-filter-view/src/_FilterDropdown.jsx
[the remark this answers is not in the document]
tomleewu: done here: https://github.com/cerner/terra-core/pull/1685/commits/53e42b9203ab32c413f61a567888ae79b090cb64#diff-a44a7546b79f6cabfce567c92a42285aR38

packages/terra-filter-view/src/_FilterOption.module.scss
tomleewu: done here: https://github.com/cerner/terra-core/pull/1685/commits/53e42b9203ab32c413f61a567888ae79b090cb64#diff-ba1c49cc02020d82360fe5d0207cd3e4

neilpfeiffer:
> I'm noticing some odd behavior with the dropdown variant. When I click on the input, the dropdown displays with options, but if I click on an option (Red, Blue, Green) it doesn't populate the input. This is different behavior compared to what we have with the default terra-form-select component.
> 
> Similarly, if I click on the input in the dropdown variant, and use the up down arrow keys, it filters throw the options, but if I press enter when on a selected option, it does not populate the input with the selected option. This is different behavior compared to what we have with the default terra-form-select component.

That was intentional to this design for the initial implementation, since the intent of a Select is to choose a value from a list to then show as the "chosen value" within the input, awaiting form submission. We thought the intent of a filter/search list is different and the behavior should be type to search/filter for an item, user select the item, which fires the onSelect and immediately the consumer app would handle the request and change the UI accordingly external to this component. (If the phrase would populate the select, then the list itself would also need to update to reduce to only show the selected item.) This may not be the correct final design but wanted to see how it works when wired up in an example, and can adjust as needed if we should switch it back.

PR #1687, opened 2018-07-10
Update jestconfig to run jest in cross-OS compatible way

### LINKED ISSUE #1686: Jest tests cannot be found on Windows
# Bug Report

## Description
When running the Jest tests on Windows, it reports that no tests are found. 

## Steps to Reproduce
1. Run `npm run jest` in the root of the project
2. Run `npm run test` inside of a package

## Expected Behavior
Jest tests run

## Possible Solution
I believe it's just because of the testMatch Regex using Unix separators. I think a combination of Jest's `testMatch` and `roots` can fix this.

```
  testMatch: [
    '**/jest/**/(*.)(spec|test).js?(x)',
  ],
  roots: [process.cwd()],
```

... but hoping to get some confirmation on the Mac side of things to make sure things still work, both when running at the root level and when running inside of a package.

## Environment
* Operating System and version (desktop or mobile): Windows 10 Enterprise

bjankord:
### Summary
Resolves #1686


mjhenkes:
This change will have to be propagated to our other repos too.

PR #1690, opened 2018-07-11
[base] Add themeable background image

### LINKED ISSUE #1200: Add theming property to base for background images
# Issue Description
Add ability to theme background-image style on the body element

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Ability to theme background-image style on the body element

## Current Behavior
N/A

Blackop778:
### Summary
Adds a themeable background image for the site.

Also adds a test for the themeable background color
### Additional Details
Closes #1200

@cerner/terra


packages/terra-base/src/Base.scss
```diff
@@ -11,6 +11,7 @@ html {
 
 body {
   background-color: var(--terra-base-background-color, #fff);
+  background-image: var(--terra-base-background-image, none);
```
bjankord: [Background image needs to go on the html element](https://github.com/cerner/terra-core/issues/1200), I've seen issues using CSS radial gradients as background-image property on body element.

Let's move the background color to the html element as well.

```css
html {
  background-color: var(--terra-base-background-color, #fff);
  background-image: var(--terra-base-background-image, none);
}
```

packages/terra-base/src/Base.scss
```diff
@@ -1,6 +1,8 @@
-// Normalize box-sizing.
-// More info: https://css-tricks.com/box-sizing/#article-header-id-6
 html {
+  background-color: var(--terra-base-background-color, #fff);
+  background-image: var(--terra-base-background-image, none);
```
StephenEsser: It's probably unlikely, but are there any concerns about this move being non-passive? It would probably require a really awkward CSS setup to cause any issues, but it might be able to cause a users workaround background color to change to white, or their image to change to none if they had one defined in a roundabout way.

bjankord: I don't see it as a concern, we can't predict customizations outside of our components.

StephenEsser: That's fair, Base is tricky because it has no component scope just global impact

yuderekyu: Should we also include `background-repeat` within base?

bjankord: Yeah, that is a good idea. I just looked at the [consumer layout/nav](http://engineering.cerner.com/terra-consumer/iframe.html?selectedKind=Layout&selectedStory=Dex&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel), they use background image, size, position, and repeat in shorthand. I think it would be good to add those as well as. Let's go with the following to cover all use-cases we have now.

```
background-attachment: var(--terra-base-background-attachment);
background-color: var(--terra-base-background-color, #fff);
background-image: var(--terra-base-background-image);
background-position:  var(--terra-base-background-position);
background-repeat:  var(--terra-base-background-repeat);
background-size: var(--terra-base-background-size);
```

The CSS custom properties set without a fallback value will default to the browser default like we use now.

bjankord:
Looks like there are some lint errors after merging with master. Let's get those updated and then this should be good to go.

PR #1692, opened 2018-07-11
[menu] Add more theming options to MenuItem

### LINKED ISSUE #1345: Menu: add new theme variables
# Issue Description
In the current Menu component, certain style attributes are not themeable. Additionally, some interaction states aren't accounted for in the current component. The items to be updated are:

- Menu item Padding (made themeable)
- Menu item Margin (made themeable)
- [EDIT] Menu item disabled color (made themeable)
- Keyboard Focus State (state defined/added, and made themeable)
- Pressed State (state defined/added, and made themeable)

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The Menu component should have themeable variables for all of the desired attributes.

## Current Behavior
"N/A"

Blackop778:
### Summary
Adds several themeable variables to the MenuItem, notably a active class which works for both mouse and keyboard input.

### Additional Details
Closes #1345

@cerner/terra


packages/terra-menu/src/MenuItem.jsx
```diff
@@ -113,9 +126,17 @@ class MenuItem extends React.Component {
       if (event.nativeEvent.keyCode === MenuUtils.KEYCODES.ENTER || event.nativeEvent.keyCode === MenuUtils.KEYCODES.SPACE) {
         this.handleSelection(event);
 
+        // If the item has sub menu items onKeyUp will never get called and item will be stuck active
```
emilyrohrbough: Do you mean something like `Keep active state for items with sub menu items`?

packages/terra-menu/src/MenuItem.jsx
```diff
@@ -113,9 +126,17 @@ class MenuItem extends React.Component {
...
+        if (!this.props.subMenuItems || this.props.subMenuItems.length === 0) {
+          this.setState({ isActive: true });
+        }
+
         if (this.props.onClick) {
           this.props.onClick(event);
         }
+      // Item will become stuck active if the user presses tab while holding space
```
emilyrohrbough: Update to `Remove active state when tab key is released while while holding the space key`

packages/terra-menu/src/MenuItem.jsx
```diff
@@ -150,6 +171,7 @@ class MenuItem extends React.Component {
     } else {
       attributes.onClick = this.wrapOnClick;
       attributes.onKeyDown = this.wrapOnKeyDown(attributes.onKeyDown);
+      attributes.onKeyUp = this.onKeyUp();
```
emilyrohrbough: See comment above, but given the update I'd expect custom prop `onKeyUp` attributes to be pass like
```
 attributes.onKeyUp = this.wrapOnKeyUp(attributes.onKeyUp);
```

packages/terra-menu/src/MenuItem.jsx
```diff
@@ -82,6 +86,14 @@ class MenuItem extends React.Component {
     }
   }
 
+  onKeyUp() {
```
emilyrohrbough: Lets rename this to wrapOnKeyUp.

We should apply the same logic as wrapOnKeyDown for the case a customer passes an onKeyUp attributes to the component.

packages/terra-menu/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Added
+* Added more themeable variables to MenuItem
```
emilyrohrbough: We should indicated we added active styling to the component

packages/terra-menu/src/MenuItem.scss
```diff
@@ -10,15 +11,29 @@
...
-    } 
+    }
+  }
+
+  .item:active:not(.is-disabled),
+  .active {
+    background-color: var(--terra-menu-item-active-color, #ebf6fd);
+
+    .title {
+      color: var(--terra-menu-item-active-text-color, #1c1f21);
```
StephenEsser: There's not a concrete standard of naming these variables but in general I believe they should be named only with the property and scoping. 

--terra-menu-item-active-**title**-color

packages/terra-menu/src/MenuItem.scss
```diff
@@ -10,15 +11,29 @@
...
+    }
   }
 
   .is-disabled {
     outline: 0;
     
     .title {
+      color: var(--terra-menu-item-disabled-text-color, #1c1f21);
```
StephenEsser: Any objections to renaming this? 

`--terra-menu-item-disabled-*title*-color`

StephenEsser: I would gather additional opinions before making a decision but because this is a new variable I would vote to name it --terra-menu-item-disabled-**title**-color or to remap the the css declaration of `.title` to be `.text`

StephenEsser: It looks like the prop name is `text` so I would lean towards the second approach

packages/terra-menu/src/MenuItem.scss
StephenEsser: The prop name is actually `text` so I would keep this variable named as --terra-menu-item-active-text-color

Although probably out of the scope of this PR I think the css class declaration should be updated to be `.text` instead of `.title`

packages/terra-menu/src/MenuItem.scss
```diff
@@ -1,7 +1,8 @@
 :local {
   .item {
     cursor: default;
-    padding: 0.286rem 0;
+    margin: var(--terra-menu-item-margin, 0);
+    padding: var(--terra-menu-item-padding, 0.286rem 0);
```
Blackop778: I suppose we want to expand these @bjankord

PR #1695, opened 2018-07-12
[search-field] Add more themeable variables

### LINKED ISSUE #1681: Add theming variables for Search Field
I'm currently looking at making the margins themeable on the input and button portions, and the border of the button.

I think we should add a themeable border property to the input so the `terra-form-input` border doesn't have to overwritten in case both components should be on the same page at the same time.

I'd also like feedback on whether we should make the border radiuses themeable on the button and input (as appropriate).

Blackop778:
### Summary
Adds more themeable variables for the Search Field.

The border radius on the input element had been `.2rem` as inherited from the `form-input` but I elected to change it to the `.25em` already present on the `search-field` button, resulting in all the screenshots needing to be updated.
### Additional Details
Closes #1681

@cerner/terra


packages/terra-search-field/src/SearchField.scss
```diff
@@ -4,24 +4,26 @@
...
     .input {
       appearance: none;
+      border-bottom-left-radius: var(--terra-search-field-border-bottom-left-radius, 0.25em);
       border-bottom-right-radius: 0;
       border-right: 0;
+      border-top-left-radius: var(--terra-search-field-border-top-left-radius, 0.25em);
       border-top-right-radius: 0;
       flex: 1 1 auto;
-      margin: 0;
+      margin: var(--terra-search-field-input-margin, 0);
```
bjankord: We've had issues using theming variables with margin/padding shorthand and bidirectional styles (LTR/RTL). Let's split this out to long-hand style:

```css
margin-bottom: var(--terra-search-field-input-margin-bottom, 0);
margin-left: var(--terra-search-field-input-margin-left, 0);
margin-right: var(--terra-search-field-input-margin-right, 0);
margin-top: var(--terra-search-field-input-margin-top, 0);
```

packages/terra-search-field/src/SearchField.scss
```diff
@@ -4,24 +4,26 @@
...
       border-bottom-left-radius: 0;
-      border-bottom-right-radius: 0.25em;
+      border-bottom-right-radius: var(--terra-search-field-border-bottom-right-radius, 0.25em);
       border-top-left-radius: 0;
-      border-top-right-radius: 0.25em;
+      border-top-right-radius: var(--terra-search-field-border-top-right-radius, 0.25em);
       display: inline-block;
       flex: 0 0 auto;
-      margin: 0;
+      margin: var(--terra-search-field-button-margin, 0);
```
bjankord: We've had issues using theming variables with margin/padding shorthand and bidirectional styles (LTR/RTL). Let's split this out to long-hand style.

packages/terra-search-field/src/SearchField.scss
```diff
@@ -4,24 +4,32 @@
 
     .input {
       appearance: none;
+      border-bottom-left-radius: var(--terra-search-field-border-bottom-left-radius, 0.25em);
       border-bottom-right-radius: 0;
       border-right: 0;
+      border-top-left-radius: var(--terra-search-field-border-top-left-radius, 0.25em);
```
StephenEsser: To avoid collisions down the road we should match the prefix used on the margin properties below: 

`--terra-search-field-input`

packages/terra-search-field/src/SearchField.scss
```diff
@@ -4,24 +4,32 @@
...
       min-width: 0;
       outline-offset: -0.143em;  //Adjust the outline offset for search input in Safari.
     }
 
     .button {
-      border: 0.071em solid #dedfe0;
+      border: var(--terra-search-field-button-border, 0.071em solid #dedfe0);
       border-bottom-left-radius: 0;
-      border-bottom-right-radius: 0.25em;
+      border-bottom-right-radius: var(--terra-search-field-border-bottom-right-radius, 0.25em);
```
StephenEsser: It will be a verbose name, but I think we should consistently scope these variables under the same name.  Similar to my comment above these should match the same prefix. 

`--terra-search-field-button`

PR #1696, opened 2018-07-13
ISSUE-1463: Fixed undefined prop type error

### LINKED ISSUE #1463: terra-props-table doesn't render always correctly render 'Type' column value correctly for props
# Issue Description

In the case that we're using an object  (`alphaPropsObj`) to store key-value pairs of valid prop options for prop `alpha`, and we attempt to use `PropTypes.oneOf(Object.values(alphaPropsObj)`, the resulting value for the `Type` field in the props-table generated by `terra-props-table` is `undefined`.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior

I would expect the table to still be able to interpret the type of the array items and yield the following props table
image

**NOTE:** This _can_ be achieved by wrapping `Object.values(...)` in a set of array brackets, and spreading the returned value into said array. Example included below.
image

## Current Behavior

Using
image
produces

image

## Steps to Reproduce

1. Define an object of key-value pairs, with values you'd expect to be type-checked using `PropTypes.oneOf()`
2. Call `Object.values(<your object)` as the argument to `PropTypes.oneOf`.
3. Observe the `terra-props-table`'s `Type` field doesn't properly reflect the expected type.

## Environment

* Component Version: 2.5.0
* Browser Name and Version: Chrome (66.0.3359.139)
* Operating System and version (desktop or mobile): Desktop

dylan-sftwr:
### Summary
Removed the ability to auto-infer types of `enum`-type objects in the `terra-prop-table` as we're not always able to do so reliably.

Fixes #1463

---
**Before:**
image

**After:**
image
 

### Additional Details
This change was cleared by @bjankord and @emilyrohrbough 

@cerner/terra


packages/terra-props-table/src/PropsTable.jsx.orig
```diff
@@ -0,0 +1,124 @@
+/* eslint-disable import/no-extraneous-dependencies */
```
bjankord: Not sure whats up with this file, looks like this wasn't meant to be checked into version control

dylan-sftwr: Whoops! looks like that was left over from my diff, f85e7b2c

packages/terra-props-table/src/PropsTable.jsx
```diff
@@ -26,14 +26,8 @@ function formatShape(shape) {
...
 
-  // Pull the first value off and use that as type.
-  // This assumes all enumerable values are the same type.
   if (typeName === 'enum') {
-    if (Number.isNaN(Number(type.value[0].value))) {
-      typeName = typeof type.value[0].value;
-    } else {
-      typeName = 'number';
-    }
+    typeName = 'enum';
```
emilyrohrbough: Ohh I think I miss understood what you meant when you had asked me about this change. `enum` would not be the correct prop type in most cases, I thought you meant you would parse the first value to determine the type. Looking in the terra-repos,`oneOf` is usually an `enum of strings` options... so this maps to a string prop essentially.

If we only indicate an `enum`, this would still require one to look in the code to verify the enum options.

Additionally, to me this indicated a more 'rigid' prop value than what will be required by most props.

emilyrohrbough: What was the reason this was displayed as undefined? I know this occurs because we want to pass `Object.keys(DEFAULTS)`, but what information is given as the type value?

bjankord: > If we only indicate an enum, this would still require one to look in the code to verify the enum options.

Listing the available enum keys in the code comment above the prop that generates the doc could help with this

> Additionally, to me this indicated a more 'rigid' prop value than what will be required by most props.

In previous meetings we've talked about not using enums as propTypes but still exporting them so people could use them if they wanted. For example, in the card props image, we'd make the propType a string and list out the values we accept while also exporting an enum that's values were strings. Users could type the string directly or use the enum.

bjankord: I've dug into the react-docgen code a bit more and it looks like they may handle the case #1463 was logged for. 

We've opted to do custom type checking, so we miss out on this feature from react-docgen, but I'd recommend taking a look at [this PR](https://github.com/reactjs/react-docgen/pull/211) and [this PR]() which seem to have added support in react-docgen for resolving `Object.keys()` in `PropType.oneOf()` and potentially resolving spreads. [Here is a util they've made to help resolve enums](https://github.com/danez/react-docgen/blob/d1abfb9bf590e4b6ec6e3335a152c24f27bbb6fa/src/utils/resolveObjectKeysToArray.js).

dylan-sftwr: > What was the reason this was displayed as undefined

This was displaying as undefined because when we would try to examine `typeof type.value[0].value`, we would in fact be trying to call `typeof` on `Object.values(<whatever the name of your enumerated object was>)`, rather than on `[/*the actual array that Object.values() returns */]` – essentially, we'd be making the call before the `Object.values` was evaluated, and hence before the objects values were converted into an array. So trying to access the first item in `type.value` and look at it's `value` would return `undefined`, so `typeName` was inadvertently getting set to `undefined` here.

Issue #1627, opened 2018-06-20
[terra-tabs] Allow additional tabs to be added after initial mount of Tabs component


tbiethman:
# Issue Description

The Tabs component has some internal state that manages the visible tab panes. If additional panes are added after the initial mounting, that state is not updated and the added panes are not visible. 

We should make changes to ensure that componentWillReceiveProps is implemented and keeping the internal component state in sync with the props.

## Issue Type

- [X] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

PR #1699, opened 2018-07-13
[terra-tabs] Allow additional tabs to be added after initial mount

sujith9:
### Summary
Currently adding any additional tabs after initial mount of Tabs component will not display the new tabs.
This fix uses the suggestion made in the [issue description](https://github.com/cerner/terra-core/issues/1627) to keep the internal component state in sync with the props.

Addresses: [#1627](https://github.com/cerner/terra-core/issues/1627)

@cerner/terra


packages/terra-tabs/tests/wdio/tabs-spec.js
```diff
@@ -185,4 +185,24 @@ describe('Tabs - Responsive', () => {
...
     });
   });
+
+  // Test additional tabs
+  describe('Additional Tabs', () => {
+    before(() => {
+      browser.setViewportSize(Terra.viewports('small')[0]);
+      browser.url('/#/raw/tests/terra-tabs/tabs/tabs/additional-tabs');
+    });
+
```
emilyrohrbough: can we add an intial screenshot as well so we know the initial state?

packages/terra-tabs/src/_CollapsibleTabs.jsx
```diff
@@ -73,6 +73,13 @@ class CollapsibleTabs extends React.Component {
     this.handleSelectionAnimation();
   }
 
+  componentWillReceiveProps(nextProps) {
```
bjankord: Let's switch this over to use `getDerivedStateFromProps` instead of `componentWillReceiveProps`. https://hackernoon.com/replacing-componentwillreceiveprops-with-getderivedstatefromprops-c3956f7ce607

You'll need to use https://github.com/reactjs/react-lifecycles-compat to polyfill `getDerivedStateFromProps` as well. Set that packages as a dependency and import `polyfill` and wrap the `CollapsibleTabs` component with it.

packages/terra-tabs/src/_CollapsibleTabs.jsx
```diff
@@ -73,6 +73,13 @@ class CollapsibleTabs extends React.Component {
...
+    if (nextProps.children.length !== this.props.children.length) {
```
StephenEsser: I understand the intention of this enhancement is to handle the specific case of adding more tabs.

Some of these **dynamic** situations may already be handled but they're worth some thought and discussions to see if they should also be handled.

1. Reordering tabs. Length remains the same.
2. Changing the content of tab(s). Length remains the same.
3. Disabling / Enabling tab(s). Length remains the same.

packages/terra-tabs/src/_CollapsibleTabs.jsx
sujith9: I had tried to avoid `componentWillReceiveProps`. The values stored in the internal state were moved out of it in this fix https://github.com/cerner/terra-core/pull/1658. Since `getDerivedStateFromProps` cannot access `this`, these values will have to be moved back inside the local state for `getDerivedStateFromProps` to work.

bjankord: Adding componentWillReceiveProps adds tech debt for us to resolve when we update to React 17 as `componentWillReceiveProps` is being deprecated in react. Instead of trying to make this component act as both and uncontrolled and controlled component via `componentWillReceiveProps/getDerivedStateFromProps` I'd suggest we go back and work on a tech design on https://github.com/cerner/terra-core/issues/1627 for making a controlled version of tabs which could support adding additional tabs.

bjankord:
Closing this in favor of working on a tech design for creating a controlled versions of tabs. [More info here](https://github.com/cerner/terra-core/pulls#discussion_r203769894).

PR #1713, opened 2018-07-18
Dst

### LINKED ISSUE #1705: Disabled Date Time Picker with DST doesn't disable the Time Clairification
# Bug Report

## Description
When a Date Time Picker with DST is toggled and then switched to disabled, the time clarification component is not disabled. It should be disabled.

## Steps to Reproduce
Run this code in Terra Core's repository.

```jsx
import React from 'react';
import Button from 'terra-button';
import DateTimePickerExampleTemplate from '../../common/DateTimePickerExampleTemplate';

export default class extends React.Component {
  constructor(props) {
    super(props);

    this.state = { active: false };
    this.toggleDateTimePicker = this.toggleDateTimePicker.bind(this);
  }

  toggleDateTimePicker() {
    this.setState({
      active: !this.state.active
    })
  }

  render() {
    return (
      <div>
        <Button
          text={this.state.active ? 'Disable' : 'Enable'}
          onClick={this.toggleDateTimePicker}
        />
        <DateTimePickerExampleTemplate
          value="2017-11-05T01:30:00"
          disabled={!this.state.active}
        />
      </div>
    )
  }
}
```

screen shot 2018-07-17 at 12 03 03 pm

## Expected Behavior
Time Clarification should be disabled

## Possible Solution
Populate the disabled prop on the buttons and other associated

JakeLaCombe:
### Summary
Resolves #1705 

If you enable a date time picker, complete the time clarification dialog and disable the input, the time clarification button is still active. This pull request resolves that issue.


packages/terra-date-time-picker/tests/wdio/date-time-picker-spec.js
```diff
@@ -129,6 +129,21 @@ describe('DateTimePicker', () => {
...
+      browser.click('input[name="terra-time-hour-input"]');
+      browser.click('input[name="terra-time-minute-input"]');
+      browser.waitForVisible('[class*="time-clarification"]', 1000);
+      browser.click('[class*="button-daylight"]');
+      browser.click('#date-time-picker-toggler');
+    });
+
+    const ignoredDisabledAlly = Object.assign({ 'color-contrast': { enabled: false } }, ignoredA11y);
+    Terra.should.beAccessible({ rules: ignoredDisabledAlly });
+    Terra.should.matchScreenshot('0');
```
Blackop778: Could this screenshot be given a better name?

JakeLaCombe: What are you thinking?

Blackop778: Disabled would be more descriptive but also kind of repetitive. If we had one overlapping describe block for Time Clarification 'disabled' would be a good name. I think `0` would be confusing, so maybe even leaving it blank and having the screenshot called default would be better.

JakeLaCombe: Oh, I see! Since we are just taking a single snapshot, we can remove this and leave it as default! I was thinking you are talking about the top name!

mmalaker:
Functional verification completed. The time clarification dialog now correctly enables/disables the time clarification button, as expected.

PR #1719, opened 2018-07-19
Update image alt default prop

bjankord:
### Summary
We used to set a default prop of an empty string for the alt prop `" "`, but that was removed when we updated the linter. This change adds the default prop back to the image component for the alt attribute.


Blackop778:
Closes cerner/terra-framework#229

Issue #1728, opened 2018-07-24
Add refCallback to SearchField Component So That It Can Be Focused


juliacalandro:
# Feature Request

## Description
My team wants the ability to place focus on the search field upon loading our search page. Currently, we don't believe this is possible, since the refCallback prop only exists on the Input component, and doesn't get passed down as a custom prop from the SearchField component. Therefore, we would like an optional inner refCallback prop to be added to the SearchField component. This would get passed straight into the refCallback prop of the inner Input component.

## Additional Context / Screenshots
I got the SearchField component to focus correctly from the terra-dev-site locally.
![screen shot 2018-07-23 at 3 28 16 pm](https://user-images.githubusercontent.com/39311706/43101118-115849b0-8e8d-11e8-88d6-9ca8b12ad1dd.png)
I have implemented the required change and will submit a pull request.

bjankord:
This has been released in [terra-search-field@2.22.0](https://www.npmjs.com/package/terra-search-field)

PR #1729, opened 2018-07-24
Add refCallback to searchField to allow it to be focused

juliacalandro:
### Summary
My team wants the ability to place focus on the search field upon loading our search page. Currently, we don't believe this is possible, since the refCallback prop only exists on the Input component, and doesn't get passed down as a custom prop from the SearchField component. Therefore, we would like an optional inner refCallback prop to be added to the SearchField component. This would get passed straight into the refCallback prop of the inner Input component. I got the SearchField component to focus correctly from the terra-dev-site locally.
screen shot 2018-07-23 at 3 28 16 pm
I have implemented the required change, added an example component to the terra-dev-site to demonstrate this new feature, and added jest and wdio tests for the change.

[Here](https://github.com/cerner/terra-core/issues/1728) is the corresponding issue.


packages/terra-search-field/tests/jest/SearchField.test.jsx
```diff
@@ -36,6 +36,12 @@ describe('Snapshots', () => {
 
     expect(searchField).toMatchSnapshot();
   });
+
+  it('renders a search field with an Input that contains a non-null refCallback prop', () => {
+    const searchField = shallow(<SearchField inputFieldRefCallback={(inputRef) => { inputRef.focus(); }} />, intlContexts.shallowContext);
+    expect(searchField.find('Input').props().refCallback).not.toBeFalsy();
+    expect(searchField).toMatchSnapshot();
+  });
```
Blackop778: We can test the callback similarly to how it was done in [Button Group](https://github.com/cerner/terra-core/blob/f757f0918ccb7b071829726cff296f389e395466/packages/terra-button-group/tests/jest/ButtonGroup.test.jsx#L32).

juliacalandro: Will make the change!

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -70,6 +70,11 @@ const propTypes = {
    * The value of search field.  Use this to create a controlled search field.
    */
   value: PropTypes.string,
+
+  /**
+   * Callback ref to pass into the inner input component.
+   */
+  inputFieldRefCallback: PropTypes.func,
```
StephenEsser: Depending on the needs of your team, another possible solution could be a first class autoFocus property that is spread onto the input. The refCallback is a wildcard for doing anything custom, but the autoFocus approach has an explicit API. 

```
<SearchField autoFocus />
```
VS
```
class Example extends React.Component {
  constructor() {
    super();
  }

  componentDidMount() {
    this.input.focus();
  }

  render() {
    return <SearchFieldExampleTemplate inputFieldRefCallback={(input) => { this.input = input; }} />;
  }
}
```

There are tradeoffs for each approach. The AutoFocus will only work when the component mounts but has an explicit API. The refCallback is more extensible but loosely couples functionality and allows many variations of behavior within an application.

A request has recently been logged to autoFocus the select component and if possible it would be nice to settle on a pattern going forward or if each component's ability to focus is case by case.

juliacalandro: My team hasn't finalized the requirements for our search feature yet, but we may want to focus it in more scenarios other than just when the component mounts, so it is nice to have the ability to implement the focusing ourselves.

Also, I initially had my example component on the terra-dev-site performing the focus on componentDidMount, but that causes the search-field site to automatically focus on that one example instead of loading the top of the page. I'm afraid an autofocus example would cause the same behavior.

juliacalandro: Yeah I'm fine with changing it to `inputRefCallback`! Will make the change.

bjankord:
Thanks for the contribution @juliacalandro! We'll merge this down after the build passes and it will be available with our next release of terra-core components.

juliacalandro:
You're welcome! Sounds good!

PR #1738, opened 2018-07-30
Assign option role to ListItems based on selectability

Blackop778:
### Summary
The [W3C specification](https://www.w3.org/TR/2010/WD-wai-aria-20100916/roles#option) states the option role should be set on

> A selectable item in a select list.

and this change aligns with that.
### Additional Details
Change comes from discussion in a [terra-clinical PR](https://github.com/cerner/terra-clinical/pull/379/files#r205246159).

@cerner/terra


packages/terra-list/tests/jest/__snapshots__/ListItem.test.jsx.snap
```diff
@@ -6,7 +6,9 @@ exports[`should mount with isSelectable 1`] = `
   isSelected={false}
 >
   <li
+    aria-selected={false}
```
emilyrohrbough: I think this value should be aria-selected=`false`?

emilyrohrbough: but maybe not after looking at other snapshots 😄

PR #1749, opened 2018-08-01
Wrap react-datepicker .scss import in :global selector to have it parse as global css

### LINKED ISSUE #1746: Update import used from react-datepicker
# Feature Request

## Description
Ran into an issue with terra-datepicker and webpacker and how it parses CSS modules. We renamed the datepicker SCSS file to include `.module` in the extension which now allows webpacker parse the file correctly as a CSS module file, but [the file we import inside of the scss file](https://github.com/cerner/terra-core/blob/master/packages/terra-date-picker/src/DatePicker.module.scss#L1) is also being treated as a CSS modules file, where currently we want it to be treated as a regular global .scss file. This seems to only be an issue with webpacker integration. 

We should be able to resolve this by importing the CSS modules based file react-datepicker provides that wraps its content in the global selector. Alterntively, we could try wrapping the import ourselves in `:global` and it should resolve as global selectors in the file.

bjankord:
### Summary
An issue came up with webpacker integration where [the .scss file we import from react-datepicker](https://github.com/cerner/terra-core/blob/master/packages/terra-date-picker/src/DatePicker.module.scss#L1) inside of our `.module.scss` file for datepicker didn't get parsed as global CSS. 
With the way we have our loaders setup in terra-toolkit, this isn't a cocern and only came up at integration time.

This update makes it more explicit this import should be parsed as global CSS.

Resolves #1746

Before screenshots:
screen shot 2018-08-01 at 11 14 06 am

screen shot 2018-08-01 at 11 14 20 am

After screenshots:
screen shot 2018-08-01 at 11 04 17 am

screen shot 2018-08-01 at 11 15 34 am

### DIFF
--- a/packages/terra-date-picker/CHANGELOG.md
+++ b/packages/terra-date-picker/CHANGELOG.md
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+Fix issue where tooling was parsing the .scss import inside of our .module.scss file as a CSS modules file
 
 2.24.1 - (July 26, 2018)
 ------------------
--- a/packages/terra-date-picker/src/DatePicker.module.scss
+++ b/packages/terra-date-picker/src/DatePicker.module.scss
@@ -1,4 +1,6 @@
-@import '~react-datepicker/src/stylesheets/datepicker';
+:global {
+  @import '~react-datepicker/src/stylesheets/datepicker';
+}
 
 /* stylelint-disable selector-max-compound-selectors, max-nesting-depth */
 :local {


packages/terra-date-picker/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+Fix issue where tooling was parsing the .scss import inside of our .module.scss file as a CSS modules file
```
StephenEsser: Are there screenshot tests for the date picker that should have broke?

bjankord: No, our webpack config parsed these files as intended. If you view [the page on terra-ui](https://engineering.cerner.com/terra-ui/#/components/terra-date-time-picker/date-time-picker/date-time-picker), it renders the datepicker correctly. 

It only appeared to come up with the way webpacker loaders parsed this file and it's import. With this change, our webpack config still parses it correctly, we've just made it more explicit the third-party styles should be global for other loader configs.

Issue #1461, opened 2018-04-30
Extra border when using ActionFooter with Dialog


dv297:
# Issue Description
Both terra-action-footer and terra-dialog define a top border for the footer. A lot of designs have modals that present users with a message and right-aligned buttons at the bottom of the modal, and the combination of modal -> dialog -> action-footer seems like the appropriate path for this.

As a side note, I also noticed there is terra-divider. There isn't much preventing a user from just implementing their own layout and use terra-divider as the border. But terra-divider is also a slightly lighter color. Perhaps it could be a recommended / not recommended route for defining layouts to use dividers. Just a thought for making intended usages a little more obvious.

## Issue Type

- [ ] New Feature
- [X] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
ActionFooter should probably not define a border. It could just be a component for laying out items without any visual artifacts of its own, as borders could be applied to the container around it. Alternatively, a prop could be used to configure if the component should have a border or be borderless.

## Current Behavior
Both terra-action-footer and terra-dialog define a top border for the footer.

bjankord:
Let's add a prop to handle this. We're open to a PR for this @dv297

dv297:
I assume tech designs are needed?

# Tech Design

Add a new prop to ActionFooter to remove the border

| Prop Name   | Description | Required? | Default Value? |
| ----------- | ----------- | --------- | -------------- |
| borderless  | Remove the border from the ActionFooter | No | false | 

Proposed API
```jsx
<ActionFooter
  start={<a href="/link/to/somewhere">A link</a>}
  end={
    <div>
      <Button text="Submit"/>
      <Button text="Cancel"/>
    </div>
  }
  borderless={true}
/>

// Or with shorthand
<ActionFooter
  start={<a href="/link/to/somewhere">A link</a>}
  end={
    <div>
      <Button text="Submit"/>
      <Button text="Cancel"/>
    </div>
  }
  borderless
/>
```

dv297:
Updated tech design to use isBorderless

bjankord:
+1 on tech design, lets add the same prop to action header.

emilyrohrbough:
+1

mjhenkes:
+1

jimipasa:
+1 on adding the same prop to the action header as well

tbiethman:
[...]

Issue #1741, opened 2018-07-30
Deprecate Terra Dialog


dkasper-was-taken:
# Feature Request
Deprecate Terra Dialog.

## Description
Terra Dialog is no longer needed, as improvements to terra-action-header, terra-action-footer, and terra-content-container have supplanted it.

This will resolve #1614 and #1461 

## OCS Requested Documentation
When Terra Dialog is removed, we need to transition the component information into an Example/Recipe section of Terra-UI docsite, so it is clear for consumers how they should construct a dialog using `terra-content-container`, `terra-action-header`, `terra-action-footer`, etc. and ways to add content including the body section, fixed alerts/notifications, and patient identification/demographic banners.

bjankord:
Plan to deprecate via npm.

PR #1754, opened 2018-08-02
Add borderless prop to ActionFooter and ActionHeader

dv297:
### Summary
Addresses #1461 , sorry it took so long!

### Additional Details
Already had name in CONTRIBUTOR.md for reviewers checking that.

@cerner/terra


packages/terra-action-footer/src/ActionFooter.jsx
```diff
@@ -16,19 +16,25 @@ const propTypes = {
...
+  /**
+   * Determines whether or not a border is shown
+   */
+  isBorderless: PropTypes.bool,
 };
 
 const defaultProps = {
   start: undefined,
   end: undefined,
+  isBorderless: false,
```
ryanthemanuel: What do people think about making this hasBorder with a default of true instead?

bjankord:
I'm fine adding this change, though we've [recently decided to deprecate this component](https://github.com/cerner/terra-core/issues/1741) in favor of combining terra-content-container and terra-action-header / terra-action-footer

packages/terra-action-footer/src/ActionFooter.jsx
emilyrohrbough: Don't we typically want default our boolean props to false?

packages/terra-action-footer/src/ActionFooter.jsx
```diff
@@ -16,19 +16,25 @@ const propTypes = {
...
 };
 
 const ActionFooter = ({
   start,
   end,
+  isBorderless,
   ...customProps
 }) => (
-  <ActionFooterContainer {...customProps} className={cx(customProps.className)}>
+  <ActionFooterContainer isBorderless={isBorderless} {...customProps} className={cx(customProps.className)}>
```
yuderekyu: Let's spread `customProps` first, otherwise there's a chance to override `isBorderless`

packages/terra-action-footer/src/CenteredActionFooter.jsx
```diff
@@ -29,6 +35,7 @@ const CenteredActionFooter = ({
 
   return (
     <ActionFooterContainer
+      isBorderless={isBorderless}
```
yuderekyu: Same here

packages/terra-action-footer/src/_ActionFooterContainer.jsx
```diff
@@ -8,21 +8,25 @@ const cx = classNames.bind(styles);
 
 const propTypes = {
   children: PropTypes.node,
+  isBorderless: PropTypes.bool,
```
emilyrohrbough: Lets add a prop description here. Even though its an internal component, its nice to have quick reference to props intention in the component itself.

packages/terra-action-footer/tests/jest/ActionFooter.test.jsx
```diff
@@ -91,6 +96,14 @@ describe('ActionFooter', () => {
         expect(actionFooter).toMatchSnapshot()
       ));
     });
+
+    describe('with no border', () => {
```
emilyrohrbough: I think we can drop one of these jest test. Seems to be testing the same thing

emilyrohrbough: With that being said, should have matching tests ActionFooter and CenterActionFooter.

packages/terra-action-footer/src/terra-dev-site/doc/example/standard/Borderless.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import Button from 'terra-button';
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import ActionFooter from '../../../../../src/ActionFooter';
```
emilyrohrbough: This needs to be `import ActionFooter from 'terra-action-footer/lib/ActionFooter`. Our hot reloading can resolve this during testing, however we need this import to be correctly resolved for the terra-ui site.

packages/terra-action-footer/src/terra-dev-site/doc/example/centered/Borderless.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import Button from 'terra-button';
+import Spacer from 'terra-spacer';
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import CenteredActionFooter from '../../../../../src/CenteredActionFooter';
```
emilyrohrbough: This needs to be `import ActionFooter from 'terra-action-footer/lib/ActionFooter`. Our hot reloading can resolve this during testing, however we need this import to be correctly resolved for the terra-ui site.

packages/terra-action-header/src/terra-dev-site/doc/example/BorderlessActionHeader.jsx
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable no-alert */
+import React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+// import ActionHeader from 'terra-action-header/lib/ActionHeader';
```
emilyrohrbough: This needs to be `import ActionHeader from 'terra-action-footer/lib/ActionHeader `. Our hot reloading can resolve this during testing, however we need this import to be correctly resolved for the terra-ui site.

emilyrohrbough: Not these only need to be updated in the example files.

neilpfeiffer:
> Talked with our UX designer this morning about this change. There is a desire to add the isBorderless prop to the dialog component instead of the action-header / action-footer components. My understanding is the dialog API is more flexible compared to the action-header / action-footer API from a design perspective.

Thanks @bjankord. The ActionHeader and ActionFooter a currently more than just purely for layout-only, UX has defined them to fully composed components, including the container and borders as part of the controlled style and usage, and so we would not like to remove them with an additional prop via the ActionHeader and ActionFooter.

To better solve for your original issue, it would be preferred to add a similar `isFooterBorderless` prop to the terra-dialog instead to removes the background-color, border, and padding from the footer area in the Dialog (so that when using the ActionFooter with-in it, the ActionFooter can correctly provide them), even though terra-dialog has been marked for depreciation with #1741 since you can now achieve the same layout with purely the content-container alone with action-header and action-footer.

\
\
On a separate note - the current ActionFooter had some default theme styles missing when originally created that need to be updated, I have listed the details here: #1764

dv297:
If that's the case, I'm going to go ahead and close this PR. Not totally sure if there is actually still a desire to add the isFooterBorderless, with all this discussion of deprecation. I'll leave it up to the Terra core team to decide, but we can probably close the original issue with the recommendation of using content-container rather than dialog.

Issue #1210, opened 2018-02-05
Add Better Error messages to components missing the IntlWrapper


JakeLaCombe:
# Issue Description
Several of our components need localization add, meaning they need to be wrapper in the I18nProvider, or our recommended approach of wrapping the component in Base. When it's not wrapped, the error message says this

`return new Error('Please add locale prop to Base component to load translations');`

However, this isn't an accurate statement. It's throwing the error because the component wasn't wrapped with intl, thus not attaching the intl to the context of component. It probably should be reworded to say

`Component requires Localization, and must be wrapped in terra-base`

## Issue Type
- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Error message says `Component requires Localization, and must be wrapped in terra-base`

## Current Behavior
Error message says `Please add locale prop to Base component to load translations`

PR #1758, opened 2018-08-02
changed table background color

amichaelparker:
### Summary
changed table background color

PR #1759, opened 2018-08-03
Error message updates

amichaelparker:
### Summary
Updated error messages with suggested text from [issue #1210 ](https://github.com/cerner/terra-core/issues/1210).


packages/terra-table/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Changed background color to transparent
```
ryanthemanuel: Should this be a separate PR?

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -69,7 +69,7 @@ const contextTypes = {
   /* eslint-disable consistent-return */
   intl: (context) => {
     if (context.intl === undefined) {
-      return new Error('Please add locale prop to Base component to load translations');
+      return new Error('Component requires Localization, and must be wrapped in terra-base');
```
emilyrohrbough: Would it make more sense to state that the component is internationalized and must be wrapped in base rather than saying it requires localization?

packages/terra-table/CHANGELOG.md
yuderekyu: it's here https://github.com/cerner/terra-core/pull/1758. looks like you branched off that branch instead of master

amichaelparker: Worked with @bjankord on reverting commit, should be correct now.

packages/terra-action-header/src/ActionHeader.jsx
amichaelparker: Spoke with @JakeLaCombe regarding error text changes, he agrees with the changes suggested.

packages/terra-table/CHANGELOG.md
```diff
@@ -3,8 +3,6 @@ Changelog
 
 Unreleased
 ----------
-### Changed
```
ryanthemanuel: Did this accidentally get removed after a merge?

amichaelparker: Yep! Worked with @bjankord on appropriate correction.

PR #1762, opened 2018-08-06
Terra.should.beAccessible, Terra-search-field

BenBoersma:
### Summary
Addresses the Terra-search-field part of https://github.com/cerner/terra-core/issues/1370
Also removes the limiting accessibility tests for terra-collapsible-menu-view

@cerner/terra


packages/terra-search-field/src/SearchField.jsx
```diff
@@ -86,6 +92,7 @@ const defaultProps = {
   placeholder: '',
   searchDelay: 250,
   value: undefined,
+  inputAttributes: undefined,
```
emilyrohrbough: @StephenEsser Should we provide a default for this? I don't remember what we had decided

packages/terra-search-field/src/terra-dev-site/test/search-field/DefaultSearchField.test.jsx
```diff
@@ -2,4 +2,4 @@ import React from 'react';
 import SearchField from '../../../SearchField';
 
 export default () =>
-  <SearchField id="searchfield" />;
+  <SearchField id="searchfield" inputAttributes={{ 'aria-label': 'Search Field' }} />;
```
emilyrohrbough: This removes the need for accessibility in our test examples, however, should we be indicating to consumers that an ariaLabel should be included?

BenBoersma: Currently in the description of inputAttributes on terra-dev-site it reads

> Custom input attributes to apply to the input field such as aria-label.

I am open to better conveying that the consumer needs to define aria-label but unsure of what would be the better way to do so.

Tangential note:
Would there be value in having the default for inputAttributes including a defined Aria-Label?

packages/terra-search-field/src/SearchField.jsx
StephenEsser: We discussed re-enabling the lint rule that enforces any non-required props to have a default so we should probably move forward with providing defaults. Eventually re-enabling this rule to enforce it. 

https://github.com/cerner/eslint-config-terra/blob/master/eslint.config.js#L15

emilyrohrbough: Should this be {} then since we are expecting an obj ?

StephenEsser: My preference is to default to `undefined`, but there is no hard rule on this and there are cases of both throughout the library.

I prefer to default to `undefined` so the internal logic for non-require props is always identical to what the consumer has declared and it works smoother with utilities like redux where there may or may not be an object.

packages/terra-search-field/src/terra-dev-site/test/search-field/DefaultSearchField.test.jsx
bjankord: I'm concerned users may not know to include  aria-labels for this, even if we document it. Instead, I'd like to move forward with providing defaults for aria-labels for the input and the button. If the user passes in `inputAttributes` with `aria-label` on it, we could use that instead of the default we provide. This way, users get a form input / button combo that has i18n labels by default, but still have the option to pass in their own input label if they want.

packages/terra-search-field/src/SearchField.jsx
JakeLaCombe: What does Object.assign do if inputAttributes is undefined? Does it make a new hash, or does it crash? If it crashes, I strongly suggest defaulting to {}, as we shouldn't require the developer to pass in any options if they don't need any

JakeLaCombe: I tested it, and it looks like undefined creates the blank hash just fine.

BenBoersma:
Currently waiting on translations for "Submit Search" so we have default aria-labels.

https://jira.cerner.com/browse/TRANS-5504

packages/terra-search-field/src/terra-dev-site/test/search-field/SearchFieldWithPlaceholder.test.jsx
```diff
@@ -2,4 +2,4 @@ import React from 'react';
 import SearchField from '../../../SearchField';
 
 export default () =>
-  <SearchField id="searchfield" placeholder="Search Text" />;
+  <SearchField id="searchfield" placeholder="Search Text" inputAttributes={{ 'aria-label': 'Search Field' }} />;
```
emilyrohrbough: Should remove these custom input attributes then?

BenBoersma: I should and have now done so. I left the Jest test in there, as I feel it still has use in testing overwriting the default value.

Issue #1727, opened 2018-07-23
Form-Select: additional themeable variable


neilpfeiffer:
# Bug Report

## Description

The themeable variable `--terra-select-padding-bottom` is currently set to have two different default values:

```
.default { 
    ...
    padding-bottom: var(--terra-select-padding-bottom, 0);
```
https://github.com/cerner/terra-core/blob/d2e933158c8a88c1cb92e0bbf1a6cfacd78961b4/packages/terra-form-select/src/_Frame.module.scss#L17

and 
 
```
.tag,
.search,
.combobox,
.multiple {
> .display {
      ...
      padding-bottom: var(--terra-select-padding-bottom, 0.21428571428571427rem);
```
https://github.com/cerner/terra-core/blob/d2e933158c8a88c1cb92e0bbf1a6cfacd78961b4/packages/terra-form-select/src/_Frame.module.scss#L195

## Possible Solution

Add new padding variables (for .tag, .search, .combobox, .multiple) to provide correct theme-ability options. 

Also may need to add new variable for input height:
https://github.com/cerner/terra-core/blob/d2e933158c8a88c1cb92e0bbf1a6cfacd78961b4/packages/terra-form-select/src/_Frame.module.scss#L156

PR #1772, opened 2018-08-14
 Fix for having additional themeable variable for form-select

m-samkc:
### Summary
Updated `_Frame.module.scss` to have theme-able variable for `tag`, `search`, `combobox` and `multiple`. This is to address [#1727](https://github.com/cerner/terra-core/issues/1727) issue.

### Additional Details
- Verified running locally. 

@cerner/terra

### DIFF
--- a/packages/terra-form-select/src/_Frame.module.scss
+++ b/packages/terra-form-select/src/_Frame.module.scss
@@ -192,7 +192,6 @@
       cursor: text;
       display: flex;
       flex: 1 1 auto; // IE 10
-      padding-bottom: var(--terra-select-padding-bottom, 0.21428571428571427rem);
       padding-left: var(--terra-select-padding-left, 0.42857rem);
       padding-right: var(--terra-select-padding-right, 0);
       padding-top: var(--terra-select-padding-top, 0);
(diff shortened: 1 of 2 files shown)


packages/terra-form-select/src/_Frame.module.scss
```diff
@@ -267,6 +266,30 @@
...
+    > .display {
+      padding-bottom: var(--terra-combobox-select-padding-bottom, 0.21428571428571427rem);
+    }
+  }
+  
+  .multiple {
+    > .display {
+      padding-bottom: var(--terra-multiple-select-padding-bottom, 0.21428571428571427rem);
+    }
+  }
```
StephenEsser: Thanks for working on this Mahesh. I have a few concerns about this solution. My initial concern is creating duplicate / identical variables for each variant. I believe the Select component should be themed by content rather than by the variant type. This solution opens the door to individual variant theming. Meaning the combobox, multiple, search, and tag could all display with different amounts of padding, color, and other styles. And requires theming 4 variables to meet a singular common goal. If we were to expand this concept to the other variables this Select component would grow to have over 600 variables to theme each variant uniquely. 

Part of the problem could potentially be solved by adding a theme-able padding bottom to the tags wrapper. https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/_Frame.module.scss#L123

This would theme the wrapping content of tags for multiple and tag variants. The other part of this problem is themeing the padding / margin for search and combobox variants. This may require adding a content div around the input or theming the input slightly differently to account for the change in margins / padding.

StephenEsser: Here is an example of another potential solution, this adds a wrapping div around the `combobox` and `search` variants and allows them to be themed together with `tag` and `multiple` variants. A better name than `content` should be chosen.

https://github.com/cerner/terra-core/commit/b75b5bcf1807128dd05e298f32aafc2d96d3eefe

bjankord:
Reading through the linked issue, it sounded like this is what UX wanted. cc/ @neilpfeiffer Though I'm open to proposed solution @StephenEsser shared https://github.com/cerner/terra-core/commit/b75b5bcf1807128dd05e298f32aafc2d96d3eefe

StephenEsser:
I interpreted the proposed solution by UX to be intended as just a possible solution to remove the duplicate theme variable with different values. Targeting specific variants of the Select creates a complicated / large theming API.

bjankord:
@m-samkc Let's move forward with the code change proposed in this commit: https://github.com/cerner/terra-core/commit/b75b5bcf1807128dd05e298f32aafc2d96d3eefe

StephenEsser:
Closing this pull request to pick up the work discussed above.

PR #1781, opened 2018-08-21
added refCallback prop to terra-button

### LINKED ISSUE #1780: Add refCallback prop to terra-button
# Feature Request

## Description

We'd like to add a refCallback prop to terra-button that gets passed into the ref of the resulting <ComponentType />

The refCallback pattern is common to other terra components, such as [terra-form-imput](https://github.com/cerner/terra-core/blob/master/packages/terra-form-input/src/Input.jsx#L46)

Matthematic:
### Summary
Added the refCallback prop to terra-button and passed it into the <ComponentType /> object during render.
fixes #1780

### DIFF
--- a/packages/terra-button/src/Button.jsx
+++ b/packages/terra-button/src/Button.jsx
@@ -78,6 +78,10 @@ const propTypes = {
    */
   onKeyUp: PropTypes.func,
   /**
+   * Callback ref to pass into the dom element.
+   */
+  refCallback: PropTypes.func,
+  /**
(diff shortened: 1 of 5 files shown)


packages/terra-button/src/Button.jsx
```diff
@@ -235,6 +241,9 @@ class Button extends React.Component {
         onClick={onClick}
         onFocus={onFocus}
         href={href}
+        ref={(inputRef) => {
+          if (refCallback) refCallback(inputRef);
```
StephenEsser: This can be shortened to `ref={refCallback}`

Matthematic: Good catch, updated

PR #1791, opened 2018-08-24
Accounting for multiple exports

m-samkc:
### Summary
Updating `terra-prop-table` to leverage `react-docgen`'s multiple component export definition.  
Resolves: #1795 

### Additional Details
If the component has multiple exports defined ( example: `named` and `default` export ) `react-docgen` default `parse` function thrown an exception. Hence uplifting the `parse` function to accept `findAllComponentDefinitions` as second argument. 

#### Verification
- Verified locally for component with one or more exports ( props table was generated fine )


JakeLaCombe:
Is there a linked issue for this PR?

m-samkc:
@JakeLaCombe : We didn't open any issue for this, however we faced this problem while addressing multiple exports from the same component.

bjankord:
@m-samkc Please add your named to https://github.com/cerner/terra-core/blob/master/CONTRIBUTORS.md

packages/terra-props-table/src/PropsTable.jsx
```diff
@@ -1,12 +1,12 @@
 /* eslint-disable import/no-extraneous-dependencies */
 import React from 'react';
 import PropTypes from 'prop-types';
-import { parse } from 'react-docgen';
```
emilyrohrbough: You are still using the parse method below. Could we keep this import and add the resolver to keep our bundle size smaller?

m-samkc: @emilyrohrbough Thanks Emily, when i imported [findAllComponentDefinitions](https://github.com/reactjs/react-docgen/blob/master/src/resolver/index.js#L18) i couldn't get resolver to work. I will take a look again.

emilyrohrbough: @m-samkc  I think you would want to import the resolver:

```
import { parse,  resolver} from 'react-docgen';
```

Then below:
```
const componentMetaData =parse(src, resolver.findAllComponentDefinitions);
```

Issue #1788, opened 2018-08-23
Switch usage of shields.io badges to badgen.net badges


bjankord:
# Feature Request

## Description
shields.io badges have a noticable delay in rendering. We should switch to a [badge service](https://badgen.net/) with improved render time.

From badgen doc's:  

> as time goes by Shields gets slower, leaves more and more broken badges in READMEs. Badgen is trying to be a fast alternative with simplicity and flexibility.  The badgen library was born as an exploration of "is it possible to generate badge svg markup directly with JavaScript (without using pdfkit/canvas/puppeteer to measure text length)?". Result is better than I expected, Verdana (the de-facto font for badges) text width can be calculated precisely with a prebuilt char-width-table, even no need to worry about kerning

PR #1793, opened 2018-08-26
Switch badges to use badgen-service

bjankord:
### Summary
Updates repo to use faster badge service.


emilyrohrbough:
Addresses https://github.com/cerner/terra-core/issues/1788

PR #1798, opened 2018-08-28
Add ability to render custom content in doc-template

### LINKED ISSUE #1796: Update doc-template component to support custom props-table
# Feature Request

## Description
Some components our users have built are not compatible with react-docgen. Either they have a list of props-types in a separate file they import across components (react-docgen parse files individual, thus does not support this use-case) or they are using more complex HOC components. In these scenarios, we'd still want people to be able to use the doc-template, however, they'd need to provide their own props documentation instead of using the autogenerated doc via react-docgen.

bjankord:
### Summary
Resolves #1796 

Deployment URL: https://terra-core-deployed-pr-1798.herokuapp.com/#/tests/terra-doc-template/doc-template/default-doc-template

You can view the custom props table at the bottom of this URL.


packages/terra-button/src/terra-dev-site/doc/button/Button.doc.jsx
```diff
@@ -65,6 +65,7 @@ const DocPage = () => (
         componentSrc: ButtonSrc,
       },
     ]}
+    customPropsTable={<div>Custom props table</div>}
```
bjankord: Removed this in a subsequent commit

packages/terra-doc-template/src/DocTemplate.jsx
```diff
@@ -47,6 +47,10 @@ const propTypes = {
     componentSrc: PropTypes.string,
     componentName: PropTypes.string,
   })),
+  /**
+   * Optional custom props table.
+   */
+  customPropsTable: PropTypes.node,
```
emilyrohrbough: Should we open this up to be children so people can any what they'd like?

bjankord: Updated to use children: https://github.com/cerner/terra-core/pull/1798/commits/1dcc90cddfc2dbb245b413627f8a16ef2894483c

packages/terra-doc-template/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ ChangeLog
 
 Unreleased
 ----------
+### Added
+* Added support for new `customPropsTable` prop to display customized prop tables
```
emilyrohrbough: add support for kiddos 👯

PR #1809, opened 2018-08-28
Clean up use of lodash/uniq in terra-props-table

### LINKED ISSUE #1803: Update lodash dependency used in terra-props-table
# Feature Request

## Description
We should look into removing this dependency, I don't think it is being used in props table. If it its used, we should update to new 17.4.10 version of lodash is available that looks to resolve the prototype pollution issues with the previous versions. We should look into updating this dependency.

bjankord:
### Summary
Resolves #1803

### DIFF
--- a/packages/terra-props-table/bin/props-table.js
+++ b/packages/terra-props-table/bin/props-table.js
@@ -4,7 +4,6 @@ const fs = require('fs');
 const path = require('path');
 const commander = require('commander');
 const glob = require('glob');
-const uniq = require('lodash/uniq');
 const { parse } = require('react-docgen');
 const generateMarkdown = require('./generateMarkdown/generateMarkdown');
 const pkg = require('../package.json');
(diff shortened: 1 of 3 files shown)


packages/terra-props-table/bin/props-table.js
```diff
@@ -27,7 +26,7 @@ let filenames = commander.args.reduce((globbed, input) => {
 }, []);
 
 // verify filenames are unique
-filenames = uniq(filenames);
+filenames = [...new Set(filenames)];
```
bjankord: This code does same thing as lodash/uniq with vanilla JS

packages/terra-props-table/package.json
```diff
@@ -31,7 +31,6 @@
     "classnames": "^2.2.5",
     "commander": "^2.9.0",
     "glob": "^7.1.2",
-    "lodash": "^4.17.4",
```
StephenEsser: I know we're removing this all together, but for anyone that needs to use lodash in the future, almost every lodash function has it's own package release which is significantly smaller in size.

https://www.npmjs.com/package/lodash.uniq

StephenEsser: https://www.npmjs.com/search?q=lodash.uniq

mjhenkes: It will help for sure.

PR #1814, opened 2018-08-30
Add new opt-in prop for custom prop-types resolution via react-docgen

### LINKED ISSUE #1812: Props table update is unable to parse various terra components
# Bug Report

## Description
The alert, clinical-item-collection, clinical-item-view, infinite-list, notification-dialog pages are not rendering correctly on the terra-ui doc site. If you check the console, each page errors out with the following error message.

```js
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at m (PropsTable.js:197)
```
## Steps to Reproduce

1. Navigate to any of these pages on the terra-ui site.

## Expected Behavior
These pages should load without issue

## Environment

* Component Name and Version: terra-props-table v2.16.0
* Browser Name and Version: Chrome / Latest

bjankord:
### Summary
Resolves #1812 

The recent update to props table has failed on miscellaneous components, this update reverts to use default behavior while making the custom resolver and opt-in via a new prop, `propsResolution`


bjankord:
I'm gonna run through all the component pages in terra-core to make sure the prop tables all render correctly before merging this down.

PR #1816, opened 2018-08-31
Terra toggle

AlexBrizi:
### Summary
Removed the extra padding on the terra-toggle component.


yuderekyu:
lets include the associated issue in your description

bjankord:
@AlexBrizi The screenshots for terra-toggle-button and terra-toggle-header will also need to be regenerated with this update to terra-toggle.

Issue #987, opened 2017-11-02
Terra Base should set directionality


mhemesath:
# Issue Description
Terra base should be set directionality via the `dir` attribute based on the passed in locale. This eliminates the common occurrence of engineers forgetting to add it on the HTML tag and also makes sense as it allows `Base` to support nested locale directionality automatically.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Using `<Base locale='en'></Base>` will result in a div wrapper of `<div dir='ltr' lang='en'></div>`
Using `<Base locale='ar'></Base>` will result in a div wrapper of `<div dir='rtl' lang='ar'></div>`

## Current Behavior
Directionality is expected to be set outside of react on the HTML document level.

ryanthemanuel:
(Comments from the PR)

I'm not a fan of this route for several reasons. In addition to Emily's thoughts (https://github.com/cerner/terra-core/pull/1833/files#r215338948):

* I don't think we should be in the business of making assumptions about ltr or rtl based on locale.
* We support custom locales, and if that custom locale happens to be rtl, there is no way to allow this.
* Even if we were to try and determine the direction based on locale, I think we'd want to relocate this change closer to other locale concepts (i.e. https://github.com/cerner/terra-toolkit/blob/master/scripts/aggregate-translations/i18nSupportedLocales.js) so that it is easier to maintain.

PR #1833, opened 2018-09-05
Set Direction on Terra Base

amichaelparker:
### Summary
Changed terra-base to set lang and direction on the created div based on the passed in locale.

Closes issue [#987](https://github.com/cerner/terra-core/issues/987)


bjankord:
Let's discuss more in our meetings on Friday.

Issue #1844, opened 2018-09-06
[terra-form-textarea] Spread custom props first.


yuderekyu:
# Bug Report
Stop custom props from overriding first class props. Pass custom props first, instead of last. 

- [textarea](https://github.com/cerner/terra-core/blob/master/packages/terra-form-textarea/src/Textarea.jsx#L206)
- [textareaField](https://github.com/cerner/terra-core/blob/master/packages/terra-form-textarea/src/TextareaField.jsx#L134)

PR #1848, opened 2018-09-10
Moved custom prop spreads in textarea and field tags

amichaelparker:
### Summary
From [issue #1844](https://github.com/cerner/terra-core/issues/1844). Custom props spreads for the textarea and Field tags were moved first.

Closes [#1844](https://github.com/cerner/terra-core/issues/1844)


packages/terra-form-textarea/src/Textarea.jsx
```diff
@@ -203,7 +204,6 @@ class Textarea extends React.Component {
         onChange={this.onChange}
         required={required}
         rows={textareaRows}
-        {...additionalTextareaProps}
```
StephenEsser: I think this is *technically* non-passive, but needed to be reorder to work appropriately

JakeLaCombe: I wouldn't worry too much about the passivity here IMO. Yeah, it technically is non passive, but I can't envision we have a lot of use cases for this scenario or needing to override classes, rows, and the other first class attributes.

packages/terra-form-textarea/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
```
emilyrohrbough: We should update this heading to be `## Fixed` since this was incorrect initially

PR #1850, opened 2018-09-10
Update aria-current for readability on screen readers

### LINKED ISSUE #1837: Make Terra Paginator more accessible
# Bug Report

## Description
Terra Paginator needs more enhancements in order for it to be fully accessible:

For every page, it should say “Goto Page #”
For the current page it should say “Current Page, Page #”
First, Previous, next and last stay the same
These rule apply for both paginates

https://medium.com/@jessebeach/beware-smushed-off-screen-accessible-text-5952a4c2cbfe

JakeLaCombe:
### Summary
Resolves #1837

Before this change, screen readers would read these elements as Page Page 1, Page Page 2, etc... With this change, the word page is left off of the screen reader, and instead shot off to the aria-label.


packages/terra-paginator/src/ControlledPaginator.jsx
```diff
@@ -96,7 +96,7 @@ class Paginator extends React.Component {
       pageButtons.push((
         <a
           aria-label={`Page ${val}`}
-          aria-current={val === selectedPage && 'page'}
+          aria-current={val === selectedPage && 'true'}
```
yuderekyu: just curious - does `aria-current` expect the string `'true'`, or does the boolean work?

JakeLaCombe: I think we may need to keep it like this. It's an enumeration actually. https://tink.uk/using-the-aria-current-attribute/

JakeLaCombe: It works! Changed it.

yuderekyu: yay!

packages/terra-paginator/src/ControlledPaginator.jsx
```diff
@@ -96,7 +96,7 @@ class Paginator extends React.Component {
       pageButtons.push((
         <a
           aria-label={`Page ${val}`}
-          aria-current={val === selectedPage && 'page'}
+          aria-current={val === selectedPage && true}
```
StephenEsser: Is it possible to use just the equivalence statement Derek's referenced here? https://github.com/cerner/terra-core/pull/1850#discussion_r217542174

JakeLaCombe:
Closing this. There are more changes that we need to make in order to get this to be accessible

PR #1851, opened 2018-09-10
Add valueText prop onto progress bar for readability.

### LINKED ISSUE #1838: Terra Progressbar doesn't read current value accurately
# Bug Report

## Description
Progress bars are leaving off the unit when they are read to the user, making it difficult to see what the progress bar is trying to say. To fix this, it's recommend to use aria-valuetext as a prop on the progress bar.

For color blindness, it's more ideal to use textures instead of straight up color. Perhaps consider using textures for some cases? That maybe better left to another Jira.

JakeLaCombe:
### Summary
Resolves #1838

Progress Bars have a hard time reading values of a progress bar to users in a human readable fashion. To resolve this, aria-valuetext can be used which is the text that is read to users through the screen reader.


packages/terra-progress-bar/src/ProgressBar.jsx
```diff
@@ -27,19 +27,25 @@ const propTypes = {
    * Boolean check for gradient.
    */
   hasGradient: PropTypes.bool,
+  /**
+   * Value passed to aria-valuetext for accessibility
+   */
+  valueText: PropTypes.string,
```
emilyrohrbough: t seems like this prop should be required if needed for accessibly. Because we need to be passive, should we
1) log an issue to make this prop required in the next MVB
2) and add a non-prod warning message saying this prop will be made required in the next MVB and should be supplied for accessibility?

StephenEsser: I don't think this attribute is always required, reading through the documentation is sounds like screen readers should be able to interpret the valueNow as a percentage and the valueText is usually used to represent non-numeric / percentage based valueNows. But it does make me wonder if we should be locking down the other aria attributes for min, max, and valueNow

https://www.w3.org/WAI/PF/aria/states_and_properties#aria-valuetext

https://www.w3.org/WAI/PF/aria/states_and_properties#aria-valuenow

JakeLaCombe: What are you thinking of locking them down to?

emilyrohrbough: If this shouldn't be required, can we beef up the description to describe when this prop should be used with a link to the accessibility information? I imagine consumers would find that useful!

emilyrohrbough:
@JakeLaCombe  Should we add an example / test to verify accessibly when this value isn't provided?

mmalaker:
Functional verification completed. The progress bar example values now can be read by screen readers.

PR #1853, opened 2018-09-10
Remove i18n post-install script

### LINKED ISSUE #1377: [terra-18n] Remove Post Install and Default Loaders
# Issue Description
The post install script and default loaders in the loaders directory were add keep passivity for consumers with the changes made in https://github.com/cerner/terra-core/pull/1350 to enhance terra's internationalization.

With these changes, the `terra-i18n` loaders were updated to pull in the dynamically generated `intlLoader` and `translationLoader` modules created by the aggregate-translations pre-build tool, instead of referencing the static intl and translation loaders. These dynamic loaders are configured to the specified locales and resolves this previously seen issue. See https://github.com/cerner/terra-core/issues/855.

To keep passivity for consumers not yet utilizing the aggregate-translations pre-build tool, i.e. using the deprecated `terra-i18n-plugin`, this script adds the previously removed static intl and translation loaders to terra-i18n's `node_modules` directory. This allows the terra-i18n loaders to reference the intlLoader and translationLoader modules. 

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [ ] Bug
- [x] Other

## Expected Behavior
Remove the post-install script and the loaders directory at the next major version bump of terra-i18n or terra-base. (React 17?)

bjankord:
### Summary
Resolves #1377

I'd like to test this in an app that is still using the [terra-i18n aggregate-translations script](https://github.com/cerner/terra-core/tree/master/packages/terra-i18n/scripts/aggregate-translations) and make sure these two files below resolve correctly. The files should resolve from the generated aggregated-translations directory where we write all the i18n loaders / translations to.

* [intlLoaders import](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/src/intlLoaders.js#L2)
* [translationLoaders import](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/src/translationsLoaders.js#L2)

Originally, I removed Jest tests around the translationLoader and intlLoader. Jest wasn't able to resolve these imports [1](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/tests/jest/intlLoaders.test.js#L3) [2](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/tests/jest/translationsLoaders.test.js#L3) after deleting the post-install script. 

I've added an npm script to terra-i18n package that will run before jest tests and will generate an aggregated-translations directory relative to the `terra-i18n/tests/jest` directory that will contain a `intlLoaders.js` and `translationsLoaders.js` file jest will be able to resolve. I've updated the jest config so its able to resolve this files.

### DIFF
--- a/packages/terra-i18n/tests/jest/generate-aggregated-translations-directory.js
+++ b/packages/terra-i18n/tests/jest/generate-aggregated-translations-directory.js
@@ -1,5 +1,6 @@
-const aggregateTranslations = require('../../scripts/aggregate-translations/aggregate-translations');
-const i18nSupportedLocales = require('../../src/i18nSupportedLocales');
+/* eslint-disable import/no-extraneous-dependencies */
+const aggregateTranslations = require('terra-toolkit/scripts/aggregate-translations/aggregate-translations');
+const i18nSupportedLocales = require('terra-toolkit/scripts/aggregate-translations/i18nSupportedLocales');
 
 const aggregateOptions = {
   baseDir: __dirname,


jestconfig.js
```diff
@@ -18,5 +18,10 @@ module.exports = {
     '\\.(svg|jpg|png|md)$': '<rootDir>/__mocks__/fileMock.js',
     '\\.(css|scss)$': 'identity-obj-proxy',
   },
+  moduleDirectories: [
+    'aggregated-translations',
+    './packages/terra-i18n/tests/jest/aggregated-translations',
```
bjankord: This allows Jest to resolve the generated `intlLoaders.js` and `translationsLoader.js` files from the aggregated-translations pre-build tool inside of the terra-i18n tests.

packages/terra-i18n/tests/jest/generate-aggregated-translations-directory.js
```diff
@@ -0,0 +1,11 @@
+const aggregateTranslations = require('../../scripts/aggregate-translations/aggregate-translations');
```
emilyrohrbough: We should use terra-toolkit's version of this as this script is deprecated

emilyrohrbough: We should run & mock this in before hooks instead of adding a pre-jest script

bjankord: Agreed, I [updated](https://github.com/cerner/terra-core/pull/1853/commits/9879ae9453e16626f24a8ef709862a70ba334596) this to use terra-toolkit's version of this script. I don't believe using before hook in test file will work as the before hook would come after [the import for the file](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/tests/jest/intlLoaders.test.js#L3) we'd need to generate in the before hook. The pre-jest script will go away, along with most all of the jest tests, once we remove the aggregate-translations script code from terra-i18n. I want to wait until we are ready to cut a major version bump on terra-i18n before we cut that though.

PR #1869, opened 2018-09-13
Remove Custom Timeouts From Tests

emilyrohrbough:
### Summary
Timeouts should be defined from the global level for wait. This allows us to update in one place if needed. 

[Our global timeout is defined as: 3000](https://github.com/cerner/terra-toolkit/blob/master/config/wdio/wdio.conf.js#L38)

More info on webdriver.io config option: http://webdriver.io/guide/getstarted/configuration.html#waitforTimeout


emilyrohrbough:
@yuderekyu I think when @mhemesath setup the wdio initially, he chose 3000 and we haven't needed to adjust it.

PR #1889, opened 2018-09-18
Add Abbreviation Nodes to Demographics Banner for Accessibility

### LINKED ISSUE #1888: Labels inside Demographics Banner need to be abbreviated.
# Bug Report

## Description
The Demographics Banner contains abbreviations for messages such as Date of Birth and Gestational Age. However, not all users may associated those abbreviations to mean those terms. To resolve this issue, we need to use the abbreviation node so that both screen readers can recognize those terms, and all other users know what those labels mean.

JakeLaCombe:
### Summary
Resolves #1888

This adds abbreviation tags to the demographics banner so that DOB, GA, and PMA are clear on what their meanings are.

### Additional Details
Since a new node was added, new styles have been introduced. May need potential design feedback.

### Deployment Link
https://terra-core-deployed-pr-1889.herokuapp.com/#/components/terra-demographics-banner/demographics-banner/demographics-banner


packages/terra-demographics-banner/src/DemographicsBannerDisplay.jsx
```diff
@@ -41,9 +45,13 @@ const propTypes = {
...
   gestationalAge: PropTypes.string,
   /**
-   * Label to display for the GestationalAge
+   * Label to display for the Gestational Age
    */
   gestationalAgeLabel: PropTypes.string,
+  /**
+   * Full text description for the Gestational Age
+   */
+  gestationalAgeFullText: PropTypes.string,
```
emilyrohrbough: Why is this opened up as a prop?

JakeLaCombe: I think we will need to refactor the demographics banner. When we did this component originally, we wanted to have a version that didn't have translations so that we could test it easier than we did if we required a translation context on the component.

JakeLaCombe: Btw, this component isn't designed to be exported by the consumer. It's still expected that they use the DemographicsBanner which contains the translations.

JakeLaCombe:
Jira has been added for translations.

packages/terra-demographics-banner/src/_sharedObjects.jsx
```diff
@@ -1,23 +1,48 @@
...
-  <span className={cx('value')}>
-    { label && <span className={cx('value-label')}>{`${label}:`}</span> }
-    <b>{value}</b>
-  </span>
-);
+const DemographicsBannerValue = ({ label, value, abbrTitle }) => {
+  let valueLabelContent;
+
+  console.log(label);
+  console.log(abbrTitle);
```
bjankord: Need to remove console.log statements.

packages/terra-demographics-banner/src/_sharedObjects.jsx
```diff
@@ -1,23 +1,46 @@
...
-    <b>{value}</b>
-  </span>
-);
+const DemographicsBannerValue = ({ label, value, abbrTitle }) => {
+  let valueLabelContent;
+
+  if (abbrTitle) {
+    valueLabelContent = (
+      <span className={cx('value-label')}>
+        <abbr className={cx('abbreviation')} title={abbrTitle} aria-hidden="true">
```
emilyrohrbough: The translations and props added to this component indicate we are adding full text labels to the component. Are we changing the default translations that are being displayed or is abbreviation the correct term here?

packages/terra-demographics-banner/src/_sharedObjects.jsx
```diff
@@ -1,23 +1,46 @@
...
+        {`${label}:`}
+      </span>
+    );
+  }
+
+  return (
+    <span className={cx('value')}>
+      {abbrTitle && (<VisuallyHiddenText text={abbrTitle} />)}
+      {valueLabelContent}
+      <b>{value}</b>
```
emilyrohrbough: Should we use css styling here instead of the `<b/>` tag

packages/terra-demographics-banner/src/DemographicsBannerDisplay.jsx
```diff
@@ -24,6 +24,10 @@ const propTypes = {
    * Label to display for the date of birth
    */
   dateOfBirthLabel: PropTypes.string,
+  /**
+   * Full text description for the date of birth
```
emilyrohrbough: Should we mentioned this is hidden text to be used by screen readers?

packages/terra-demographics-banner/src/_sharedObjects.jsx
```diff
@@ -1,23 +1,46 @@
...
-const DemographicsBannerValue = ({ label, value }) => (
-  <span className={cx('value')}>
-    { label && <span className={cx('value-label')}>{`${label}:`}</span> }
-    <b>{value}</b>
-  </span>
-);
+const DemographicsBannerValue = ({ label, value, abbrTitle }) => {
+  let valueLabelContent;
+
+  if (abbrTitle) {
```
emilyrohrbough: Should we still be checking that the `label` prop is valid here as well?

packages/terra-demographics-banner/src/_sharedObjects.jsx
JakeLaCombe: The displayed translations should be what's currently displayed, and the default ones should be what they are. For the abbreviation, we want to display the abbreviated text, and have the full text displayed when this is hovered over or read through a screen reader.

packages/terra-demographics-banner/src/_sharedObjects.jsx
JakeLaCombe: I mostly assumed that if abbrTitle is provided, that the label has to be since that's the full text descriptor of the abbrTitle, but I can be more strict with the prop checking

packages/terra-demographics-banner/src/_sharedObjects.jsx
yuderekyu: yeah, we'll prob need to convert this to a themeable font-weight: https://github.com/cerner/terra-core/issues/1897

packages/terra-demographics-banner/src/_sharedObjects.jsx
emilyrohrbough: hmmm 🤔 I don't have a strong opinion. I just saw we checked for it previously before rendering it

JakeLaCombe: I'll make it more strict. I think that's what is best for our components

packages/terra-demographics-banner/src/_sharedObjects.jsx
JakeLaCombe: Also, I think it's unnecessary to have screenshot behavior for the hover effect. That's a native browser implementation, and I think wdio should only test functionality terra provides on top of what browsers do.

JakeLaCombe:
I have added the Arabic translations. Google Translate was unable to translate the abbreviations, so I resulted to using long string text for those specific labels. I think we may want to revisit this component in the future to get better direction on how to handle the case when a language doesn't have short handed abbreviations available for the text.

mmalaker:
Functional verification completed. The demographics abbreviation examples now display and function correctly.

PR #1890, opened 2018-09-18
Added strictMode prop to terra-base

### LINKED ISSUE #1886: Add optional `strictMode` prop to terra-base
# Feature Request

## Description
Adding the `strictMode` prop to terra-base would allow teams to opt-in to this feature of React and start auditing their apps for errors related to the [new strict mode](https://reactjs.org/docs/strict-mode.html). By default, we'd set this as a boolean value to false. This would also help us audit our components in different repos.

bjankord:
### Summary
Resolves #1886


packages/terra-base/src/Base.jsx
```diff
@@ -73,20 +75,25 @@ class Base extends React.Component {
       children,
       locale,
       customMessages,
+      strictMode,
       translationsLoadingPlaceholder,
       ...customProps
     } = this.props;
 
     const messages = Object.assign({}, this.state.messages, customMessages);
+    const renderStrictMode = strictMode ? (<React.StrictMode>{children}</React.StrictMode>) : children;
```
emilyrohrbough: Should we add a jest test for this scenario?

packages/terra-base/src/Base.jsx
```diff
@@ -23,6 +23,7 @@ const propTypes = {
       return new Error(`Missing locale prop for ${propName} in ${componentName} props`);
     }
   },
+  strictMode: PropTypes.bool,
```
yuderekyu: want to add a descriptive comment? something like `activates additional checks and warnings for descendants`

packages/terra-base/src/Base.jsx
yuderekyu: nit: would it make more sense to name this variable `renderChildren`, since we won't always be rendering in strict mode?

packages/terra-base/src/Base.jsx
bjankord: Updated: https://github.com/cerner/terra-core/pull/1890/commits/05a761b6f58b7a4e600044b670c7a6071f00aa48

packages/terra-base/src/Base.jsx
bjankord: Updated: https://github.com/cerner/terra-core/pull/1890/commits/90c983b9fb7a4492f6805d111c31f8ba02d91a61

bjankord: Updated: https://github.com/cerner/terra-core/pull/1890/commits/ff2991ba0e7d0e67067af930892a713b2fded5b1

PR #1910, opened 2018-09-20
New show hide component

### LINKED ISSUE #1563: Create show hide component
# Issue Description
Create show hide component that follows Cerner's design standard

## Issue Type

- [x] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The show hide component should be developed following Cerner's design standard for showing and hiding content.

gabeparra01:
### Summary
Implementation of the new terra-show-hide component. Closes #1563 
### Additional Details
I had to use a different strategy than the other Terra components for writing the Jest tests- The component requires intl values for the button label. I had to wrap the component in <IntlProvider> (in the Jest tests) every time that it was rendered.

https://terra-core-deployed-pr-1910.herokuapp.com/#/components/terra-show-hide/show-hide/show-hide

@cerner/terra


tb5748:
This pull request doesn't look quite ready.  All three checks show as failed.

bjankord:
There are some lint errors but I'm sure @gabeparra01 can get those cleaned up. If you go to https://travis-ci.org/cerner/terra-core/jobs/431239769 and view the log, it will show the lint error. Or you can run `npm run lint` on the command line and it will list the lint errors.

packages/terra-show-hide/src/ShowHide.jsx
```diff
@@ -0,0 +1,156 @@
...
+   */
+  expandedButtonText: PropTypes.string,
+  /**
+   * Icon displayed next to the button text
+   */
+  icon: PropTypes.element,
+  /**
+   * The intl object to be injected for translations. Provided by the injectIntl function.
+   */
+  intl: intlShape.isRequired,
```
JakeLaCombe: We actually are using these as part of the context instead of with the injectIntl. We don't want to expose the to the props table. https://github.com/cerner/terra-core/blob/master/packages/terra-form-field/src/Field.jsx#L81

packages/terra-show-hide/src/ShowHide.jsx
```diff
@@ -0,0 +1,156 @@
...
+  icon: <IconChevronDown />,
+  isAnimated: false,
+  isInitiallyOpen: false,
+};
+
+class ShowHide extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      isOpen: props.isInitiallyOpen,
```
JakeLaCombe: Should we be doing this as an uncontrolled component? I've seen apps in Cerner that would have a button that would open and expand all of the menus and panels at the same time. If we don't want that, we should at least have a way to change this state value from outside the component. https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#new-lifecycle-getderivedstatefromprops

packages/terra-show-hide/src/ShowHide.jsx
gabeparra01: @StephenEsser Thanks for the example. I will give this a try

gabeparra01: @StephenEsser @JakeLaCombe using this.context is marked as deprecated. are we still ok with using it?
https://reactjs.org/docs/legacy-context.html

StephenEsser: That's what I would recommend until we have a strategy to uplift the other components, unless you're interested in pioneering the new context API, but that may have a large scope of changes to all internationalized components.

gabeparra01: Alright thanks for the update :)

tbiethman: I think what you have (using injectIntl, getting intl through props) is what you want to do. This insulates us from the Context API entirely. If we interface with the legacy context APIs here (and I know we do elsewhere), we'll just have more changes to make in the future. react-intl should be able to passively switch to the new Context API without us noticing.

packages/terra-show-hide/src/ShowHide.jsx
tbiethman: +1 to making this component uncontrolled. I think if you do so, you could also go from having two callbacks (onOpen on onClose) to a single callback (onChange/onToggle/onSomething?). The component could decide what needs to happen in that callback based on the isOpen state (which they would own).

packages/terra-show-hide/src/ShowHide.jsx
```diff
@@ -0,0 +1,156 @@
...
+      customProps.className,
+    ]);
+
+    let previewText = '';
+
+    if (preview) {
+      if (preview.slice(-1) === '.') {
+        previewText = `${preview}..`;
+      } else {
+        previewText = `${preview}...`;
```
tbiethman: Can we rely on the consumer to provide ellipses with their preview text instead?

packages/terra-show-hide/src/ShowHide.jsx
```diff
@@ -0,0 +1,156 @@
...
+        aria-expanded={this.state.isOpen}
+        text={buttonText}
+        onClick={this.handleOnClick}
+      />
+    );
+
+    let isAnimatedStyle = {};
+
+    if (this.props.isAnimated) {
+      isAnimatedStyle = {
```
tbiethman: You could define this style block as a static on the class (or as a const outside the class) to keep a consistent object reference between renders. Doing so should (minutely :)) increase rendering performance.

packages/terra-show-hide/src/ShowHide.jsx
```diff
@@ -0,0 +1,158 @@
...
+  icon: <IconChevronDown />,
+  isAnimated: false,
+  isInitiallyOpen: false,
+};
+
+const animatedStyle = {
+  // fix for the toggle oscillation when setting isAnimated to true
+  transitionDelay: '0.3s',
+  WebkitTransitionDelay: '0.3s', /* Safari */
+};
```
StephenEsser: Does this work in IE10? Can we do the transitions in CSS?

packages/terra-show-hide/src/ShowHide.jsx
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    const button = (
+      <Button
+        icon={<span className={cx('icon')}>{icon}</span>}
+        aria-expanded={this.state.isOpen}
+        text={buttonText}
+        onClick={this.handleOnClick}
+      />
+    );
```
StephenEsser: This is up to your preference, do you have any objections to moving the button down into the render function?

```
    return (
      <div {...customProps} className={showHideClassName}>
        {!this.state.isOpen && previewText}
        <Toggle isOpen={this.state.isOpen} isAnimated={this.props.isAnimated} style={isAnimatedStyle}>
          {this.props.children}
        </Toggle>
        <Button
          icon={<span className={cx('icon')}>{icon}</span>}
          aria-expanded={this.state.isOpen}
          text={buttonText}
          onClick={this.handleOnClick}
        />
      </div>
    )
```

packages/terra-show-hide/src/ShowHide.jsx
emilyrohrbough: Here is an example of using a class. https://github.com/cerner/terra-core/blob/e9f0aad0f345ea10030211a477aa130c4191a3c4/packages/terra-toggle/src/Toggle.module.scss#L14

packages/terra-show-hide/src/ShowHide.jsx
```diff
@@ -0,0 +1,158 @@
...
+    super(props);
+    this.state = {
+      isOpen: props.isInitiallyOpen,
+    };
+
+    this.handleOnClick = this.handleOnClick.bind(this);
+  }
+
+  handleOnClick(e) {
+    e.preventDefault();
```
StephenEsser: What default action do we need prevented?

packages/terra-show-hide/src/ShowHide.jsx
gabeparra01: which css attribute/selector would I use for ``` .toggle[] ```? 

``` .toggle[aria-hidden='false']``` works because aria-hidden is a valid selector. I don't see one for animated

File: packages/terra-show-hide/src/ShowHide.jsx (hunk 6 above)
Comment (gabeparra01) [reply packages/terra-show-hide/src/ShowHide.jsx]: No objections from me. this change added a little bit more space between the link/button and the text so I had to generate new screenshots for the webdriver tests: https://github.com/cerner/terra-core/pull/1910/commits/f11dc6a8475a61b39ddc9f204be8da50e511fa48

File: packages/terra-show-hide/src/ShowHide.jsx (hunk 7 above)
Comment (gabeparra01) [reply packages/terra-show-hide/src/ShowHide.jsx]: I copied this function from terra-toggle-button. I assumed it was a bug fix for something. It might not be necessary now though. @bjankord do you recall why this line was added to terra-toggle-button?

File: packages/terra-show-hide/README.md [hunk 8]
```diff
@@ -0,0 +1,26 @@
+# Terra Show Hide
+
+
+[![NPM version](http://img.shields.io/npm/v/terra-show-hide.svg)](https://www.npmjs.org/package/terra-show-hide)
```
emilyrohrbough: Update to be:
```
[![NPM version](https://badgen.net/npm/v/terra-show-hide)](https://www.npmjs.org/package/terra-abstract-modal)
[![Build Status](https://badgen.net/travis/cerner/terra-core)](https://travis-ci.org/cerner/terra-core)
```

packages/terra-show-hide/docs/README.md
```diff
@@ -0,0 +1,30 @@
...
+import ShowHide from 'terra-show-hide';
+
+  <ShowHide preview="Lorem ipsum dolor sit amet">
+    {Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.}
+  </ShowHide>
+```
+
+## Component Features
+
+<!-- Uncomment supported features.
```
emilyrohrbough: Add the supported features & update links to point to terra-ui: https://engineering.cerner.com/terra-ui/#/getting-started/terra-ui/features

packages/terra-show-hide/package.json
```diff
@@ -0,0 +1,48 @@
...
+    "react": "^16.2.0",
+    "react-dom": "^16.2.0",
+    "terra-base": "^3.1.0"
+  },
+  "dependencies": {
+    "classnames": "^2.2.5",
+    "prop-types": "^15.5.8",
+    "terra-doc-template": "^1.2.0",
+    "terra-icon": "^2.19.0",
+    "react-intl": "^2.4.0",
```
emilyrohrbough: What are `react-intl` & `react-test-renderer` included as dependencies?

tbiethman: react-intl definitely should not be

packages/terra-show-hide/README.md
gabeparra01: https://github.com/cerner/terra-core/pull/1910/commits/3123a966699da53c55bd73f6c69a7aad5d33b8f4

packages/terra-show-hide/docs/README.md
gabeparra01: Which features will be supported?

packages/terra-show-hide/src/ShowHide.jsx
JakeLaCombe: I think I confused you as well. Sorry about that.

Right now, the component is uncontrolled as the isOpen status is maintained inside this component. For a controlled component, the parent would store the state of isOpen, and this state I believe wouldn't even be set. The request is to go with the controlled route, where instead of this.state.isOpen, you would look at this.props.isOpen

JakeLaCombe: If you would like an example of the different components, check out our paginator code. https://github.com/cerner/terra-core/tree/master/packages/terra-paginator

packages/terra-show-hide/src/_ShowHideButton.module.scss
```diff
@@ -0,0 +1,103 @@
...
+    color: var(--terra-link-color, #0059a8);
+    cursor: pointer;
+    display: inline-block; // prevents variant icon from breaking onto its own line
+    font-family: inherit;
+    font-size: var(--terra-button-font-size, 1rem);
+    line-height: var(--terra-button-line-height, 1.429);
+    outline: none;
+    padding: 0;
+    text-align: center;
+    text-decoration: var(--terra-link-underline, none);
```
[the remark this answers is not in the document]
gabeparra01: I am not sure what you mean by this? Also, I used the same underline styling as terra-hyperlink for consistency

packages/terra-show-hide/src/_ShowHideButton.module.scss
```diff
@@ -0,0 +1,103 @@
...
+  input[type='reset'].button,
+  input[type='submit'].button {
+    appearance: button;  // autoprefix fix into -webkit-appearance
+  }
+  // stylelint-enable selector-no-qualifying-type
+
+  .button {
+    background-color: transparent; // Remove the gray background on active links in IE 10
+    border: 0;
+    color: var(--terra-link-color, #0059a8);
```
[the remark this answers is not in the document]
gabeparra01: ok sounds good: https://github.com/cerner/terra-core/pull/1910/commits/08cebcf168467e2e8ed5ca98ef37e512c319a164

packages/terra-show-hide/tests/wdio/show-hide-spec.js
```diff
@@ -0,0 +1,116 @@
...
+
+describe('Show Hide', () => {
+  before(() => browser.setViewportSize(Terra.viewports('medium')[0]));
+
+  describe('Default', () => {
+    before(() => browser.url('/#/raw/tests/terra-show-hide/show-hide/default-show-hide'));
+
+    Terra.should.matchScreenshot('collapsed');
+    Terra.should.beAccessible();
+
```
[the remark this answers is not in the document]
gabeparra01: I never knew about that one before, thanks for pointing that out: https://github.com/cerner/terra-core/pull/1910/commits/529a63611311c9da626377a49fc7140f8c490160

packages/terra-show-hide/src/ShowHide.jsx
emilyrohrbough: The class name can differ, but the style can be applied via stylesheets once without adding the webkit extension

gabeparra01: So if I'm understanding you correctly, instead of having ``` if (this.props.isAnimated) ```, I would just add an additional class to the class name declaration? Like this: ```
if (this.props.isAnimated) {
const showHideClassName = cx([
      'button',
      { 'is-open': this.state.isOpen },
      { 'animated': this.props.isAnimated },
      customProps.className,
    ]);```

packages/terra-show-hide/src/ShowHide.jsx
StephenEsser: Interesting, this change should not have caused any screenshot updates.

emilyrohrbough: @StephenEsser  Do you think this was caused from https://github.com/cerner/terra-core/pull/1816?

packages/terra-show-hide/src/ShowHide.jsx
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * Sets the initial state of the component to open and expanded
+   */
+  isInitiallyOpen: PropTypes.bool,
+  /**
+   * Callback function triggered when the component is collapsed
+   */
+  onClose: PropTypes.func,
+  /**
+   * Callback function triggered when the component is expanded
```
StephenEsser: Minor nit, to be consistent can we add period punctuation to each of these descriptions?

packages/terra-show-hide/src/ShowHide.jsx
gabeparra01: @emilyrohrbough helped me with the refactor for the styling as a CSS class: https://github.com/cerner/terra-core/pull/1910/commits/6979d941a0a0749df9f9ffeae82def7c0b9f34c5
I still need to test it in IE10

gabeparra01: I can confirm that the transition-delay styling works in IE10. I used the latest deployment to test: https://terra-core-deployed-pr-1910.herokuapp.com/#/components/terra-show-hide/show-hide/show-hide

And here is a screenshot of the Windows 7 virtual machine I am running through VirtualBox to test IE10: 
image

packages/terra-show-hide/src/ShowHide.jsx
gabeparra01: No worries about the terminology mixup, I do it too. I am still confused about this thread though. @tb5748 said in his comment that after this refactor, the component should not have internal state. But the terra-paginator example that @JakeLaCombe provided manages its own state variables

tbiethman: The ControlledPaginator is what I think Jake was referring to: https://github.com/cerner/terra-core/blob/master/packages/terra-paginator/src/ControlledPaginator.jsx

gabeparra01: @tbiethman thanks for the clarification!

gabeparra01: @JakeLaCombe @tbiethman  the refactor is ready for review: https://github.com/cerner/terra-core/pull/1910/commits/6979d941a0a0749df9f9ffeae82def7c0b9f34c5

packages/terra-show-hide/docs/README.md
gabeparra01: @emilyrohrbough https://github.com/cerner/terra-core/pull/1910/commits/d21cec6ba55386326e3b4650a3dbb95935e9c0e6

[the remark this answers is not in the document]
gabeparra01: https://github.com/cerner/terra-core/pull/1910#discussion_r220578044

packages/terra-show-hide/src/ShowHide.jsx
gabeparra01: @tbiethman looks like the mockups in Zeppelin were updated and the ellipses were removed: https://github.com/cerner/terra-core/pull/1910/commits/ba4503889c2bc08e4b305ae11dea6738987a40e3

packages/terra-show-hide/src/ShowHide.jsx
bjankord: I don't, might test removing it and see if things still work as intended.

packages/terra-show-hide/src/_ShowHideButton.jsx
```diff
@@ -0,0 +1,192 @@
+/* eslint-disable react/button-has-type */
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import 'terra-base/lib/baseStyles';
+import styles from './_ShowHideButton.module.scss';
+
+const cx = classNames.bind(styles);
+
+const KEYCODES = {
```
ryanthemanuel: Can't remember if we've talked about standardizing these codes somewhere common across our components, or using a 3rd party library, if it's worth it, etc..

bjankord: I'm fine leaving as is for now. At some point, we'll want to package this up into a utility library.

packages/terra-show-hide/src/ShowHide.jsx
gabeparra01: Alright I tested it and it can be removed

packages/terra-show-hide/src/ShowHide.jsx
```diff
@@ -0,0 +1,113 @@
...
+    const showHideClassName = cx([
+      'button',
+      { 'is-open': this.props.toggle },
+      customProps.className,
+    ]);
+
+    return (
+      <div {...customProps} className={showHideClassName}>
+        {!this.props.toggle && preview}
+        <Toggle isOpen={this.props.toggle} isAnimated={this.props.isAnimated} className={cx([{ animated: this.props.isAnimated }])}>
```
tbiethman: Should we change the `toggle` prop to `isOpen`? I think `toggle` feels kind of weird as a boolean prop.

Also, since you've pulled the prop values out above, you can just use those variables directly instead of going through this.props again.

gabeparra01:
I am about to start my second redesign on this component since opening this pull request. The first design was based loosely off of terra-toggle-button under the guidance of the Terra team. In order to use that design and still have Jest tests, I had to research the best way to use react-intl with Jest tests. The Terra team approved that new test design in this pull request. The Terra team later decided they needed a controlled component without internal state- which is not the same design that terra-toggle-button uses. That first redesign required me to change 28 files.

After making these changes, the Terra team decided that I no longer need the last remaining instance method of the ShowHide.jsx class- handleOnClick(). Which means that component could now be written as a pure function instead of a class. This change will require a second redesign from me and I will need to rewrite the tests and examples. This will also include me reverting the new Jest test design back to the previous one. In an effort to prevent additional redesigns and work, I wanted to make sure everyone is agreed to this latest redesign before I move forward. Please confirm: @bjankord @JakeLaCombe @ryanthemanuel @tbiethman @StephenEsser @emilyrohrbough

JakeLaCombe:
I'm ok with it. +1

tbiethman:
+1, though I'm not sure why examples or tests would need to be updated solely based on the transition to a functional component

gabeparra01:
> +1, though I'm not sure why examples or tests would need to be updated solely based on the transition to a functional component

@tbiethman I will be able to use intl through context this time now that it is a purely functional component. From what I see in terra-form-field, it looks like I will be setting up the tests differently: 
https://github.com/cerner/terra-core/blob/master/packages/terra-form-field/tests/jest/Field.test.jsx#L4
https://github.com/cerner/terra-core/blob/master/packages/terra-form-field/tests/jest/intl-context-setup.js

bjankord:
Talked to @gabeparra01 about this. Plan is to post a new tech design on #1563 
Before we move that new design into code review, I'd like the reviewers on this PR to look at the tech design and get that landed so @gabeparra01 can work on the code without a lot of churn happening in the next PR.

tbiethman:
@gabeparra01 I'd still recommend using injectIntl and getting intl as a prop vs. reading from the context parameter on functional components. Based on the new context API and from what I see in the documentation, I'm not sure that parameter will continue to exist after the legacy API documentation.

PR #1922, opened 2018-09-24
Removed DefaultFieldset.test.jsx

### LINKED ISSUE #1843: [terra-form-fieldset] Remove default test example
# Bug Report
There is a [form-fieldset test example](https://engineering.cerner.com/terra-core/#/tests/terra-form-fieldset/form-fieldset/default-fieldset) that isn't tested in the [spec](https://github.com/cerner/terra-core/blob/master/packages/terra-form-fieldset/tests/wdio/fieldset-spec.js#L6). We can remove this.

AlexBrizi:
### Summary
Removed the DefaultFieldset.test.jsx file from the terra-form-fieldset package.

Resolves #1843


packages/terra-form-fieldset/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ ChangeLog
 Unreleased
 ----------
 
+### Changed
+* Removed DefaultFieldset.test.jsx
```
emilyrohrbough: could we add a reason :) it seems weird to only we removed a test

AlexBrizi: I wasn't given a reason :(

emilyrohrbough: @AlexBrizi I think its because we aren't using this test examples in our tests so its an extra, unused example. You will need to check the test urls we use in the test spec for wdio.

bjankord: On the [issue](https://github.com/cerner/terra-core/issues/1843), @yuderekyu mentioned, "There is a form-fieldset test example that isn't tested in the spec. We can remove this."

bjankord: Updated changelog comment noting the file was unnecessary: https://github.com/cerner/terra-core/pull/1922/commits/edae2b9f027b540d50b0dfe10dadc468f243c529

PR #1923, opened 2018-09-24
Exports for oneOf proptypes

amichaelparker:
### Summary
Added exports on several components for proptypes using oneOf enumeration.

See issue Closes https://github.com/cerner/terra-core/issues/1044


packages/terra-badge/src/Badge.jsx
```diff
@@ -6,6 +6,10 @@ import styles from './Badge.module.scss';
 
 const cx = classNames.bind(styles);
 
+const BadgeScheme = ['default', 'primary', 'secondary', 'info', 'warning', 'positive', 'negative'];
```
bjankord: Let's change the name on this const to `BadgeIntent`

packages/terra-responsive-element/src/ResponsiveElement.jsx
```diff
@@ -4,6 +4,7 @@ import ResizeObserver from 'resize-observer-polyfill';
 import 'terra-base/lib/baseStyles';
 import breakpoints from './breakpoints.module.scss';
 
+const ResponsiveElementResponsiveTo = ['window', 'parent'];
```
yuderekyu: hmmmm, want to rename this? maybe `dependentViewport`

amichaelparker: name changed!

bjankord:
We'll want to keep using the original values for `PropTypes.oneOf`.

Currently, in the PR, the following type of change was made.
```diff
- intent: PropTypes.oneOf(['default', 'primary', 'secondary', 'info', 'warning', 'positive', 'negative']),
+ intent: PropTypes.oneOf([BadgeIntent.DEFAULT, BadgeIntent.PRIMARY, BadgeIntent.SECONDARY, BadgeIntent.INFO, BadgeIntent.WARNING, BadgeIntent.POSITIVE, BadgeIntent.NEGATIVE]),
```

We'll want to go with `PropTypes.oneOf` where we keep the string values or numeric values they were prior to the PR like so.
```jsx
intent: PropTypes.oneOf(['default', 'primary', 'secondary', 'info', 'warning', 'positive', 'negative']),
```

This allows people to keep using strings as values if the want.
e.g. 
```jsx
<Badge intent="primary" text="Badge text" /> or <Badge intent={BadgeIntent.PRIMARY} text="Badge text" />
```

PR #1925, opened 2018-09-24
Improve the accessibility of the Alert component

### LINKED ISSUE #1822: Make Alert message an entire paragraph
# Feature Request
Terra Alert is constructed from two generic divs: one for the title and the other for the message. This is hard on screen readers, as it requires one key press to read the title, and another to read the error content. To make this more accessible, we should use a paragraph to encapsulate both the title and the message.

JakeLaCombe:
### Summary
Resolves #1822

This makes the alert message more readable on a screen reader. No longer do you need to use multiple down arrows to hear an entire alert message. Just press down once, and the whole thing is read!

Some notes: VoiceOver on mac does not like floated styles that well, as it takes elements out of the flow of your current content. As a result, the screen reader would break when it was on the title message. Thus, I made it inline instead of floated. It looks like this defect was carried in the wdio screenshots as well, as some of the existing ones were defective. Thus, I had to update those, even though stylistically they look the same before and after the changes in a browser.


packages/terra-alert/src/Alert.jsx
```diff
@@ -164,12 +164,10 @@ const Alert = ({
   }
 
   const alertMessageContent = (
-    <div className={alertSectionClassName}>
-      <strong className={cx('title')}>{title || defaultTitle}</strong>
-      <div>
-        {children}
-      </div>
-    </div>
+    <p className={alertSectionClassName}>
```
yuderekyu: just curious, what's the difference between a div and paragraph for screen readers?

JakeLaCombe: It's more semantic. It reads it as text as opposed to a generic container, and I'm finding that it reads better when you have a variety of spans and other nodes for text decorating.

mmalaker:
Functional verification completed. The alert examples now function correctly with screen readers.

PR #1928, opened 2018-09-25
Add Visually Hidden Text to Badge Element

### LINKED ISSUE #1823: Terra Badge should have an aria-label attribute prop
# Feature Request

Terra Badge has a visual indicator noting it's importance. However, screen readers will not know what to do with this element, and do nothing else besides read it as generic text. To improve screen reader support, an aria label prop should be added to the badge that identifies it's purpose.

JakeLaCombe:
### Summary
Fixes #1823 

When terra badge is used with a screen reader, there isn't any clear indication of what the element is. It's based on a visual display, which gives no purpose to screen readers. This pull request adds visually hidden text so that context can be provided to a badge when it is needed.


packages/terra-badge/src/terra-dev-site/doc/example/BadgeVisuallyHiddenText.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import Badge from 'terra-badge/lib/Badge';
+
+const BadgeVisuallyHiddenText = () => (
+  <div>
+    <Badge text="Low" visuallyHiddenText="Rick Factor" />
```
bjankord: Nit: `visuallyHiddenText` should probably be `Risk Factor`

JakeLaCombe: We could just call this a Rick & Morty badge :)

packages/terra-badge/src/Badge.jsx
```diff
@@ -57,8 +65,9 @@ const Badge = ({
     customProps.className,
   );
 
-  const textContent = text ? <span className={styles.text}>{text}</span> : null;
-  const content = isReversed ? [textContent, icon, customProps.children] : [icon, textContent, customProps.children];
+  const textContent = text ? <span className={styles.text} aria-label={visuallyHiddenText}>{text}</span> : null;
```
bjankord: Let's remove the aria-label. This results in the visually hidden text being read twice on VoiceOver on desktop, it also prevents VoiceOver on desktop from reading the `{text}` content as aria-label is read instead.

VoiceOver on iOS does not read the aria-label so the use of `<VisuallyHiddenText />` component is needed.

mmalaker:
Functional verification completed. The badge examples function correctly with screen readers.

PR #1929, opened 2018-09-25
Make cards more accessible

### LINKED ISSUE #1826: Make Terra-Card either a section or an article
# Feature Request
Terra Cards have visual indicators to not their presence. However, screen readers are not capable of distinguishing those indicators. We need to make the card either a section or an article so that screen readers better understand the intent of a card.

JakeLaCombe:
### Summary
Resolves #1826 

Out of the box, it's hard to determine the purpose of a card for screen readers. The only indication of a card is a visually one, making it hard to determine the importance of a card vs the content around the card. This pull request adds the article card to make it easier to read.


packages/terra-card/src/Card.jsx
```diff
@@ -22,6 +23,11 @@ const propTypes = {
    * Sets the card variant to change the style for different use cases. One of `default`,  `raised`.
    */
   variant: PropTypes.oneOf([...Object.values(CardVariants)]),
+  /**
+   * Text that describes the badge to a screen reader. Use this
+   * for creating an accessible badge.
+   */
+  visuallyHiddenText: PropTypes.string,
```
StephenEsser: Some copy pasta left over

PR #1937, opened 2018-09-28
Issue 1935: Use absolute positioning in modals.

### LINKED ISSUE #1935: iOS Fixed Position Modal Bug
# Bug Report

## Description
Editable fields in fixed content (ie Modals) do not behave as expected.  iOS accounts for the keyboard by natively shifting up the view.  When quickly switching between editable fields, it's possible this shifting can get off.  See the attached image below.

## Steps to Reproduce
1.  Have an editable field on the page, in a modal
2. Select the editable field
3. Quickly deselect, then reselect the field

Note: this doesn't happen every time, and may be more prone to UIWebview and not Safari.  Programmatically blurring and refocusing the element could help reproduce the issue as well.

image

## Expected Behavior
The modal and modal overlay do not shift up when changing focus between editable fields.

## Possible Solution
Switching to absolute positioning for the modal and modal overlay resolves the problem.  There are many documented bugs around fixed positioning and how it responds to the keyboard on iOS.  

## Environment
* Component Name and Version: N/A
* Browser Name and Version: UIWebview, iOS 11.2, iOS 11.4
* Node/npm Version: 9.2.0, 5.5.1
* Webpack Version: 4.20.0
* Operating System and version (desktop or mobile): Mobile iOS

## @ Mentions
@dkasper-was-taken, @tbiethman

dkschoonover:
### Summary
Use absolute positioning instead of fixed for modals.
Fixes #1935 

### Additional Details
There are many open bugs for iOS detailing how editable content inside of fixed elements does not behave as expected when the iOS keyboard is open (see the screenshot on the linked issue for a particularly annoying example). Using absolute positioning for modals resolves some of these issues and will behave the same since the modals are being shown with react-portal and being attached to the document body.  

This would be non-passive for anybody that is explicitly setting the body's position to relative and the body is scrollable.

### Deployment URL
https://terra-core-deployed-pr-1937.herokuapp.com/#/components/terra-abstract-modal/abstract-modal/abstract-modal

@cerner/terra

### DIFF
--- a/packages/terra-abstract-modal/CHANGELOG.md
+++ b/packages/terra-abstract-modal/CHANGELOG.md
@@ -8,6 +8,7 @@ Unreleased
 ------------------
 ### Changed
 * Minor dependency version bump
+* Use absolute position for modal and modal overlay for iOS stability
 
 1.17.1 - (September 6, 2018)
 ------------------
--- a/packages/terra-abstract-modal/src/AbstractModal.module.scss
+++ b/packages/terra-abstract-modal/src/AbstractModal.module.scss
@@ -6,7 +6,7 @@
     max-width: calc(100% - 2 * var(--terra-abstract-modal-horizontal-inset, 10px));
     outline: none;
     overflow: hidden;
-    position: fixed;
+    position: absolute;
     top: 50%;
     transform: translate(-50%, -50%);
 
--- a/packages/terra-abstract-modal/src/ModalOverlay.module.scss
+++ b/packages/terra-abstract-modal/src/ModalOverlay.module.scss
@@ -6,7 +6,7 @@
     color: var(--terra-abstract-modal-overlay-foreground-color);
     left: 0;
     opacity: var(--terra-abstract-modal-overlay-opacity);
-    position: fixed;
+    position: absolute;
     right: 0;
     top: 0;
 


dkschoonover:
We're targeting to code cut this week, it'd be awesome if this change could make it in to Terra's next release.

bjankord:
Tested this across our supported desktop browsers and didn't see any issues. Testing this on our Android and iOS devices now.

bjankord:
Tested on our Android and iOS devices and didn't see any issues. I think this is good to go.

dkasper-was-taken:
I'm not necessarily against the change, I just want to make sure we are 100% certain as a platform that we don't want to support body scrolling. I had similar issues with fixed vs absolute with popup.

dkasper-was-taken:
Also if this is non-passive we should definitely make it a major version bump.

dkschoonover:
When I was testing this, modals displayed correctly when the body was scrollable as long as the body isn't set to "position: relative."  Still think that warrants a major version bump?

dkasper-was-taken:
Sorry for the late edit, it's probably fine to not major version bump in that case.

bjankord:
> Also if this is non-passive we should definitely make it a major version bump.

It's non-passive if you've modified code outside of terra-ui, specifically if you've set set the body's position to relative and the body is scrollable. It's passive with our styles we set on `body` element.

bjankord:
If we get feedback from users that this breaks things with styles they've added on top of terra-ui styles, we can revisit these styles.

PR #982, opened 2017-10-31
Implement Textarea for Theming

JakeLaCombe:
### Summary
This updates the Textarea to be in it's own package, and also to be ready for OCS.

### Additional Details
New additions to textarea
- Added prop of isInvalid to indicate textarea is invalid.
- Added ability to auto resize textarea based on input.
- Added set sizes that adjusts the rows attribute of the textarea.
- Added the ability to create a textarea that is full width and height of container (Ex: Full Page Size)

Deployment link: 
https://terra-core-deployed-pr-982.herokuapp.com/static/#/site/form-textarea

Thanks for contributing to Terra. 
@cerner/terra

[CONTRIBUTORS.md]: ../blob/master/CONTRIBUTORS.md
[License]: ../blob/master/LICENSE

PR #1938, opened 2018-09-30
Textarea docs

### LINKED ISSUE #1019: Textarea - adjust documentation examples
# Issue Description
- add headers above each example 
- add/reorder examples to show 1.resizable at small size (two rows), 2.fixed-small size, 3.fixed-medium size, 4.fixed-large size, 5.fill-containter (same as current), 6.textarea-small size that becomes invalid/error when empty, 7.disabled (same as current), 8.disabled with error (same as current)

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Introduced by 
PR #982 
Merge Comments: https://github.com/cerner/terra-core/pull/982#issuecomment-344742186

sharynneazhar:
### Summary
Updated Textarea documentation with the following examples:
1. resizable at small size (two rows), 
2. fixed-small size, 
3. fixed-medium size, 
4. fixed-large size, 
5. fill-containter (same as current), 
6. textarea-small size that becomes invalid/error when empty, 
7. disabled (same as current), 
8. disabled with error (same as current)

Closes #1019 

@cerner/terra


packages/terra-form-textarea/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ ChangeLog
 Unreleased
 ----------
 
+2.27.0 - (September 30, 2018)
+------------------
+### Changed
```
emilyrohrbough: This should be moved under the `Unreleased` header. Once this is merged to master and the next release happened, this will be updated with the release version/date.

packages/terra-form-textarea/src/terra-dev-site/doc/example/FixedLargeSize.jsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import Textarea from 'terra-form-textarea/lib/Textarea';
+
+const FixedLargeSizeExample = () => (
+  <Textarea
+    size="large"
+    style={{ resize: 'none' }}
```
emilyrohrbough: Why do we need the custom style?

packages/terra-form-textarea/src/terra-dev-site/doc/example/DisabledWithError.jsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import Textarea from 'terra-form-textarea/lib/Textarea';
+
+const DisabledWithErrorExample = () => (
```
emilyrohrbough: Would this be a valid example we would want to show? I would think a disabled textarea shouldn't technically be able to reach the invalid state

packages/terra-form-textarea/src/terra-dev-site/doc/example/FixedLargeSize.jsx
sharynneazhar: I understood "fixed" as non-resizable and the Textarea component did not have a prop for that. May be an misunderstanding. Should I go ahead and remove this?

packages/terra-form-textarea/src/terra-dev-site/doc/example/DisabledWithError.jsx
sharynneazhar: Thoughts @neilpfeiffer?

packages/terra-form-textarea/CHANGELOG.md
sharynneazhar: 508d10c

packages/terra-form-textarea/src/terra-dev-site/doc/example/FixedLargeSize.jsx
bjankord: cc/ @neilpfeiffer

packages/terra-form-textarea/src/terra-dev-site/doc/example/DisabledWithError.jsx
neilpfeiffer: From a usability perspective, no, an input or text area "should" never be disabled and invalid since you would never be able to resolve unless it was only temporary while focus was on another item/field/input then returned to active once the other item was resolved. 

But from a basic UI perspective, since it is something that could exist in technical implementation by combining props/styles, we should both be sure to handle it, and we should have a test and screenshot for that case, but I don't think we need it as a main example on the docsite page.

sharynneazhar: @emilyrohrbough  Do you want me to go ahead and remove this example from the site or leave it? Let me know what other changes you'd like here.

emilyrohrbough: @sharynneazhar Lets remove this from the documentation pages. Since this isn't a desirable use-case we don't want to confuse consumers by providing an example we don't want people to implement.

packages/terra-form-textarea/src/terra-dev-site/doc/example/Resizable.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import Textarea from 'terra-form-textarea/lib/Textarea';
+
+const ResizableExample = () => (
+  <Textarea
+    isAutoResizable
+    size="small"
+    value="I automatically resize as you type in more information (except in mobile)."
```
emilyrohrbough: I am unable to type more text into this example on the deployed site

packages/terra-form-textarea/src/terra-dev-site/doc/example/Invalid.jsx
```diff
@@ -0,0 +1,31 @@
...
+    this.setState({ area: event.target.value });
+  }
+
+  render() {
+    return (
+      <Textarea
+        value={this.state.area}
+        isInvalid={this.state.area === ''}
+        onChange={this.handleAreaChange}
+        size="small"
```
emilyrohrbough: Can we start this example as invalid? We can add `Leaving this as blank makes this input invalid.` to the example description.

packages/terra-form-textarea/src/terra-dev-site/doc/example/FixedLargeSize.jsx
emilyrohrbough: I think @neilpfeiffer meant fixed as in `isAutoResizable = false`

packages/terra-form-textarea/src/terra-dev-site/doc/example/Resizable.jsx
JakeLaCombe: yeah. Use that here instead.

packages/terra-form-textarea/src/terra-dev-site/doc/example/FixedLargeSize.jsx
JakeLaCombe: Size full has what you are looking for here.  All sizes except for that one should be resizable.

As for that prop, that dictates if the textarea should be resized automatically as you type in more lines of text instead of hiding the gripper. Should we add another prop that dictates hide gripper? (Note: resize none doesn't work in Internet Explorer)

packages/terra-form-textarea/src/terra-dev-site/doc/example/DisabledWithError.jsx
sharynneazhar: 729b016

packages/terra-form-textarea/src/terra-dev-site/doc/example/Resizable.jsx
sharynneazhar: 303ac6e

packages/terra-form-textarea/src/terra-dev-site/doc/example/Invalid.jsx
sharynneazhar: 8502c90

packages/terra-form-textarea/src/terra-dev-site/doc/example/Invalid.jsx
```diff
@@ -0,0 +1,31 @@
...
+
+  handleAreaChange(event) {
+    this.setState({ area: event.target.value });
+  }
+
+  render() {
+    return (
+      <Textarea
+        value={this.state.area}
+        isInvalid={this.state.area === ''}
```
emilyrohrbough: This isn't displaying the invalid state still. lets try

```
constructor(props) {
    ....
    this.state = {
          isInvalid: true,
    }
}
...
handleAreaChange(event) {
    this.setState({ isInvalid: event.target.value.length < 3 });
}
...
<Textara
    ....
    isInvalid={this.state.isInvalid}
/>
```

emilyrohrbough: @JakeLaCombe actually, looking at the text example for invalid state, is there any difference in the view when invalide? https://terra-core-deployed-pr-1938.herokuapp.com/#/tests/terra-form-textarea/form-textarea/invalid-textarea

If not, should we remove this example?

packages/terra-form-textarea/src/terra-dev-site/doc/example/FixedLargeSize.jsx
neilpfeiffer: Catching up and refreshing myself with specification details posted https://zpl.io/V09j49E & @emilyrohrbough was correct in my original (poorly worded) intent on the issue:
> I think @neilpfeiffer meant fixed as in `isAutoResizable = false`


\
\
To recap desired state of component functionality:
- Textarea maintains current size prop which is enum one of `small`, `medium`, `large`, `full`
- If size is set to `small`, `medium`, or `large`  -- the gripper always visible, if `isAutoResizable=true` then will grow in height unless user resizes (then stops auto-resize and keeps the user-moved-to size, same as current), and all three sizes are always user-resizable.
- If size is set to `full` -- the gripper is hidden, always grows to width+height of it's container, a user *cannot* resize and *ignores* if `isAutoResizable=true` (since can never grow bigger than it's container).

\
\
To recap & update the desired state for docsite examples:
1. `size="small'` (should be default of `isAutoResizable=false`), 
2. `size="medium"` (should be default of `isAutoResizable=false`), 
3. `size="large"` (should be default of `isAutoResizable=false`), 
4. `isAutoResizable=true` and `size="small"`, 
5.  `size="full"`, 
6. `size="small"` (default `isAutoResizable=false`) that becomes invalid/error when empty, 
7. `size="small"` and `disabled=true`

\
And not captured on the original issue, but I think will help in user understanding of these details, we should update the prop descriptions for isAutoResizable and size: 
```
/**
   * Whether the textarea can be auto-resized vertically. _(Will be ignored if size attribute is set to "full".)_
   */
isAutoResizable: PropTypes.bool,
```

```
/**
   * Standard sizes for a textarea. One of `small`, `medium`, `large`, or `full`.
   * _(Sizes the textarea by setting the rows attribute a corresponding preset value.)_ 
   */
size: PropTypes.oneOf(['small', 'medium', 'large', 'full']),
```

packages/terra-form-textarea/src/terra-dev-site/doc/example/Invalid.jsx
sharynneazhar: I was wondering that too but saw this comment on an earlier PR: https://github.com/cerner/terra-core/pull/982#pullrequestreview-73310815

emilyrohrbough: 😢 lol I had the same question then as I do now. I don't like it because it feels broken but I guess it's valid!

packages/terra-form-textarea/src/terra-dev-site/doc/example/FixedLargeSize.jsx
bjankord: Will implement these updates in a separate PR.

PR #1939, opened 2018-10-01
[terra-form-select] - Fixed duplicate theme variable

### LINKED ISSUE #1727: Form-Select: additional themeable variable

StephenEsser:
### Summary
- Removed duplicate theme value
- Added theme variable for the input height

The Select had a theme variable defined twice with different values making theming difficult. 

This change enables the combobox, multiple, and tag variants to have a common theme-able wrapper and removes the duplicate variable.

### Additional Details
Adds theme-able padding variables to the red outline below.
image

Resolves #1727 

@cerner/terra


packages/terra-form-select/src/_Frame.module.scss
```diff
@@ -120,14 +120,17 @@
     padding: var(--terra-select-toggle-padding, 0 0.42857rem);
   }
 
-  .tags {
+  .content {
```
StephenEsser: I'm struggling to find a good name for this content region, if anyone has any ideas let me know. This content region is the wrapper around the input and selected values for tag, multiple, and combobox variants. The default variant does not have this wrapper.

bjankord: `.input-container` is what comes to mind. I'm cool with `.content`.

yuderekyu: if you want to be more specific, maybe `non-default-content`?

neilpfeiffer:
Tested: PR #1939
Resolves: Issue #1727

**Reviewed:**
- Looks good, only alternate naming suggestion (could consider for the future) for the Frame-module might be something like:  `.display  -->  . outer` and `.content  -->  .inner`

\
\

**Pass** |
-----------|
UX Review|

PR #1942, opened 2018-10-01
Pass down custom props for ResponsiveElement

### LINKED ISSUE #1930: ResponsiveElement doesn't pass down custom props
# Bug Report
ResponsiveElement doesn't pass down custom props

## Description
I have Aggregator wrapped around ResponsiveElement. Aggregator sends down aggegatorDelegate prop and that is not being passed to the components in the responsiveElement since the responsive element doesn't send down custom props

## Steps to Reproduce
```
<Aggregator> #sends down aggregatorDelegate prop
   <ResponsiveElement>
       tiny={<someComponent>//this component doesn't get aggregatorDelegate prop }
   </ResponsiveElement>
</Aggregator>
```

## Expected Behavior
Responsive Element should send down customProps. So that the user doesn't have to use wrapper around responsiveElement to get the props and again pass the custom props down to responsive element.

## Possible Solution
Modify ResponsiveElement to send down customProps

corymcdonald:
### Summary
Resolved #1930
 
### Additional Details
I think that @tbiethman had reservations about this but I really want a hacktober shirt.

@cerner/terra

### DIFF
--- a/packages/terra-responsive-element/src/ResponsiveElement.jsx
+++ b/packages/terra-responsive-element/src/ResponsiveElement.jsx
@@ -111,7 +111,12 @@ class ResponsiveElement extends React.Component {
 
   render() {
     if (this.state.element) {
-      return this.props[this.state.element] || null;
+      const Tag = this.props[this.state.element] || null;
+      if (Tag) {
+        return <Tag {...this.props} />;
+      } else {
+        return null;
+      }
     }
 
     return <div ref={this.setContainer} />;


packages/terra-responsive-element/src/ResponsiveElement.jsx
```diff
@@ -111,7 +111,12 @@ class ResponsiveElement extends React.Component {
 
   render() {
     if (this.state.element) {
-      return this.props[this.state.element] || null;
+      const Tag = this.props[this.state.element] || null;
+      if (Tag) {
+        return <Tag {...this.props} />;
```
StephenEsser: Can you write a few tests that add custom properties into the component and verify they are present upon render?

corymcdonald: Oh duh, of course. That makes a lot more sense than what i'm doing

StephenEsser:
I also have a few reservations with passing down custom properties onto each element in the Responsive Element. 

There's a little bit of a risk with the custom properties during cloning. Because the responsive element itself has an API there will be a blacklist of properties the children can never receive if supplied in a flat API.

IE: The children can never receive props already defined for the ResponsiveElement. `tiny`, `small`, `responsiveTo`, etc... Unless they are passed down in a separate prop deliberately to be spread onto each element. There's also a few concerns about cloning the elements every time they render.

packages/terra-responsive-element/src/ResponsiveElement.jsx
StephenEsser: This would also need to merge existing classNames with custom classNames similar to how many of our components are currently doing this:

https://github.com/cerner/terra-core/blob/master/packages/terra-card/src/Card.jsx#L44

StephenEsser:
After discussions with the Terra team we are closing this issue.

There are concerns around cloning and overriding props on each element defined within the responsive element. The custom property pattern is intended to be applied only to a top-level component wrapper that has a defined set of PropTypes. Spreading custom properties onto unknown elements poses a risk of introducing unsupported properties and overriding existing properties.

PR #1945, opened 2018-10-02
Export intlProvider from terra-base

### LINKED ISSUE #1944: Export intlProvider from terra-base
# Feature Request

## Description
Currently we export injectIntl, and a lot of the other things from react-intl. In order to completely removed the dependency on `react-intl` for consumers it would be nice to be able to `export intlProvider`. 

Introduced by: https://github.com/cerner/terra-framework/pull/293

corymcdonald:
### Summary
Resolves #1944 🎉 

@cerner/terra

### DIFF
--- a/packages/terra-i18n/src/I18n.js
+++ b/packages/terra-i18n/src/I18n.js
@@ -1,4 +1,6 @@
-import { injectIntl, intlShape, FormattedMessage } from 'react-intl';
+import {
+  injectIntl, intlShape, FormattedMessage, intlProvider,
+} from 'react-intl';
 import I18nProvider from './I18nProvider';
 import i18nLoader from './i18nLoader';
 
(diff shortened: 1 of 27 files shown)


packages/terra-i18n/src/I18n.js
```diff
@@ -1,4 +1,6 @@
-import { injectIntl, intlShape, FormattedMessage } from 'react-intl';
+import {
+  injectIntl, intlShape, FormattedMessage, intlProvider,
```
tbiethman: Does this work? The IntlProvider is a component and starts with a capital I.

corymcdonald: Well this is awkward, I didn't actually test because I couldn't get `npm link` to work. But you are correct, this is incorrect https://github.com/cerner/terra-core/pull/1945/commits/7e190b372fc66e5bd6f646d189a2e481908bcdee

PR #1955, opened 2018-10-04
[terra-overlay] Various bug fixes

### LINKED ISSUE #963: Allow Overlay above SlidePanel
# Issue Description

CareAware Connect would like to use LoadingOverlay above all parts of SlidePanel, including the `panel` part. This is currently not possible without manually overriding the z-index of the overlay.

This is because Overlay has a [z-index of 100](https://github.com/cerner/terra-core/blob/999788849093ec06856881af96d3442014537a4e/packages/terra-overlay/src/Overlay.scss#L15) and SlidePanel's `panel` has a [z-index of 500](https://github.com/cerner/terra-core/blob/999788849093ec06856881af96d3442014537a4e/packages/terra-slide-panel/src/SlidePanel.scss#L30).

I'd like to propose introducing a prop to dictate whether Overlay has a larger z-index. Considerations will need to be made for whether Overlay covers modals in this scenario.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior

Overlay covers both single and multi-panel SlidePanel mode

## Current Behavior

Overlay is hidden in single slide panel mode

## Steps to Reproduce

N/A

## Environment

* Component Version:
* Browser Name and Version: Chrome 61.0.3163.100 (Official Build) (64-bit)
* Operating System and version (desktop or mobile): macOS 10.12.6 (16G29)

### LINKED ISSUE #1340: [terra-overlay] React Portal version of Loading Overlay
# Issue Description
The terra-core LoadingOverlay (even general Overlay) needs to support a 'React Portal' variant. This is helpful in several scenarios where we need to have a common place to show a full screen overlay with and without the loading indicator. The ideal solution that will help us manage the overlays will be an 'OverlayManager'. Any efforts to bring that will be appreciated. But, at least until that's planned, a 'React Portal' version of LoadingOverlay will be useful.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
It should show up over the page full even with transformed elements like SlidePanels.

### LINKED ISSUE #1450: Overlay does not restore document overflow if isOpen not first set to false
# Issue Description
The Overlay component will capture the [document's overflow style when it is first mounted](https://github.com/cerner/terra-core/blame/master/packages/terra-overlay/src/Overlay.jsx#L78) due to setting `isOpen=true`. However, if you do not first set the Overlay's `isOpen=false` because you choose not to render the Overlay, then the document's overflow style is not restored. 

This is because `resetBeforeOverlay` is called only in the [render method](https://github.com/cerner/terra-core/blame/master/packages/terra-overlay/src/Overlay.jsx#L140). Disregarding that it is really an anti-pattern to have side-effects in render methods, the Overlay should really be cleaning up after itself before the entire component is unmounted. The solution should be to call `resetBeforeOverlay` during the `componentWillUnmount` event so that it ensures that everything is the way it was before it was mounted. 

You could argue that we should just set `isOpen=false` instead of not rendering the Overlay but my thinking is that there really isn't a need for `isOpen` at all and if you really didn't want an Overlay to be displayed you should not render it. It feels weird to always render an Overlay and thus add it to the React component tree if it is not supposed to be doing anything. 

I plan on submitting a PR to address this issue soon.

## Issue Type
- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
The expected behavior is for the Overlay to restore the document's overflow styling whenever it is unmounted regardless of whether the `isOpen` property is set to false first. 

## Current Behavior
The current behavior is the document's overflow styling is not restored to previous styling prior to the Overlay being displayed if the Overlay does not have it's `isOpen` set to false.

## Steps to Reproduce
1. Note document's overflow style on initial render of component where Overlay is not rendered. 
2. In component, conditionally render Overlay with `isOpen=true` based on some internal component state
3. Component state changes so the Overlay should not be rendered
4. Instead of setting `isOpen=false` on the Overlay, component chooses to not render the Overlay.
5. Note the document's overflow style is now still set to `hidden`.

## Environment
* Component Version: 2.7.0
* Browser Name and Version: All
* Operating System and version (desktop or mobile): All

### LINKED ISSUE #1880: Overlay fullscreen scrolling IE11 and Edge
# Bug Report

## Description
Fullscreen overlay allows for background document to scroll in IE11/Edge which is unintended.

## Steps to Reproduce

1. Using IE11 or Edge, go to [Terra-ui overlay example page](https://engineering.cerner.com/terra-ui/#/components/terra-overlay/overlay/overlay)
2. Click "Trigger Fullscreen Overlay" button.
3. Attempt to scroll the page up and down. 
4. If the pages moves - it is broken.

## Expected Behavior
Base page doesn't scroll when overlay is fullscreen.

## Solution
1. Need to switch the fullscreen overlay to be a [portaled](https://reactjs.org/docs/portals.html) version using `ReactDOM.createPortal(...)`, similar to request #1340, will also resolve #963
2. Need to remove the isScrollable prop on the children contents of the overlay (shouldn't be an issue since no reason to add scrollable content with overlay, abstract modal is better use if needing long/custom scrolling content).
3. Need to capture scroll handling similar to Terra-Popup, then update Terra-Popup to import and use Terra-overlay first class, may require z-index to be prop.
4. Should update example page to use a relative positioned parent for the container example, but not for fullscreen example.
5. Also should include resolution for #1450 and #1804

## Environment

* Component Name and Version: terra-overlay 2.26.1 (current)
* Browser Name and Version: IE11 & Edge
* Operating System and version (desktop or mobile): Windows 10+

jmsv6d:
### Summary
* Changed overlay to be portalled
* Added zIndex prop
* Added hidden div to catch scrolling behind overlay
* Fixed issue with scrolling not being restored on unmount

Fixes #1880 , fixes #1340, fixes #963, fixes #1450 

### Deployment Link
https://terra-core-deployed-pr-1955.herokuapp.com/#/components/terra-overlay/overlay/overlay

@cerner/terra


packages/terra-overlay/tests/jest/intl-context-setup.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable react/jsx-boolean-value, import/no-extraneous-dependencies */
+import { IntlProvider, intlShape } from 'react-intl';
```
bjankord: You can import these from terra-base now which exports them from react-intl. This helps us better match which version of react-intl is used.

packages/terra-overlay/src/Overlay.module.scss
```diff
@@ -11,7 +11,39 @@
     justify-content: center;
     outline: none;
     overflow: hidden;
-    z-index: 100;
+
+    .inner {
+      height: 200%;
```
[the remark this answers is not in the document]
dkasper-was-taken: I think if this 200% is going to work, the parent needs to use the overflow: scroll;
The larger content is the means to prevent scrolling behind overflow that occurs on mobile devices by having it's own scrolling, just with hidden scrollbars. So if we want this fix we'll need updates.

mmalaker:
I've found a few accessibility issues with the overlay and our examples:
- On the Components page, screen readers are not detecting that there is any text being displayed within the overlay once it is triggered. The user needs to understand that an overlay has appeared within their workflow and the displayed text should be read aloud.
- On the Tests pages, keyboard navigation is blocked by several of the overlays. It is not an issue with the overlay itself, but how we arrange our examples. We should always enable keyboard navigation of our site. I'd recommend providing only one fullscreen example and putting it as the last item in the navigation. You could also have that as a loading overlay so that it is automatically dismissed.

packages/terra-overlay/src/LoadingOverlay.jsx
```diff
@@ -32,6 +32,10 @@ const propTypes = {
   * Indicates if the overlay is relative to the triggering container.
   */
   isRelativeToContainer: PropTypes.bool,
+  /**
+  * Z-Index layer to apply to the ModalContent and ModalOverlay. Valid values are the standard modal layer: '6000', and the max layer: '8000'.
```
tbiethman: This comment is missing descriptions of the other supported values.

packages/terra-overlay/src/Overlay.module.scss
```diff
@@ -11,7 +11,31 @@
     justify-content: center;
     outline: none;
     overflow: hidden;
-    z-index: 100;
```
bjankord: Should we keep this `z-index: 100` as a default? It seems this may be a breaking change if we remove the default z-index.

jmsv6d: I still have it as the default: https://github.com/cerner/terra-core/pull/1955/files#diff-5c1dfe9c477f29f0ae7f7738c5cb13e9R62

bjankord: ah ok, perfect!

mmalaker:
Functional verification completed. The overlay examples and tests display and function correctly. Overlay screen reader focus issue to be addressed in #1983

PR #1959, opened 2018-10-05
Fix alert padding

### LINKED ISSUE #1859: [terra-alert] The text needs padding on the right
# Bug Report

## Description

If the text entered in the alert is too long it gets too close to the border. It needs padding on the right.

screen shot 2018-09-12 at 1 01 00 pm

## Steps to Reproduce

1. Add lengthy text in the alert
2. Reduce the size of the screen

screen shot 2018-09-12 at 1 03 29 pm

## Expected Behavior

There should be space between the text and the alert banner border.

## Possible Solution

Add padding-right css.
## Environment

* Component Name and Version: Terra-Alert
* Browser Name and Version: Chrome, Version 68.0.3440.106 (Official Build) (64-bit)
* Node/npm Version: [e.g. Node 8/npm 5] v8.11.1 (npm v5.6.0)
* Operating System and version (desktop or mobile): mac desktop

bjankord:
### Summary
Resolves #1859


packages/terra-alert/src/Alert.module.scss
```diff
@@ -42,7 +42,7 @@
 
   .body-std {
     margin-bottom: 0.7143rem;
-    margin-right: 0;
+    margin-right: 0.6667rem;
```
StephenEsser: I'm surprised this didn't fail any wdio tests, the text must not extend all the way to the edge in any of the test examples

mjhenkes: Should we add a test that would fail if this changed?

bjankord: Added a test case which would shift the letters in the alert around if padding is adjusted. https://github.com/cerner/terra-core/pull/1959/commits/b79aa2e7f86e47908b37e6cdf93eeb179f001545

Issue #1893, opened 2018-09-19
Avatar accessibility issues with VoiceOver on iOS


bjankord:
# Bug Report

## Description
The avatars that use an SVG background are non-interactable and do not convey information to the user when using VO on iOS. The image avatar is interactable and conveys the alt text and the initial avatars convey the initials.

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-core/#/tests/terra-avatar/avatar/facility-avatar
2. VO does not convey meaning of avatar with SVG background, the aria-label on the div is never read

## Expected Behavior
Avatars should convey their meaning to VO users

## Possible Solution
Move aria-lable text to visually hidden text that VO would be able to pick up.
another option would be looking into setting role="img" on the div with the aria-label on it.

## Environment

* Component Name and Version: terra-avatar
* Browser Name and Version: Mobile Safari 12 w/ VoiceOver
* Node/npm Version: [e.g. Node 8/npm 5]
* Webpack Version:
* Operating System and version (desktop or mobile): iOS 12

Issue #1831, opened 2018-09-05
[terra-avatar] Enhancements


yuderekyu:
# Feature Request
* Split avatar into three variants: Person, Facility, and Multiple (name pending). 
* Make other changes that will be listed in the tech design below.

yuderekyu:
# Avatar Tech Design
`terra-avatar` will be split into three sub-packages: `user`, `location`, and `multi-user` (names pending) and be exported similar to [this](https://github.com/cerner/terra-framework/blob/master/packages/terra-application-utility/src/ApplicationUtility.js). They shall share themeable variables and common utils for determining color.
## Enhancements
### Color Support
* Support 10 color variants.
* Add color prop
	* Defaults to `auto`. Consumers can override this by passing `neutral` for a theme specific base color, or  one of 10 color variants: `one`, `two,`, `three`, `four`, `five`, `six`, `seven`, `eight`, `nine`, `ten`.
* Create shared color utils, utilized when `color="auto"`. This will hash the `alt` prop to determine  a color from the color variants. [Sample hash function.](https://werxltd.com/wp/2010/05/13/javascript-implementation-of-javas-string-hashcode-method/) 

### Other
* `alt` prop shall be required.
* Extract box-shadow to an `:after` selector. Addresses https://github.com/cerner/terra-core/issues/1845.
*  Support single initials for a user avatar.
* Spread custom props onto inner span. This allows an outer component to set `aria-hidden` on avatar.
* Make size in `em`'s - avatar shall scale based on its container's font-size.

## Removals
* Remove support for three initials from user avatar.
* Remove `ariaLabel` prop. Instead, use `alt` prop to set the `aria-label` attribute.
## Code Samples
### User
```jsx
import React from 'react';
import { User } from 'terra-avatar/lib/Avatar';

const AvatarUser = () => (
  <User alt="John Doe" initials="JD" />
);

export default AvatarUser;
```
### Facility
```jsx
import React from 'react';
import { Location } from 'terra-avatar/lib/Avatar';

const AvatarLocation = () => (
  <Location alt="Cerneeeer" />
);

export default AvatarLocation;
```
### MultiUser
```jsx
import React from 'react';
import { MultiUser } from 'terra-avatar/lib/Avatar';

const AvatarMultiUser = () => (
  <MultiUser alt="Squad" />
);

export default AvatarMultiUser;
```


## User Props

| Prop Name | Type | Is Required | Default Value | Description |
|-|-|-|-|-|
[...]

PR #1965, opened 2018-10-10
Add screen reader compatibility for Avatar component

amichaelparker:
### Summary
Added image role and alt text attributes to facility and user avatar placeholders for screen reader compatibility. Dev site tests updated with alt text for testing.

Closes [#1893](https://github.com/cerner/terra-core/issues/1893)

Tested with voiceover on Mac and iOS 12.

### Deployment Link
https://terra-core-deployed-pr-1965.herokuapp.com/#/components/terra-avatar/avatar/avatar


mmalaker:
I tested this with VoiceOver and JAWS. Neither is reading anything other than "image".

bjankord:
Let's handle the deprecate notice in a separate PR and move forward with if the user has both alt and ariaLabel filled out for now, match those prop to prop. If they provide just one of these, populate both aria-label and alt of the span with that value.

bjankord:
Closing out this PR in favor of https://github.com/cerner/terra-core/issues/1831. We'll want to incorporate the accessibility improvements we were working on per @JakeLaCombe's [comment](https://github.com/cerner/terra-core/pull/1965#pullrequestreview-166675575) into the new avatar update. cc/ @yuderekyu

PR #1970, opened 2018-10-15
Fix terra-image docs usage example

eawww:
### Summary
* Corrections to the following errors in `terra-image` documentation example. 
  * Uses an image source string for the `placeholder` prop where an `element` type is expected.
  * Uses `"placeholder image"` as `alt` prop where a alt text of the primary image is expected.

### DIFF
--- a/packages/terra-image/docs/README.md
+++ b/packages/terra-image/docs/README.md
@@ -18,11 +18,13 @@ import placeholderSrc from './mock/valid/path/placeholder.png';
 import invalidImageSrc from './mock/invalid/path/image.png';
 import validImageSrc from './mock/valid/path/image.png';
 
-// Providing an invalid or unresolvable path with cause the placeholder image to be displayed.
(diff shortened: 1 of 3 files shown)


packages/terra-image/docs/README.md
```diff
@@ -18,11 +18,13 @@ import placeholderSrc from './mock/valid/path/placeholder.png';
 import invalidImageSrc from './mock/invalid/path/image.png';
 import validImageSrc from './mock/valid/path/image.png';
 
-// Providing an invalid or unresolvable path with cause the placeholder image to be displayed.
-<Image src={invalidImageSrc} placeholder={placeholderSrc} alt="placeholder image" />
+const placeholderNode = (<img src={placeholderSrc} alt="placeholder image" />);
+
+// Providing an invalid or unresolvable path with cause the placeholder node to be displayed.
```
StephenEsser: Would you mind also updating this comment? `with` Does not seem to fit into this sentence appropriately. Maybe `will` was the intention

PR #1971, opened 2018-10-15
Fix Image flicker #1748

### LINKED ISSUE #1748: Image flickering placeholder on ever rerender
# Bug Report
In our application, we are using an Image. It is flickering the placeholder image everytime our component rerenders.

## Description

https://github.com/cerner/terra-core/blob/master/packages/terra-image/src/Image.jsx#L68
It is showing the placeholder, we believe, because it fails on this line. Using === on two nodes that appear the same fails.

## Steps to Reproduce
Using demographics manager with our application, if we have a profile image and the component containing the demographics manager rerenders, we see the placeholder image once again.

## Expected Behavior
The placeholder image would not be reshown on rerender once the image has loaded

## Possible Solution

I believe you'll have to change the === check on the placeholder image.

## Environment

* Component Name and Version: terra-image 2.17.0
* Browser Name and Version: Chrome Version 67.0.3396.99 
* Node/npm Version: [e.g. Node 8/npm 5] Node: 8.4.0, npm 5.6.0
* Webpack Version: ^4.4.1
* Operating System and version (desktop or mobile):  Desktop--Mac

eawww:
### Summary
* Prevent `terra-image` from flickering due to entering `isLoading` state on any new props.
  * Fixes #1748 

### Additional Details
**Rationale:**
Image was flickering to placeholder element each time new props were given. This was due to the inequality condition between two objects always evaluating to true, specifically on the `placeholder` node prop, which caused the component to render with `isLoading` set to true on any change in props. 
```js
componentWillReceiveProps(newProps) {
    // If a new image is being loaded, reset the state to loading
    if (newProps.src !== this.props.src || newProps.placeholder !== this.props.placeholder) {
      this.setState({ isLoading: true, isError: false });
    }
  }
```

I just removed that condition because, not only was it the source of this issue, but it didn't seem to be necessary. Entering the loading state when a new `src` prop is passed makes sense, but, when a new placeholder is passed, there would be no reason to fall back to the placeholder while the new placeholder loads. This would only cause a properly loaded primary image to flicker even if the check were working as intended.

This should fix 'er right up! Thanks for making contributing to this project a pleasant experience! 😁

### DIFF
--- a/packages/terra-image/CHANGELOG.md
+++ b/packages/terra-image/CHANGELOG.md
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Fix flickering on new props #1748.
 
 2.24.0 - (October 8, 2018)
 ------------------
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -72,6 +72,7 @@ Cerner Corporation
 - Alexander Brisimitzakis [@AlexBrizi]
 - Kyle Roush [@kyleroush]
 - Sharynne Azhar [@sharynneazhar]
+- Eric Wilson [@eawww]
 
 [@ryanthemanuel]: https://github.com/ryanthemanuel
 [@Matt-Butler]: https://github.com/Matt-Butler
@@ -146,3 +147,4 @@ Cerner Corporation
 [@AlexBrizi]: https://github.com/AlexBrizi
 [@kyleroush]: https://github.com/kyleroush
 [@sharynneazhar]: https://github.com/sharynneazhar
+[@eawww]: https://github.com/eawww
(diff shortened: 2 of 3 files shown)


packages/terra-image/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Fix flickering on new props #1748.
```
StephenEsser: There may not be any official documentation on this, but there are a few main groups of headings for the changelog.

Changed, Added, Breaking Change, Fixed

Would you mind updating this to ### Fixed ?

eawww: Absolutely! Fixed.

PR #1975, opened 2018-10-16
UPDATE copyright date from 2017 to 2018.

mndrsn:
### Summary
- UPDATE copyright date from 2017 to 2018 on the README.


README.md
```diff
@@ -167,7 +167,7 @@ npm run test
 
 ## LICENSE
 
-Copyright 2017 Cerner Innovation, Inc.
+Copyright 2018 Cerner Innovation, Inc.
```
emilyrohrbough: Should this be an inclusive date? 2017-2018?

emilyrohrbough: Also, we likely need to update the Copyright date of each component's readme as well

bjankord: @emilyrohrbough We should check with legal about the date. We may be able to go with something like `2017 - Present`

bjankord: We'll want to update this to `2017 - 2018`. The Copyright years listed here match [Apache Foundation's recommendation for year(s) of distribution of the current and past versions of the product](https://www.apache.org/legal/src-headers.html#notice)

Issue #1913, opened 2018-09-21
VoiceOver on iOS does not communicate content in container overlay


bjankord:
# Bug Report

## Description
When a container overlay is opened, it may contain content within it. VO on iOS does not read this content when the overlay is displayed, nor does it pick up that this is an element that can be interacted with. This is not an issue with the full screen overlay.

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-core/#/components/terra-overlay/overlay/overlay
2. Open the container overlay example

## Expected Behavior
VoiceOver on iOS reads the container overlay content when its displayed in overlay.

## Environment

* Component Name and Version: terra-overlay
* Browser Name and Version: Safair Mobile 12 with VoiceOver on iOS 12

bjankord:
Talked with @emilyrohrbough and @jmsv6d about this. We want to keep [this code](https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/Overlay.jsx#L90-L107). It is used to disable focusing on children in the container that is being overlayed. 

It looks like right after [this line](https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/Overlay.jsx#L83), we need to shift focus to the overlay.

bjankord:
We need to explore making it optional to shift focus to relative container based overlays. In some cases, in makes sense, in others it may not. Maybe you have multiple on the page, maybe they only appear for a brief moment while loaded content, e.g. infinite scroll component. In these cases users likely don't want the focus to be moved around, but there are also scenarios it may make sense to move focus to a relative container based overlay. 

To handle this, I propose adding a callback function that will execute when a relative container based overlay is about to open, maybe we expand this to just and overlay (relative container and full screen) which would allow users to control side effects like shifting focus if it made sense for their app.

bjankord:
Resolved in #2249

PR #1983, opened 2018-10-22
Adjustments to terra-overlay for screen reader compatibility

amichaelparker:
### Summary
Focus is moved to the container overlay so screen readers like Voiceover will read the overlay close text. The fullscreen overlay was adjusted to prevent a screen reader from reading the content underneath the overlay.

'base' class was added directly to terra-base component to ensure targeting compatibility.

Closes [#1913](https://github.com/cerner/terra-core/issues/1913)

### Deployment Link
https://terra-core-deployed-pr-1983.herokuapp.com/#/components/terra-overlay/overlay/overlay/


packages/terra-base/src/Base.jsx
```diff
@@ -87,14 +87,14 @@ class Base extends React.Component {
 
     if (locale === undefined) {
       return (
-        <div {...customProps}>
+        <div className={['base', customProps.className]} {...customProps}>
```
emilyrohrbough: Is there styling associated with the `base` class? If not, when this is packed by webpack, the class name will not be hashed and will display as `base`, which means this class name is not locally scoped to this react module. If there are no styles associated, I would suggest a data attribute `terra-base-main`

bjankord: Let's update this to:
```suggestion
        <div data-terra-base {...customProps}>
```

packages/terra-base/src/Base.jsx
```diff
@@ -87,14 +87,14 @@ class Base extends React.Component {
...
           {renderChildren}
         </div>);
     }
 
     if (!this.state.areTranslationsLoaded) return <div>{this.props.translationsLoadingPlaceholder}</div>;
     return (
-      <I18nProvider {...customProps} locale={this.state.locale} messages={messages}>
+      <I18nProvider {...customProps} className={['base', customProps.className]} locale={this.state.locale} messages={messages}>
```
bjankord: Let's update this to:
```suggestion
      <I18nProvider data-terra-base {...customProps} locale={this.state.locale} messages={messages}>
```

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -81,9 +82,10 @@ class Overlay extends React.Component {
       this.container = node.parentNode;
       this.containerChildren = this.container.children;
       this.disableContainerChildrenFocus();
+      this.containerChildren[0].focus();
     } else {
       document.documentElement.style.overflow = 'hidden';
-      this.container = null;
+      document.documentElement.getElementsByClassName('base')[0].setAttribute('aria-hidden', 'true'); // prevent screen reader from moving to hidden content
```
bjankord: Let's update this to:
```suggestion
      document.querySelector(selector).setAttribute('aria-hidden', 'true'); // prevent screen reader from moving to hidden content
```

Where `selector` is a variable based on a new overlay prop, `rootSelector`.

```jsx
const selector = this.props.rootSelector;
```

We'll default this prop to `data-terra-base`, but this allows users to pass in their own "root selector" if they are not using terra-base.

packages/terra-base/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Added
+* 'base' class added to Base component
```
emilyrohrbough: We need to update this changelog to reflect the data attribute

packages/terra-overlay/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Focus changes to container overlay text for screen reader compatibility
+* Main page content hidden from screen reader while fullscreen overlay is deployed
```
emilyrohrbough: Lets add : 
```
### Added
- Added react-portal dependency for ...
- Added `rootSelector` prop for ...
```

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -125,21 +133,24 @@ class Overlay extends React.Component {
...
   render() {
     const {
-      children, isOpen, backgroundStyle, isScrollable, isRelativeToContainer, onRequestClose, ...customProps
+      children, isOpen, backgroundStyle, isScrollable, isRelativeToContainer, onRequestClose, rootSelector, ...customProps
     } = this.props;
     const type = isRelativeToContainer ? 'container' : 'fullscreen';
 
     if (!isOpen) {
-      this.resetBeforeOverlay();
+      this.resetBeforeOverlay(this.props.rootSelector);
```
yuderekyu: we can remove `this.props` and just use `rootSelector` - we have already unpacked `this.props` into distinct prop variables on line 148

amichaelparker: Change made!

packages/terra-overlay/package.json
```diff
@@ -32,9 +32,10 @@
     "classnames": "^2.2.5",
     "focus-trap-react": "^4.0.1",
     "prop-types": "^15.5.8",
-    "terra-button": "^2.28.0",
-    "terra-doc-template": "^1.19.0",
-    "terra-icon": "^2.26.0"
+    "react-portal": "^4.1.2",
+    "terra-button": "^2.27.0",
+    "terra-doc-template": "^1.18.0",
+    "terra-icon": "^2.25.0"
```
emilyrohrbough: Whoops, button, doc-template and icon versions need bumped to match latest released versions

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -39,19 +40,24 @@ const propTypes = {
   * Indicates if the overlay is relative to the triggering container.
   */
   isRelativeToContainer: PropTypes.bool,
+  /**
+   * Allows assigning of root element custom data attribute for easy selecting.
```
JakeLaCombe: This documentation feels a little vague. What do we mean by the root element exactly? Root of the application, or root container that triggered the overlay to go.

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -76,14 +82,16 @@ class Overlay extends React.Component {
...
 
     if (this.props.isRelativeToContainer) {
       this.container = node.parentNode;
       this.containerChildren = this.container.children;
       this.disableContainerChildrenFocus();
+      this.containerChildren[0].focus();
     } else {
       document.documentElement.style.overflow = 'hidden';
-      this.container = null;
+      document.querySelector(selector).setAttribute('aria-hidden', 'true'); // prevent screen reader from moving to hidden content
```
JakeLaCombe: Does this need to be removed when the overlay is hidden or unmounted.

JakeLaCombe: I do see that iit's remove on the resteBeforeOverlay method, but I'm curious if we should add it to the componentWillUnmount lifecycle step. Anyone else have any thoughts?

bjankord: I don't think it would hurt to check in componentWillUnmount lifecycle step and make sure its removed.

amichaelparker: I talked to @bjankord, and most of these items are related to fixing a bug with the fullscreen overlay that was actually fixed at the same time in [#1955](https://github.com/cerner/terra-core/pull/1955), and this stuff was in addition to the main scope of this PR, so I'm going to strip it all out in favor of the already merged changes. :)

amichaelparker: For checking this attribute, the merged changes from master run this check again in componentWillUnmount, so this seems to be addressed in master.

packages/terra-overlay/src/Overlay.jsx
amichaelparker: I changed the description a little bit in one of the more recent commits. The goal is to give an easy way to target whatever the user's version of the base component is... If the updated comment doesn't work, let me know what you suggest!

packages/terra-overlay/CHANGELOG.md
amichaelparker: Added during the slash and burn master rebase! :)

packages/terra-base/CHANGELOG.md
amichaelparker: Changed!

packages/terra-overlay/src/LoadingOverlay.jsx
```diff
@@ -32,6 +32,10 @@ const propTypes = {
   * Indicates if the overlay is relative to the triggering container.
   */
   isRelativeToContainer: PropTypes.bool,
+  /**
+   * Allows assigning of root element custom data attribute for easy selecting of document base component.
```
bjankord:
```suggestion
   * Query selector used to hide the DOM that is not part of the overlay which prevents focus from moving outside of the overlay.
```

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -134,21 +142,24 @@ class Overlay extends React.Component {
     }
   }
 
-  resetBeforeOverlay() {
+  resetBeforeOverlay(selector) {
     document.documentElement.style.overflow = this.overflow;
+    if (document.querySelector(selector)) {
```
bjankord: Let's also add this conditional check inside a componentWillUnMount lifecycle method.

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -79,20 +84,23 @@ class Overlay extends React.Component {
...
 
   setContainer(node) {
     if (!node) { return; } // Ref callbacks happen on mount and unmount, element is null on unmount
     this.overflow = document.documentElement.style.overflow;
+    const selector = this.props.rootSelector;
 
     if (this.props.isRelativeToContainer) {
       this.container = node.parentNode;
       this.containerChildren = this.container.children;
       this.disableContainerChildrenFocus();
```
bjankord: Let's add this to set the container to aria-hidden when using overlay relative to a container.
```suggestion
      this.disableContainerChildrenFocus();
      this.container.setAttribute('aria-hidden', 'true');
```

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -134,21 +142,24 @@ class Overlay extends React.Component {
...
   }
 
-  resetBeforeOverlay() {
+  resetBeforeOverlay(selector) {
     document.documentElement.style.overflow = this.overflow;
+    if (document.querySelector(selector)) {
+      document.querySelector(selector).setAttribute('aria-hidden', 'false');
+    }
     if (this.props.isRelativeToContainer) {
       this.enableContainerChildrenFocus();
```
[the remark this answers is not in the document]
amichaelparker: Added the aria-hidden logic to the container overlay

bjankord:
@amichaelparker Looks like we need to add a conditional check before we set aria-hidden attribute, there's a console error in the browser around not being able to setAttribute on undefined.

mmalaker:
Functional verification completed. The overlay examples now display and function as expected. 

Note that one issue was found during testing where the z-index wasn't set properly on some of the test examples. That issue will be addressed separately.

PR #1993, opened 2018-10-25
Removed typo in ButtonVariants.

### LINKED ISSUE #1289: Remove DE-EMPSHASIS from ButtonVariants
# Issue Description
The Button exports its supported variant strings which includes, 'DE-EMPSHASIS', that contains an extra 's'. This should be removed in the next major release of button.

https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L18

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
Button variant props API are spelled correctly

## Current Behavior
Button variant props API are spelled incorrectly

jeremyfuksa:
### Summary
Resolves #1289.

### DIFF
--- a/packages/terra-button/src/Button.jsx
+++ b/packages/terra-button/src/Button.jsx
@@ -15,8 +15,6 @@ const KEYCODES = {
 const ButtonVariants = {
   NEUTRAL: 'neutral',
   EMPHASIS: 'emphasis',
-  // TODO: this should be removed on the next major version bump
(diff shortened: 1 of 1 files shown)


bjankord:
Hey Jeremy, thanks for the PR. We're not quite ready to do a major version bump on this component yet. When we are ready to roll out the major version bump for this, will incorporate this change and any other major version bump changes like changing the variable names to match the version scheme were working on.

PR #1995, opened 2018-10-27
Improve code examples for terra-form-input

noahbenham:
### Summary
Cleans up & makes example code more copy & pastable.

### Additional Details
N/A

@cerner/terra

### DIFF
--- a/packages/terra-form-input/src/terra-dev-site/doc/common/BlankExample.jsx
+++ b/packages/terra-form-input/src/terra-dev-site/doc/common/BlankExample.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import Input from '../../../Input';
+import Input from 'terra-form-input'
 
 const BlankExample = () => (
   <div>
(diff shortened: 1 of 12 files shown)


packages/terra-form-input/src/terra-dev-site/doc/common/BlankExample.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 
-import Input from '../../../Input';
+import Input from 'terra-form-input'
```
emilyrohrbough: @mjhenkes Does this lose hot reloading by defining the module?

bjankord: I believe so.

bjankord: Talked with @mjhenkes and the current format is used to enable hot reloading. Changing the format is something we're not interested in updating at this time.

noahbenham: Thanks @bjankord. It appears most of your other examples define the module instead of using a relative import. Could you explain the discrepancy? Does `terra-form-input` have a special need for hot reloading that other examples don't?

bjankord: The other examples likely haven't been updated to a format that supports hot-reloading with the webpack config used for this repo.

Issue #1580, opened 2018-06-06
Provide options to resize the InputField without using custom style


Mahmud-Khan:
# Issue Description
For some use case, the user may need to resize the width of the InputField to a custom size. At present setting min and max, input attributes can resize the InputField. However, this solution works for chrome & safari, but not for Firefox and ie. Please check the screenshots below.
chrome: 
<img width="355" alt="screenshot_chrome" src="https://user-images.githubusercontent.com/21693381/41059170-d016149a-6999-11e8-9b50-d1f14d82f429.png">
Firefox:
<img width="360" alt="screenshot_firefox" src="https://user-images.githubusercontent.com/21693381/41059194-db14605e-6999-11e8-8d71-7081683cfd65.png">

Setting max-width through a CSS class works. But it would be convenient to provide a prop to resize the InputField width.
## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
User will be able to resize the InputField to a custom size passing a prop value.

## Current Behavior
User can resize the InputField that works for all browsers through custom styling.

## Steps to Reproduce

1. Add InputField in a form or page.
2. Resize the InputField by passing values to inputAttr prop
3. InputField is resized from chrome & safari,but not for Firefox & ie

## Environment

* Component Version: [InputField](http://engineering.cerner.com/terra-core/#/components/terra-form-input/form-input/input-field)
* Browser Name and Version:
* Operating System and version (desktop or mobile): Both

PR #2005, opened 2018-10-31
Add maxWidth prop to terra-form-input field

amichaelparker:
### Summary
Based on request in [1580](https://github.com/cerner/terra-core/issues/1580), adds maxWidth prop to Field input container. Allows for both customprop style or maxwidth prop to set container width.

Closes [#1580](https://github.com/cerner/terra-core/issues/1580)


packages/terra-form-input/src/InputField.jsx
```diff
@@ -115,14 +119,17 @@ const InputField = (props) => {
...
+    onChange,
     refCallback,
     required,
     showOptional,
-    onChange,
     value,
     ...customProps
   } = props;
 
+  const styles = 'style' in props ? props.style : { maxWidth: props.maxWidth };
```
StephenEsser: `maxWidth` will need to be merged into style properties if either one exists, otherwise all previous styles will be overwritten. 

```
<Input style={{ color: 'red' }} maxWidth="200px" />
```

packages/terra-form-input/src/InputField.jsx
```diff
@@ -137,6 +144,7 @@ const InputField = (props) => {
       isInline={isInline}
       isLabelHidden={isLabelHidden}
       htmlFor={inputId}
+      style={styles}
```
StephenEsser: The Field has a class applying a maxWidth, this will overwrite the class style. This might be something to check with UX on to make sure it's okay to extend beyond the maxWidth set in the Field itself.

https://github.com/cerner/terra-core/blob/50a0e01af972252c657d020790e1b0f94dc02be4/packages/terra-form-field/src/Field.module.scss#L5

packages/terra-form-input/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Added
+* maxWidth prop added to input field component
```
StephenEsser: Should this change be applied to `terra-form-field` so all fields have the ability to have a maxWidth. As is this functionality will only work for `terra-form-input` and no other Field components. (SelectField, RadioField, TextArea)

amichaelparker: Logic moved to terra-form-field!

packages/terra-form-input/src/InputField.jsx
bjankord: Yeah, I think the behavior of having `style={{maxWidth: 100%}}` or `maxWidth="100%" will override the default set by the class is fine.

packages/terra-form-field/src/Field.jsx
```diff
@@ -61,6 +65,11 @@ const propTypes = {
    * Whether or not to append the 'optional' label to a non-required field label.
    */
   showOptional: PropTypes.bool,
+  /**
+   * Allows for custom inline styles in addition to (or separate from) the maxWidth style prop.
+   */
+  // eslint-disable-next-line react/forbid-prop-types
+  style: PropTypes.object,
```
StephenEsser: I'm not sure if we should expose style as a first class property, this will generate documentation for it within the example site props table of this component. Style can be read from props without exposing it as a prop-type

packages/terra-form-field/src/Field.jsx
```diff
@@ -100,11 +111,25 @@ const Field = (props, { intl }) => {
...
+    maxWidth,
     required,
     showOptional,
+    style,
     ...customProps
   } = props;
 
+  let customStyles;
+
+  if (props.style) {
```
StephenEsser: Each occurrence of `props.style` can be simplified to `style` because it was deconstructed from props above.

packages/terra-form-field/src/Field.jsx
```diff
@@ -100,11 +111,25 @@ const Field = (props, { intl }) => {
...
+    if (!props.style.maxWidth) {
+      customStyles = { ...props.style, maxWidth };
+    } else if (props.style.maxWidth) {
+      customStyles = { ...props.style };
+    }
+  } else {
+    customStyles = { maxWidth };
+  }
```
StephenEsser: I think this can be simplified to
```
const customStyles = Object.assign({}, { maxWidth }, style);
```

packages/terra-form-field/src/Field.jsx
```diff
@@ -132,7 +141,7 @@ const Field = (props, { intl }) => {
   );
 
   return (
-    <div {...customProps} className={fieldClasses}>
+    <div style={customStyles} {...customProps} className={fieldClasses}>
```
StephenEsser: I think style needs to be declared after custom properties to prevent maxWidth from being overwritten. 

You can double check before making the change with something like this:
```
<Field maxWidth="300px" style={{ color: 'red'}} />
```

```suggestion
    <div {...customProps}  style={customStyles} className={fieldClasses}>
```

amichaelparker: It seems to apply maxWidth regardless of order (either swapping around the props on the test case, swapping style / customProps on the component, or both).

StephenEsser: That makes sense, style was deconstructed so it doesn't get included with the customProps spread.

packages/terra-form-field/src/Field.jsx
```diff
@@ -53,6 +53,10 @@ const propTypes = {
    */
   // eslint-disable-next-line react/forbid-prop-types
   labelAttrs: PropTypes.object,
+  /**
+   * Set max width of input field. Passed in custom inline styles take precedence over this prop.
+   */
```
neilpfeiffer: @amichaelparker Wording update suggestion:
```
  /**
   * Set the max-width of a field using `length` or `%`.  Best practice recommendation to never exceed
   * a rendered value of 1020px. _(Note: Providing custom inline styles will take precedence.)_
   */
```

neilpfeiffer:
Tested: PR #2005
Resolves: Issue #1580


**Reviewed:**
- allows for ability to use a form-field as a structural element with set width instead of forcing to be included in grid or other container. 
(form-field normally grows with 100% width and it is bound by parent)

\
\
Verified by @neilpfeiffer, ready for release after applying comment for docsite wording update: https://github.com/cerner/terra-core/pull/2005#discussion_r233092193


**Pass** |
-----------|
UX Review|

packages/terra-form-field/src/Field.jsx
amichaelparker: Changed!

StephenEsser:
The danger script was added last night. You can add a new tar file by running `npm run compile` inside of `terra-form-field`

packages/terra-form-field/src/Field.jsx
```diff
@@ -100,11 +106,15 @@ const Field = (props, { intl }) => {
...
     label,
     labelAttrs,
+    maxWidth,
     required,
     showOptional,
+    style,
     ...customProps
   } = props;
 
+  const customStyles = Object.assign({ maxWidth }, style);
```
bjankord: Let's update this to:
```suggestion
  const customStyles = maxWidth ? Object.assign({ maxWidth }, style) : style;
```

This will remove the following from the jest snapshots:
```
style={
    Object {
      "maxWidth": undefined,
    }
  }
```

PR #2017, opened 2018-11-02
Make Abstract Modal More Accessible

### LINKED ISSUE #2013: Default modal role to dialog and add aria-modal to abstract modal.
# Feature Request

## Description

While doing accessibility testing with voiceover iOS, we found that focus wasn't being trapped on the modal. This is due to the role not being dialog, and leaving off aria-modal="true" from the modal. To resolve this issue, I believe it's best to default the role prop to dialog, and set aria-modal to true.

JakeLaCombe:
### Summary
Resolves #2013 

iOS was having issues focusing on modal content. After researching online, I found that the best approach was to default all the modals to be dialog, and ensure the aria-modal attribute is set on them. You can read more about this approach at the following issue. A lot of great sources are listed here. 

https://github.com/reactjs/react-modal/issues/611

### Additional Details

Test in VoiceOver for iOS and Mac, and also tested with JAWS in IE11.


packages/terra-abstract-modal/CHANGELOG.md
```diff
@@ -3,6 +3,14 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Defaulted role prop to dialog.
+* Added aria-modal to the modal.
+
+1.23.0 - (November 9, 2018)
```
tbiethman: Duplicate entry for this feller

mmalaker:
Functional verification completed. The abstract modal examples display and function correctly as confirmed through screen reader testing.

Issue #1909, opened 2018-09-20
Add example icon that uses aria-label to test page to verify usage with screen readers


bjankord:
# Feature Request

## Description
By default, we hide icons from screen readers and encourage users to pair them with visual text. In the case that visual text can't be paired to convey the icon message, users can set an aria-label. We should set up a test page to ensure this works as expected.

PR #2029, opened 2018-11-07
Added a test page for the icon compenent using aria-label

AlexBrizi:
### Summary
Added a test page for the icon component that utilizes aria-label to convey the message of the icon to screen readers.

Resolves [#1909](https://github.com/cerner/terra-core/issues/1909)


packages/terra-icon/tests/wdio/icon-spec.js
```diff
@@ -20,4 +20,11 @@ describe('Icon', () => {
     Terra.should.beAccessible();
     Terra.should.matchScreenshot();
   });
+
+  describe('AriaLabel', () => {
+    before(() => browser.url('/#/raw/tests/terra-icon/icon/icon-aria-label'));
```
JakeLaCombe: Does this warrant a new screenshot test? All we are really doing is adding aria-label to the alert, which would make beAccessible pass whether it's set or not.

JakeLaCombe:
I see @yuderekyu  comment on this. I'm fine with the accessible testing on it, but I don't think it warrants a new screenshot.

packages/terra-icon/tests/wdio/icon-spec.js
bjankord: Yeah, let's drop this line:

```
 Terra.should.matchScreenshot();
```

PR #2054, opened 2018-11-14
Correct action footer theme tests.

dkasper-was-taken:
### Summary
Correct terra-action-footer theme test routing.

@cerner/terra


packages/terra-action-footer/tests/wdio/theme/block-action-footer-theme-spec.js
```diff
@@ -5,13 +5,13 @@ const {
...
   describe('No Actions', () => {
-    beforeEach(() => browser.url('/#/raw/tests/terra-action-footer/action-footer/without-actions-action-footer-container'));
+    beforeEach(() => browser.url('/#/raw/tests/terra-action-footer/action-footer/without-actions-block-action-footer'));
 
     Terra.should.themeEachCustomProperty(withoutActionsThemeables);
   });
 
   describe('With Actions', () => {
-    beforeEach(() => browser.url('/#/raw/tests/terra-action-footer/action-footer/with-actions-action-footer-container'));
+    beforeEach(() => browser.url('/#/raw/tests/terra-action-footer/action-footer/with-actions-block-action-footer'));
```
StephenEsser: These routes were a direct copy paste from the original file, are they also wrong here?

https://github.com/cerner/terra-core/blob/master/packages/terra-action-footer/tests/wdio/block-action-footer-spec.js

PR #2060, opened 2018-11-15
Fix terra-icon global icon animation

dev-cprice:
### Summary
When using the default `@rails/webpacker` css-loaders, the Icon animation global classes were being hashed. This resolves that issue by wrapping the spin class in `:global()`

@cerner/terra


bjankord:
To regenerate the tar file, cd into terra-core/packages/terra-icon and run the following npm script. `npm run compile`, then commit the updated .tgz file.

PR #2073, opened 2018-11-27
Update vertical centering styles for buttons

### LINKED ISSUE #1383: [terra-button] Inline SVGs do not align vertically in buttons
# Issue Description
I noticed in every case an inline svg is used as the background for a button, the svg does not align vertically. Some current implementations work around this by adding margins to control the centering. Some components do not center the svg at all.

# Examples
## [terra-dialog](http://engineering.cerner.com/terra-core/#/site/components/site/dialog/index)
image

## [terra-toggle-button](http://engineering.cerner.com/terra-core/#/site/components/site/toggle-button/index)
image

## terra-action-header (In development)
image

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [X] Bug
- [ ] Other

## Expected Behavior
Inline SVGs vertically align in terra-button.

## Current Behavior
Inline SVGs are not vertically centered without setting vertical margins.

bjankord:
### Summary
Fixes #1383


bjankord:
@yuderekyu I think so, running wdio tests now. Wanted to send out a PR and see if any failed but it looks like I have some lint to clean up.

bjankord:
I wont have much time to take a look at this for a while, but @neilpfeiffer did give me some styles to test out with this. Will check that out after break.

neilpfeiffer:
If you run the PR and tests, the current styles break the ability for long button labels to correctly wrap with appropriate line-height (the current code has the new text lines just overlap all overtop each other), so my suggestions should resolve the issues [(gist posted here)](https://gist.github.cerner.com/NP032453/ecfc050bde7f01b95385dad9a4e4408d).

bjankord:
Closing this, will pick up when I get capacity again.

Issue #1884, opened 2018-09-17
Update components to work in react strict mode without triggering warnings


bjankord:
# Feature Request

## Description
In react 16.3, they introduced a `StrictMode` component to warn about 

<ul>
<li><a href="https://reactjs.org/docs/strict-mode.html#identifying-unsafe-lifecycles">Identifying components with unsafe lifecycles</a></li>
<li><a href="https://reactjs.org/docs/strict-mode.html#warning-about-legacy-string-ref-api-usage">Warning about legacy string ref API usage</a></li>
<li><a href="https://reactjs.org/docs/strict-mode.html#detecting-unexpected-side-effects">Detecting unexpected side effects</a></li>
<li><a href="https://reactjs.org/docs/strict-mode.html#detecting-legacy-context-api">Detecting legacy context API</a></li>
</ul>

https://reactjs.org/docs/strict-mode.html

By updating components to render in strict mode without triggering warnings, the components will be able to work in newer versions of react where outdated lifecycle hooks are removed.

It doesn't look like we are using `componentWillUpdate` anywhere throughout the terra ecosystem of components we own. I also didn't see any use of the legacy string ref API in our code base.

bjankord:
Starting up the local dev server and hitting the hompage, the following warnings are listed in the console when enabling strict mode:

```
componentWillMount: Please update the following components to use componentDidMount instead: Redirect, Route, Switch

componentWillReceiveProps: Please update the following components to use static getDerivedStateFromProps instead: AbstractModal, ApplicationLayout, ApplicationTabs, Image, Layout, NavigationLayout, Route, RoutingStack, Switch, TabMenu
```

Some of the components that need updated are from react-router. There is [an issue logged](https://github.com/ReactTraining/react-router/pull/6256) to enable react router usage in react strict mode without triggering warnings. For the scope of this issues, we'll just be focused on our components.

bjankord:
Here are the places we're currently using `componentWillMount`
* https://github.com/cerner/terra-framework/search?q=componentWillMount&unscoped_q=componentWillMount
* https://github.com/cerner/terra-core/search?q=componentWillMount&unscoped_q=componentWillMount

[...]

PR #2078, opened 2018-11-30
Remove componentWillReceiveProps from menu

bjankord:
### Summary


packages/terra-menu/src/Menu.jsx
```diff
@@ -71,8 +71,9 @@ class Menu extends React.Component {
     this.state = { stack: [this] };
   }
 
-  componentWillReceiveProps(nextProps) {
-    if ((this.props.isOpen && !nextProps.isOpen) || this.props.children.length !== nextProps.children.length) {
+  componentDidUpdate(prevProps) {
+    if ((!this.props.isOpen && prevProps.isOpen) || this.props.children.length !== prevProps.children.length) {
+      /* eslint-disable react/no-did-update-set-state */
```
bjankord: More info on this rule here: https://github.com/airbnb/javascript/issues/1875

packages/terra-menu/src/_MenuContent.jsx
```diff
@@ -115,14 +115,6 @@ class MenuContent extends React.Component {
     return { isSelectableMenu: this.isSelectable() };
   }
 
-  componentWillReceiveProps(newProps) {
```
bjankord: I don't believe this was doing anything impactful. I tested removing and went through open and closing nested menus with mouse / keyboard and didn't notice any impact. If we think this should be preserved, let me know and I can update it.

packages/terra-menu/src/terra-dev-site/test/menu/menu/DynamicMenuChildren.test.jsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
```
bjankord: Created this test page to test `this.props.children.length !== prevProps.children.length` logic in `Menu.jsx`

packages/terra-menu/src/_MenuContent.jsx
bjankord: I figured out why this is needed. Going to incorporate it into the refactor. Thank goodness for visual regression tests.

PR #2080, opened 2018-11-30
Remove use of componentWillReceiveProps from date-picker

bjankord:
### Summary


packages/terra-date-picker/src/DatePicker.jsx
```diff
@@ -109,6 +109,7 @@ class DatePicker extends React.Component {
 
     this.state = {
       selectedDate: DateUtil.createSafeDate(props.selectedDate),
+      prevPropsSelectedDate: props.selectedDate,
```
tbiethman: The pattern of keeping a copy of props in state makes me sad :( but implementing componentDidUpdate and dealing with two renders doesn't seem great either.

bjankord: Agreed, for this update, I'm following a pattern from the react docs on [getDerivedStateFromProps](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#alternative-1-reset-uncontrolled-component-with-an-id-prop) where I'm using `prevPropsSelectedDate` as the "id" prop. The [concept of storing previous props in state looks like it will be carried over to react hooks](https://reactjs.org/docs/hooks-faq.html#how-do-i-implement-getderivedstatefromprops).

Issue #2043, opened 2018-11-13
Document usage of maxWidth prop on input-field, radio-field, checkbox-field, select-field, textarea-field


bjankord:
# Feature Request

## Description
We recently added the prop, `maxWidth` to terra-form-field. We need to expose and document this prop on all the form components that interface with terra-form-field.

This should be a prop we expose on the field wrapper components that passes a value directly to the `maxWidth` prop on terra-form-field.

## @ Mentions
cc/ @neilpfeiffer

PR #2084, opened 2018-12-03
Add maxWidth prop to various components for passthrough to form-field

amichaelparker:
### Summary
This adds a maxWidth prop to various form components (input, textarea and select) for passthrough to the same in terra-form-field.

Closes [#2043](https://github.com/cerner/terra-core/issues/2043)


packages/terra-form-select/src/SelectField.jsx
```diff
@@ -140,6 +146,8 @@ const SelectField = ({
     isInline={isInline}
     isLabelHidden={isLabelHidden}
     htmlFor={selectId}
+    maxWidth={maxWidth}
+    {...customProps}
```
StephenEsser: Why was the customProps spread moved to last?

amichaelparker: Oh yeah, was testing order to max sure I didn't screw up priority.

emilyrohrbough:
I think you'll need to update your jest snapshots.

Issue #2089, opened 2018-12-04
Enhance Input to accept pattern prop as first class attribute


Jmacek:
# Feature Request

## Description
I would like to request that the HTML5 [`pattern`](https://www.w3schools.com/tags/att_input_pattern.asp) attribute be a first class attribute due to the fact that not all components have `customProps` to be able to pass the field into. I also think this is a commonly enough used attribute that it should be assignable directly rather than through `customProps`.

## @ Mentions
@emilyrohrbough

PR #2091, opened 2018-12-05
Add Pattern Prop to Input Component

amichaelparker:
### Summary
This adds the html pattern attribute as a first class prop to terra-form-input. 

Closes [#2089](https://github.com/cerner/terra-core/issues/2089)


packages/terra-form-input/src/Input.jsx
```diff
@@ -40,6 +40,10 @@ const propTypes = {
    * Content to be displayed as the name.
    */
   name: PropTypes.string,
+  /**
+   * Pattern to match when input type is used.
```
JakeLaCombe: This comment is a little confusing with the is used ending. I would just say Pattern for the input value.

JakeLaCombe:
I would expect a jest test to be added and updated

packages/terra-form-input/src/Input.jsx
amichaelparker: Comment changed to suggested text.

packages/terra-form-input/tests/jest/Input.test.jsx
```diff
@@ -16,6 +16,12 @@ describe('Input', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  it('should render with a type of password and pattern prop value', () => {
+    const input = <Input type="password" pattern=".{6,}" />;
+    const wrapper = render(input);
```
StephenEsser: Is there a specific reason to do a full render vs a shallow render here?

amichaelparker: I updated that test (and the others) to run shallow, and the snaps all seem to look good. Look them over and let me know!

packages/terra-form-input/src/Input.jsx
```diff
@@ -40,6 +40,10 @@ const propTypes = {
    * Content to be displayed as the name.
    */
   name: PropTypes.string,
+  /**
+   * Pattern for the input value.
```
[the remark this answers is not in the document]
emilyrohrbough: According to the definition, this pattern applies to certain input types. Should we also add the type prop so it is also first class?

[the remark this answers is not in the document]
bjankord: Sure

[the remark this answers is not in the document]
bjankord: Let's get an issue logged for adding type as a first-class prop and handle it in a new PR.

PR #2097, opened 2018-12-05
Remove use of componentWillReceiveProps from toggle button

bjankord:
### Summary


packages/terra-toggle-button/src/ToggleButton.jsx
```diff
@@ -75,14 +75,6 @@ class ToggleButton extends React.Component {
     this.handleOnClick = this.handleOnClick.bind(this);
   }
 
-  componentWillReceiveProps(nextProps) {
```
bjankord: This allowed the prop `isInitiallyOpen` to be used in was we didn't intend. Now, this prop is only used in the constructor to set initial state.

PR #2098, opened 2018-12-05
Remove use of componentWillReceiveProps from toggle section header

bjankord:
### Summary


packages/terra-toggle-section-header/src/ToggleSectionHeader.jsx
```diff
@@ -65,12 +65,6 @@ class ToggleSectionHeader extends React.Component {
     this.wrapOnClick = this.wrapOnClick.bind(this);
   }
 
-  componentWillReceiveProps(nextProps) {
```
bjankord: This allowed the prop `isInitiallyOpen` to be used in was we didn't intend. Now, this prop is only used in the constructor to set initial state.

PR #2101, opened 2018-12-06
Remove use of componentWillReceiveProps from signature

bjankord:
### Summary


packages/terra-signature/src/Signature.jsx
```diff
@@ -79,10 +79,11 @@ class Signature extends React.Component {
     this.canvas.addEventListener('resize', this.updateDimensions);
   }
 
-  componentWillReceiveProps(nextProps) {
-    if ((this.props.lineSegments !== nextProps.lineSegments) || (this.props.lineWidth !== nextProps.lineWidth)) {
-      this.setState({ lineSegments: nextProps.lineSegments });
-      this.drawSignature(nextProps.lineSegments, nextProps.lineWidth);
+  componentDidUpdate(prevProps) {
+    if ((this.props.lineSegments !== prevProps.lineSegments) || (this.props.lineWidth !== prevProps.lineWidth)) {
+      // eslint-disable-next-line react/no-did-update-set-state
```
bjankord: Wanted to keep the state call and the method `drawSignature` update together so I opted for componentDidUpdate instead of splitting it across `componentDidUpdate` and `getDerivedStateFromProps`.

PR #2103, opened 2018-12-06
Remove use of componentWillReceiveProps from base

bjankord:
### Summary


packages/terra-base/src/Base.jsx
```diff
@@ -73,6 +73,17 @@ class Base extends React.Component {
     }
   }
 
+  componentDidUpdate(prevProps) {
```
yuderekyu: oops, forgot to remove `componentWillReceiveProps`

bjankord: Ah, removed in this commit: https://github.com/cerner/terra-core/pull/2103/commits/aa9d59465e0678efb018e9c4d05b24aae47634d5

PR #2107, opened 2018-12-07
Updating terra-date-picker examples/tests to use DisclosureManager v3

tbiethman:
### Summary
The terra-date-picker has examples/tests that include date pickers in modals. I'm updating those examples to use the new APIs provided by DisclosureManager v3.


tbiethman:
@bjankord Regarding a major version bump, only examples and tests are impacted here; the DatePicker API has not been altered in any way. The way this change would be non-passive is if someone was incorrectly and transitively dependent on the modal manager dependency in this package.

bjankord:
> Regarding a major version bump, only examples and tests are impacted here; the DatePicker API has not been altered in any way. The way this change would be non-passive is if someone was incorrectly and transitively dependent on the modal manager dependency in this package.

Ok, I think I'm fine with that. My main concern is around apps pulling in / bundling 2 versions of modal manager without them really knowing it.

e.g. App defines modal-manager `^3.0.0` in package.json, and date-picker now pulls in modal-manager `^4.0.0`, the bundle now ends up with 2 versions of modal manager.

Node modules directory ends up looking something like this after install:

```sh
my-app/
  node_modules/
    terra-modal-manager (v3)
    terra-date-picker (v2)
      node_modules/
        terra-modal-manager (v4)
```

If we roll out this change as a major release, consumers can still get mixed versions by simply pulling modal manager v3 and datepicker v3 together, but if they are concerned about the bundle size, they can either roll back to datepicker v2 (requires no development work) or update to modal manager v4 (requires development work).

With rolling this change out as a passive release in datepicker, if consumers have the example above, the only way to remove the mixed versions of modal manager is to update their main depdency on modal manager to v4, which requires some additional development time.

I don't think it's a big concern, but something worth thinking through.

tbiethman:
I think my wires got crossed somewhere. I updated the Changlog for the DatePicker but made changes for the DateTimePicker. There's also a few DatePicker examples that need updating, so I'll go through and do that real quick.

PR #2112, opened 2018-12-10
Add Focus Fallback to Abstract Modal

### LINKED ISSUE #2069: Support Fallback Focus in Terra Abstract Modal
# Bug Report

## Description
Focus Trap as of 3.0.0[1] requires at least one focusable element or it fails with error:  "Error: You can't have a focus-trap without at least one focusable element"

## Steps to Reproduce

Example code:
```
<AbstractModal
  isOpen={state.isOpenModal}
  onRequestClose={closeModalFunc}
  ariaLabel="Hi"
>
  Have a nice day.  Click anywhere to continue
</AbstractModal>

When modal attempts rendered, error occurs.

There was a related change which removes modal from tabs only on iOS: https://github.com/cerner/terra-core/pull/2017 

## Expected Behavior
<!-- A clear and concise description of what you expected to happen. -->

## Possible Solution
* Set 'fallbackFocus' to modal container
* OR expose 'fallbackFocus' to clients
* OR next another div with tabindex=0 (if that plays ok with accessibility)

## Environment
<!-- Include as many relevant details about the environment you experienced the bug in -->
* Component Name and Version: Terra Abstract Modal >= 1.20
* Browser Name and Version: iOS => 11.2
* Node/npm Version: [e.g. Node 8/npm 5] Node 8.9.2 / Npm 5.5.1
* Webpack Version: Webpack 4.26.0

## @ Mentions
@bjankord via Teams

### PR DESCRIPTION
**Pull request by JakeLaCombe**
### Summary
Resolves #2069

If you open an Abstract Modal without providing a focusable element, it will error out. This solution exposes the fallbackFocu option that FocusTrap provides for developers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,6 @@ Please review [Terra's Internationalization documentation](https://engineering.c
 - [terra-search-field](https://github.com/cerner/terra-core/tree/master/packages/terra-search-field)
 - [terra-status-view](https://github.com/cerner/terra-core/tree/master/packages/terra-status-view)
 - [terra-tabs](https://github.com/cerner/terra-core/tree/master/packages/terra-tabs)
-- [terra-time-input](https://github.com/cerner/terra-core/tree/master/packages/terra-time-input)
 
 ## Contributing
 

### CODE REVIEW

File: packages/terra-abstract-modal/tests/jest/AbstractModal.test.jsx [hunk 1]
```diff
@@ -67,6 +67,30 @@ it('should render the correct snapshot for iPads', () => {
...
 });
 
+it('should allow us to pass in a fallbackFocus element', () => {
+  const modal = mount(<AbstractModalExample fallbackFocus="#focusable-element" />);
+  expect(modal).toMatchSnapshot();
+  global.navigator.platform = null;
+});
+
+it('should allow us to pass in a dom selector as a fallbackFocus', () => {
+  const modal = mount(<AbstractModalExample fallbackFocus="#focusable-element" />);
```
emilyrohrbough: Whats different about this test to the previous test?

packages/terra-abstract-modal/src/AbstractModal.jsx
```diff
@@ -31,6 +31,15 @@ const propTypes = {
    * If set to true, the modal will close when a mouseclick is triggered outside the modal.
    */
   closeOnOutsideClick: PropTypes.bool,
+  /**
+   * Element to fallback focus on if the FocusTrap can not find any focusable elements. Valid values are a valid
+   * dom selector string, a dom element, or a function that returns a dom element.
```
emilyrohrbough: We should specify what types of selectors are supported. I know xpath selectors are not supported in IE out-of-the-box.

emilyrohrbough: Are there concerns on what happens if selectors are not unique enough? Should this be limited to data attributes? I wonder if we should allow dom selector at all

packages/terra-abstract-modal/src/terra-dev-site/test/abstract-modal/AbstractModalFallbackFocus.test.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+  render() {
+    return (
+      <div>
+        <AbstractModal
+          ariaLabel="Modal disable close on outside click"
+          isOpen={this.state.isOpen}
+          closeOnOutsideClick
+          onRequestClose={this.handleCloseModal}
+          fallbackFocus={() => (this.focusTrapRef.current)}
```
emilyrohrbough: Would be valuable to add an test example for each prop type? String, function and node reference?

packages/terra-abstract-modal/tests/wdio/abstract-modal-spec.js
```diff
@@ -115,6 +115,16 @@ describe('Abstract Modal', () => {
...
   });
 
+  describe('Custom Focusable Fallback', () => {
+    before(() => browser.url('/#/raw/tests/terra-abstract-modal/abstract-modal/abstract-modal-fallback-focus'));
+
+    it('clicks to open modal', () => {
+      browser.click('button');
+    });
+
+    Terra.should.matchScreenshot({ selector: 'div[role="dialog"]' });
```
emilyrohrbough: Is there a way to determine that the focus fallback element actually has focus? Webdriver.io has this command: http://webdriver.io/api/state/hasFocus.html

packages/terra-abstract-modal/src/AbstractModal.jsx
JakeLaCombe: I'll document it as dom supported selectors. I don't want to get too involved with what browsers support which selectors, but I have no problem specifying a caution noting that you should ensure the browser selector is cross platform.

JakeLaCombe: I'd also like to keep the support, as it's the same prop types listed here. https://github.com/davidtheclark/focus-trap#usage

packages/terra-abstract-modal/tests/wdio/abstract-modal-spec.js
JakeLaCombe: I'm not sure. This is mostly iOS this issue is appearing up, as that doesn't focus on the modal like it does on desktop chrome.

JakeLaCombe:
I'm going to remove the dom element prop type. I'm not finding a good way to use that prop, as elements are not created until the modal is open, meaning it's not possible to pass in a dom element that is a part of the modal unless it's an id reference or a function returning the dom node.

packages/terra-abstract-modal/tests/wdio/abstract-modal-spec.js
JakeLaCombe: It worked. https://github.com/cerner/terra-core/pull/2112/commits/66602e5d456163d54d146893b281b7a046ee74b3

packages/terra-abstract-modal/src/terra-dev-site/test/abstract-modal/AbstractModalFallbackFocus.test.jsx
JakeLaCombe: Added the tests.

packages/terra-abstract-modal/src/AbstractModal.jsx
JakeLaCombe: Kept the dom selector with documentation, and the function. Had to drop the dom element

packages/terra-abstract-modal/tests/jest/AbstractModal.test.jsx
JakeLaCombe: Removed it.

PR #2135, opened 2018-12-21
Adding support for private prop suppression to PropsTable

tbiethman:
### Summary
As we consume more and more Context-based components, we're running into scenarios where a component may receive a prop automatically. In these cases, it's a little weird to expose those props in our consumer-facing documentation when consumers should never be providing them manually.

Please chime in if you have better ideas for marking a prop as private.


packages/terra-props-table/src/terra-dev-site/test/props-table-test/MockPrivatePropsComponent.jsx
```diff
@@ -0,0 +1,24 @@
...
+// These props are used to test react-doc-gen, but they only need to be mocked
+/* eslint-disable react/no-unused-prop-types */
+const propTypes = {
+  /**
+   * An example of type string markup.
+   * `This component contains a very long section of code in it's description column. This component contains a very long section of code in it's description column.`
+   */
+  string: PropTypes.string.isRequired,
+  /**
+   * [Private] A private prop that should not show up in the rendered markup.
```
tbiethman: This is what a private prop would look like in our other components.

StephenEsser: This looks clean. It's a little uglier but JSDoc has a tag for private variables. 

http://usejsdoc.org/tags-private.html

```
/**
 * @private
 */
privateProp: PropTypes.string,
```

bjankord: +1 for `@private`

PR #2138, opened 2019-01-01
fixes hold and scroll issue of form-select in IE #1863

### LINKED ISSUE #1863: Hold and scroll through options doesn't work in Select box in IE 11 and IE 10
# Bug Report
Scroll bar doesn't work let a user scroll through select options in select field.
## Description

I am using this component - https://github.com/cerner/terra-core/blob/728a5106f1e3fb113b3956cefeca996468b0ffb8/packages/terra-form-field/src/Field.jsx - This works fine in Chrome, however it doesn't work as expected with IE 11 and IE 10 browsers. Also, this doesn't work inside Powerchart.

Scrolling through select field with options with a mouse with a wheel lets a user scroll through select options. However, when trying to hold the scroll bar and drag it down makes the select options disappear. I am attaching a video recording of this.
## Steps to Reproduce

1. Open an application that has **field** react component.
2. Have many select options in it.
3. Click to open the select field.
4. Try to scroll down by holding the scroll bar

[Terra Form Field Issue.pptx](https://github.com/cerner/terra-core/files/2380022/Terra.Form.Field.Issue.pptx)

## Expected Behavior

User should be able to click and hold the scroll bar on select field and navigate through all the select options.
## Possible Solution

NA
## Environment

* Component Name and Version: Terra Form Select - https://terra-ui.herokuapp.com/#/components/terra-form-select/form-select/select
* Browser Name and Version: IE 10 and IE 11
* Node/npm Version: [e.g. Node 8/npm 5]  Node 8 and npm 6
* Webpack Version: 3

supreethmr:
### Summary
Prevent terra-form-select from closing onClick of scroll bar in IE.
Resolves #1863

### Additional Details

Hold and Scroll issue exist only on IE. When ever user tries to hold and scroll or try to click on arrow icons of scrollbar on form-select-dropdown, dropdown window was getting disappear.

IE was not able to recognize scroll bar as part of form-select-dropdown div hence it was removing focus from the component on click of scroll bar and functionality has been written to hide the dropdown when it loses focus by calling close dropdown on blur event.

To prevent IE from removing the focus from component on click of scroll bar, below Condition has been added to see on onBlur event whether dropdown is still active element or any part of dropdown is active before calling close drop-down method in onBlur event.

` if (this.dropdown && (this.dropdown === document.activeElement && 
    this.dropdown.contains(document.activeElement))) {
       this.focus();
       return;
    } `

Onblur event of form-select-dropdown will fire only when it loses focus on all other browsers hence this issue was not seen in any of the browsers like chrome, firefox, edge and safari. 

Only IE was firing Onblur event on click of scrollbar of dropdown hence added check on Onblur event to prevent dropdown from getting hidden.


package.json
```diff
@@ -93,5 +93,8 @@
     "terra-dev-site": "^2.0.0",
     "terra-toolkit": "^4.5.0",
     "webpack-merge": "^4.1.2"
+  },
+  "dependencies": {
+    "eslint-plugin-react": "7.11.0"
```
supreethmr: Updated eslint-config-terra to latest version 2.1.0 since 2.0.0 was giving error on npm run lint:js. 
Added dependencies eslint-plugin-react : 7.11.0 explicitly since by default it was referring to 7.12.0 which was leading to below error : 
`AssertionError [ERR_ASSERTION]: Node must be provided when reporting error if location is not provided
    at assertValidNodeInfo (/Users/sm051274/Documents/GitHub/terra-core/node_modules/eslint/lib/util/report-translator.js:96:9)`

emilyrohrbough: I have a PR for this to lock down the eslint-plugin-react version in the eslint-config-terra module: https://github.com/cerner/eslint-config-terra/pull/19

packages/terra-form-select/src/_Frame.jsx
```diff
@@ -228,6 +228,12 @@ class Frame extends React.Component {
    * Handles the blur event.
    */
   handleBlur(event) {
+    // The check for dropdown.contains(activeElement) is necessary to prevent IE11's from closing dropdown on click of scrollbar in certain contexts.
+    if (this.dropdown && (this.dropdown === document.activeElement && this.dropdown.contains(document.activeElement))) {
+      this.focus();
```
StephenEsser: `this.focus()` throws an error in IE. Focus will need to be called on a ref to the appropriate element. ( `this.select` or `this.input` depending on the variant type )

One concern I have about this approach is the focus callbacks. `onFocus` is a first class prop on the Select. I don't think we'll want the users onFocus callback fired with this approach. 

```<Select onFocus={() => { console.log('focus'); }} />```

File: package.json (hunk 1 above)
Comment (bjankord) [reply package.json]: Just released estlint-config-terra v2.2.0 which has the change @emilyrohrbough mentioned. We can switch this to `^2.2.0`.

Comment (supreethmr) [reply package.json]: Updated version with ^2.2.0 : https://github.com/cerner/terra-core/pull/2138/files#diff-32052204d6a35b2810d45b5391d555ed

File: packages/terra-form-select/src/_Frame.jsx [hunk 3]
```diff
@@ -243,12 +248,12 @@ class Frame extends React.Component {
...
-    if (!this.state.isFocused) {
-      this.setState({ isFocused: true });
+    if (this.props.onFocus && !this.state.isFocused) {
+      this.props.onFocus(event);
     }
 
-    if (this.props.onFocus) {
-      this.props.onFocus(event);
+    if (!this.state.isFocused) {
+      this.setState({ isFocused: true });
```
supreethmr: This code changes were made to prevent callback to users onFocus that was firing after onBlur event. 

Since isFocused state variable was meant to check Focus state of the control used it has condition in handleFocus() before doing callback to [users Onfocus()|https://github.com/cerner/terra-core/blob/5864418275108bfa739d907cd1790dad44514def/packages/terra-form-select/src/_Frame.jsx#L252]. 

IsFocus state was set to true on openDropdown() which will be invoked before invocation of handleFocus. Since it was set to true before execution of handleFoucs() condition (!this.state.isFocused) will always fails and prevents even the initial callback to users onFocus() which is supposed to fire when control gets Focus. 

Hence initializing isFocused to false on openDropDown and setting to true in handleFocus after Focus callback has been made to users code. 

Since isFocused is set to true once after initial callback has been it restricts if any attempt made to callback Onfocus again.

packages/terra-form-select/CHANGELOG.md
```diff
@@ -3,7 +3,10 @@ ChangeLog
 
 Unreleased
 ----------
+### Fixed
+* Fixed Dropdown closing issue on Hold and Scroll in IE
 
+------------------
```
StephenEsser: This divider can be removed, it adds an extra border to the generated markdown file

supreethmr: divider has been removed.

ShettyAkarsh:
Hi All,

Requesting you guys to review this PR.

packages/terra-form-select/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ ChangeLog
 
 Unreleased
 ----------
+### Fixed
+* Fixed Dropdown closing issue on Hold and Scroll in IE
+=======
```
StephenEsser: These === may be a result of resolving conflicts. They are appearing in the generated readme and should be removed.

image

package.json
StephenEsser: Is the version change still necessary for this PR?

packages/terra-form-select/CHANGELOG.md
supreethmr: removed.

package.json
StephenEsser: cc: @supreethmr

supreethmr: it's not need. It was added to pass travis build previously.  
Reverted back to ^2.0.0 :  https://github.com/cerner/terra-core/pull/2138/commits/6ee41f00060f7fb20696ea4c157433122b468450

supreethmr:
Requesting to please review the code changes for Issue #1863.

Issue #1891, opened 2018-09-18
[terra-responsive-element] Issue with repetitive resizing


JS062512:
# Bug Report

## Description
When resizing a page with multiple view types, such as different display types for small and large displays, there is a range of widths that can cause the formatting to become inconsistent. While within this width range the object will switch between view types rapidly. Once the view loads, the page seems to update and decide it should switch views, after switching views it then updates again and determines it should switch back to the previous view - creating an infinite loop.
This is a video showing one such page stuck in a loop of switching between two views [Video of problem](https://gfycat.com/TotalSaneGreyhounddog)

## Steps to Reproduce
1. Slowly resize the width of a page that has a compressed and non-compressed view
2. Once the form switches views stop resizing and see if the view switches once loaded
3. If the view does not switch resize until you go back to the previous view
(As we are unsure what combination of elements cause this issue it may not work on all pages that use terra-responsive-element)

## Additional Context / Screenshots
Our code uses terra-responsive-element and uses terra-table for the expanded version and terra-SingleSelectList for the collapsed version.
For resizing we use default and 'small'

## Expected Behavior
Resizing the width to change view types would make this change a single time instead of creating a loop of switching views.

## Possible Solution
A possible source of the issue may be how the terra-responsive-element is getting the width for the terra-table and for the terra-SingleSelectList may be inconsistent, causing the view type to switch between the two. It also may be a problem with the scroll bar. Watching the video it seems the expanded version has a scroll bar that includes the header while the collapse version does not.

## Environment

* Component Name and Version: terra-responsive-element@3.9.1, terra-table@2.23.1, terra-list@2.23.1
* Browser Name and Version:
* Node/npm Version: [e.g. Node 8/npm 5]
* Webpack Version:
* Operating System and version (desktop or mobile):

bjankord:
@JS062512 Can we get a reduced test case to help us debug this further.

tbiethman:
I worked with @JS062512 to dig into this a little more. The issue is indeed caused by the scrollbar, particularly when the scrollbar is rendered as always visible (due to a mouse being detected or other settings being enabled).

I'm not sure how tenable it will be to have the ResponsiveElement account for the presence of the scrollbar in its width-detection logic. The various browser implementations make that tricky, especially so given that a lot of that logic is itself encapsulated within the ResizeObserver.  

[...]

Issue #2048, opened 2018-11-13
[terra-responsive-element] Consume terra-breakpoints and update size props


tbiethman:
# Feature Request

## Description
The ResponsiveElement has historically defined its own breakpoint values. However, the terra-breakpoints package was recently created to contain any and all things breakpoint related. The ResponsiveElement should consume breakpoint values from that package and remove any files that are defining breakpoint values internally.

The creation of terra-breakpoints also saw the redefinition of some of our breakpoint names. The ResponsiveElement's props should be updated to reflect those new values.

These will be breaking changes, but they are necessary to bring the component into alignment with our breakpoint strategy.

PR #2141, opened 2019-01-03
Fix for Repetitive Resizing issue when scrollbar appears between two breakpoints  #1891

ShettyAkarsh:
### Summary

Resolves the Repetitive resizing issue when scrollbar appears between two breakpoints while the browser window is  minimised and user tries to change the browser window width.

The fix is to pass the width of the parent container along with its scrollbar width to the handle resize method.


packages/terra-responsive-element/src/ResponsiveElement.jsx
```diff
@@ -61,7 +61,7 @@ class ResponsiveElement extends React.Component {
       this.resizeObserver = new ResizeObserver((entries) => {
         this.animationFrameID = window.requestAnimationFrame(() => {
           this.animationFrameID = null;
-          this.handleResize(entries[0].contentRect.width);
+          this.handleResize(entries[0].target.getBoundingClientRect().width);
```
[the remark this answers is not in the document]
tbiethman: I'm not sure if getting the parentNode of the target is what we want. The target should already be the container for the responsive elements, and that container is what is what is (conditionally) rendering the scrollbar. If we go a level higher, we're in uncharted waters.

I wouldn't have expected the width from getBoundingClientRect to differ from the offsetWidth of the same element.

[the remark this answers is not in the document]
ShettyAkarsh: @StephenEsser In above code example there is outer div for responsive element hence the target value is set as <div> and the target.parentnode is set as parentcontainer. so we are getting the issue of repetitive resizing using my initial code. if we add one more outerdiv to the current div the issue still exists with newly suggested code as the target.parentnode will be the outermost div and target will be innerdiv.

Observerresize.target will always hold the immediate parent control of responsiveelement. if the div elements are removed **entries[0].target.getBoundingClientRect().width** works fine as the target is always set to immediate parent of the responsive element. without any outer div elements the target.parentnode value will be set to parent container of responsive element parent container.

i even tried creating sample react app consuming the responsive element without any blank outer div elements. in this case the target value is set to root div element where as target.parentnode is html body.

so my question here is will there be a real time scenario where blank div be there around responsive element ?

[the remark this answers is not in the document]
ShettyAkarsh: @bjankord , @StephenEsser request you guys to review this PR. please let me know if i need to test anything else.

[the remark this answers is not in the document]
StephenEsser: Is there a scenario where the code change above prevents an infinite resize?

@tbiethman @bjankord and I spoke offline and there is not currently a known solution that fixes the infinite resizing in all cases and it's possible there isn't one. The infinite resizing is caused by the available space fluctuating by a set amount of pixels that pushes each resize into the other breakpoint threshold.  

Unless the solution above fixes any of the infinite resizing use-cases I would vote to close this PR and replace it with a PR that documents some of the caveats and recommended use cases of the ResponsiveElement. Specifically the caveat that if any ancestor's overflow can be affected by the contents within the ResponsiveElement it can result in an infinite resize. 

@tbiethman  @bjankord Do you guys have any thoughts on this?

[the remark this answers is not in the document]
ShettyAkarsh: @StephenEsser Regarding your question "Is there a scenario where the code change above prevents an infinite resize?"

Yes, the above code change prevents the infinite resize during the scrollbar display while shifting to next immediate breakpoint of responsive element. the scrollbar is due to parent control which has the overflow property set to 'Auto'. 

the previous code **'entries[0].contentRect.width'** used to not take scrollbar width into account which would make the breakpoint of responsive element shift to one level below. while it should actually not shift the breakpoints. hence using parent control width which includes scrollbar width also.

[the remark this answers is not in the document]
StephenEsser: I believe the width returned from `entries[0].target.getBoundingClientRect().width` also includes the padding and borders. The ResponsiveElement is currently implemented to shift between breakpoints based on the available space for the content. There are a few notes about this on the [ReadMe](https://github.com/cerner/terra-core/blob/master/packages/terra-responsive-element/docs/README.md):

> The ResponsiveElement observes the available width. Meaning padding, borders, and margins set on the parent are not included in the calculation. Modifying box-sizing will impact how these properties are used in an elements width calculation. For more details see: https://css-tricks.com/almanac/properties/b/box-sizing/

I think the ResponsiveElement should continue to honor the available space for calculations and not the raw width of the parent including paddings / borders. 

Here's a short example.

```
<div id="parent" style="padding: 20px; border: 5px solid black; width: 100px;">
  <div>ResponsiveElement</div>
</div>
```

In this example the bounding rect width would report 150px of space, but there is actually only 100px of available space for the ResponsiveElement.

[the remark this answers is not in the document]
bjankord: Had a chance to look over this PR and the related issue in more detail. @StephenEsser created a [branch with a reduced test case](https://github.com/cerner/terra-core/tree/InfiniteResizeExamples) which was helpful to reason about the issue. I'm okay with updating the responsive element behavior to report the **raw width of the parent including paddings / borders** instead of the **available space** as it was prior, but I would consider this to be a non-passive change.

I'd like to see some documentation updates noting the behavior change as the docs currently mention reporting the available space.

 e.g. Before this PR, if a div has `width: 120px` and `padding: 10px`, it would report `100px` of available space. With the change in this PR, the responsive element would now report `120px` of available space even though there is `10px` padding on the sides. I think this concern can be mitigated with some documentation.

In regards to this being a non-passive change, I'd like to hold off on merging this PR until we are ready to incorporate [the other non-passive changes we have planned in this issue](https://github.com/cerner/terra-ui/issues/117). That work will likely happen sometime in Q1.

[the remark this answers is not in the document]
ShettyAkarsh: Updated the documentation.

bjankord:
In addition to this PR, we'll also want to complete the work for https://github.com/cerner/terra-core/issues/2048 in a separate PR. The goal will be to release the breaking changes in this PR and the breaking changes for https://github.com/cerner/terra-core/issues/2048 in 1 major release version.

packages/terra-responsive-element/docs/README.md
```diff
@@ -7,7 +7,7 @@ The viewport can be set to the immediate parent or window.
 
 It is not necessary to set each breakpoint. If a breakpoint is not set the component will look at each successive smaller breakpoint until one is found.
 
-The ResponsiveElement observes the available width. Meaning padding, borders, and margins set on the parent are not included in the calculation. Modifying box-sizing will impact how these properties are used in an elements width calculation. For more details see: https://css-tricks.com/almanac/properties/b/box-sizing/
+The ResponsiveElement observes the total width. Meaning padding, borders except the margins set on the parent are included in the calculation. Modifying box-sizing will impact how these properties are used in an elements width calculation. For more details see: https://css-tricks.com/almanac/properties/b/box-sizing/
```
bjankord: Do modifying the box-sizing still impact how these properties are used in an elements width calculation? cc/ @StephenEsser

ShettyAkarsh: Yes, we need this to be mentioned in the documentation. as the default value of box-sizing property for the elements will be ' 'content-box' this includes the content, padding and the border in width calculation. if we change box-sizing value to 'border-box' then only the content width will get retrieved without border and padding.

emilyrohrbough:
@ShettyAkarsh We need to add an upgrade guide for these breaking changes

ShettyAkarsh:
> @ShettyAkarsh We need to update the upgrade guide for these breaking changes for this MVB

Changes updated for upgrade guide.

packages/terra-responsive-element/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ Changelog
 Unreleased
 ----------
 
+### Fixed
+* Repititive Resizing when scrollbars appear between two breakpoints
+=======
```
emilyrohrbough: whoops, remove `=====` from merge conflict

ShettyAkarsh: done.

packages/terra-responsive-element/CHANGELOG.md
```diff
@@ -24,6 +24,9 @@ Unreleased
 ### Changed
 * Made required updates to consume terra-toolkit v5 and terra-dev-site v5
 
+### Fixed
+* Repititive Resizing when scrollbars appear between two breakpoints
```
ryanthemanuel: Repetitive

package.json
```diff
@@ -79,6 +79,9 @@
     "enzyme-adapter-react-16": "^1.1.1",
     "enzyme-to-json": "^3.2.2",
     "eslint": "^5.0.0",
+    "eslint-config-terra": "^2.2.0",
+    "gh-pages": "^1.1.0",
```
bjankord: This line can be removed, we have gh-pages defined on line 86.

package.json
```diff
@@ -79,6 +79,9 @@
...
+    "glob": "^7.1.1",
```
bjankord: This line can be removed, we have glob defined on line 87

package.json
```diff
@@ -79,6 +79,9 @@
     "enzyme-adapter-react-16": "^1.1.1",
     "enzyme-to-json": "^3.2.2",
     "eslint": "^5.0.0",
+    "eslint-config-terra": "^2.2.0",
```
bjankord: We can keep this line and remove the eslint-config-terra defined on line 85

packages/terra-responsive-element/CHANGELOG.md
StephenEsser: We'll also want to move this changelog entry into the unreleased section

packages/terra-responsive-element/docs/README.md
```diff
@@ -6,6 +6,8 @@ The Responsive Element can be set to be responsive to the parent of the componen
 
 ## Additional Notes
 
+The ResponsiveElement observes the total width. Meaning padding, borders except the margins set on the parent are included in the calculation. Modifying box-sizing will impact how these properties are used in an elements width calculation. For more details see: https://css-tricks.com/almanac/properties/b/box-sizing/
+
 For consistency, breakpoint ranges are inherited from [terra-breakpoints](https://engineering.cerner.com/terra-ui/#/components/terra-breakpoints/breakpoints/about).
 
 The Responsive Element observes the available width. Meaning padding, borders, and margins set on the parent are not included in the calculation. Modifying box-sizing will impact how these properties are used in an elements width calculation. For more details see: <https://css-tricks.com/almanac/properties/b/box-sizing/>
```
StephenEsser: We'll need to delete this section, it's the opposite of the newly added section

packages/terra-responsive-element/docs/README.md
```diff
@@ -6,6 +6,8 @@ The Responsive Element can be set to be responsive to the parent of the componen
 
 ## Additional Notes
 
+The ResponsiveElement observes the total width. Meaning padding, borders except the margins set on the parent are included in the calculation. Modifying box-sizing will impact how these properties are used in an elements width calculation. For more details see: https://css-tricks.com/almanac/properties/b/box-sizing/
```
StephenEsser: We might clean this sentence up, it is a little confusing as it is. Maybe something along the lines of:

The ResponsiveElement observes the total width of the bound container. Padding and borders are included in the calculation to determine the available width. Margins are not included. For breakpoint accuracy it is not recommended to set padding on the parent container when responsiveTo is set to parent.

bjankord:
Merging down as is to get theme repos unblocked. @StephenEsser will work on the code review comments he has in a new PR.

PR #2149, opened 2019-01-07
Correcting Errors from Upcoming 'eslint-plugin-react' Version Lift

amichaelparker:
### Summary
Ref: [this issue](https://github.com/cerner/eslint-config-terra/issues/21). Unlocking the eslint-plugin-react version in eslint-config-terra causes lint errors due to rule changes in the plugin, particularly parentheses formatting around JSX elements (react/jsx-wrap-multilines) and some prop type validation failures (react/prop-types). This PR corrects those errors ahead of the plugin version change.


packages/terra-date-picker/src/DateInput.jsx
```diff
@@ -43,6 +43,10 @@ const propTypes = {
    * A callback function to request focus from the containing component (e.g. modal).
    */
   requestFocus: PropTypes.func,
+  /**
+   * @private Internal prop for showing date picker.
+   */
+  shouldShowPicker: PropTypes.bool,
```
amichaelparker: This comment may need changing, noting it here.

packages/terra-form-field/src/Field.jsx
```diff
@@ -66,6 +66,11 @@ const propTypes = {
    * Whether or not to append the 'optional' label to a non-required field label.
    */
   showOptional: PropTypes.bool,
+  /**
+   *  Provides first class prop for custom inline styles
+   */
+  // eslint-disable-next-line react/forbid-prop-types
+  style: PropTypes.object,
```
amichaelparker: Also this one.

packages/terra-date-picker/src/DateInput.jsx
emilyrohrbough: What changing would it need?

amichaelparker: Not sure! :) Usually my comments can be worded better, so I tagged them here for easier input.

emilyrohrbough: Ah gottchya! Maybe add for test / doc examples??

Issue #138, opened 2017-03-16
Update examples to use icons


Matt-Butler:
We have a few examples that have been using svgs. We need to update these examples to use the icon component.

For example:
https://github.com/cerner/terra-ui/blob/master/packages/terra-badge/examples/BadgeIcon.jsx#L4

PR #2162, opened 2019-01-12
Added test for accessibilty check

AlexBrizi:
### Summary
Added a node-based test for checking for accessibility accommodations on the Terra doc site.

Resolves [#138](https://github.com/cerner/terra-ui/issues/138)


bjankord:
This should be added to the terra-ui repo.

PR #2163, opened 2019-01-14
Screen reader no results

### LINKED ISSUE #1905: No matching results for string in select search variant is not communicated via VoiceOver on iOS
# Bug Report

## Description
When typing in a search value where there are no results, the No matching results for ${string} is not announced to the user when it is displayed.

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-core/#/tests/terra-form-select/form-select/controlled-search
2. Type in random string, "asdf"
3. No matching results for asdf is displayed visually but not communicated via VO.

## Expected Behavior
The result of a search should be communicated to screen reader.

## Possible Solution
Might look into using `aria-live` to [communicate search results on the fly](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)

## Environment

* Component Name and Version: terra-select search variant
* Browser Name and Version: Mobile Safari iOS 12 w/ VoiceOver

JakeLaCombe:
### Summary

Resolves #1905

In order to read the no results found while typing, we need to make use of a live region that sends text updates to the user's screen reader.

Verification Steps:

Go to https://terra-core-deployed-pr-2163.herokuapp.com/#/components/terra-form-select/form-select/select, go to the search box, and type in a value that doesn't exist.


packages/terra-form-select/src/_Frame.jsx
```diff
@@ -374,6 +390,15 @@ class Frame extends React.Component {
       customProps.className,
     ]);
 
+    const { searchValue } = this.state;
+
+    let ariaLiveText = '';
+    const children = MenuUtil.filter(options, searchValue, optionFilter);
```
StephenEsser: This might be worth looking into a little more. This will cause a double filter each render for searchable variants. The first filter will happen here, the second filter will happen in the Menu itself.

JakeLaCombe: I did see both filters here, and the first thing I was thinking about is that if the list isn't that large, I don't think double filtering is a large concern. The trick to aria-live is that if it's not already mounted on the screen before it's used, then it won't send the user's results to a screen reader.

JakeLaCombe: I'll see what it takes to do the filtering at the frame level, that pass it on through to the dropdown.

StephenEsser: The `Frame` was originally designed so that it would be possible for users to create a custom wrapper and render content into the dropdown. Similar to the implementation in this package where the `Select.jsx` file is a wrapper around `Frame.jsx`. The `Select` uses the `DropdownMenu` as the content.

I am not aware of any consumers using the Frame this way currently and discussions may lead to suggest they shouldn't anyway. But if we move any of the Menu logic into Frame it will break the generic API the Frame was built on top of. 

It would be worth talking through, if we want to abandon the API of the Frame it should be dissolved. The `Select.jsx` and `Frame.jsx` would merge into a single file.

JakeLaCombe: Here's one option we can do.

When we mount the select box, we can create a new node that is independent of react, and append that to the dom. We can give it visually hidden styles so that it does not display at all on the screen. From there, when we do the no results filter in the menu, we can use vanillaJS to update the results of the live region with the new text. That will cause the screen reader to update results

JakeLaCombe: There is also the consideration of making an aria-live region for terra-base, and provide an api through terra base that allows for developers to communicate results like that to a screen reader

JakeLaCombe:
New solution posted! See if that works for you.

packages/terra-form-select/src/_Menu.jsx
```diff
@@ -18,10 +20,18 @@ const propTypes = {
...
+   */
+  intl: PropTypes.object,
   /**
    * Content to display when no results are found.
    */
   noResultContent: PropTypes.node,
+  /**
+   * String to send to screen readers to read out no results found.
+   */
+  noResultScreenReaderText: PropTypes.string,
```
StephenEsser: Is this used anywhere?

packages/terra-form-select/src/_Menu.jsx
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable react/forbid-prop-types */
```
JakeLaCombe: This is for the intl prop. To remove this lint, we would have to make terra-base a required dependency of terra-form-select.

packages/terra-form-select/src/_Frame.module.scss
```diff
@@ -287,4 +287,16 @@
     background-size: contain;
     display: inline-block; // IE 10
   }
+
+  .visually-hidden-component {
```
bjankord: Let's use the https://www.npmjs.com/package/terra-visually-hidden-text component instead

packages/terra-form-select/src/_Menu.jsx
```diff
@@ -18,10 +20,18 @@ const propTypes = {
    * The content of the menu.
    */
   children: PropTypes.node,
+  /**
```
bjankord: We'll want to add `@private` to the description for this prop, it doesn't need to be publically documented and this string will prevent it from being rendered on the docsite.

```
/**
 * @private Translation object to use for interpreting strings
 */
```

packages/terra-form-select/src/_Menu.jsx
bjankord: It doesn't look like it is.

packages/terra-form-select/src/_Frame.module.scss
JakeLaCombe: I actually had trouble using that component because it's a stateless one. I'm not able to access refs for those, so if we wanted to do that, we would have to convert over terra-visually-hidden-text over to one. I'm not completely opposed to the idea though, as I don't see a large number of technical concerns with it.

JakeLaCombe:
Try my latest changes. The debouncing is working pretty well

packages/terra-form-select/src/_Frame.jsx
bjankord: I like the idea of creating a new node that is independent of react, and append that to the dom when we mount the select box for the aria-live text. I also think we should look into corporating that into terra-base as it is a good candidate for handling global concerns which I think aria-live makes sense to handle at a global app level.

packages/terra-form-select/src/_Menu.jsx
JakeLaCombe: Done. https://github.com/cerner/terra-core/pull/2163/commits/207f487d49ce1516167e38a806510a7584848084

packages/terra-form-select/src/_Menu.jsx
JakeLaCombe: https://github.com/cerner/terra-core/pull/2163/commits/be2f020b6c4484501b6d68bbde493987529c2899

packages/terra-form-select/src/_Menu.jsx
JakeLaCombe: Not even needed anymore. https://github.com/cerner/terra-core/pull/2163/commits/c80a4bd1bd23db67fc55748bccc0e1bde34e9ce3

packages/terra-form-select/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ Unreleased
 ### Changed
 * Updated theme variables
 
+### Added
+* Create dropdown that announces no search results to screen reader
```
StephenEsser: It looks like the change above has already been released, this entry should be moved into the unreleased section.

mmalaker:
Functional verification completed. The select search example now enables no results text to be read by screen readers. Note that there are still accessibility issues with the example page itself that interfere with keyboard navigation through the page examples. Site accessibility will be addressed separately and is an ongoing effort.

PR #2164, opened 2019-01-14
Allow for screen readers to output disabled select boxes to their users

### LINKED ISSUE #1906: Disabled select does not communicate disabled state to screen reader with using VoiceOver on iOS
# Bug Report

## Description
When interacting with disabled select, VoiceOver doesn't provide any communication to the user that the select is disabled. VoiceOver on iOS will usually say "Dimmed" after interacting with an input that is disabled.

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-core/#/tests/terra-form-select/form-select/disabled
2. Interact with Select component
3. VO will read, "Select a color" without indicated the input is disabled. Though user is unable to select a color.

## Expected Behavior
If select is disabled, it should communicate that information to screen reader.

## Possible Solution
Likely need to take a look at the markup structure and rework some part of it.

## Environment

* Component Name and Version: terra-select
* Browser Name and Version: Mobile Safari 12 on iOS w/ VoiceOver

JakeLaCombe:
### Summary
Resolves #1906

Voiceover on iOS has difficulty picking up that a select element was disabled. In addition, JAWS had some difficulties as well in picking up a disabled state. This pull request resolves that. The main issue is that when role is set to combobox, that is taking presidence over the disabled message when read to a screen reader. Setting the role to undefined for a disabled select element resolves this issue.

### Deployment Link
https://terra-core-deployed-pr-2164.herokuapp.com/#/components/terra-form-select/form-select/select


mmalaker:
Functional verification completed. The disabled select box examples now display and function as expected with screen readers.

Issue #2090, opened 2018-12-04
[terra-form-select] Top corner radius on select is not round.


benbcai:
# Bug Report

## Description
When the select drop-down is opened _above_ the select input field and an option is selected, the top-left and top-right corner radiuses do not change back to a rounded radius of 3px. 

 is correctly styled from a rounded radius to a square radius ([3px to 0px](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/_Frame.module.scss#L55-L56)). When an option is selected, the dropdown will close. However, the top-left and top-right corner radius remains square (0px). It should be changed back to a rounded corner (3px). 

## Steps to Reproduce
1. Go to the examples on https://engineering.cerner.com/terra-ui/#/components/terra-form-select/form-select/select
2. Scroll to a select input such that is is near the bottom of the browser to force the dropdown to display above the select.
3. Click on the select input.
4. Select an option.
5. Note that the top-left and top-right corner radiuses do not change to a rounded radius of 3px.
6. Click any where on the page and the corner radiuses then rounded.

## Additional Context / Screenshots
![form-select corner radius](https://user-images.githubusercontent.com/22618655/49472084-2aeb6f00-f7d4-11e8-9692-c2cdb14c27b4.gif)

## Expected Behavior
After an option is selected, the corner radius should change back to rounded just like if the dropdown was presented below the select input.

## Possible Solution
In the [handleSelect](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/_Frame.jsx#L324-L334) method in Frame.jsx, the isAbove state needs to be set to false (if it doesn't allow multiple selections) so that the `is-above` classes is removed from the markup once an option is selected.

## Environment
* Component Name and Version: terra-form-select (4.25.1)
* Browser Name and Version: Any
* Node/npm Version: [e.g. Node 8/npm 5] Node 8.12.0; npm 6.4.1
* Webpack Version:
* Operating System and version (desktop or mobile): Desktop

PR #2170, opened 2019-01-17
Fix Issue With Corner Styling in Form Select

amichaelparker:
### Summary
Resolves an issue with form-select where lists opened above the select box would not clear styling from the select box upper corners when an item is selected. Does not apply to multi-select lists.

Closes [#2090](https://github.com/cerner/terra-core/issues/2090)

### DIFF
--- a/packages/terra-form-select/src/_Frame.jsx
+++ b/packages/terra-form-select/src/_Frame.jsx
@@ -322,19 +322,17 @@ class Frame extends React.Component {
    * @param {ReactNode} option - The option that was selected.
    */
   handleSelect(value, option) {
+    const { isAbove } = this.state;
+    const isOpen = Util.allowsMultipleSelections(this.props);
+
     this.setState({
       searchValue: '',
       hasSearchChanged: false,
-      isOpen: Util.allowsMultipleSelections(this.props),
+      isOpen,
+      isAbove: isOpen ? isAbove : false,
     });
 
     if (this.props.onSelect) {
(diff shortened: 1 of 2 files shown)


packages/terra-form-select/src/_Frame.jsx
```diff
@@ -329,6 +329,12 @@ class Frame extends React.Component {
     });
 
     if (this.props.onSelect) {
+      if (!Util.allowsMultipleSelections(this.props)) {
```
StephenEsser: We should avoid calling setState twice within the same function / render lifecycle and pull this logic out of the logic block for `this.props.onSelect` which is an optional prop and shouldn't influence the outcome of isAbove.

isAbove should always be false if the drop down is closed, and not change if the dropdown remains open.

Here's an example of something that might work.

```
  handleSelect(value, option) {
    const { isAbove } = this.state;
    const isOpen = Util.allowsMultipleSelections(this.props);

    this.setState({
      searchValue: '',
      hasSearchChanged: false,
      isOpen,
      isAbove: isOpen ? isAbove : false,
    });

    if (this.props.onSelect) {
      this.props.onSelect(value, option);
    }
  }
```

amichaelparker: Definitely works better. I'll get this at some point, probably. :) - [41280d9](https://github.com/cerner/terra-core/pull/2170/commits/41280d9331e938f7d2b94d76bf58c9b495142fb6)

PR #2171, opened 2019-01-18
Custom properties scripts

bjankord:
### Summary
Resolves https://github.com/cerner/terra-ui/issues/140

* Add script from terra-toolkit to aggregate all CSS custom properties in a give glob path and convert it to JSON
* Use the JSON file to render CSS custom properties on the component example doc page
* Add script from terra-toolkit to spell check each word in CSS custom property 
* Add a custom dictionary to aid with spell check

THIS IS A WORK IN PROGRESS. 

I've only added theme variable documentation to the [terra-button page](https://terra-core-deployed-pr-2171.herokuapp.com/#/components/terra-button/button/button).


JakeLaCombe:
The themeable variables page is really long. This maybe out of scope, but how do you feel about putting the themeable variables in a dropdown of some sort that's expanded? I would also be fine with if the themeable variables is always the last item on the page.

bjankord:
@JakeLaCombe Yeah I think the variables would work better in an expandable content section.

packages/terra-doc-template/src/DocTemplate.jsx
```diff
@@ -127,6 +132,31 @@ const DocTemplate = ({
...
+                <tr>
+                  <th>Property Name</th>
+                  <th>Property Value</th>
+                </tr>
+              </thead>
+              <tbody>
+                {Object.keys(themeVariables).map(key => (
+                  <tr>
+                    <td>{key}</td>
+                    <td>{themeVariables[key]}</td>
```
yuderekyu: not a big deal, but for readability can we encase theme variable values in back-ticks?

bjankord:
Closing this based on team sync we had today. More info here: https://github.com/cerner/terra-ui/issues/140

PR #2172, opened 2019-01-18
Fixes issue #2074 DatetimePicker throws NPE if manually entered valid datetime is outside of allotted datetime range

### LINKED ISSUE #2074: [DateTimePicker] throws NPE if manually entered valid date is outside range of allotted dates
# Bug Report

## Description
If you manually enter a _valid_ date into the `Date` field that is outside the range allowed by a `minDate`, `maxDate` or any other type of date filtering then enter a time; an exception is thrown when attempting to format the time into the existing date. (which is because the existing date, while valid, does not exist due to it being outside the allotted dates)

## Steps to Reproduce

1. Navigate to the [`terra-date-picker component` page](https://engineering.cerner.com/terra-ui/#/components/terra-date-time-picker/date-time-picker/date-time-picker)
2. Open browser `dev tools` and navigate to console
3. Scroll down to an example that filters specific dates (I used `Min Max Selection Options`)
4. Manually enter a `date` outside the allotted dates (I used `11/11/1999`)
5. Enter a `time`
6. See console output `Uncaught TypeError: Cannot read property 'format' of null`

image

Null field is [`updatedDateTime`](https://github.com/cerner/terra-core/blob/master/packages/terra-date-time-picker/src/DateTimePicker.jsx#L245)

## Expected Behavior
Should not throw NPE.

## Possible Solution
Preferred: Add condition to `isValidDate` to ensure date is within the bounds of the allotted dates.
Quick and dirty: Check for null before using `updatedDateTime` 

## Environment

* Component Name and Version: terra-date-time-picker ^2.36.0
* Node/npm Version: v10.13.0/6.4.1
* Operating System and version (desktop or mobile): Desktop OS X High Sierra 10.13.6

## @ Mentions
I brought up the issue to @benbcai

ShettyAkarsh:
### Summary
Resolves the NPE error when user tries to enter time along with valid date outside of the allotted datetime range.

The fix is to validate whether the date entered is within the datetime range after the date value is validated.

Resolves #2074

### DIFF
--- a/packages/terra-date-time-picker/src/DateTimePicker.jsx
+++ b/packages/terra-date-time-picker/src/DateTimePicker.jsx
@@ -230,7 +230,7 @@ class DateTimePicker extends React.Component {
 
   handleTimeChange(event, time) {
     this.timeValue = time;
-    const validDate = DateTimeUtils.isValidDate(this.dateValue, this.state.dateFormat);
+    const validDate = DateTimeUtils.isValidDate(this.dateValue, this.state.dateFormat) && this.isDateTimeWithinRange(this.convertDateTimeStringToMomentObject(this.dateValue, this.timeValue));
     const validTime = DateTimeUtils.isValidTime(this.timeValue);
     const previousDateTime = this.state.dateTime ? this.state.dateTime.clone() : null;
 
@@ -305,13 +305,17 @@ class DateTimePicker extends React.Component {
   }
 
   validateDefaultDate() {
+    return this.isDateTimeWithinRange(this.state.dateTime);
+  }
+
+  isDateTimeWithinRange(newDateTime) {
     let isAcceptable = true;
 
-    if (DateUtil.isDateOutOfRange(this.state.dateTime, DateUtil.createSafeDate(this.props.minDateTime), DateUtil.createSafeDate(this.props.maxDateTime))) {
+    if (DateUtil.isDateOutOfRange(newDateTime, DateUtil.createSafeDate(this.props.minDateTime), DateUtil.createSafeDate(this.props.maxDateTime))) {
       isAcceptable = false;
     }
 
-    if (DateUtil.isDateExcluded(this.state.dateTime, this.props.excludeDates)) {
+    if (DateUtil.isDateExcluded(newDateTime, this.props.excludeDates)) {
       isAcceptable = false;
     }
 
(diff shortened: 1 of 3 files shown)


packages/terra-date-time-picker/src/DateTimePicker.jsx
```diff
@@ -346,6 +350,10 @@ class DateTimePicker extends React.Component {
     this.setState({ isTimeClarificationOpen: false });
   }
 
+  convertDateTimeStringToMomentObject(date, time) {
+    return DateTimeUtils.updateTime(DateUtil.createSafeDate(DateUtil.convertToISO8601(date, this.state.dateFormat)), time);
```
JakeLaCombe: The previous check just did createSafeDate by itself without the need for updateTime. How come that's needed here?

ShettyAkarsh: The string passed to the previous check had time along with the date in iSO format. so we had to only convert the string to moment object using createsafedate.  in our case we get only the date string from date control which i am converting into moment object using createsafedate and later appending time value retrieved from time control to the moment object using updatetime.

Here the MIN and MAX datetime will have time value along with the date. hence we have to use both date and time value entered to check whether it falls within the range.

JakeLaCombe: I'm still not seeing it though. maxDateTime and minDateTimes are already in ISO8601, and convertToISO8601 is converting the current date to that format, so I would definitely expect createSafeDate for both. However, since neither the maxDateTime or minDateTime need to use updateTime, I would expect the same to occur with this function. What happens if you leave of updateTime?

ShettyAkarsh: This new method convertDateTimeStringToMomentObject0 is only called for the dates entered by the user in handleTimeChange event. for MIN and MAX dates this method is not called.

emilyrohrbough:
FYI this component has been moved to terra-framework: https://github.com/cerner/terra-core/pull/2122

bjankord:
@ShettyAkarsh We've moved this package into the terra-framework repo to better manage dependency flow across the monorepos. Can you open a new PR with this change in terra-framework.

ShettyAkarsh:
> @ShettyAkarsh We've moved this package into the terra-framework repo to better manage dependency flow across the monorepos. Can you open a new PR with this change in terra-framework.

Sure, will log the new PR in terra-framework.

PR #2179, opened 2019-01-23
Add the ability for Screen Readers to read the list of options in terra-form-select.

### LINKED ISSUE #1887: [terra-form-select] - Options are not read by VoiceOver on iOS
# Bug Report

## Description
When focusing on the terra-select and trying to interact with it, VoiceOver on iOS skips reading the options. It seems there is no detection that options exist when using VO.

## Steps to Reproduce

1. Open up https://engineering.cerner.com/terra-ui/#/components/terra-form-select/form-select/select on iOS device
2. Tap on home button 3 times to turn on VoiceOver
3. Shift focus to Select (swipe right until terra-select is focused)
4. Double click to interact with select

## Expected Behavior
terra-select should communicate and interact with the select options when using VoiceOver on iOS.

## Environment

* Component Name and Version: terra-select v3.0.0+
* Browser Name and Version: Safari on iOS
* Webpack Version: v4

JakeLaCombe:
### Summary
Resolves #1887 

Screen Readers are having difficulty reading the selected options for terra-form-select. In order to resolve this, it is recommended to use the aria-activedescendent attribute on the input element, and switch the id as the user moves from one item to another. This will cause the screen reader to announce the text to the user.

To prevent double rendering, I passed a ref into the menu component, and updated the attribute whenever the current active element updated.


packages/terra-form-select/src/_Menu.jsx
```diff
@@ -128,15 +134,23 @@ class Menu extends React.Component {
     document.addEventListener('keydown', this.handleKeyDown);
   }
 
-  componentDidUpdate() {
+  componentDidUpdate(prevProps, prevState) {
     this.updateLiveRegion();
     this.scrollIntoView();
+
+    if (prevState.active !== this.state.active && this.props.searchInput) {
```
StephenEsser: What is the check for the searchInput for?

packages/terra-form-select/src/_Option.jsx
```diff
@@ -68,17 +68,23 @@ const Option = ({
   ]);
 
   return (
+    // Voiceover is having issues in a multiselect dropdown where once it reads the first
```
bjankord: Is this resolved or is this still an issue? If it is still an issue, we should log a GitHub issue for this and reference it in these comments.

bjankord:
Issue #1887 was for select options not being read by VoiceOver on iOS. In testing with Voice Over on iOS, I was unable to get the virtual indicator into the select dropdown.

On desktop, using Voice Over in Safari and using the arrow keys to cycle through the options, none of them were read aloud by Voice Over.

I think it would be good to sync up on scope of this and desired functionality with screen readers for this component and it's variations.

@StephenEsser shared some links he was referencing when originally working on select that I think would should look to when discussing desired functionality in regards to screen reader usage of this component:

* https://dequelabs.github.io/combobo/demo/
* https://a11y.nicolas-hoffmann.net/autocomplete-list/
* https://weboverhauls.github.io/demos/autosuggest/

packages/terra-form-select/src/_Option.jsx
```diff
@@ -68,17 +68,23 @@ const Option = ({
...
       role="option"
       disabled={disabled}
       className={optionClassNames}
-      aria-selected={isSelected}
+      aria-checked={isSelected}
       aria-disabled={disabled}
     >
       {(isCheckable || isAddOption) && <span className={cx('icon')} />}
       <span className={cx('display')}>{display}</span>
     </li>
```
StephenEsser: One thing to double check is whether or not these options are announced when they are contained within an OptGroup. When I initially investigated this issue nested lists had trouble with Voice Over.

packages/terra-form-select/src/_Frame.jsx
```diff
@@ -429,6 +445,7 @@ class Frame extends React.Component {
                  onDeselect,
                  optionFilter,
                  noResultContent,
+                 menuTriggerElement: variant === Variants.DEFAULT ? this.select : this.input,
```
StephenEsser: This name feels a little awkward to me. I don't have a great alternative but maybe something along the lines of `focusTarget`, `focusRegion`, `focusTrap`, etc... or something that implies the inverse of aria-owns / aria-controls. `parent`

packages/terra-form-select/src/_Frame.jsx
```diff
@@ -385,15 +395,21 @@ class Frame extends React.Component {
       customProps.className,
     ]);
 
+    let role;
+
+    if (!disabled) {
+      role = 'combobox';
```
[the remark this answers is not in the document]
JakeLaCombe: I'm actually getting accessibility errors with that, as there is no discernable text present with that role inside the component. That may take revisiting the placeholder text potentially.

JakeLaCombe:
I think I'm going to close this. After further investigation, I don't think we can rely on the aria attributes to output the correct messages to the user. In order to get a good experience for the user, we will need to use an aria-live region, and make the messages assertive.

PR #2187, opened 2019-01-24
Remove paragraph tag from alert body

### LINKED ISSUE #2075: Remove Paragraph Tag from Alert Body
# Bug Report

## Description

Currently the terra-alert module wraps the body of the alert in a `<p>` tag. This proves problematic if a consumer uses HTML elements (which the docs state is valid input) in body of the alert as the react DOM validator outputs a warning. I have a attached a screenshot of the warning from the console below.

## Steps to Reproduce

1. Create a Terra Alert in a test site.
2. Place a `<div>` element (or a Terra element that uses a `<div>` inside the alert as a child element.
3. Observe the warning in the console.

screen shot 2018-11-28 at 3 44 15 pm

## Expected Behavior

I would expect that when I use HTML elements (besides a `<span>`, for some reason it does not complain about a `<span>` tag) or other Terra elements that use divs under the hood to not print an error.

## Possible Solution

Change the Terra Alert component to use a `<div>` for the body instead of a `<p>` tag.

## Environment

* Component Name and Version: terra-alert 2.33.0
* Browser Name and Version: Google Chrome 70.0.3538.110
* Node/npm Version: Node 8.4 npm 5.3
* Webpack Version: 4
* Operating System and version (desktop or mobile): macOS desktop

## @ Mentions

@tbiethman

gabeparra01:
### Summary
The TerraAlert component wraps its children in a p tag. This causes a console error if the consumer passes in a div tag (or a Terra component that creates a div tag) as a child. Bug Report and recommended solution posted by @andrewnottaviano

During development, I discovered that the "More Information" link in the TerraAlert README page produces a 404 error. I was not able to figure out if the file was actually being pulled into the dev-site. Instead, I replaced the relative link address with a link to the MoreInformation.md file in the terra-core GitHub repo.

### Additional Details
 Resolves #2075

@cerner/terra


packages/terra-alert/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Fixed console error
```
emilyrohrbough: Lets update this changelog entry to be specific to the error that was fixed. Maybe something like: change children wrapper to be div instead of p tag given children accepts type node

gabeparra01: Sounds like a good description to me: https://github.com/gabeparra01/terra-core/commit/18802ce604470dd9a7af6bdb6fbc534641567b96

PR #2200, opened 2019-01-30
Update README to include terra-show-hide

CompSciLauren:
### Summary
Run `npm run docs` to generate docs and include terra-show-hide.


bjankord:
👍Thanks for the contribution @CompSciLauren

PR #2203, opened 2019-01-30
Fixes issue #1752 Status View glyph does not always display when it should.

### LINKED ISSUE #1752: Status View glyph does not always display when it should
# Bug Report

## Description
The Status View component contains logic that attempts to determine if the component has enough space to display the glyph.  In some cases, the glyph is not displayed even when there should be enough room for it.  This has been observed specifically in cases when the browser zoom level is changed.  The following link can be used to recreate the issue:

https://pages.github.cerner.com/healtheintent/action_plan/?knob-icon=plan&knob-title=Activities%20To%20Do&knob-subheading=5%20of%208%20activities%20need%20to%20be%20completed&knob-content=Card%20content&selectedKind=Components%2FNoActivitiesCard&selectedStory=default&full=0&addons=1&stories=1&panelRight=1&addonPanel=storybooks%2Fstorybook-addon-knobs

## Steps to Reproduce
1. Using Chrome browser on desktop, set browser zoom level to 100%
2. Click on link above [a component using Status View should display with a glyph]
3. Set browser zoom level to 125%
4. Refresh browser [the component is displayed again this time without the glyph]

noactivitiescard snapshot - with icon
noactivitiescard snapshot - no icon

## Expected Behavior
The glyph should alway be displayed regardless of the browser zoom level when there is no restriction placed on the height of the StatusView component.

## Possible Solution
Change the logic that determines whether to display the glyph to be a little more forgiving.  Allow the calculated height of the StatusView contents to be 1 to 2 px greater than the height of the container.  At different zoom levels the calculated height of nodes can change slightly, probably due to the fact that offsetHeight used in the calculations is a rounded value.

## Environment
* Component Name and Version: terra-status-view v2.20.0
* Browser Name and Version: Chrome Version 68.0.3440.84 (Official Build) (64-bit)
* Node/npm Version: Node: 8.9.2 / yarn 1.3.2
* Webpack Version: 3.12.0
* Operating System and version (desktop or mobile): OS X El Capitan version 10.11.6

ShettyAkarsh:
### Summary

Resolves the glyph not being displayed issue when it is expected to be displayed.  which is basically due to inner view height which holds the components and the glyph has the height lesser than the total height of components, glyph. As per the existing logic for the glyph to display the inner view height should be always greater than the total height of components, glyph. the difference between the heights will always be small fractional values in the case where the issue exists.

The fix is to check whether the difference in height is lesser than equal to fractional values let say 0.0001. then in that case add this value to inner view height and then compare the heights to check whether to display the glyph or not.

Resolves #1752


packages/terra-status-view/src/StatusView.jsx
```diff
@@ -98,29 +98,43 @@ class StatusView extends React.Component {
    * with 40% of the remaining space at the top with 60% on the bottom.
    */
   showAndCenterItems() {
-    const viewHeight = this.innerNode.offsetHeight;
+    const viewHeight = this.innerNode.getBoundingClientRect().height;
+    let newViewHeight = viewHeight;
```
yuderekyu: do we need to reasign `viewHeight` to a new variable?

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -98,29 +98,43 @@ class StatusView extends React.Component {
...
 
-    if (this.glyphNode && viewHeight >= componentsHeight + this.glyphNode.offsetHeight && this.innerNode.offsetWidth >= this.glyphNode.offsetWidth) {
+    let totalComponentsHeight = componentsHeight;
+    if (this.glyphNode) {
+      totalComponentsHeight += this.glyphNode.getBoundingClientRect().height;
+      // if inner view node height is lesser than the components along with glyph inside the inner view.
+      // and the height difference is very less (i.e 0.0001) then add that comparision value to the
+      // the inner view height.
+      if (viewHeight < totalComponentsHeight) {
+        if (totalComponentsHeight - viewHeight <= 0.0001) {
```
yuderekyu: do we need the conditional on line 126? or can we just check ` if (totalComponentsHeight - viewHeight <= 0.0001)`?

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -98,29 +98,43 @@ class StatusView extends React.Component {
...
     }
     if (this.dividerNode) {
-      componentsHeight += this.dividerNode.offsetHeight;
+      componentsHeight += this.dividerNode.getBoundingClientRect().height;
     }
 
-    if (this.glyphNode && viewHeight >= componentsHeight + this.glyphNode.offsetHeight && this.innerNode.offsetWidth >= this.glyphNode.offsetWidth) {
+    let totalComponentsHeight = componentsHeight;
+    if (this.glyphNode) {
+      totalComponentsHeight += this.glyphNode.getBoundingClientRect().height;
```
yuderekyu: related to my second comment, can we hoist and group the variable definitions together at the top?

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -98,29 +98,43 @@ class StatusView extends React.Component {
...
-    const viewHeight = this.innerNode.offsetHeight;
+    const viewHeight = this.innerNode.getBoundingClientRect().height;
+    let newViewHeight = viewHeight;
 
     let paddingTop = 0;
     let paddingBottom = 0;
     let showGlyph = false;
 
     // get the content heights of the nodes that are to be shown if have content and cannot be hidden
-    let componentsHeight = this.titleNode.offsetHeight;
```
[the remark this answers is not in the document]
ShettyAkarsh: Moved all the variables to the beginning. regarding the refactor suggestion we might have to declare more variables  in that case and that will make this function more complex.

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -98,29 +98,43 @@ class StatusView extends React.Component {
...
-      componentsHeight += this.messageNode.offsetHeight;
+      componentsHeight += this.messageNode.getBoundingClientRect().height;
     }
     if (this.dividerNode) {
-      componentsHeight += this.dividerNode.offsetHeight;
+      componentsHeight += this.dividerNode.getBoundingClientRect().height;
     }
 
-    if (this.glyphNode && viewHeight >= componentsHeight + this.glyphNode.offsetHeight && this.innerNode.offsetWidth >= this.glyphNode.offsetWidth) {
+    let totalComponentsHeight = componentsHeight;
```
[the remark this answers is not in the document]
ShettyAkarsh: Yes we need, as the componentsHeight variable value which does not include glyph height will be used to calculate the padding when glyph will not be displayed.  Renamed  totalComponentsHeight to contentHeight.

packages/terra-status-view/src/StatusView.jsx
ShettyAkarsh: Yes we need actual viewheight to calculate the padding of the inner view.

packages/terra-status-view/src/StatusView.jsx
ShettyAkarsh: Removed

packages/terra-status-view/src/StatusView.jsx
ShettyAkarsh: Yes, Moved all the variable declaration to the beginning of the method.

packages/terra-status-view/src/StatusView.jsx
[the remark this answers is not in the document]
yuderekyu: Adding more variables does not add to cyclomatic complexity. To me, the refactor adds clarity for other developers by grouping size values into a single block and removing nested dot notation scattered through the function. This eliminates multiple if statements and gives clear intent on what each value is.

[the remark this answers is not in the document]
ShettyAkarsh: .Added the changes for inline conditional statements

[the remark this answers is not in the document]
yuderekyu: Sweet. Sorry if there was a misunderstanding, but we don't have to initialize every variable at the top. We're already using inline conditionals, so they'll be set to `0` if the nodes don't exist. 

We can add a `let` to each variable followed by the inline conditional, similar to the above suggestion. This helps remove redundancy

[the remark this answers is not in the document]
ShettyAkarsh: Updated the changes for conditional statement variable declaration.

packages/terra-status-view/CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ Changelog
 Unreleased
 ----------
 
+### Fixed
+* Status View glyph does not always displays when it should
+=======
```
bjankord: Replace this line with a blank new line.
e.g.

```
### Fixed
* Status View glyph does not always displays when it should

3.5.0 - (February 12, 2019)
------------------
### Changed
```

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -98,35 +98,42 @@ class StatusView extends React.Component {
...
+    const actionNodeHeight = this.actionsNode ? this.actionsNode.getBoundingClientRect().height : 0;
+    const messageNodeHeight = this.messageNode ? this.messageNode.getBoundingClientRect().height : 0;
+    const dividerNodeHeight = this.dividerNode ? this.dividerNode.getBoundingClientRect().height : 0;
+    contentHeight = (titleNodeHeight + actionNodeHeight + messageNodeHeight + dividerNodeHeight);
+
+    contentWithGlyphHeight = contentHeight + (this.glyphNode ? this.glyphNode.getBoundingClientRect().height : 0);
+    // if inner view node height is lesser than the components along with glyph inside the inner view.
+    // and the height difference is very less (i.e 0.0001) then add that comparision value to the
+    // the inner view height.
+    if (contentWithGlyphHeight - viewHeight <= 0.0001) {
```
yuderekyu: nit: want to create constant to store `0.0001`, then reuse it on the next two lines? something like `precision` or `roundingPrecision`

PR #2217, opened 2019-02-07
Add the ability for Screen Readers to read current active option

### LINKED ISSUE #1769: [terra-form-select] screen reader is unable to read options in Select component
# Bug Report

## Description

Using a screen reader such as JAWS or Narrator for Windows, the reader is unable to read options in the Select component when navigating through the select options using a keyboard. 

## Steps to Reproduce

1. Open screen reader (JAWS, Narrator for Windows, etc)
2. Using the keyboard tab into a Select dropdown component (https://engineering.cerner.com/terra-ui/#/components/terra-form-select/form-select/select)
3. Use the arrow keys to move through the options of the dropdown
4. Expect the highlighted option to be read by the screen reader

## Expected Behavior

A screen reader should be able to read options in the Select component.

## Environment

* Component Name and Version: terra-form-select 4.13.0
* Browser Name and Version: IE 11
* Webpack Version:

### LINKED ISSUE #1887: [terra-form-select] - Options are not read by VoiceOver on iOS

JakeLaCombe:
### Summary
Resolves #1769

The previous pull request didn't accurately depict the results to the user. This time, we are using the aria-live region to directly communicate not only the active element, but the active elements that are selected to screen readers.


JakeLaCombe:
There will be a translation request sent when more people look at this PR.

packages/terra-form-select/package.json
```diff
@@ -43,7 +43,7 @@
     "compile:clean": "rm -rf lib",
     "compile:build": "babel src --out-dir lib --copy-files",
     "lint": "npm run lint:js && npm run lint:scss",
-    "lint:js": "eslint --ext .js,.jsx . --ignore-path ../../.eslintignore",
+    "lint:js": "eslint --fix --ext .js,.jsx . --ignore-path ../../.eslintignore",
```
yuderekyu: nice - should we set the `fix` flag across all components then?

packages/terra-form-select/src/_MenuUtil.js
```diff
@@ -22,7 +22,7 @@ class MenuUtil {
    * @return {boolean} - True if the values are equal.
    */
   static isEqual(a, b) {
-    if (a === undefined || b === undefined) {
+    if (a === undefined || a === null || b === undefined || b === null) {
```
yuderekyu: should we be checking for all falsy values instead?
```suggestion
    if (!a || !b) {
```

packages/terra-form-select/package.json
JakeLaCombe: oops. I mean to remove that.

packages/terra-form-select/src/_MenuUtil.js
StephenEsser: Javascript [falsey](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) includes empty string and zero which can be valid values.

When is the value null?

packages/terra-form-select/translations/en.json
```diff
@@ -1,5 +1,6 @@
 {
   "Terra.form.select.add": "Add \"{text}\"",
   "Terra.form.select.defaultDisplay": "- Select -",
-  "Terra.form.select.noResults": "No matching results for \"{text}\""
+  "Terra.form.select.noResults": "No matching results for \"{text}\"",
+  "Terra.form.select.selected": "{text} Selected"
```
emilyrohrbough: All translation files need updated, not just english

JakeLaCombe:
I think that's because there are two messages there. The first is that it's announcing the user input to the user, and the next is to alert the name of the selected option. I'll see if I can find ways around that, but I think we are getting thin on possible options for this.

packages/terra-form-select/src/_MenuUtil.js
JakeLaCombe: Type in something that returns no results found, then press down on the keyboard. It will throw the air.

JakeLaCombe: Here's the line that returns null. https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/_MenuUtil.js#L184

JakeLaCombe: actually, we need to leave this how it is. If we do a check on !a or !b and both a and b are empty strings '', '', then this would incorrectly return false. We essentially need to call toString on the method, which is done through null and undefined checks.

packages/terra-form-select/src/_Frame.jsx
```diff
@@ -385,15 +394,21 @@ class Frame extends React.Component {
       customProps.className,
     ]);
 
+    let role;
+
+    if (!disabled) {
+      role = 'combobox';
```
benbcai: For the sake of consistency you could make this a ternary operator just like the rest below.

StephenEsser: +1 to ternary or creating the aria attributes in a utility method so they are easier to reason about. Something similar to this:

https://github.com/cerner/terra-core/compare/SelectAccessibility#diff-00944ece91b08443a6cd856f48385ee6R10

JakeLaCombe:
Translations have been addd, and it should be ready to be fully reviewed

bjankord:
The stutter when reading options with VoiceOver on Safari desktop seems like it still exists. In my opinion, this creates a jarring exprience. 

It makes me wonder if we should ditch the role=listbox / role=option to avoid the stutter and rely on aria-live soley for communicating the options. We made need to add tabIndex="-1" to enable navigating between the options on iOS with VoiceOver as I believe these roles offer some heuristics that VoiceOver picks up to enable navigating through them.

packages/terra-form-select/src/_Frame.jsx
```diff
@@ -389,11 +398,11 @@ class Frame extends React.Component {
       <div
         {...customProps}
         role={!disabled ? 'combobox' : undefined}
-        aria-controls={!disabled && this.state.isOpen ? 'terra-select-dropdown' : undefined}
+        aria-controls={!disabled && this.state.isOpen ? 'terra-select-menu' : undefined}
```
bjankord: With testing I've been doing with VoiceOver on iOS, to enable usage with the screen reader on mobile, I needed to shift the role and aria attributes from this div to the div with role="textbox" on it on line 406. 

This change matches with [other](https://github.com/cerner/terra-core/issues/1887#issuecomment-460730393) [custom](https://github.com/cerner/terra-core/issues/1887#issuecomment-460734805) [selects](https://github.com/cerner/terra-core/issues/1887#issuecomment-460739869) we've looked. The aria attributes in those code examples are placed on the input element. For this code, the input is essentially the `div[role="textbox"]` element.

Can you test this change and see how it works with JAWS and VoiceOver on desktop.

JakeLaCombe: Is that to deal with the focus management? It seems to be reading the results fine on the list on my device with the current role settings.

bjankord: I've switched the code I had back to match similar to DOM structure in this PR. I had run into some issues with aXe with the DOM structure I had tried out.

bjankord:
I've migrated the code in this PR into https://github.com/cerner/terra-core/pull/2236. Closing in favor of https://github.com/cerner/terra-core/pull/2236

Issue #1847, opened 2018-09-07
[terra-status-view] SVG size display consistency


neilpfeiffer:
# Bug Report

## Description
Minor visual bug which causes different icons to sometimes shift text in status view

## Additional Context / Screenshots
![image](https://user-images.githubusercontent.com/1355485/45238801-ad819280-b2a8-11e8-8682-e560954c7ff3.png)

## Expected Behavior
Switching between variants should not cause text contents of a status view to shift due to different icons being provided. 

## Possible Solution
Set display on glyph SVG element to block.
```
.glyph {
    ...
    > svg {
      ...
      height: 100px;
      width: 100px;
      display: block;
    }
    ...
}
```

PR #2218, opened 2019-02-07
[terra-status-view] Add display block to SVG

yuderekyu:
### Summary
Closes https://github.com/cerner/terra-core/issues/1847

The SVG's display attribute defaults to `inline` - this ignores `height` and `width` values, which makes the parent div's size inconsistent, intermittenly pushing text downward.

Setting `display: block` on the SVG fixes this. We have to regenerate reference screenshots for this change.


packages/terra-status-view/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Added
```
emilyrohrbough: I think this should be `### Fixed` since this is a bug fix. Also, might be beneficial to add updated style to use display block for inline bug fix (or something with more context)

yuderekyu: https://github.com/cerner/terra-core/pull/2218/commits/05004c807b09dd8f7d0dfcf9744a33030f53f3ff

neilpfeiffer:
Tested: PR #2218
 
**Reviewed:**
- resolves icon size consistency
 
**Pass** |
-----------|
UX Review|

Issue #2221, opened 2019-02-11
[Terra-Avatar] - Fall Back Icon(USER VARIANT) not coming up in case of Image load failure.


vaibhavmsv:
# Bug Report

## Description
I am trying to load Terra-Avatar(2.6.0 version) component with Image, hash value and size props. 
In case of image load failure, the fall back icon user variant should come up.  Instead a Blank circle is coming up. If we dont use Image and initials as props, then the user variant is coming up. But in case of Image Load Failure, User Variant is not coming as fall back icon.

## Steps to Reproduce

1. Build an Avatar Component. 
<Avatar
        image={URL of image}
        hashValue={give hash value}
        size="0.6em"
      />
2. Make sure URL of image is invalid.
3. Now when we load that avatar component, User Variant will not be coming up as fall back icon. 

## Additional Context / Screenshots

![image](https://user-images.githubusercontent.com/18640970/52560678-199b9000-2e1f-11e9-874f-688490a7b2ca.png)

![image](https://user-images.githubusercontent.com/18640970/52560747-549dc380-2e1f-11e9-9008-f31ca9e89756.png)

![image](https://user-images.githubusercontent.com/18640970/52560767-61221c00-2e1f-11e9-84f4-fb08a822ce5e.png)

## Expected Behavior
User Variant should come up incase of image load failure.

## Possible Solution
Not Sure

## Environment

* Component Name and Version: terra-avatar and version is 2.6.0
* Browser Name and Version: Chrome 
* Node/npm Version: Node 8

yuderekyu:
The icon renders correctly on fallback - the issue is that a color variant is not attached to avatar when an `image` is specified. The color variant styles control the `background-color` and `box-shadow` - because the icon is white, it appears to be missing. 

The fix is to change [this](https://github.com/cerner/terra-core/blob/master/packages/terra-avatar/src/variants/Avatar.jsx#L73) to set a color variant. This will also need to be done in the `Facility` variant. I'll issue a PR to address this.

PR #2224, opened 2019-02-11
[terra-avatar] Generate color variant for invalid image

yuderekyu:
### Summary
Closes https://github.com/cerner/terra-core/issues/2221

Previously, the color variant was not generated for an image. So, if someone passes in an invalid image or the image fails to load, the fall back icon fails to render with colors.

This change implements a color variant for an image, if this edge case occurs. Other changes were necessary to keep this passive.

### Notable Changes
* Converted `Avatar` and `Facility` to stateful components. This is necessary to keep track of the internal image's error state - if the image renders a fallback, apply the correct styles.
* Always generate a color variant; if an invalid src url is provided, the fallback should display a color variant.
* Added CSS classes to cover edge cases
   * Valid image = override color variant
   * Invalid image = fallback icon with color variant
   * Deceased Image = override color variant
   * Deceased icon/initials = override color variant

### Test Links
https://terra-core-deployed-pr-2224.herokuapp.com/#/tests/terra-avatar/avatar/avatar/invalid-image-avatar


neilpfeiffer:
@yuderekyu - We will need to fix this with a future issue, but fine for now as long as consumer provides a perfectly square image src, since this forces any image to be square and squished incorrectly if not a square.

https://github.com/cerner/terra-core/blob/27e320862b7f1cae3c6135515c10eb6acd3f2289/packages/terra-avatar/src/common/Avatar.module.scss#L64-L68

Which is noticeable here when using the wide image file here:
https://github.com/cerner/terra-core/blob/27e320862b7f1cae3c6135515c10eb6acd3f2289/packages/terra-avatar/src/terra-dev-site/test/avatar/avatar/UpdateAvatar.test.jsx#L4

neilpfeiffer:
Tested: PR #2224
 
**Reviewed:**
- resolves correctly handling bad images with fallback to the other avatar presentations. 
 
**Pass** |
-----------|
UX Review|

PR #2225, opened 2019-02-11
[terra-form-select] Add clear option.

### LINKED ISSUE #1736: Form-Select: Need to be able to deselect
The problem:
Currently with any given select field, the user can not clear out their selection.  The placeholder text is not included in the list.

screen shot 2018-07-27 at 11 08 01 am
 
The enhancement:
For form selects, there are use cases where the user needs to be able to deselect the previously chosen selection (ie - clearing a filter, etc).

benbcai:
### Summary
- Added new props `allowClear` to show an option to allow clearing the selected item.

### Additional Details
Closed #1736. 
When the `allowClear` prop is set, an option is added to the top of the menu to allow clearing the selected item and reset the select back to the unselected state. The placeholder passed in is used as the display text for the clear option. If the placeholder is not provided or is an empty string, the default placeholder ("- Select -") would be used as the clear option text. The `allowClear` prop will be ignored for the `Mulitple` or `Tag` select variant.

Still requires some UX input on what margins to use for the clear option when displayed in an Option Group and accessibility when selecting the clear option.

@cerner/terra


benbcai:
This PR is not auto deploying. Closing and replacing with #2240.

Issue #2023, opened 2018-11-05
Shift focus to select popup when using voice over on iOS


bjankord:
# Bug Report

## Description
When select popup is opened on iOS with VoiceOver, focus is not shifted to popup. This makes it difficult for mobile users with using voice over to interact with the select.

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-ui/#/components/terra-form-select/form-select/select
2. Try to interact with select on iOS with VoiceOver turned on

## Expected Behavior
When select menu is opened, the focus is shifted to is on iOS when using VO.

## Possible Solution
Add `aria-modal="true` and `role="dialog"` to hookshot container in select.

bjankord:
Was running into some issue with keydown on this and thought I might need to use [`event.nativeEvent.stopImmediatePropagation()`](https://gist.github.com/ggregoire/ce7bc946212920c0a6bad8125567d001) but I don't think it is needed anymore.

bjankord:
Got this in a working state, hoping to get a PR out soonish.

Issue #1907, opened 2018-09-20
No mapping relationship between select field label and select field input


bjankord:
# Bug Report

## Description
There is no for attribute on the label in the <SelectField /> component mapping to the input (role=textbox) in select field.

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-core/#/tests/terra-form-select/form-select/select-field
2. Shift focus to div with role="textbox"
3. Screen readers are unable to read label that should be associated to the select.

## Expected Behavior
When focused on the select, screen readers should be able to read the associated label

## Possible Solution
Set for attribute on select field label that maps to id on div with role="textbox"

## Environment

* Component Name and Version: terra-select select field component

PR #2236, opened 2019-02-17
Screen reader improvements when using select component

bjankord:
### Summary
Add improvements to enhance screen reader experience with terra-form-select
* Resolves https://github.com/cerner/terra-core/issues/1887
* Resolves https://github.com/cerner/terra-core/issues/1907
* Resolves https://github.com/cerner/terra-core/issues/2023
* Resolves https://github.com/cerner/terra-core/issues/1769

## Validation link
https://terra-core-deployed-pr-2236.herokuapp.com/#/components/terra-form-select/form-select/select

## Component file changes:
* packages/terra-form-select/src/Select.jsx
* packages/terra-form-select/src/SelectField.jsx
* packages/terra-form-select/src/_Frame.jsx
* packages/terra-form-select/src/_Menu.jsx
* packages/terra-form-select/src/_Option.jsx
* packages/terra-form-select/src/_Frame.module.scss
* packages/terra-form-select/src/_Option.module.scss

## WDIO Test additions
I've added a lot of wdio tests to cover these scenarios:
 * should be closed initially
 * should gain focus when tabbed to
 * should open dropdown by spacebar key press
 * should open dropdown by down arrow key press
 * should not open dropdown by enter key press
 * should close when clicking off of the select
 * should close when clicking off of the select after being opened by toggle icon
 * should close when pressing tab key to shift focus away from select
 * should close when clicking on toggle icon when select is open
 * should open and close the dropdown by clicking on the select
 * should open and close the dropdown by clicking on toggle icon
 * should select an option by keyboard interaction
 * should select an option by click after clicking on toggle icon
 * should select an option by alternative keyboard interaction
 * should select an option by pressing enter
 * should select second option by keyboard interaction


packages/terra-form-select/src/_Frame.jsx
```diff
@@ -403,12 +485,23 @@ class Frame extends React.Component {
...
+          aria-controls={!disabled && this.state.isOpen ? 'terra-select-dropdown' : undefined}
+          aria-disabled={!!disabled}
+          aria-owns={this.state.isOpen ? 'terra-select-dropdown' : undefined}
+          aria-expanded={!!disabled && !!this.state.isOpen}
+          aria-haspopup={!disabled ? 'true' : undefined}
+          aria-label="Search"
+          aria-describedby="screen-reader-text"
+          className={cx('display')}
+        >
+          <span id="screen-reader-text" className={cx('visually-hidden-component')}>
```
JakeLaCombe: Should this id be more specific to the select element?

bjankord: Updated in [this commit](https://github.com/cerner/terra-core/pull/2236/commits/77e44ebdb23637b8a9fdc33e3278ba3d0937bb73) to use 
```js
const ariaDescribedById = `terra-select-screen-reader-description-${uniqueid()}`;
```

packages/terra-form-select/src/Select.jsx
```diff
@@ -206,13 +206,14 @@ class Select extends React.Component {
...
         {...otherProps}
+        data-terra-select
         value={Util.value(this.props, this.state)}
         display={this.display()}
         onDeselect={this.handleDeselect}
         onSelect={this.handleSelect}
         placeholder={selectPlaceholder}
         dropdown={dropdownProps => (
-          <DropdownMenu intl={intl} {...dropdownProps}>
+          <DropdownMenu {...dropdownProps}>
```
bjankord: The menu is now wrapped in injectIntl and so it no longer needs to be passed down as a prop here.

packages/terra-form-select/src/_Menu.jsx
```diff
@@ -163,14 +207,35 @@ class Menu extends React.Component {
         return;
       }
 
-      if (hasNoResults && searchValue) {
```
JakeLaCombe: I take it that if searchValue is null then we assume that hasNoResults is false?

packages/terra-form-select/src/_Menu.jsx
```diff
@@ -313,19 +402,22 @@ class Menu extends React.Component {
...
+       * shift the virtual indicator to this DOM element when dropdown is rendered.
+       */
       <ul
+        id="terra-select-menu"
         role="listbox"
         className={cx('menu')}
         ref={(menu) => { this.menu = menu; }}
-        aria-activedescendant={`terra-select-option-${this.state.active}`}
+        {...('ontouchstart' in window ? { 'aria-activedescendant': `terra-select-option-${this.state.active}` } : {})}
+        data-activedescendant={`terra-select-option-${this.state.active}`}
```
JakeLaCombe: should we null this if nothing is active?

packages/terra-form-select/src/_Menu.jsx
```diff
@@ -313,19 +402,22 @@ class Menu extends React.Component {
...
         tabIndex="0"
+        style={{ outline: 'none' }}
```
JakeLaCombe: Could this be attached to the css class menu?

packages/terra-form-select/src/_Menu.jsx
```diff
@@ -126,20 +149,41 @@ class Menu extends React.Component {
...
     this.scrollIntoView();
   }
 
   componentWillUnmount() {
     this.clearSearch();
     this.clearScrollTimeout();
     document.removeEventListener('keydown', this.handleKeyDown);
+
+    if (this.props.focusRegion) {
+      this.props.focusRegion.removeAttribute('data-activedescendant');
```
JakeLaCombe: Where is data-activedescendant used? Should this be aria-activedescendant?

bjankord: Updated in this commit: https://github.com/cerner/terra-core/pull/2236/commits/7ed989860674b09c4c4c49b168d745eb37edc8f5

packages/terra-form-select/src/_Menu.jsx
bjankord: Updated in this commit: https://github.com/cerner/terra-core/pull/2236/commits/7ed989860674b09c4c4c49b168d745eb37edc8f5

packages/terra-form-select/src/_Menu.jsx
bjankord: Updated in this commit: https://github.com/cerner/terra-core/pull/2236/commits/7ed989860674b09c4c4c49b168d745eb37edc8f5

packages/terra-form-select/src/_Menu.jsx
bjankord: Yeah, I pulled this code change from this PR: https://github.com/cerner/terra-core/pull/2217/files

JakeLaCombe:
Forgot I wrote that, lol. Yeah, it should be fine I believe.

Get Outlook for iOS<https://aka.ms/o0ukef>

________________________________

packages/terra-form-select/src/_Menu.jsx
```diff
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import { polyfill } from 'react-lifecycles-compat';
+import { injectIntl, intlShape } from 'react-intl';
```
JakeLaCombe: Does terra-i18n include this? It feels like this would be better pulled from that than react-intl directly?

packages/terra-form-select/src/_Menu.jsx
JakeLaCombe: Forgot I wrote that. Yeah, I think this line of code is valid.

packages/terra-form-select/src/_Menu.jsx
bjankord: We want to pull these from react-intl. Pulling in injectIntl and intlShape from terra-i18n has caused some issue in the past, we have this issue logged which covers the issue in more detail: https://github.com/cerner/terra-core/issues/1982

JakeLaCombe: sounds good

bjankord:
Meet with @scottwilmarth and Emily Ogle today to review the screen reader user experience of this component. Received feedback on the following:

* Announce when items are selected or deselected and what the option is as the user selects or deselects options.
* Disabled options are counted in the list of total options which may be confusing. Scott and Emily were going to discuss some more on what to do with disabled options from a screen reader perspective noting that more thought is needed.
* Prepend text to description text on variants that include an input element. "Enter text or use up and down arrow..."
* Don't count "clear" as an option in the list of total options.
* Provide text description on how to select items. Users will have to guess what key to press if we dont provide that info.

packages/terra-form-select/tests/jest/__snapshots__/Frame.test.jsx.snap
```diff
@@ -27,11 +27,23 @@ exports[`Frame should render a clear option 1`] = `
...
+        "Terra.form.select.dimmed": "Dimmed.",
+        "Terra.form.select.disabled": "Disabled.",
+        "Terra.form.select.listOfTotalOptions": "List of {total} options.",
         "Terra.form.select.maxSelectionHelp": "Limit {text} items.",
         "Terra.form.select.maxSelectionOption": "Maximum number of {text} items selected",
+        "Terra.form.select.mobileButtonUsageGuidance": "Tap to navigate to options.",
+        "Terra.form.select.mobileUsageGuidance": "Swipe right to navigate to options.",
+        "Terra.form.select.multiSelectUsageGuidance": "Enter text or use up and down arrow keys to navigate through options. Press enter to select or unselect an option.",
         "Terra.form.select.noResults": "No matching results for \\"{text}\\"",
-        "Terra.form.select.selected": "{text} Selected",
```
emilyrohrbough: How does changing the intention of a translation key's meaning affect downstream consumers if they are providing custom translations?

I see here we updated the `Terra.form.select.selected` key meaning to be `Terra.form.select.selectedText` and then changed `Terra.form.select.selected`to have a different representation & string translations.

I don't think updating translations are a concern, but wasn't sure of the key intent.

bjankord: `Terra.form.select.selected` is a new key added in this PR. I have another PR which adds only the new translation keys so people can see what new translations are being added. https://github.com/cerner/terra-core/pull/2266/files

packages/terra-form-select/src/_Frame.jsx
```diff
@@ -213,13 +256,40 @@ class Frame extends React.Component {
...
+      && (event.target.hasAttribute('data-terra-form-select-toggle-button')
+      || event.target.hasAttribute('data-terra-form-select-toggle-button-icon'))) {
+      this.setState({ isOpen: true, isPositioned: false });
+
+      // Allows time for state update to render select menu DOM before shifting focus to it
+      setTimeout(() => {
+        if (document.querySelector(this.selectMenu)) {
+          document.querySelector(this.selectMenu).focus();
+        }
+      }, 10);
```
StephenEsser: gamble

bjankord: Focus management is tricky.

PR #2240, opened 2019-02-18
[terra-form-select] Add clear option

### LINKED ISSUE #1736: Form-Select: Need to be able to deselect

benbcai:
### Summary
- Added new props `allowClear` to show an option to allow clearing the selected item.

### Additional Details
Closed #1736. 
When the `allowClear` prop is set, an option is added to the top of the menu to allow clearing the selected item and reset the select back to the unselected state. The placeholder passed in is used as the display text for the clear option. If the placeholder is not provided or is an empty string, the default placeholder ("- Select -") would be used as the clear option text. The `allowClear` prop will be ignored for the `Mulitple` or `Tag` select variant.

Still requires some UX input on what margins to use for the clear option when displayed in an Option Group and accessibility when selecting the clear option.

### Example Deployment Links:
https://terra-core-deployed-pr-2240.herokuapp.com/#/tests/terra-form-select/form-select/allow-clear

@cerner/terra


StephenEsser:
Do the UX callouts mention a selection state for the clear option? Is the clear option always supposed to be selected by default?

image

benbcai:
The terra-opensource-theme doesn't have the clear option selected by default. The other themes do have it in the selected state by default.

Open Source Theme:
screen shot 2019-02-19 at 12 11 05 pm

Clinical Theme:
screen shot 2019-02-19 at 12 11 24 pm

@neilpfeiffer Should the default selection state of the clear option be themeable? Or should we just choose one of the above and not make it themeable?

benbcai:
@neilpfeiffer I'm still in need of a couple of things from UX.
 - margin-top and margin-bottom for the clear option in OptGroup.
 - After making the clear option to be not selected by default, the color contrast between the color and the hover background no longer meets the WCAG 2 AA contrast ratio thresholds. The color used to be #70787c, which is the same as the place holder color. I need a color that is a bit darker. I'm just using #666F70 for the time being.

screen shot 2019-03-01 at 2 00 29 pm

packages/terra-form-select/src/Select.jsx
```diff
@@ -197,11 +203,20 @@ class Select extends React.Component {
...
-      children, defaultValue, onChange, placeholder, value, ...otherProps
+      allowClear, children, defaultValue, onChange, placeholder, value, ...otherProps
     } = this.props;
 
     const defaultPlaceholder = intl.formatMessage({ id: 'Terra.form.select.defaultDisplay' });
     const selectPlaceholder = placeholder === undefined ? defaultPlaceholder : placeholder;
+    let clearOptionDisplay;
+
+    if (allowClear) {
+      if (selectPlaceholder.length === 0) {
```
tbiethman: If they give us an empty string, do we really want to override that?

benbcai: We use the placeholder as the display for the clear option. If they provide an empty placeholder, we don't want the display for the clear option to also be empty. We are not overriding their placeholder, we are just falling back to our default placeholder (" -Select- ") as the clear option display if the placeholder they provide is really empty.

packages/terra-form-select/src/_ClearOption.module.scss
```diff
@@ -0,0 +1,6 @@
+:local {
+  .clear-option {
+    color: var(--terra-form-select-clear-option-color, #666f70);
+    font-style: var(--terra-form-select-clear-option-font-style);
+  }
```
neilpfeiffer: Add margin:
```suggestion
    margin: var(--terra-form-select-clear-option-margin, 0 0 0.17857rem 0);
  }
```

packages/terra-form-select/src/Select.jsx
```diff
@@ -9,6 +9,11 @@ import Tag from './_Tag';
 import Util from './_SelectUtil';
 
 const propTypes = {
+  /**
+   * Whether a clear option is available to clear the selection.
```
neilpfeiffer:
```suggestion
   * Whether a clear option is available to clear the selection, will use placeholder text if provided.
```

neilpfeiffer:
Couple of docsite adjustments: 
1. lets move `Clear` example before `Controlled`
https://github.com/cerner/terra-core/blob/5b9edded6d4cb2ed4a2fdcd21453d2b72e6c32db/packages/terra-form-select/src/terra-dev-site/doc/form-select/Select.1.doc.jsx#L101

2. Let update these example titles, so the set is more understandable which are types/variants, and which are ala carte features in addition to the variants:
- [ ]  `title: 'Disabled',` -> 'Disabled (applies to all)',
- [ ]  `title: 'Invalid',` -> 'Invalid (applies to all)',
- [ ]  `title: 'OptGroup',` -> 'Ability to add Option Groups (applies to all)'
- [ ]  `title: 'Custom Max Height',` -> 'Ability to set a Maximum Height (applies to all)'
- [ ]  `title: 'Clear',` -> ' Ability to Clear Selection (applies to only \'default\', \'combobox\', and \'search\')',
- [ ]  `title: 'Controlled',` -> 'Create a Controlled Select (applies to all)'

mmalaker:
Functional verification completed. The select example functions as expected with the clear option. +1 to UX changes.

packages/terra-form-select/src/_ClearOption.module.scss
```diff
@@ -0,0 +1,6 @@
+:local {
+  .clear-option {
+    color: var(--terra-form-select-clear-option-color, #666f70);
```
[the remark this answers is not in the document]
benbcai: Updated: https://github.com/cerner/terra-core/pull/2240/commits/d059316429c0bc319635619bbbc456c09589e49b

packages/terra-form-select/src/terra-dev-site/doc/form-select/Select.1.doc.jsx
```diff
@@ -95,6 +97,11 @@ const Index = () => (
       description: 'An example of implementing a controlled Select.',
       example: <ControlledExample />,
       source: ControlledExampleSrc,
+    }, {
+      title: 'Clear',
+      description: 'An example with an option to clear the selected item.',
```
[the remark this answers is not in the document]
benbcai: Updated: https://github.com/cerner/terra-core/pull/2240/commits/d059316429c0bc319635619bbbc456c09589e49b

packages/terra-form-select/src/_ClearOption.module.scss
benbcai: Updated: https://github.com/cerner/terra-core/pull/2240/commits/d059316429c0bc319635619bbbc456c09589e49b

packages/terra-form-select/src/Select.jsx
benbcai: Updated: https://github.com/cerner/terra-core/pull/2240/commits/d059316429c0bc319635619bbbc456c09589e49b

benbcai:
@neilpfeiffer 
- The Clear example has been moved to be before the Controlled example.
- The title and description on those examples are updated.

neilpfeiffer:
Tested: PR #2240 
 
**Reviewed:**
- Resolves adding ability to soft "clear" a select option from the single select variants in  `terra-form-select`

**Related:**
- will need to add new variables and values into theme files 

\
\
Verified by @neilpfeiffer, ready for release.
 
**Pass** |
-----------|
UX Review|

PR #2249, opened 2019-02-28
Added overlay prop to container to separate the overlay from container content

### LINKED ISSUE #2024: When container overlay is opened on iOS, VoiceOver users can focus on content behind container overlay
# Bug Report

## Description
When container overlay is opened on iOS, VoiceOver users can focus on content behind container overlay

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-ui/#/components/terra-overlay/overlay/overlay on iOS using VoiceOver
2. Open container based overlay
3. Swipe left or right
4. Focus can be shifted to buttons behind the container overlay

## Expected Behavior
VoiceOver users on iOS cannot focus on elements behind container overlay when it is open.

## Possible Solution
We need a way to mark children in the OverlayContainer with `aria-hidden`. Currently, the `<OverlayContainer />` component just takes children. We could loop through the children, set aria-hidden on all of them which (are not an `Overlay` component or already include aria-hidden attribute). This has complexitity to it as we'd need to keep track of these so when we remove aria-hidden, only the children we originally targeted have the attribute reset.

Another option would be to add a prop to the `OverlayContainer` called `overlay`. This would change the API of `OverlayContainer` where we would expect children to be only the content to render behind the overlay when it is open, and the overlay prop on `OverlayContainer` would take take the `Overlay` component. This would make it a lot easier to split the 2 up and hide the content intended to render behind the overlay when it is open from screen readers.

## Environment

* Component Name and Version:  terra-overlay
* Browser Name and Version: iOS 12 Mobile Safari w/ Voice Over

supreethmr:
### Summary
Issue : Voice over was able to access the content behind the overlay when it is opened.
Issue seems to happening only with 'container overlay' type. Since Overlay container was accepting single prop called children both the overlay and the other content of OverlayContainer were rendered within children div, making it available for screen readers. 

To fix this issue Added overlay prop to OverlayContainer to Separate the overlay from the container content, which will be displayed behind the overlay. So that overlay will have only overlay and its content and children will be having rest of container content. ARIA-HIDDEN attribute has been set to true in disableContainerChildrenFocus() method for children on Open of Overlay to prevent screen readers from reading the content. On Close of overlay it will be reset to false on enableContainerChildrenFocus() method.

### Additional Details
Added className variable to get className of overlay, To use it as condition to distinguish between overlay and other components of OverlayContainer. 

### Deployment URL
Shortend url so its easier to type in on mobile: https://bit.ly/2T21DN9

Resolves #2024


bjankord:
Tested with VoiceOver on iOS and this now works as intended.

packages/terra-overlay/src/OverlayContainer.jsx
```diff
@@ -7,24 +7,30 @@ import styles from './Overlay.module.scss';
 const cx = classNames.bind(styles);
 
 const propTypes = {
-/**
- * The overlay and the elements the overlay should hide when overlay isRelativeToContainer and isOpen.
- */
+  /**
+  * The Overlay and the content to be displayed within the overlay.
+  */
+  overlay: PropTypes.node,
```
emilyrohrbough: Should this only accept overlay or loading overlay elements?

emilyrohrbough:
We should add documentation on the read me for the use of this change for accessibility

packages/terra-overlay/src/OverlayContainer.jsx
supreethmr: this prop can be used for both overlay or loading overlay.  the content which is to be displayed on overlay or loading overlay will be passed as overlay to Overlay Container.

supreethmr:
> We should add documentation on the read me for the use of this change for accessibility
do we need to add documentation about overlay container on read me..? my confusion is that any new prop changes made to component are not documented in read me files before so are you referring to adding documentation for overlay container to read me or documentation for new prop that is been added to overlay container..?

emilyrohrbough:
@supreethmr if you look at the existing documentation, the usage sections and face-up terra-doc examples are using the OverlayContainer different. Since we added this props for accessibility, we should update what exists to reflect this pattern:

- https://github.com/cerner/terra-core/tree/master/packages/terra-overlay/docs
- https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/terra-dev-site/doc/example/LoadingOverlayExample.jsx#L36
- https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/terra-dev-site/doc/example/OverlayExample.jsx#L33
- https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/terra-dev-site/test/overlay/LoadingOverlay/ContainerLoadingOverlay.test.jsx#L7
- https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/terra-dev-site/test/overlay/Overlay/ContainerOverlay.test.jsx#L7
-https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/terra-dev-site/test/overlay/Overlay/ClearOverlay.test.jsx#L7
- https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/terra-dev-site/test/overlay/Overlay/CustomContentOverlay.test.jsx#L43

There are probably other places as well, but if we are inconsistently using it, it could be confusing for consumers

supreethmr:
@emilyrohrbough  updated the read-me files and examples of overlay container : [code changes](https://github.com/cerner/terra-core/pull/2249/commits/35aca247290abcc2386c13fc831c3dd9157e32ee).

packages/terra-overlay/src/terra-dev-site/test/overlay/Overlay/OnRequestCloseOverlay.test.jsx
```diff
@@ -32,8 +32,10 @@ class OverlayExample extends React.Component {
           <br />
           <p>Close by clicking inside the overlay or pressing the ESC key.</p>
         </Overlay>
-        <button type="button" id="trigger_container" onClick={this.handleTriggerOverlay}>Trigger Container Overlay</button>
-        <button type="button" id="trigger_fullscreen" onClick={this.handleTriggerFullScreenOverlay}>Trigger Fullscreen Overlay</button>
+        <div>
```
emilyrohrbough: I am confused by this change. I thought this PR added an overlay prop such that overlays should be constructed like:
```diff
+ const overlayComponent = (
+    <Overlay isOpen={this.state.show} isRelativeToContainer={this.state.relative}>
+          // content displayed in overlay
+    </Overlay>
+  );

- <OverlayContainer>
+ <OverlayContainer overlay={overlayComponent}>
-    <Overlay isOpen={this.state.show} isRelativeToContainer={this.state.relative}>
-          // content displayed in overlay
-    </Overlay>
      <button>Trigger Container Overlay</button>
      <button>Trigger Fullscreen Overlay</button>
 </OverlayContainer>
```

packages/terra-overlay/docs/README.md
```diff
@@ -1,7 +1,7 @@
 # Terra Overlay
 The Overlay component is a component that creates an semi-transparent overlay screen that blocks interactions with the elements underneath the display. There are two types of overlays: fullscreen and relative to its container.
 
-When creating as an overlay relative to its container, place the overlay component and the element(s) which should be blocked within the Overlay Container wrapper subcomponent.
+When creating as an overlay relative to its container, place the overlay component within the overaly wrapper and the element(s) which should be hidden by overlay within the content(children) wrapper of Overlay Container subcomponent.
```
JakeLaCombe: Typo: overlay

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -53,6 +53,12 @@ const propTypes = {
   zIndex: PropTypes.oneOf(zIndexes),
 };
 
+const KEYCODES = {
```
JakeLaCombe: we actually switched over to a node package for these not too long ago. I would go ahead and change that here. https://github.com/cerner/terra-core/commit/45efddb9acfa894b3e7a094b0995fbc7dd1d74a3

packages/terra-overlay/src/terra-dev-site/doc/example/OverlayExample.jsx
```diff
@@ -33,8 +33,10 @@ class OverlayExample extends React.Component {
         <Overlay isOpen={this.state.show} isRelativeToContainer={this.state.relative} onRequestClose={this.handleOnRequestESC} zIndex="6000">
           <p>Close by clicking the overlay or pressing the ESC key.</p>
         </Overlay>
-        <Button text="Trigger Container Overlay" onClick={this.handleTriggerOverlay} />
-        <Button text="Trigger FullScreen Overlay" onClick={this.handleTriggerFullScreenOverlay} />
+        <div>
```
JakeLaCombe: What is the div needed for?

packages/terra-overlay/docs/README.md
supreethmr: Thank you.  [updated](https://github.com/cerner/terra-core/pull/2249/commits/4dd8524ffaaec55fcb3f9c13fedde57d4f555908)

packages/terra-overlay/src/terra-dev-site/doc/example/OverlayExample.jsx
supreethmr: I added This div wrapper to group the elements which are not part of overlay. By doing this the number elements that will be iterated in for loop of method [disableContainerChildrenFocus()](disableContainerChildrenFocus) will be reduced since container will have only one child element div to loop. 
without wrapper div method will loop through all the elements to set attribute aria-hidden property.  
with @emilyrohrbough  comment got to know div should be added around children prop at OverlayContainer itself instead of adding it on consuming code hence will be removing div from here.

packages/terra-overlay/src/terra-dev-site/test/overlay/Overlay/OnRequestCloseOverlay.test.jsx
supreethmr: updated the examples and read me files with use of [overlay prop](https://github.com/cerner/terra-core/pull/2249/commits/4dd8524ffaaec55fcb3f9c13fedde57d4f555908)

packages/terra-overlay/src/OverlayContainer.jsx
emilyrohrbough: oh I was thinking
```
overlay: PropTypes.oneOf([PropTypes.instanceOf(Overlay), PropTypes.instanceOf(LoadingOverlay)]),
```

supreethmr: this requires importing of Overlay and LoadingOverlay which will violate the Lint rule : [import/no-cycle](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-cycle.md). still we can bypass this rule error by doing it in this way 
`overlay: PropTypes.oneOf([PropTypes.instanceOf('./Overlay'), PropTypes.instanceOf('./LoadingOverlay')]),`  but am not sure is it preferable to do it in this way or not.

emilyrohrbough: yeah lets not override the linter here. The cyclical dependency issues can get weird

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -106,6 +107,10 @@ class Overlay extends React.Component {
       for (let i = 0; i < this.containerChildren.length; i += 1) {
         prevTabIndex.push(this.containerChildren[i].tabIndex);
         this.containerChildren[i].tabIndex = -1;
+        // childern with calss name Overlay is the overlay component and it's content which should not be disabled for screen readers.
```
bjankord: nit: Change `calss name` to `class name`

supreethmr: Thank you : corrected the [typo]([url](https://github.com/cerner/terra-core/pull/2249/commits/9dee2d1f607a345680c9525867030403e74fbee6))

bjankord:
Tested this on iOS with VoiceOver. Resolves #2024. In testing, I found it also resolves #1913

PR #2263, opened 2019-03-07
[terra-list] Update documentation to better cover sections and headers.

dkasper-was-taken:
### Summary
Update documentation to better cover sections and headers.

@cerner/terra


packages/terra-list/docs/guides/MultiSelectList.md
```diff
@@ -1,6 +1,6 @@
 # Terra List - Implementing a Multi Select List
 
-In previous versions of the terra-list a multi select style list could be created as either a controlled version of SelectableList or uncontrolled version of MultiSelectList.  These implementations suffered from inflexibility and performance concerns. Going foward terra-list is more granular, though this puts more reponsibility on the consumer to properly update their list items with the appropriate state. The following is a guide to addressing those concerns in your implementation.
+The terra-list implementation requires controlled state if selections are required. As a result selections are applied at child items generation from HOC state. The following guide show you how to implement that state within a multiple items selection variant of list, as well as implement the additional narrowing requirements of a max selection for items.
```
JakeLaCombe: Just a few typos in this paragraph.

- I think the second sentence is trying to say 'As a result, selections are applied when child items are generated from HOC state.
- The following guide will show you

packages/terra-list/docs/guides/MultiSelectList.md
```diff
@@ -25,14 +25,14 @@ As a precaution we can prevent default on the event, in case the list has an anc
     event.preventDefault();
   }
 ```
-Terra list comes with additional helpers to manage state, in this case we want to determine if the selection has selected or unselected the list item key in our state. So we use the utility method "updatedMulitSelectedKeys", which returns an array of the keys following the addition or removing of the key passed. We then set this as our state.
+Terra list comes with additional helpers to manage state, in this case we want to determine if the selection has selected or unselected the list item key in our state. So we use the utility method `updatedMulitSelectedKeys`, which returns an array of the keys following the addition or removing of the key passed. We then set this as our state.
```
JakeLaCombe: I think the middle of this linke should be `manage state. In this case,`

packages/terra-list/docs/guides/MultiSelectList.md
```diff
@@ -90,7 +90,7 @@ Finally we need to check if the item is selected. As we support IE10 & 11, we ca
     );
   }
 ```
-Then we can implement a method to loop through our data and create the list item with our methods and call it from our render method. Making special note to assign the aria role of "listbox" for the list, as it is required for accessibility with selectable list options. Ensuring that we add the appropriate aria label for multiple selections.
+Then we can implement a method to loop through our data and create the list item with our methods and call it from our render method. Making special note to assign the aria role of `"listbox"` for the list, as it is required for accessibility with selectable list options. Ensuring that we add the appropriate aria label for multiple selections, `aria-multiselectable`.
```
JakeLaCombe: Can we rephrase the second sentence to say `In addition, we need to assign the aria role`

packages/terra-list/docs/guides/SectionWithinSection.md
```diff
@@ -1,6 +1,6 @@
 # Terra List - Sections with Subsections
 
-With the inclusion of sections in the list, there are two recommended patterns for created nested collapsible sections. First is collapsible sections and static subsections, and the second is static sections and collapsible subsections. The guidance from UX is to not have both sections and subsections to be collapsible, have one or the other be collapsible.
+With the inclusion of sections in the list, there are two recommended patterns for created nested collapsible sections. First is collapsible sections and static subsections, and the second is static sections and collapsible subsections. The guidance from UX is to only use collapsible sections or collapsible subsections in a list, but not both.
```
JakeLaCombe: Double check with UX to see that we include design standards in these docs? We did that in one component, and the UX team had concerns with that. @neilpfeiffer

packages/terra-list/docs/guides/SingleSelectList.md
```diff
@@ -1,6 +1,6 @@
 # Terra List - Implementing a Single Select List
 
-In previous versions of the terra-list a single select style list could be created as either a controlled version of SelectableList or uncontrolled version of SingleSelectList.  These implementations suffered from inflexibility and performance concerns. Going forward terra-list is more granular, though this puts more responsibility on the consumer to properly update their list items with the appropriate state. The following is a guide to addressing those concerns in your implementation.
+The terra-list implementation requires controlled state if selections are required. As a result selections are applied at child item generation from HOC state. The following guide show you how to implement that state within a single row selection variant of list.
```
JakeLaCombe: shows*

packages/terra-list/docs/guides/MultiSelectList.md
dkasper-was-taken: Updated.

packages/terra-list/docs/guides/MultiSelectList.md
dkasper-was-taken: Updated.

packages/terra-list/docs/guides/SingleSelectList.md
dkasper-was-taken: Updated.

packages/terra-list/docs/guides/MultiSelectList.md
dkasper-was-taken: Updated.

packages/terra-list/docs/guides/SectionList.md
```diff
@@ -0,0 +1,119 @@
+# Terra List - Sections
+
+List sections are available in both static and collapsible varieties. A list section that is collapsed will be remove its children items from the DOM. This improves performance and accessiblilty, but may remove potentional state associated to a mounted list item. If your list item content has additional state, it needs to be stored externally and reapplied when thawed for the next render that shows its section open. Since this render occurs at the same time that a collapsible section state changes, for additional optimization, collapsed sections do not necessarily need to be provided with child items.
```
bjankord: nit: Change `that is collapsed will be remove its children items` 
to `that is collapsed will remove its children items`

packages/terra-list/docs/guides/SectionWithinSection.md
```diff
@@ -1,6 +1,6 @@
...
```
bjankord: nit: `creating` instead of `created`

dkasper-was-taken: Updated.

packages/terra-list/docs/guides/SectionList.md
dkasper-was-taken: Updated.

Issue #2178, opened 2019-01-23
[terra-status-view] Remove opt object and add proptype check


yuderekyu:
# Feature Request
We export variants with `Statusview.opts.variants`, let's remove the opts object, create a named export, and add a proptype check.

```jsx
const StatusViewVariants = {
  NODATA: 'no-data',
  NOMATCHINGRESULTS: 'no-matching-results',
  NOTAUTHORIZED: 'not-authorized',
  ERROR: 'error',
  CUSTOM: 'custom',
};

variant: PropTypes.string.onOf([ 'no-data', 'no-matching-results', etc]);
...
...
export default StatusView;
export {StatusViewVariants}

PR #2264, opened 2019-03-08
[terra-status-view] Remove opt object and add proptype check

amichaelparker:
### Summary
Removes Opts object from status view component, and exports status view variants directly. See [#2178](https://github.com/cerner/terra-core/issues/2178). Jest tests updated to reflect variant access changes.

Closes [#2178](https://github.com/cerner/terra-core/issues/2178).


packages/terra-status-view/src/StatusView.jsx
```diff
@@ -53,7 +53,7 @@ const propTypes = {
    * `no-matching-results`, `not-authorized`,
    * `error`, or `custom`.
    */
-  variant: PropTypes.string,
+  variant: PropTypes.oneOf(['no-data', 'no-matching-results', 'not-authorized', 'error', 'custom']),
```
JakeLaCombe: Instead of hardcoding these values, could we do Object.values(StatusViewVariants)?

amichaelparker: Makes sense, I will check it out.

yuderekyu:
@emilyrohrbough good point. Maybe we should hold off on merging and bundle with other major changes, when they come up?

packages/terra-status-view/src/StatusView.jsx
amichaelparker: Changed: [846faa8](https://github.com/cerner/terra-core/pull/2264/commits/846faa8916a54c4b847be71184220fb9a6fc14b3)

packages/terra-status-view/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Updated status view variants object, added proptype check
```
StephenEsser: This might be worth splitting into a couple different entries when resolving the conflicts. 

Something similar to this

### Breaking Change
* Removed `StatusView.Opts` export

### Added
* Added additional variants export

amichaelparker: Changes made! [fd9c33d](https://github.com/cerner/terra-core/pull/2264/commits/fd9c33d8e3034b0c78d362541b68921ca293de36)

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -252,6 +252,5 @@ class StatusView extends React.Component {
 StatusView.propTypes = propTypes;
 StatusView.contextTypes = contextTypes;
 StatusView.defaultProps = defaultProps;
-StatusView.Opts = {};
```
bjankord: I've been thinking on this and I think it would be preferable to keep this a passive change and add documentation that StatusView.Opts is deprecated and will be removed in a future major version bump. To do that, I think we keep `StatusView.Opts.variants` in addition to exporting `StatusViewVariants`.

e.g.

```jsx
StatusView.Opts = {};
StatusView.Opts.variants = StatusViewVariants;
export default StatusView;
export { StatusViewVariants };
```

amichaelparker: Updated: [0a2be6a](https://github.com/cerner/terra-core/pull/2264/commits/0a2be6aacc3a2b6c4015aee5729091bb9562a2ab). I also modified the [changelog](https://github.com/cerner/terra-core/pull/2264/commits/47005c670792c89f5279cadb903074eef080d60c) with a note about deprecating the original object; let me know if that could use adjustment.

Issue #1510, opened 2018-05-15
Improve compile time


bjankord:
# Issue Description
Currently, we run compile commands through lerna. We could potentially save some time using globbing patterns and running babel in a script rather than using lerna's exec script feature.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Compile times are improved

## Current Behavior
Compile times have potential to be improved

bjankord:
For the command that removes lib to ensure we're always copying to clean directory, we use the following npm script: `"compile:clean": "rimraf packages/*/lib",`

I was hoping we could use globbing with babel, however when I checked with the babel team, babel-cli only outputs to 1 directory. It supports glob entry but not glob output.

To get around this we could loop through all directories and then run the cli command.

Something like, 
```
"compile:babel": "lerna ls | cut -d ' ' -f1 | xargs -I {} sh -c './node_modules/.bin/babel packages/{}/src --out-dir packages/{}/lib --copy-files'",
```
The question that needs tested will be if this is faster or if running the compile command via lerna faster.

PR #2269, opened 2019-03-14
Updated Compile Command in Package.json

amichaelparker:
### Summary
This updates the compile and postinstall commands to use the script suggested by @bjankord in the linked issue. This speeds up both individual compiles and install times.

Closes [1510](https://github.com/cerner/terra-core/issues/1510).

## Additional Information
Screenshots are on a fresh clone / install for each branch.

Pre-change:
Screen Shot 2019-03-14 at 11 29 03 AM

Post-change:
Screen Shot 2019-03-14 at 11 30 30 AM


package.json
```diff
@@ -32,7 +32,9 @@
     "clean:all": "lerna clean --yes && rm -rf node_modules",
     "clean:install": "npm run clean:all && npm install",
     "clean:obsolete-snapshots": "npm test -- -u",
-    "compile": "lerna run compile",
+    "compile": "npm run compile:clean && npm run compile:build",
+    "compile:build": "lerna ls | cut -d ' ' -f1 | xargs -I {} sh -c './node_modules/.bin/babel packages/{}/src --out-dir packages/{}/lib --copy-files'",
```
emilyrohrbough: could you explain this?

bjankord: This lists out all the terra packages, loops through them and runs babel on the src directory and outputs to the lib directory. Might be helpful to post time comparisons of running `compile` before and after this change.

amichaelparker: I'll add some screenshots to the issue.

emilyrohrbough: I agree it is nice to speed this up. out of curiosity, how does the following compare?
```
"compile": "lerna run compile --parallel"
```
For me, I notice it took 62 ms to run this vs running the original script which took 86 sec on my machine.

bjankord: Running `compile` in master today, these are the times I see on my device:
```
npm run compile  129.51s user 20.61s system 210% cpu 1:11.27 total
```

Running `"compile": "lerna run compile --parallel"`, these are the times I see on my device:
```
npm run compile  161.44s user 26.73s system 419% cpu 44.868 total
```

Running `compile` with the changes in this PR, these are the times I see on my device:
```
npm run compile  52.09s user 6.54s system 121% cpu 48.364 total
```

amichaelparker: I had looked into the --parallel option previously, but I got about what @bjankord was getting above, immediately following a fresh clone / npm install. The times are actually even higher if it runs as part of the post install step.

bjankord: I'm fine with the code as-is in the PR, it looks to be the fastest/least CPU intensive.

Issue #2210, opened 2019-02-05
[terra-form-select] onClick does not fire on initial click in multiple select box variant when an option already chosen


prmeznar:
# Bug Report

## Description
When working with a multiple variant of a select form that already has one value selected, when first selecting the component in the 'search area' (not the dropdown or the currently selected value) the dropdown generates but the onClick event does not fire.

## Steps to Reproduce
1. Build a Select component
```
 <Select variant="multiple" onClick={() => console.log('on click')}>
    <Select.Option value="blue" display="Blue" />
    <Select.Option value="green" display="Green" />
    <Select.Option value="purple" display="Purple" />
    <Select.Option value="red" display="Red" />
    <Select.Option value="violet" display="Violet" />
 </Select>
```
2. Select an initial value
3. Move off the component to lose focus
4. Click next to the selected value - dropdown shows, onclick is not printed

## Additional Context / Screenshots
The area with the -Select- placeholder text is what I'm referring to:
<img width="1067" alt="image" src="https://user-images.githubusercontent.com/47360296/52285397-db3c3780-292b-11e9-9629-1b95c2c13b97.png">

Note: A subsequent click in the same position will generate onClick

## Expected Behavior
onClick to occur on the initial call

## Possible Solution
Not sure.

## Environment

* Component Name and Version: terra-form-select
* Browser Name and Version: Chrome

## @ Mentions
@StephenEsser

PR #2270, opened 2019-03-14
[terra-form-select] - Fix onClick not firing

BenBoersma:
### Summary
Fixes https://github.com/cerner/terra-core/issues/2210.

@cerner/terra


StephenEsser:
I did notice one strange behavior with Firefox. The onClick will fire if you click inside of the dropdown, but the other browsers will not. This is probably a separate issue.

Issue #2271, opened 2019-03-14
Remove terra-form-select dependency from terra-signature and terra-avatar


bjankord:
# Issue Description
terra-form-select is used in doc examples for terra-signature and terra-avatar. We should decouple this and switch to using native selects for these examples to lower the dependencies these packages need.

* https://github.com/cerner/terra-core/blob/7de1f94fb3312ef46ef1250c1623094fda9bf216/packages/terra-signature/package.json#L32
* https://github.com/cerner/terra-core/blob/7de1f94fb3312ef46ef1250c1623094fda9bf216/packages/terra-avatar/package.json#L32

PR #2272, opened 2019-03-14
Removed terra-form-select dependencies

AlexBrizi:
### Summary
Removed terra-form-select dependencies from terra-avatar and terra-signature components.

Resolves [#2271](https://github.com/cerner/terra-core/issues/2271)

### DIFF
--- a/packages/terra-avatar/package.json
+++ b/packages/terra-avatar/package.json
@@ -29,7 +29,6 @@
     "classnames": "^2.2.5",
     "prop-types": "^15.5.8",
     "terra-doc-template": "^2.5.0",
-    "terra-form-select": "^5.9.0",
     "terra-image": "^3.3.0",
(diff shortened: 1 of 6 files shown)


packages/terra-avatar/package.json
```diff
@@ -29,7 +29,6 @@
     "classnames": "^2.2.5",
     "prop-types": "^15.5.8",
     "terra-doc-template": "^2.5.0",
-    "terra-form-select": "^5.9.0",
```
StephenEsser: There are a couple examples in each of these packages that are using the terra-form-select that will need updated to use the native select.

AlexBrizi: Yup, that's the plan. Was just pushing the dependency removal first, sorry for the confusion!

PR #2273, opened 2019-03-14
Form field aria describedby update

### LINKED ISSUE #2267: Add additional ARIA attributes for help text and error text to InputField, TextAreaField, and SelectField
# Issue Description
This is related to #1515. This issue contains work that can be done passively now without requiring a major version bump.

Add additional ARIA attributes like `aria-labelledby` and `aria-describedby` to helptext and other field elements as needed in all specifically created Form Field Compositions (InputField, TextAreaField, SelectField, etc) for better accessibility. Key area of focus are the Label, Required, Help Text, and Error Text.  

(Note: doesn't apply to Terra-Form-Field since component is abstract without context of which input might be provided, unless using a RenderProps method -- future consideration)

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Should be similar to existing work already done for form-fields in Legacy Terra et. al. 
Help Text References: 
- https://www.w3.org/WAI/tutorials/forms/instructions/
- https://www.w3.org/TR/WCAG20-TECHS/ARIA1.html
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/forms/Basic_form_hints
- https://material.io/develop/web/components/input-controls/text-field/helper-text/

Error Text References:
- https://www.w3.org/WAI/GL/wiki/Using_ARIA_role_of_alert_for_Error_Feedback_in_Forms
- https://www.w3.org/WAI/GL/wiki/Using_Aria-Invalid_to_Indicate_An_Error_Field

bjankord:
### Summary
Resolves #2267


packages/terra-form-select/docs/README.md
```diff
@@ -7,19 +7,6 @@ The Select component provides a dropdown of selectable options.
 - Install with [npmjs](https://www.npmjs.com):
   - `npm install terra-form-select`
 
-## Usage
```
bjankord: Removing this as it is covered in the code samples on the page.

packages/terra-form-select/docs/SelectField.md
```diff
@@ -7,19 +7,6 @@ A convenience wrapper assembling a [terra-form-select](https://github.com/cerner
 - Install with [npmjs](https://www.npmjs.com):
   - `npm install terra-form-select`
 
-## Usage
-
-```jsx
-import React from 'react';
```
bjankord: Removing this as it is covered in the code samples on the page.

packages/terra-form-textarea/docs/README.md
```diff
@@ -7,19 +7,6 @@ Element for building out textareas in a form.
 - Install with [npmjs](https://www.npmjs.com):
   - `npm install terra-form-textarea`
 
-## Usage
-
-```jsx
```
bjankord: Removing this as it is covered in the code samples on the page.

packages/terra-form-textarea/docs/TextareaField.md
```diff
@@ -8,22 +8,6 @@
 - Install with [npmjs](https://www.npmjs.com):
   - `npm install terra-form-textarea`
 
-## Usage
-
-```jsx
-import React from 'react';
```
bjankord: Removing this as it is covered in the code samples on the page.

packages/terra-form-field/src/Field.jsx
```diff
@@ -131,6 +133,48 @@ const Field = (props, { intl }) => {
     labelAttrs.className,
   ]);
 
+  // Checks to run when not in production
+  if (process.env.NODE_ENV !== 'production') {
```
tbiethman: Are there webpack configuration requirements for this to work?

JakeLaCombe: I take it we will remove this in the next major version of this component. Should we leave a comment noting to remove this section at a later date?

packages/terra-form-field/src/Field.jsx
```diff
@@ -131,6 +133,48 @@ const Field = (props, { intl }) => {
...
+  let helpTextId;
+  let errorTextId;
+
+  if (htmlFor) {
+    /**
+     * IE + JAWS has trouble reading aria-describedby content with our form components.
+     * In that browser, we don't want an ID for the aria-describedby content so we have a
+     * Microsoft specific feature detect to flex on if the ID is undefined or valid.
+     */
+    helpTextId = !window.navigator.msPointerEnabled ? `${htmlFor}-help` : undefined;
```
JakeLaCombe: This looks like its going to be removed in the future. What is our next plan? https://technet.microsoft.com/en-us/windows/hh972895(v=vs.60)

packages/terra-form-field/src/Field.jsx
bjankord: Hmm, I believe so to some extent. If you run webpack without the NODE_ENV var being set to production, you'll see these warnings. We have a similar checks in the following components:
* https://github.com/cerner/terra-core/blob/45efddb9acfa894b3e7a094b0995fbc7dd1d74a3/packages/terra-section-header/src/SectionHeader.jsx#L91-L94
* https://github.com/cerner/terra-core/blob/96bcacb151ac6738df253d0ac1d7d28c1ec98139/packages/terra-status/src/Status.jsx#L35
* https://github.com/cerner/terra-core/blob/728a5106f1e3fb113b3956cefeca996468b0ffb8/packages/terra-i18n/src/i18nLoader.js#L10
* https://github.com/cerner/terra-core/blob/728a5106f1e3fb113b3956cefeca996468b0ffb8/packages/terra-base/src/baseStyles.js#L4

bjankord: Yeah thats a good idea, I'll log an issue as well to reference in the code comment.

tbiethman: 👍 I've been looking to do something similar for dev-only warnings for some of the Context problems people were running into

bjankord: Update in this commit: https://github.com/cerner/terra-core/pull/2273/commits/10798ad56d452ec1cec05272bbe5314cefc07517
Logged to follow up: https://github.com/cerner/terra-core/issues/2290

packages/terra-form-field/src/Field.jsx
bjankord: Updated this logic to be more explicit: https://github.com/cerner/terra-core/pull/2273/commits/c56e155b4d07f3cdd3b2574815aea38b2139af11

packages/terra-form-field/src/Field.jsx
bjankord: I think we picked this pattern of using `if (process.env.NODE_ENV !== 'production')` for these types of warnings [from react](https://github.com/facebook/react/blob/baff5cc2f69d30589a5dc65b089e47765437294b/packages/react-dom/npm/index.js#L31). It has seemed to work well so far. I think using it for the context problems makes perfect sense.

PR #2284, opened 2019-03-15
Add terra enzyme intl

bjankord:
### Summary
* Updated terra-core Jest tests to use terra-enzyme-intl.
* Updated snapshots accordingly


packages/terra-action-header/package.json
```diff
@@ -32,7 +32,8 @@
     "terra-button-group": "^3.6.0",
     "terra-doc-template": "^2.5.0",
     "terra-icon": "^3.4.0",
-    "terra-mixins": "^1.28.0"
+    "terra-mixins": "^1.28.0",
+    "terra-enzyme-intl": "git://github.com/cerner/terra-enzyme-intl#add-enzyme-intl-helpers"
```
emilyrohrbough: This is a test dependency. We should be able to pull this in at the root level

bjankord: Updated in this commit: https://github.com/cerner/terra-core/pull/2284/commits/509422cc01a6d69da40e5f73ea76feddee645f81

PR #2285, opened 2019-03-18
added aria-label for selected rows to resolve the voice over issue #1839 

### LINKED ISSUE #1839: Terra Table doesn't explicitly say what rows are selectable on a screen reader.
# Bug Report

## Description
Screen readers read selected table rows just fine. Selectable rows however are missed by the screen reader. We need to add the approrpriate labels and rows to prevent this.

supreethmr:
### Summary
Voice-over was not able to recognize selected rows while reading the table contents due to missing roles on Table rows and headers. To resolve this issue ARIA Label has been added to selectable rows along with appropriate ARIA Roles to table row and table header. 
content for Aria-Label is being populated dynamically with cell values of the selected row . By this approach now screen reader reads pre-selected row as 'pre-selected row  along with its cell contents ' and also reads the selected rows as ' row selected '.

### Additional Details
Since there are no pre-defined roles to support interactions with table while using voice-over it is being handled with the below methods :

1. [handleOnChange()](https://github.com/cerner/terra-core/blob/f4e469fc37e008087364c2d01dfcabee2fe1b83f/packages/terra-table/src/SelectableTableRows.jsx#L72) : code as been added to add aria label for row on handleOnChange event. Row can be selected by mouse click and also with enter and space keys since both wrappedClick and wrappedKeyDown calls handleOnChange() event code to add aria label to selected row has been added [here](https://github.com/cerner/terra-core/blob/f4e469fc37e008087364c2d01dfcabee2fe1b83f/packages/terra-table/src/SelectableTableRows.jsx#L72)

2. [wrappedOnBlur()](https://github.com/cerner/terra-core/blob/f4e469fc37e008087364c2d01dfcabee2fe1b83f/packages/terra-table/src/SelectableTableRows.jsx#L85) is added to remove the aria-label from row when it is unselected. since screen reader picks the content from aria-label it should be removed from the row to make the content of cells to be available for screen reader.

aria-live is used to identify Dynamic changes ( adding aria-label attribute ) to table row.

Resolves #1839


packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -31,24 +35,77 @@ const propTypes = {
...
+    for (let rowIndex = 0; rowIndex < rowGroup.children.length; rowIndex += 1) {
+      const preselectedRow = rowGroup.children[rowIndex];
+      let ariaLabel = 'pre-selected row';
+      if ((preselectedRow.hasAttribute('aria-selected')) && (preselectedRow.getAttributeNode('aria-selected').value === 'true')) {
+        for (let columnIndex = 0; columnIndex < preselectedRow.children.length; columnIndex += 1) {
+          ariaLabel += ` ${preselectedRow.children[columnIndex].innerText}`;
+        }
+        rowGroup.children[rowIndex].setAttribute('aria-label', ariaLabel);
+      }
+    }
```
ShettyAkarsh: You can have this logic in a separate sub method and call that method from here.

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -31,24 +35,77 @@ const propTypes = {
...
+  wrappedOnBlur(row, index) {
+    const initialOnBlur = row.props.onBlur;
+    // removes aria-label from unselected row
+    return (event) => {
+      if (row.props.isSelectable !== false) {
+        this.handleOnBlur(event, index);
+      }
+
+      const rowGroup = document.querySelector(`#${rowsSelector}`);
+      const previousRow = rowGroup.children[index];
```
ShettyAkarsh: You can rename the variable as selectedRow and rowGroup as SelectedRowGroup.

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -31,24 +35,77 @@ const propTypes = {
...
     this.handleOnChange = this.handleOnChange.bind(this);
+    this.handleOnBlur = this.handleOnBlur.bind(this);
+  }
+
+
+  componentDidMount() {
+    // Adds ARIA-Label for Pre-selected rows
+    const rowGroup = document.querySelector(`#${rowsSelector}`);
+    for (let rowIndex = 0; rowIndex < rowGroup.children.length; rowIndex += 1) {
+      const preselectedRow = rowGroup.children[rowIndex];
```
nramamurth: Should this simply be called `row`? There is no guarantee that it is pre-selected at this point.

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -31,24 +35,77 @@ const propTypes = {
...
     super(props);
     this.handleOnChange = this.handleOnChange.bind(this);
+    this.handleOnBlur = this.handleOnBlur.bind(this);
+  }
+
+
+  componentDidMount() {
+    // Adds ARIA-Label for Pre-selected rows
+    const rowGroup = document.querySelector(`#${rowsSelector}`);
+    for (let rowIndex = 0; rowIndex < rowGroup.children.length; rowIndex += 1) {
```
nramamurth: I'd calculate the length before the loop and avoid calculating it during each iteration unless doing it this way gives an advantage.

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -122,12 +181,12 @@ class SelectableTableRows extends React.Component {
...
   render() {
     const {
-      children, disableUnselectedRows, onChange, selectedIndexes, ...customProps
+      children, disableUnselectedRows, onChange, onBlur, selectedIndexes, ...customProps
     } = this.props;
     const clonedChildItems = this.clonedChildItems(children);
 
     return (
-      <TableRows {...customProps}>
+      <TableRows id={rowsSelector} {...customProps}>
```
nramamurth: Please correct me if I am wrong - if the page has multiple tables, then are we adding the same id to the `TableRows` in all the tables?

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -31,24 +35,77 @@ const propTypes = {
...
+  }
+
+
+  componentDidMount() {
+    // Adds ARIA-Label for Pre-selected rows
+    const rowGroup = document.querySelector(`#${rowsSelector}`);
+    for (let rowIndex = 0; rowIndex < rowGroup.children.length; rowIndex += 1) {
+      const preselectedRow = rowGroup.children[rowIndex];
+      let ariaLabel = 'pre-selected row';
+      if ((preselectedRow.hasAttribute('aria-selected')) && (preselectedRow.getAttributeNode('aria-selected').value === 'true')) {
```
nramamurth: I'd directly use `getAttribute()` and that way avoid performing two checks if it works out.

packages/terra-table/src/SelectableTableRows.jsx
StephenEsser: Base level components within Terra should rarely have IDs because there can generally be multiple on the same page. Those components that do have them generally require a unique identifier to be generated unless specific constraints enforce that only one is ever present on a single page. 

It should be possible to add the aria labels prior to rendering the table into the dom. Potentially within the `clonedChildItems` function.

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -31,24 +35,77 @@ const propTypes = {
...
   constructor(props) {
     super(props);
     this.handleOnChange = this.handleOnChange.bind(this);
+    this.handleOnBlur = this.handleOnBlur.bind(this);
+  }
+
+
+  componentDidMount() {
+    // Adds ARIA-Label for Pre-selected rows
+    const rowGroup = document.querySelector(`#${rowsSelector}`);
```
StephenEsser: This will query the entire dom, if there are multiple tables this logic will affect all of them. It will also run for every table on the page each time a new table mounts. 

We should investigate a solution that adds the aria labels before the table is rendered into the dom.

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -31,24 +35,77 @@ const propTypes = {
...
   }
 
   handleOnChange(event, index) {
     if (this.props.onChange) {
       this.props.onChange(event, index);
     }
+
+    if (!document.activeElement.hasAttribute('aria-label')) {
+      // Aria-Label is being populated with the cell contents for screen reader to read it as whole instead of cell value.
+      const ariaLabel = 'row selected';
```
StephenEsser: If this is the text the screen reader will announce it will need to be translated.

This PR is a little large, but it has good examples of dynamically announcing selections: https://github.com/cerner/terra-core/pull/2236

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -31,24 +35,77 @@ const propTypes = {
...
+      }
+    }
   }
 
   handleOnChange(event, index) {
     if (this.props.onChange) {
       this.props.onChange(event, index);
     }
+
+    if (!document.activeElement.hasAttribute('aria-label')) {
```
nramamurth: Are we doing more than what `handleOnChange` is intended to do? I'm just curious if this work could instead be part of `wrappedOnClickForRow`.

nramamurth: What happens if I select a row that already has some other value set to `aria-label`?

packages/terra-table/src/SelectableTableRows.jsx
supreethmr: having methods without `this` : throws lint error : class-methods-use-this. hence having this line of code inside componentDidMount() method and only static class methods inside class can be declared without `this` .

packages/terra-table/src/SelectableTableRows.jsx
supreethmr: wrappedOnClickForRow will invoke only for mouse click but there are other ways for selecting rows like by enter key / space key press. since handle on change will called by both onclick and onkeydown adding aria-label row selected on handle on change will work for both the cases.

supreethmr: this check is added to prevent adding aria-label for pre-selected rows  and for the row which has already got aria-label with it. This will happen when click on same row again. 

Since there is no default aria-label is added to rows initially other than for pre-selected rows, row will not be having aria-label with some other value as of now. 

Aria-Label for preselected-rows will not be overridden since pre-selected rows cannot be selected again on click or on key down.

packages/terra-table/src/SelectableTableRows.jsx
supreethmr: @StephenEsser  I tried adding aria-label attribute in clonedChildItems but there are no ref added to row to access dom element. it throws below warning when i try to add ref with [newProps](https://github.com/cerner/terra-core/blob/f4e469fc37e008087364c2d01dfcabee2fe1b83f/packages/terra-table/src/SelectableTableRows.jsx#L163)  ' Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()? ' it will be helpful to know if there are any other ways to add ARIA attribute to react constants or to access dom element from row prop in [newPropsForRow()](https://github.com/cerner/terra-core/blob/f4e469fc37e008087364c2d01dfcabee2fe1b83f/packages/terra-table/src/SelectableTableRows.jsx#L138) method

packages/terra-table/src/SelectableTableRows.jsx
supreethmr: Fix is suppose to add aria-label for all the tables in page which has pre-selected rows. I agree that it will run for every table on page each time a new table mounts but that is how it is supposed to be since it should add aria-label to every table on page having pre-selected row within it. 

As you mentioned in other comment this can be done at [clonedChildItems](https://github.com/cerner/terra-core/blob/f4e469fc37e008087364c2d01dfcabee2fe1b83f/packages/terra-table/src/SelectableTableRows.jsx#L169) to add aria attribute before tables are mounted. 

I think this can be done by adding aria attribute as prop to react component. but I would like to know if possible how this can be done or if there are any better way to do this.

packages/terra-table/src/SelectableTableRows.jsx
supreethmr: [updated to selected row](https://github.com/cerner/terra-core/blob/153a90093ca106ce109d4c4d0ab861ec611985b2/packages/terra-table/src/SelectableTableRows.jsx#L77) 👍

packages/terra-table/src/SelectableTableRows.jsx
supreethmr: adding aria-labels prior to rendering the table in [clonedChildItems](https://github.com/cerner/terra-core/blob/153a90093ca106ce109d4c4d0ab861ec611985b2/packages/terra-table/src/SelectableTableRows.jsx#L165).

replaced ID with data-attribute in [selected rows](https://github.com/cerner/terra-core/blob/153a90093ca106ce109d4c4d0ab861ec611985b2/packages/terra-table/src/SelectableTableRows.jsx#L181)

 @StephenEsser Thank you for suggesting

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -31,24 +35,60 @@ const propTypes = {
...
+
+  wrappedOnBlur(row, index) {
+    const initialOnBlur = row.props.onBlur;
+    // removes aria-label from unselected row
+    return (event) => {
+      if (row.props.isSelectable !== false) {
+        this.handleOnBlur(event, index);
+      }
+
+      const rowGroup = document.querySelector('[data-terra-row-group]');
```
nramamurth: This affects the first table with this data attribute on the page and there is no guarantee that it is operating on the table that we want it to operate upon.

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -21,6 +21,10 @@ const propTypes = {
    * A callback function to execute when the selection state changes. The first parameter is the event. The second parameter is the selectedIndexes.
    */
   onChange: PropTypes.func,
+  /**
+   * A callback function to execute when the selected row loses focus. The first parameter is the event. The second parameter is the index of the row losing focus.
+   */
+  onBlur: PropTypes.func,
```
nramamurth: What is the need to introduce this new prop? Could we have done we did for `onClick`?

packages/terra-table/src/SelectableTableRows.jsx
supreethmr: it has been removed and being referred as row [here](https://github.com/cerner/terra-core/blob/cc6ee3fad6bea01cc24bb5ce812e6483291d781d/packages/terra-table/src/SelectableTableRows.jsx#L173)

packages/terra-table/src/SelectableTableRows.jsx
supreethmr: i was not sure that it calculates the length at each iteration. assumed that it may be called only one time during initial loop. 
added const to hold length: [here](https://github.com/cerner/terra-core/blob/cc6ee3fad6bea01cc24bb5ce812e6483291d781d/packages/terra-table/src/SelectableTableRows.jsx#L68)

packages/terra-table/src/SelectableTableRows.jsx
supreethmr: added translations for  [on click](https://github.com/cerner/terra-core/blob/cc6ee3fad6bea01cc24bb5ce812e6483291d781d/packages/terra-table/src/SelectableTableRows.jsx#L67) and for [pre selected row](https://github.com/cerner/terra-core/blob/cc6ee3fad6bea01cc24bb5ce812e6483291d781d/packages/terra-table/src/SelectableTableRows.jsx#L172)

packages/terra-table/src/SelectableTableRows.jsx
supreethmr: yes..! found that while testing with multiple tables it picks the first row group always. fixed this by moving code to handleOnBlur which provides event parameter which holds the selected row as value for 'currentTarget' Property. so that current row can be retrieved by referring to event.currentTarget : [code](https://github.com/cerner/terra-core/blob/cc6ee3fad6bea01cc24bb5ce812e6483291d781d/packages/terra-table/src/SelectableTableRows.jsx#L81)

packages/terra-table/src/SelectableTableRows.jsx
supreethmr: we need blur event to get unselected rows because onChange will be called in onClick method previously selected row loses its focus when new row is selected so calling on-change from wrappedBlur event will gives the row which is newly selected instead of row which lost focus.

packages/terra-table/package.json
```diff
@@ -20,9 +20,14 @@
     "url": "https://github.com/cerner/terra-core/issues"
   },
   "homepage": "https://github.com/cerner/terra-core#readme",
+  "devDependencies": {
+    "react-intl": "^2.3.0"
+  },
   "peerDependencies": {
     "react": "^16.8.5",
     "react-dom": "^16.8.5"
```
jmsv6d: You're missing a comma here.

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -30,24 +34,73 @@ const propTypes = {
...
+};
+
 class SelectableTableRows extends React.Component {
   constructor(props) {
     super(props);
     this.handleOnChange = this.handleOnChange.bind(this);
+    this.handleOnBlur = this.handleOnBlur.bind(this);
   }
 
+
```
jmsv6d: nit: extra blank line

supreethmr: updated 👍

mmalaker:
I've tested all of the deployed examples: https://terra-core-deployed-pr-2285.herokuapp.com/#/components/terra-table/table/table
I'm not seeing any change in the selected state behavior on screen readers. VoiceOver does not announce selected rows. VO also ignores text in table headers and sub-headers. JAWS will read the selected state of rows on the first pass through the table; however, when navigating back to a selected item, the selected state is not announced.

supreethmr:
> I've tested all of the deployed examples: https://terra-core-deployed-pr-2285.herokuapp.com/#/components/terra-table/table/table
> I'm not seeing any change in the selected state behavior on screen readers. VoiceOver does not announce selected rows. VO also ignores text in table headers and sub-headers. JAWS will read the selected state of rows on the first pass through the table; however, when navigating back to a selected item, the selected state is not announced.

Voice over picking the aria-label changes bit slow. it is reading the row selected after reading the whole content row. In the below screenshots we can see that voice over initially just reads the content of row and then it reads aria-label content 'row selected'  which has been added in onSelectedRowChange event. 

Screen Shot 2019-04-03 at 5 19 13 PM

Screen Shot 2019-04-03 at 5 19 22 PM

Voice over and JAWS treats aria-label differently when aria-label is present on row voice over reads both cell content and aria-label content but Jaws will read only the aria-label content. 

Since I am adding cell content of selected row to aria-label voice over reading the same content twice. By just adding text 'Row selected' to aria label without cell contents would resolve the voice over reading contents twice but then In jaws it reads just the content of aria-label by ignoring cell contents that results in jaws being reading rows as just 'row selected' on each time it is selected.

JakeLaCombe:
I also noticed that when I use voiceover, I am unable to select a row.

I think to get this working, we are going to need to rewrite several components inside the table.

- Every table row needs to have a subheader added with a checkbox to indicate that the row is selectable, and we should add on visually hidden text so that the screen reader can describe the row as selectable.
- Every table column needs to be given the role="button" so that it can be triggered with the enter key, and maybe the space. When I used the table navigator on my test code, it felt that even in table mode, pressing space would scroll you on the screen. Tables are non interactive elements, so I would not expect space to do anything interactionable.
- Table will need a non passive change that passed in isSelectable at the top level. That way, an extra table header can be added to the table header row that is the selectable column. 

One thing I was not able to get working right with the table is that when I move between different rows on the table that my visual cursor does not move with it. I find this very strange behavior with vioceover, but it could be because the table is not interactible that is behavior occurs. You could also maybe add tabIndex of 0 to all the rows, but that could throw off keyboard navigation for non screen reader keyboard users.

ryanthemanuel:
This one seems to be more complicated than initially thought and will likely require UX accessibility input. Let’s hold up on this one for now.

bjankord:
> This one seems to be more complicated than initially thought and will likely require UX accessibility input. Let’s hold up on this one for now.

Agreed, I think this is going to be more challenging than originally anticipated especially since HTML tables don't have any semantics of "selected" row. I am optimistic that we can find a solution but I agree, let's hold up on this. I think we need to do some spike work investigating other tables that have selection state across the web and see how they work with screen readers to see if there is any techniques we can pick up that work well for this pattern.

Planning to close the PR for now but will leave the branch in case we want to use it as a starting point when were ready to get back to this.

Issue #2282, opened 2019-03-15
Lerna hoist warnings


bjankord:
# Issue Description
The following is a warning we see after running npm install related to how lerna is hoisting dependencies versions.

```
lerna WARN EHOIST_ROOT_VERSION The repository root depends on babel-jest@^23.0.1, which differs from the more common babel-jest@^19.0.0.
lerna WARN EHOIST_PKG_VERSION "terra-i18n" package depends on babel-jest@^19.0.0, which differs from the hoisted babel-jest@^23.0.1.
lerna WARN EHOIST_ROOT_VERSION The repository root depends on glob@^7.1.1, which differs from the more common glob@^7.1.2.
lerna WARN EHOIST_PKG_VERSION "terra-props-table" package depends on glob@^7.1.2, which differs from the hoisted glob@^7.1.1.
lerna WARN EHOIST_ROOT_VERSION The repository root depends on react-test-renderer@^16.4.2, which differs from the more common react-test-renderer@^16.5.2.
lerna WARN EHOIST_PKG_VERSION "terra-show-hide" package depends on react-test-renderer@^16.5.2, which differs from the hoisted react-test-renderer@^16.4.2.
lerna WARN EHOIST_PKG_VERSION "terra-search-field" package depends on react-intl@^2.3.0, which differs from the hoisted react-intl@^2.4.0.
lerna WARN EHOIST_PKG_VERSION "terra-status-view" package depends on react-intl@^2.3.0, which differs from the hoisted react-intl@^2.4.0.
```

We should update our dependencies across terra-core so hoisting them is consistent.

bjankord:
This has been resolved in terra-base@5.0.0/terra-i18n@4.0.0.

Issue #1982, opened 2018-10-22
Improve usage of passed along react-intl modules


bjankord:
# Feature Request

## Description
We've started exposing react-intl modules from terra-base. One of the issues with this is related to how we expose the react-intl modules. We import modules from react intl and then export them from [this file](https://github.com/cerner/terra-core/blob/master/packages/terra-base/src/Base.jsx). This is then consumed like so: `import { injectIntl, intlShape } from 'terra-base';`.

This pulls in and executes more code than just `injectIntl` and `intlShape`, but it's not immediately obivous. This executes [this file](https://github.com/cerner/terra-core/blob/master/packages/terra-base/src/Base.jsx), which executes [this line](https://github.com/cerner/terra-core/blob/master/packages/terra-base/src/Base.jsx#L4), which exectus [this file from terra-i18n](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/src/I18n.js) which executes our i18n logic.

If we want to continue exporting react-intl modules from terra-base, we should look at ways to do this that don't require the full terra-base and terra-i18n files to execute when importing the react-intl modules.

## Additional Context / Screenshots
Consumers noticed this issue in a recent release of terra-application-header-layout where we added this line `import { injectIntl, intlShape } from 'terra-base';`.

bjankord:
# Tech Design
Split react-intl into separate file exports under terra-base.

This allows terra-base to remain 1 dependency used to manage react-intl while allowing individual modules to be imported without pulling in all of terra-base.

bjankord:
Discussed with @tbiethman about this issue. We think a better path forward would be to avoid passing react-intl modules through terra-i18n / terra-base and instead manage the version of react-intl through peerDependencies in components that need intl similar to how we manage the version of react that components use.

With this, each package that needs modules from react-intl would import directly from react-intl and set a peerDependency on react-intl. The react-intl modules that are passed through terra-i18n and terra-base should be removed in a major version bump.

emilyrohrbough:
+1 on peer-dependencies.

noahbenham:
Peer deps would also allow consumers to begin using the new context API sooner once [this PR](https://github.com/yahoo/react-intl/pull/1186) merges.

Issue #1941, opened 2018-10-01
Remove post install script from terra-i18n


bjankord:
# Feature Request

## Description
The post-install script in terra-i18n is deprecated. Users should switch to the [terra-toolkit aggregate-translations pre-build tool](https://github.com/cerner/terra-toolkit/blob/master/docs/AggregateTranslations.md).

## Additional Context / Screenshots
Removing this will be a major version bump to terra-i18n

We've removed this once and it caused breaking issues with some of our consumers, namely around Jest testing. We'll want to document how to test with Jest once we no longer write files via post-install lifecycle hook.

@emilyrohrbough had good ideas around looking into running aggregate-translations pre-build tool in Jest setup and providing a terra Jest config to resolve the files generated from the  aggregate-translations pre-build tool

bjankord:
This has been resolved in terra-base@5.0.0/terra-i18n@4.0.0.
If consumers run into issues with Jest tests, they should [consult our recommendations on testing terra components that have translations with Jest here].(https://engineering.cerner.com/terra-ui/#/getting-started/terra-ui/internationalization/testing-components-which-include-translations)

Issue #1940, opened 2018-10-01
Remove aggregate-translations script from terra-i18n


bjankord:
# Feature Request

## Description
The aggregate-translations script in terra-i18n is deprecated. Users should switch to the [terra-toolkit aggregate-translations pre-build tool](https://github.com/cerner/terra-toolkit/blob/master/docs/AggregateTranslations.md).

## Additional Context / Screenshots
Removing this will be a major version bump to terra-i18n

bjankord:
This has been resolved in terra-base@5.0.0/terra-i18n@4.0.0.

PR #2288, opened 2019-03-19
Update terra-base and terra-i18n packages

bjankord:
### Summary
BREAKING CHANGES

* Resolves https://github.com/cerner/terra-core/issues/1149 - Error when trying to use terra-base without locale / translations
* Resolves https://github.com/cerner/terra-core/issues/1720 - Uplift i18n code to use dynamic imports
* Resolves https://github.com/cerner/terra-core/issues/1941 - Remove post install script from terra-i18n
* Resolves https://github.com/cerner/terra-core/issues/1940 - Remove aggregate-translations script from terra-i18n
* Resolves https://github.com/cerner/terra-core/issues/1982 - Improve usage of passed along react-intl modules 
* Resolves https://github.com/cerner/terra-core/issues/2282 - Lerna hoist warnings


packages/terra-base/package.json
```diff
@@ -22,7 +22,8 @@
   "homepage": "https://github.com/cerner/terra-core#readme",
   "peerDependencies": {
     "react": "^16.4.2",
-    "react-dom": "^16.4.2"
+    "react-dom": "^16.4.2",
+    "react-intl": "^2.4.0"
```
emilyrohrbough: We should bump to ^2.8.0

packages/terra-i18n/docs/AggregateTranslations.md
```diff
@@ -1,5 +1,5 @@
 # Aggregating Translations
 
-_The terra-i18n `aggregate-translations` pre-build tool is deprecated._
+_The terra-i18n `aggregate-translations` pre-build tool has been removed from this package._
```
emilyrohrbough: Lets remove all docs & link to terra-toolkit. It risk of becoming obsolete (it probably already is).

packages/terra-i18n/docs/UPGRADEGUIDE.md
```diff
@@ -1,5 +1,36 @@
...
 
+## Changes from 3.0 to 4.0
+
+The terra-i18n `aggregate-translations` pre-build tool has been removed from this package.
+
+Along with this, the post-install script which wrote default intl and translation loaders has been removed.
+
+Use the [`aggregate-translations` pre-build tool from terra-toolkit instead.](https://github.com/cerner/terra-toolkit/blob/master/docs/AggregateTranslations.md)
+
+If you require translations in your Jest tests, create a `jestglobalsetup.js` file in your app.
```
emilyrohrbough: I'm not sure if the jest setup information is relevant here. It seems like this should be included in the terra-ui testing i18n components page.

packages/terra-i18n/src/i18nSupportedLocales.js
```diff
@@ -1 +1 @@
-module.exports = ['ar', 'en', 'en-US', 'en-GB', 'es', 'es-US', 'es-ES', 'de', 'fi-FI', 'fr', 'fr-FR', 'nl', 'nl-BE', 'pt', 'pt-BR', 'sv', 'sv-SE'];
+module.exports = ['en', 'en-US', 'en-GB', 'en-AU', 'es', 'es-US', 'es-ES', 'de', 'fi-FI', 'fr', 'fr-FR', 'nl', 'nl-BE', 'pt', 'pt-BR', 'sv', 'sv-SE'];
```
emilyrohrbough: Do we need this file now? I know we used it to aggregate translations, but do we use it directly in the i18n Component?

packages/terra-i18n/src/intl.js
```diff
@@ -0,0 +1,4 @@
+import IntlPolyfill from 'intl';
```
emilyrohrbough: Does this dependency need to be added to the package.json? I am curious, what does this Polyfill provide? Is this something we should be adding to webpack instead of a global? https://www.npmjs.com/package/intl

packages/terra-i18n/src/intl.js
```diff
@@ -0,0 +1,4 @@
...
+
+const exportedIntl = typeof (Intl) === 'undefined' ? IntlPolyfill : global.Intl;
```
emilyrohrbough: @tbiethman you and I have been working on removing the global XFC var in apps. Do you think a global is desired here?

packages/terra-base/src/Base.jsx
```diff
@@ -107,6 +105,3 @@ Base.propTypes = propTypes;
 Base.defaultProps = defaultProps;
```
emilyrohrbough: With these changes, weren't we going to remove the optional locale prop? IMO we make 'en' the default locale since its our fallback

bjankord: [Last we had talked on the issue, the plan was to make the `locale` prop required.](https://github.com/cerner/terra-core/issues/1149#issuecomment-408193958)

packages/terra-i18n/src/intl.js
bjankord: `global` could be replaced with `window`.

packages/terra-i18n/src/intl.js
bjankord: This dependency is [currently listed in the package.json](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/package.json#L39).

From the Intl.js readme:
> In December 2012, ECMA International published the first edition of Standard ECMA-402, better known as the ECMAScript Internationalization API. This specification provides the framework to bring long overdue localization methods to ECMAScript implementations.
All modern browsers, have implemented this API. Intl.js fills the void of availability for this API in legacy browsers. It will provide the framework as described by the specification, so that developers can take advantage of the native API in environments that support it, or Intl.js for legacy or unsupported environments.

My preference would be to leave the intl dependency in terra-i18n as-is.

packages/terra-i18n/src/i18nSupportedLocales.js
bjankord: I debated with myself what to do with this. It is still used directly in the i18n component and it's related tests. With the code changes on the PR so far, this module is only used in the following files:
* Used to throw a warning the locale is not supported - https://github.com/cerner/terra-core/blob/728a5106f1e3fb113b3956cefeca996468b0ffb8/packages/terra-i18n/src/i18nLoader.js
* Used to test fallback behavior of unsupported locales - https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/src/terra-dev-site/test/i18n/DefaultI18n.test.jsx
* Used to test fallback behavior along with supported locale behavior -https://github.com/cerner/terra-core/blob/74e0af611b806303864132a53546ae054643ee3c/packages/terra-i18n/tests/wdio/i18n-spec.js

I feel like this module is kinda out of sight, out of mind, I'd prefer to have a singular place for our list of supported locales. I've kept this file for now but I'm open to ideas on how to handle it.

packages/terra-i18n/docs/UPGRADEGUIDE.md
bjankord: That is a good point, I'll move this over to the terra-ui doc site.

packages/terra-base/package.json
bjankord: Updated in this commit: https://github.com/cerner/terra-core/pull/2288/commits/76423fcbbd7d3f2301452db3667e6f69db6960f0

packages/terra-i18n/docs/AggregateTranslations.md
bjankord: Updated in this commit: https://github.com/cerner/terra-core/pull/2288/commits/6245e266ef54ae055263c4b8679b3c2dca060b12

packages/terra-i18n/docs/UPGRADEGUIDE.md
bjankord: Updated in this commit: https://github.com/cerner/terra-core/pull/2288/commits/6245e266ef54ae055263c4b8679b3c2dca060b12

I've started a new branch in terra-ui with this info. I'll send out a PR for that later on once terra-enzyme-intl is released as I want to include that info in the terra-ui i18n testing doc update.

packages/terra-base/src/Base.jsx
emilyrohrbough: That means lines 87-93 can be removed right?

bjankord: Yep, good call. Those can be removed.

bjankord: Updated in this commit: https://github.com/cerner/terra-core/pull/2288/commits/8c66e70a2ea75ed114e3a71d338ac81542c04435

packages/terra-i18n/docs/UPGRADEGUIDE.md
bjankord: Here is a the PR for moving this to terra-ui doc site: https://github.com/cerner/terra-ui/pull/158

packages/terra-i18n/src/i18nLoader.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable */
+import Intl from './intl';
 import loadIntl from './intlLoaders';
 import loadTranslations from './translationsLoaders';
-
 import supportedLocales from './i18nSupportedLocales';
 
 const hasIntl = typeof (Intl) !== 'undefined';
```
tbiethman: If we're guaranteed to have Intl defined at this point, we should be able to remove the branching logic around this

packages/terra-i18n/src/intl.js
tbiethman: This is exporting the global Intl that already exists in some browsers, not making a new global as part of the terra-i18n consumer API. The problem with the XFC implementations we're seeing are the global values that the various wrappers are making for direct consumer usage.

packages/terra-action-footer/package.json
```diff
@@ -21,8 +21,8 @@
   },
   "homepage": "https://github.com/cerner/terra-core#readme",
   "peerDependencies": {
-    "react": "^16.4.2",
-    "react-dom": "^16.4.2",
+    "react": "^16.8.4",
+    "react-dom": "^16.8.4",
     "terra-base": "^4.0.0"
```
tbiethman: @bjankord I haven't had a chance to look into the ramifications of this since we last talked about it, but would you be interested in getting rid of these terra-base peer-dependencies with this major version bump if we can determine they're not necessary?

bjankord: If its something we decide to do, now would be a good time to do it. Let's get an issue logged to track thoughts on the change of removing base as a peerDep.

tbiethman: Logged #2291

packages/terra-i18n/src/i18nLoader.js
bjankord: I had thought about this some, I believe the branching logic is still needed. My understanding of it is that while we do have Intl at this point (either from the native browser implementation or the polyfilled Intl), this check determines if we need to load the polyfilled Intl locale data which is needed if we not using the native browser implementation for Intl)

packages/terra-i18n/src/intl.js
bjankord: Update to `global` to `window`: https://github.com/cerner/terra-core/pull/2288/commits/6d9d55dc9ed7cf2fdaac9134e2be73841d26163d

packages/terra-action-footer/package.json
bjankord: Removed terra-base peer dependency across terra-core packages: https://github.com/cerner/terra-core/pull/2288/commits/adc84acfdb1919c67d9140e0cd5780ee2d1f97ac

packages/terra-i18n/src/i18nSupportedLocales.js
emilyrohrbough: I wonder if we need to worry about calling out that we are loading non-supported locales form terra-i18n anymore. We will log messages during aggregate-translations if the locale is not supported by terra but we will both aggregate and load whatever the user has defined.

since we do provided `ltr` and `rtl` styling for our components, I don't necessarily think it matters at this level if we are loading site assets we don't support. I am curious on what others think about this.

I do think it would be nice to drop these test fixtures: https://github.com/cerner/terra-core/tree/update-base-and-i18n-packages/packages/terra-i18n/tests/jest/fixtures. Our consumers ended up with these translations files aggregated when they run aggregate-translations and it logs warning messages for content we are shipping with the component.

bjankord: That's a good point about no longer needing to call out loading non-supported locales from terra-i18n. I'm fine dropping [the check from i18nLoader.js](https://github.com/cerner/terra-core/blob/728a5106f1e3fb113b3956cefeca996468b0ffb8/packages/terra-i18n/src/i18nLoader.js)

I'm on the fence about dropping the test fixtures. I see value in them but I'm also curious what others think as well.

packages/terra-action-footer/package.json
bjankord: Removed baseStyles import in this commit: https://github.com/cerner/terra-core/pull/2288/commits/01e5365ac3356bb3916420ad8b9ae14b3a9bab23

packages/terra-i18n/src/i18nSupportedLocales.js
bjankord: I tested removing the `fixtures` directory and all our Jest tests still pass. Now I'm not sure what they were used for, I assumed they were used for the Jest tests. It does seem like the `fixtures` directory is no longer needed. @emilyrohrbough do you recall what it was used for?

bjankord: Talked with @emilyrohrbough offline and decided the `fixtures` were used for aggregate-translation jest tests which have been removed in this branch so they are no longer needed. 

I've removed the `fixtures` directory in this commit: https://github.com/cerner/terra-core/pull/2288/commits/5a752387b5a83132ceaec80fe541b0909377d56f
I've removed the i18nSupportedLocales import and logic from `i18nLoaders.js` in this commit: https://github.com/cerner/terra-core/pull/2288/commits/aa792964b721b27641fe881781c8135f98ca7a13

There are 2 remaining places that the i18nSupportedLocales.js module is used in the terra-i18n package. 

* Used to test fallback behavior of unsupported locales - https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/src/terra-dev-site/test/i18n/DefaultI18n.test.jsx
* Used to test fallback behavior along with supported locale behavior -https://github.com/cerner/terra-core/blob/74e0af611b806303864132a53546ae054643ee3c/packages/terra-i18n/tests/wdio/i18n-spec.js

I'm open to idea on how to remove the i18nSupportedLocales.js module from those files.

bjankord: Talked with @mjhenkes about moving `i18nSupportedLocales.js` into the test directory to show that it is only needed for tests.

bjankord: Talked with @mjhenkes about just pulling this from toolkit.

bjankord: Removed this file and replaced it with an import from terra-toolkit in the 2 test files that still need the list of supported locales in this commit: https://github.com/cerner/terra-core/pull/2288/commits/954f2514715cd9e1d180a87da6c6a528435e9974

packages/terra-action-header/package.json
```diff
@@ -21,8 +21,8 @@
   },
   "homepage": "https://github.com/cerner/terra-core#readme",
   "peerDependencies": {
-    "react": "^16.4.2",
-    "react-dom": "^16.4.2"
+    "react": "^16.8.5",
+    "react-dom": "^16.8.5"
```
emilyrohrbough: Would it be possible to move these changes to a different PR?

bjankord: 👍I've split off the react/react-dom changes into a separate PR here: https://github.com/cerner/terra-core/pull/2298

packages/terra-overlay/src/LoadingOverlay.jsx
```diff
@@ -46,7 +46,7 @@ const defaultProps = {
   isOpen: false,
   backgroundStyle: BackgroundStyles.LIGHT,
   isRelativeToContainer: false,
-  rootSelector: '[data-terra-base]',
+  rootSelector: '#root',
```
mjhenkes: It seems we're making an assumption that the there is a root selector present. That assumption works for [terra-dev-site](https://github.com/cerner/terra-dev-site/blob/master/src/index.html#L11) and most orion apps, but it's not guaranteed that apps will have the root selector. If we want to change this from the base selector we might want to open it up to a prop but default to root and document it on the readme.

bjankord: We have [a prop for this already named rootSelector](https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/LoadingOverlay.jsx#L35-L37). This line sets the default value based on that assumption that there is an `id="root"` in the DOM. For apps that don't have that id attribute, they can change it by passing a different selector to the `rootSelector` prop. I can update the prop type documentation to be more clear about this.

mjhenkes: If we already have a prop for this i don't think you have to document more. I should have looked more deeply for the doc.

bjankord: Updated the doc, the doc referenced a data attribute that was on terra-base div that used to be rendered. We've removed that div to [fix terra-base fill issues](https://github.com/cerner/terra-framework/issues/124), but with that, we needed to update the rootSelector we default to. https://github.com/cerner/terra-core/pull/2288/commits/c7e816f812722711252e67b6cb35d93fcb13e4bd

Issue #1876, opened 2018-09-14
[terra-divider] Add text in divider


neilpfeiffer:
# Feature Request

## Description
Add enhanced features to the terra-divider:
- Ability to display text in the center of a divider, inline with that divider
- themeable left and right padding of the text
- themable bottom and top margin

PR #2296, opened 2019-03-22
Allow Text in Divider

amichaelparker:
### Summary
Updates terra-divider to use a div that can be themed, with text centered inside the divider line.


emilyrohrbough:
Is there an issue for this?

amichaelparker:
There is: [1876](https://github.com/cerner/terra-core/issues/1876). That’s my goof, I threw up a draft just to get a deployment that @neilpfieffer could take a look at, and forgot to reference it in the body.

packages/terra-divider/src/Divider.jsx
```diff
@@ -1,17 +1,28 @@
...
+const Divider = (props) => {
+  const { text, ...customProps } = props;
+
   const dividerClassNames = cx([
     'divider',
     customProps.className,
   ]);
 
-  return <hr {...customProps} className={dividerClassNames} aria-hidden="true" />;
+  return <div {...customProps} className={dividerClassNames} aria-hidden="true" content={text} />;
```
bjankord: Let's change `content` attribute to to `data-content`.

packages/terra-divider/src/Divider.module.scss
```diff
@@ -1,12 +1,33 @@
...
-    border-bottom-width: 0;
-    border-color: var(--terra-divider-border-color, #dedfe0);
-    border-left-width: 0;
-    border-right-width: 0;
-    border-style: solid;
-    border-top-width: var(--terra-divider-border-top-width, 1px);
-    height: 0;
-    margin: 0;
+    border: 0;
+    height: 1.5em;
```
bjankord: Adding height and line-height styles make this version of the divider render more space vertically when there is no text compared to the version that is in master. 
### Master
Screen Shot 2019-03-25 at 6 35 17 PM

### PR
Screen Shot 2019-03-25 at 6 35 30 PM

Need to sync up with @neilpfeiffer and see if this is an intentional height increase. If so, I'd consider this a non-passive change as it will impact spacing where prior to this change, the divider only rendered at 1px height.

amichaelparker: I put in some changes to hopefully correct this a little bit: [f614149](https://github.com/cerner/terra-core/pull/2296/commits/f6141490bc5feede0c8c51b8f9513ce833957c3d). The custom text styling is now conditional, so the changes to the standard divider should hopefully be more passive now.

packages/terra-divider/src/Divider.jsx
amichaelparker: Updated! [f614149](https://github.com/cerner/terra-core/pull/2296/commits/f6141490bc5feede0c8c51b8f9513ce833957c3d)

packages/terra-divider/src/Divider.module.scss
```diff
@@ -1,12 +1,37 @@
...
-    height: 0;
-    margin: 0;
+    border: 0;
+    height: 1px;
+    line-height: 1em;
+    outline: 0;
+    position: relative;
+    text-align: center;
+
+    &::before {
```
bjankord: I'd like to avoid removing the current variables if possible as that would be non-passive. Instead of setting a background color on the div, I think it would be worthwhile to take a look at https://css-tricks.com/line-on-sides-headers/ and see if we can use the border styles we previously had and apply them to `:before` and `:after` based on the technique in that CSS tricks article.

packages/terra-divider/src/Divider.module.scss
```diff
@@ -1,12 +1,37 @@
...
+      width: 100%;
+    }
+  }
+
+  .divider-text {
+    height: 1.5em;
+
+    &::after {
+      background-color: #fff;
+      color: var(--terra-divider-after-text-color, #000);
```
yuderekyu: once this is merged, are we planning on opening a PR into the clinical theme?

neilpfeiffer:
css for text (default theme):
```
color: var(-add-theme-var, #4E5558);
font-size: var(--add-theme-var, 0.9285714286rem);
font-style: var(--add-theme-var, normal);
font-weight: var(--add-theme-var, bold);
line-height: var(--add-theme-var, 1.231);
/* remove margins */
padding-left: var(--add-theme-var, 0.7142857143rem);
padding-right: var(--add-theme-var, 0.7142857143rem);
```

and we'll need to toggle between aria-hidden for non-with-text-version and when there is text, make it screen reader accessible, suggestions cc: @JakeLaCombe

packages/terra-divider/src/Divider.module.scss
neilpfeiffer: I agree - we should not update the themeing variables - best option is to refactor to with two templates (old hr with no text, plus new div with 3 spans for with text) and use the border trick on new template to satisfy linter rules without any variable changes (and not force exploring a linter override).

amichaelparker: Updated! [d6ecd19](https://github.com/cerner/terra-core/pull/2296/commits/d6ecd19a5b489194f21b5e4576f27779bf82ca05)

These changes are the amalgamation of @bjankord, @neilpfeiffer, and @JakeLaCombe's suggestions for better accomplishing the intended functionality. Text centering still looks odd, but this seems to at least be going in the right direction.

packages/terra-divider/CHANGELOG.md
```diff
@@ -3,6 +3,12 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Divider migrated from horizontal rule to div
```
bjankord: We should be fine dropping this "Changed" section since we rendering the HR now in the default case.

packages/terra-divider/src/terra-dev-site/doc/example/DividerWithText.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import Divider from 'terra-divider/lib/Divider';
+import DividerExampleTemplate from './DividerExampleTemplate';
+import Moment from '../../../../../../node_modules/moment/moment';
```
bjankord: Let's avoid pulling in moment here. My preference would be we just use a string for this demo.

e.g. `<Divider text="01/01/2019" />`

JakeLaCombe:
@neilpfeiffer Maybe it's just my eyes, but I'm looking at the sample long text that Adam posted, and it  feels unnatural to me when its long. It maybe because so little of the border is there once the long text is applied. Should there be a max width applied to the text for a divider?

packages/terra-divider/CHANGELOG.md
amichaelparker: Changed: [81179c5](https://github.com/cerner/terra-core/pull/2296/commits/81179c5398cb4eeb8c820f4f76ec2d946aaea09b)

packages/terra-divider/src/terra-dev-site/doc/example/DividerWithText.jsx
amichaelparker: Thought that might have been pushing it - changed! [81179c5](https://github.com/cerner/terra-core/pull/2296/commits/81179c5398cb4eeb8c820f4f76ec2d946aaea09b)

packages/terra-divider/src/Divider.module.scss
bjankord: These changes look good to me.

packages/terra-divider/src/Divider.jsx
```diff
@@ -1,16 +1,37 @@
...
+    ? <hr {...customProps} className={dividerClassNames} aria-hidden="true" />
+    : (
+      <div className={cx(['divider-container'])}>
+        <span className={cx(['divider-text'])}>{text}</span>
+      </div>
+    );
+
+  return (
+    dividerType
+  );
```
StephenEsser: This is up to your preference, leave it as is if you prefer it this way.  

Personally I think it's easier to read components without ternaries within the render logic.


```jsx
  if (!text) {
    return <hr {...customProps} className={dividerClassNames} aria-hidden="true" />;
  }

  return (
    <div className={cx(['divider-container'])}>
      <span className={cx(['divider-text'])}>{text}</span>
    </div>
  );
```

Or

```jsx
  if (text) {
    return (
      <div className={cx(['divider-container'])}>
        <span className={cx(['divider-text'])}>{text}</span>
      </div>
    );
  }

  return <hr {...customProps} className={dividerClassNames} aria-hidden="true" />;
```

packages/terra-divider/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ ChangeLog
 
 Unreleased
 ----------
+### Added
+* Separate div, props, and styling to allow inserting text into divider
+* CSS variables added for theming text and divider
+* Tests added for text divider
```
StephenEsser: When reading through the changelog it may be easier for consumers if we limit the implementation details and just add an entry for the API changes. 

```md
### Added
* Optional `text` property to display within the center of the divider
* Themeable SCSS variables for the text and divider
```

packages/terra-divider/src/Divider.jsx
```diff
@@ -1,16 +1,37 @@
 import React from 'react';
 import classNames from 'classnames/bind';
+import PropTypes from 'prop-types';
 import styles from './Divider.module.scss';
 
+const propTypes = {
+  /**
+  * String to be displayed in-line with the divider.
```
StephenEsser: Should `inline` be one word instead of hyphenated?

packages/terra-divider/src/terra-dev-site/doc/example/DividerExampleTemplate.jsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+/* eslint-disable react/prop-types */
+const DividerTemplate = props => (
+  <div>
+    <p>
+      Lorem Ipsum is simply dummy text of the printing and typesetting industry.Lorem Ipsum has been the industrys standard
+      dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen
+      book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially
+      unchanged.
```
StephenEsser: No changes are needed since this is just dummy text. 

I just thought it was a little funny that this isn't Lorem Ipsum text but rather the definition of Lorem Ipsum text. 

https://www.lipsum.com/

packages/terra-divider/src/Divider.jsx
amichaelparker: Honestly I don’t really have a preference. Either the ternary or the suggestions above are good for me, so if it makes it easier to parse for others, I’m good with changing it.

packages/terra-divider/src/terra-dev-site/doc/example/DividerExampleTemplate.jsx
amichaelparker: I hadn’t noticed that. :D I peeled that out of the original test file.

packages/terra-divider/src/Divider.jsx
amichaelparker: Maybe.
.
.
.
.
.
.
.
Yes.

packages/terra-divider/src/Divider.jsx
amichaelparker: Changes added! [7ad1cc8](https://github.com/cerner/terra-core/pull/2296/commits/7ad1cc8ca8af947a4b40c5cbf02c9a0bf49c534f)

packages/terra-divider/CHANGELOG.md
amichaelparker: Changes added! [7ad1cc8](https://github.com/cerner/terra-core/pull/2296/commits/7ad1cc8ca8af947a4b40c5cbf02c9a0bf49c534f)

packages/terra-divider/src/Divider.jsx
amichaelparker: Changes added! [7ad1cc8](https://github.com/cerner/terra-core/pull/2296/commits/7ad1cc8ca8af947a4b40c5cbf02c9a0bf49c534f)

packages/terra-divider/tests/jest/Divider.test.jsx
```diff
@@ -15,4 +15,10 @@ describe('Divider', () => {
     const wrapper = shallow(<Divider id="testDivider" />);
     expect(wrapper).toMatchSnapshot();
   });
+
+  // In-line Custom Text Text
```
jmsv6d: "Inline text test"

packages/terra-divider/tests/wdio/divider-spec.js
```diff
@@ -10,8 +10,29 @@ describe('Divider', () => {
     Terra.should.themeCombinationOfCustomProperties({
       testName: 'themed',
       properties: {
-        '--terra-divider-border-color': 'blue',
-        '--terra-divider-border-top-width': '10px',
```
jmsv6d: Why was this line removed? It also looks like this screenshot was deleted.

mmalaker:
Functional verification completed. The divider examples display and function as expected.

packages/terra-divider/tests/jest/Divider.test.jsx
amichaelparker: Whoops! Updated: [1deb07c](https://github.com/cerner/terra-core/pull/2296/commits/1deb07c772b96ffb855e2e09c0cff61d7c2d4d00)

packages/terra-divider/tests/wdio/divider-spec.js
amichaelparker: Updated: [1deb07c](https://github.com/cerner/terra-core/pull/2296/commits/1deb07c772b96ffb855e2e09c0cff61d7c2d4d00). The screenshot is still there (Default[themed]), but shrank because of the missing style. I added the style back in! It was a victim of an earlier version of the changes.

Issue #2278, opened 2019-03-15
Decouple terra-markdown from github-markdown-css


bjankord:
# Issue Description
I'd like to fork the styles from github-markdown-css. We used them originally as they offer a nice baseline set of styles for markdown content. However, I don't want to keep updating the markdown package. It would be preferable to have a set of core markdown styles that need little maintenance/updating.

bjankord:
Resolved in https://github.com/cerner/terra-core/pull/2335

PR #2300, opened 2019-03-25
Highlight markdown codeblocks

mjhenkes:
### Summary
Are you tired of bland boring old code blocks? Constantly miss reading and straining your eyes?

There has to be a better way!
Introducing prismJS formatting for markdown code blocks! No longer will your diff removals look like additions!
Featuring:
* Supports javascript, jsx, scss, and more! It's even extensible!
* Github-like theme, but more accessible!
* A theme variable!

Compare and be amazed!
Github: https://github.com/cerner/terra-core/blob/highlight-markdown-codeblocks/packages/terra-markdown/docs/README.md
terra-dev-site: https://terra-core-deployed-pr-2300.herokuapp.com/#/components/terra-markdown/markdown/markdown

### Additional Details
Theming? We had a single theme variable before that I carried forward, but I don't think it was getting the job done before and it definitely isn't getting the job done now. The "right" way to do theming would be to pull in the github markdown and make that theme-able as well, but I see diminishing returns to completing that work now, if ever, since this is a dev component.

I based the theme on the default prismjs theme https://github.com/PrismJS/prism/blob/master/themes/prism.css but that wasn't accessible so I borrowed colors from the github theme and cranked theme darker to meet color contrast (I am a designer!)
Pay some closer attention to the css because:

@cerner/terra


packages/terra-markdown/src/terra-dev-site/test/markdown/Markdown.test.jsx
```diff
@@ -1,11 +1,9 @@
 import React from 'react';
 import Markdown from '../../../../lib/Markdown';
-
-const markdownSrc = `### Test
-> This is a quote.`;
```
mjhenkes: I'm going to add this back to the test, I realized I wasn't covering this any more.

bjankord:
syntax-highlight-theme-meme

bjankord:
> The "right" way to do theming would be to pull in the github markdown and make that theme-able as well, but I see diminishing returns to completing that work now, if ever, since this is a dev component.

Agreed, I see value in us fully owning the styles that are used in terra-markdown. I logged [this issue](https://github.com/cerner/terra-core/issues/2278) to tackle that work at a later date.

packages/terra-markdown/src/terra-dev-site/test/markdown/MockREADME.md
```diff
@@ -1,23 +1,210 @@
...
++ This is an important
++ notice! It should
++ therefore be located at
++ the beginning of this
++ document!
+
+! compress the size of the
+! changes.
+
+  It is important to spell
```
tbiethman: Is it reely?

packages/terra-markdown/src/Markdown.scss
```diff
@@ -1,11 +1,114 @@
-.markdown-body {
-  margin-bottom: 1.25rem;
 
-  code {
-    background-color: #ebebef;
+
+:local(.markdown) {
```
tbiethman: So this will locally scope `.markdown` and leave the internal selectors global?

packages/terra-markdown/src/terra-dev-site/test/markdown/MockREADME.md
mjhenkes: I borrowed all of my code examples from here:
https://highlightjs.org/static/demo/

packages/terra-markdown/src/Markdown.scss
mjhenkes: Yep, that's what happening.

packages/terra-markdown/package.json
```diff
@@ -27,6 +27,7 @@
   "dependencies": {
     "classnames": "^2.2.5",
     "github-markdown-css": "^2.4.1",
+    "prismjs": "~1.15.0",
```
mjhenkes: Another dependency, refractor, locks in prismjs like this, I followed suit to avoid duplicate dependencies in terra dev site. I've logged an issue to refractor to update, once they update so can we, but until then this works.

packages/terra-markdown/tests/wdio/markdown-spec.js
```diff
@@ -1,18 +1,59 @@
 describe('Markdown', () => {
-  before(() => browser.setViewportSize(Terra.viewports('large')[0]));
+  before(() => browser.setViewportSize({ width: 1000, height: 2000, name: 'large' }));
   describe('Default', () => {
     before(() => browser.url('/#/raw/tests/terra-markdown/markdown/markdown'));
 
-    Terra.should.beAccessible();
-    Terra.should.matchScreenshot();
+    Terra.should.validateElement();
```
mjhenkes: I went through and added individual tests for the main code themes to catch non passive changes when prismjs updates.

PR #2311, opened 2019-03-28
Update to OCS icons v1.21.1

bjankord:
### Summary
Update to OCS icons v1.21.1


packages/terra-icon/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Updated to one-cerner-style-icons v1.21.0
```
emilyrohrbough: Should we include:
```
### Added
- IconCircleLinesSymbol
- IconCatalogueNumber
- IconCaution
- IconConsultInstructionsForUse
- IconDateOfManufacture 
- IconManufacturer 
```

bjankord:
In one-cerner-style-icons v1.21.0, they removed an icon and replaced it with a different icon.

From the [v1.21.0 release notes](https://github.com/cerner/one-cerner-style-icons/blob/master/CHANGELOG.md#v1210---1292019):

>Replaced 'Reload' icon with 'Undo' - src/icons/0152_circleArrowLeft_a.svg
Having a reload and refresh icon was redundant.
Replaced icon with the concept of 'Restore'. Functions similarly to 'Undo'

This feels like a non-passive change. Anyone who was using that reload icon now how to go and update it to the new icon. 
I've found there are some issues it seems like with are automated icon generation and this change as well. Need to investigate some more.

packages/terra-icon/CHANGELOG.md
bjankord: Updated changelog: https://github.com/cerner/terra-core/pull/2311/commits/acbab240cc223ac16853a5f15fed0d4cb84e749b

PR #2314, opened 2019-03-28
[terra-form-radio] Null check for child component so that it won't cr…

yfaney:
…ash if it's null

### Summary
I added the null check for child components so that it won't crash if it's null.

@cerner/terra


packages/terra-form-radio/src/RadioField.jsx
```diff
@@ -123,7 +123,7 @@ const RadioField = (props, { intl }) => {
   );
 
   const content = React.Children.map(children, (child) => {
-    if (child.type.isRadio) {
+    if (child && child.type.isRadio) {
```
tbiethman: Can you include a Jest test for this scenario?

yfaney: Added: b53561c

packages/terra-form-radio/package.json
```diff
@@ -28,6 +28,7 @@
     "classnames": "^2.2.5",
     "lodash.uniqueid": "^4.0.1",
     "prop-types": "^15.5.8",
+    "raf-polyfill": "^1.0.0",
```
StephenEsser: Why was this added?

yfaney: No idea. Removed: https://github.com/cerner/terra-core/pull/2314/commits/6374930ee51190515c85b4726ae6514a15f42c67#diff-afd97ad7b73244896cf178257a620ee2L31

yfaney:
Added changelog as per a bot check: https://github.com/cerner/terra-core/pull/2314/commits/6374930ee51190515c85b4726ae6514a15f42c67#diff-2acab889c3500ac761bbaa2fd7a19b4fR6

packages/terra-form-radio/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Adding a null check to the child components to avoid a crash
```
yuderekyu: nitpick - your change protects against all falsey values. Not just null.

packages/terra-form-radio/tests/jest/RadioField.test.jsx
```diff
@@ -9,6 +9,13 @@ it('should render a default radio field', () => {
   expect(wrapper).toMatchSnapshot();
 });
 
+it('should render a default radio field without crashing if it has a null child', () => {
```
yuderekyu: either change this to 
```suggestion
it('should render a default radio field without crashing if it has an undefined child', () => {
```

packages/terra-form-radio/tests/jest/RadioField.test.jsx
```diff
@@ -9,6 +9,13 @@ it('should render a default radio field', () => {
...
+  const nullChild = undefined;
```
yuderekyu: or
```suggestion
  const nullChild = null;
```

packages/terra-form-radio/CHANGELOG.md
yfaney: Updated: 0bf88b3a0

packages/terra-form-radio/tests/jest/RadioField.test.jsx
yfaney: Updated: efc1d732b

packages/terra-form-radio/src/RadioField.jsx
yfaney: @tbiethman I tried to run jest to create the snapshot but it gives me an error. What do I need to do to fix this?

```
● Validation Error:

  Module raf/polyfill in the setupFiles option was not found.
         <rootDir> is: /Users/yj048444/git/terra-core

  Configuration Documentation:
  https://jestjs.io/docs/configuration.html
```

packages/terra-form-radio/tests/jest/RadioField.test.jsx
yfaney: I picked undefined: efc1d73

packages/terra-form-radio/src/RadioField.jsx
tbiethman: raf is a dependency of terra-toolkit: https://github.com/cerner/terra-core/blob/master/package.json#L91. Can you verify that those dependencies are present in your node_modules directory?

yfaney: I found the cause. I needed to run compile first LOL
Updated: d95158091

packages/terra-form-radio/tests/jest/RadioField.test.jsx
```diff
@@ -9,6 +9,13 @@ it('should render a default radio field', () => {
   expect(wrapper).toMatchSnapshot();
 });
 
+it('should render a default radio field without crashing if it has an undefined child', () => {
+  const undefinedChild = undefined;
+  const radioField = (<RadioField legend="Default RadioField">{ undefinedChild }</RadioField>);
```
emilyrohrbough: Should this example just be
```<RadioField legend="Default RadioField" />```
or are we actually wanting to prevent  passing undefined as the children value?

File: packages/terra-form-radio/tests/jest/RadioField.test.jsx [hunk 7]
```diff
@@ -9,6 +9,13 @@ it('should render a default radio field', () => {
   expect(wrapper).toMatchSnapshot();
 });
 
+it('should render a default radio field without crashing if it has an undefined child', () => {
```
emilyrohrbough: nit: lets remove without crashing. all of these tests shouldn't crash the component 😃

packages/terra-form-radio/tests/jest/RadioField.test.jsx
yfaney: Prevent from passing undefined as the children value. In other words, we are testing the RadioField component with a child component which is undefined.

packages/terra-form-radio/tests/jest/RadioField.test.jsx
yfaney: Removed: 56b2ee965

danplubell: There is also a potential for crashing if the child does not have the type property.

yfaney: @danplubell K I will cover that as well.

yfaney: @danplubell Just curious, is that possible?

danplubell: Yes, if the child is not a Radio element or anything else that might not have the type property.  For instance, there isn't anything to prevent someone from having a mix of Radio elements and Select elements as children of the RadioField.

yfaney: Added: 0f1b14558

mjhenkes: @danplubell & @yfaney I discussed this with the terra team and we're ok allowing non-radio elements in radio form field. Namely the case we discussed was adding some kind of image next to a radio button. Radio form field will add screen reader accessibility to direct radio children, any non radio children will need to be evaluated for accessibility. We don't recommend adding any other form elements to the radio field but should not prevent it.

yfaney: @mjhenkes @danplubell Reverted: 192dab011

packages/terra-form-radio/src/RadioField.jsx
bjankord: We should update [checkboxfield](https://github.com/cerner/terra-core/blob/master/packages/terra-form-checkbox/src/CheckboxField.jsx#L126) to work the same as well. @yfaney can you update that as part of this PR as well.

yfaney: @bjankord Added: 9b8324a7a

Issue #1522, opened 2018-05-17
Clear button in search field doesn't appear


sujith9:
# Issue Description
The clear button used to clear input text in terra-search-field doesn't appear. Observed in IE 10. 

## Issue Type
- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
The button to clear input text should appear in the search field. Clicking the button should clear the input text.
The clear button appears in other browsers and IE 11.
![search_field_clear_button](https://user-images.githubusercontent.com/1754143/40188148-e202cd82-59be-11e8-9f6e-829e19ec971b.JPG)

## Current Behavior
The button to clear input text doesn't appear in the search field (in IE 10).

## Steps to Reproduce
1. Add terra-search-field to (any) app.
2. Render the search field in IE 10.
3. Input a search text in the search field.
4. Hover over right corner of search field and see no clear button.

## Environment
* Component Version: 2.6.0
* Browser Name and Version: IE 10
* Operating System and version (desktop or mobile): Windows 10 desktop.

noahbenham:
We're seeing this on iOS as well. Tested on iOS 11.3 mobile Safari.

As @dkasper-was-taken mentioned, this is a result of different browsers handling the ["search" input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search) type differently. To mitigate this, Terra would essentially have to add a custom clear button on the right side of the input for browsers that don't display a native clear button.

bb3284:
Just as an FYI, this issue was logged in CareAware Connect as a defect as there are requirements dictating the existence of a "clear" option.

bjankord:
This PR adds a cross-browser compatible solution for clearing in the select component: https://github.com/cerner/terra-core/pull/2240

bjankord:
We are adding a clear option to the select in this PR: https://github.com/cerner/terra-core/pull/2240
It may be worth investigating adding a cross-browser clear option to search.

ShantiiP:
@bjankord - Is there an estimated completion date for this?
Thanks.

StephenEsser:
There's no firm date when this will be completed, but the PR looks like it is getting pretty close.

https://github.com/cerner/terra-core/pull/2317

If you want to follow along with the PR, the changes are usually available within ~ 1 week after being merged into master.

PR #2317, opened 2019-03-29
[terra-search-field] Implement Cross-Browser Compatible Clear Button on Search Field

amichaelparker:
### Summary
This implements a cross-browser compatible clear button to terra-search-field. 

### Additional Details
Webkit native input clear functionality is hidden for uniformity. Uncontrolled search field behavior is preserved. The clear button is placed via flex and css to appear seamless, but is also adjustable.

Test in Firefox 67, Chrome 73, IE10, Edge, MacOS Safari, iOS Safari.

Closes [#1522](https://github.com/cerner/terra-core/issues/1522)

### Deployment link
https://terra-core-deployed-pr-2317.herokuapp.com/#/components/terra-search-field/search-field/search-field


packages/terra-search-field/src/SearchField.jsx
```diff
@@ -145,6 +184,9 @@ class SearchField extends React.Component {
     if (event.nativeEvent.keyCode === KeyCode.KEY_RETURN) {
       this.handleSearch();
     }
+    if (event.nativeEvent.keyCode === KeyCode.KEY_ESCAPE) {
```
bjankord: Nice, I like the addition of this keyboard interaction.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -183,36 +225,57 @@ class SearchField extends React.Component {
...
           refCallback={inputRefCallback}
+          value={defaultValue ? undefined : this.state.searchText}
           {...additionalInputAttributes}
         />
+        { this.previouslyFocusedInput && this.state.searchText
+          ? (
+            <Button
+              className={cancelButtonClassNames}
+              onClick={this.handleClear}
+              text="cancel"
```
bjankord: We'll want to get translations for this text. I can work with you to get those from our translation team.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -183,36 +225,57 @@ class SearchField extends React.Component {
...
+              icon={Clear}
```
bjankord: We'll want this icon to be a background image we can theme. Take a look at [how the icon is set on this button](https://github.com/cerner/terra-core/blob/master/packages/terra-dialog/src/Dialog.jsx#L54) and [the associated styles for it](https://github.com/cerner/terra-core/blob/master/packages/terra-dialog/src/Dialog.module.scss#L31-L38) as an example.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -183,36 +225,57 @@ class SearchField extends React.Component {
...
+              isIconOnly
+              aria-label="cancel"
```
bjankord: Since we are setting the `text` prop and the `isIconOnly` prop, we don't need to set the aria-label. The `<Button />` component [has internal logic to set an aria-label if both the `text` and `isIconOnly` props are used](https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L79).

packages/terra-search-field/src/SearchField.module.scss
```diff
@@ -1,13 +1,34 @@
 :local {
   .searchfield {
-    display: inline-flex;
+    display: flex;
```
bjankord: I believe we'll want to keep this as `inline-flex`.

packages/terra-search-field/src/SearchField.module.scss
```diff
@@ -43,11 +83,12 @@
   }
 
   .block {
-    display: flex;
+    width: var(--terra-search-field-block-width, 100%);
```
bjankord: We should be ok setting this to `100%` and not making it themable.

packages/terra-search-field/src/SearchField.jsx
amichaelparker: I remember it being talked about, but later noticing that it is apparently also one of those behaviors tied to the WebKit clear button functionality, like in Chrome; like the button, hitting escape also didn’t work in Firefox.

packages/terra-search-field/src/SearchField.module.scss
```diff
@@ -1,13 +1,34 @@
...
+    min-width: 0;
+    width: var(--terra-search-field-max-width, 15.31em);
+
+    .clear {
+      background: transparent;
+      border-bottom-left-radius: var(--terra-search-field-clear-border-bottom-left-radius, 0);
+      border-bottom-right-radius: var(--terra-search-field-clear-border-bottom-right-radius, 0);
+      border-color: var(--terra-search-field-clear-border-color, #dedfe0);
+      border-left: var(--terra-search-field-clear-border-left, 0);
+      border-right: var(--terra-search-field-clear-border-right, 0);
```
JakeLaCombe: I'm seeing the X jump and you focus in and out of the input. This is due to changing of the border from 0 to 1. We will need to start border-right with a width of 1px to prevent the jump.

bjankord: Let's set this default value to `1px solid rgba(0,0,0,0)`. This should help prevent a pixel shift when we add the focus style.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -122,7 +127,41 @@ class SearchField extends React.Component {
...
+  }
+
+  handleInputFocus(event) {
+    if (this.previouslyFocusedInput !== event.currentTarget) {
+      this.previouslyFocusedInput = event.currentTarget;
+    }
+  }
+
   handleTextChange(event) {
+    this.handleInputFocus(event);
```
jmsv6d: Why are we doing this here instead of using a ref?

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -183,36 +225,57 @@ class SearchField extends React.Component {
...
+          className={inputClassNames}
           type="search"
           placeholder={placeholder}
           onChange={this.handleTextChange}
+          onFocus={this.handleInputFocus}
           disabled={isDisabled}
           aria-disabled={isDisabled}
           onKeyDown={this.handleKeyDown}
           refCallback={inputRefCallback}
+          value={defaultValue ? undefined : this.state.searchText}
```
jmsv6d: Shouldn't we honor the value prop here?

packages/terra-search-field/src/SearchField.module.scss
JakeLaCombe: I don't think that will work. Borders come in at a diagonal, so you will this gap with the style. 
Screen Shot 2019-04-08 at 4 28 44 PM

Most likely, it needs to be the same color as the rest of the input border

packages/terra-search-field/src/SearchField.jsx
amichaelparker: I forgot to change the text when I changed other stuff from 'cancel' to 'clear'. I went back and forth on this, but clear struck me as more indicative of the intended functionality. As @bjankord said, however, if anyone has any thoughts on this, let me know!

packages/terra-search-field/src/SearchField.jsx
amichaelparker: I'm honestly not super familiar with the actual use of refs yet. I had attempted it, but I had some trouble getting it to work as intended in this case. If you can provide some guidance on how to incorporate it here, it would be appreciated!

packages/terra-search-field/src/SearchField.jsx
amichaelparker: Updated: [c5496c8](https://github.com/cerner/terra-core/pull/2317/commits/c5496c8d26b6ca995c3a7213a493c1d6169b626b)

packages/terra-search-field/src/SearchField.module.scss
amichaelparker: Updated: [83eee68](https://github.com/cerner/terra-core/pull/2317/commits/83eee6882b20fa5c793d89e7eea8d429e46ebd52).

packages/terra-search-field/src/SearchField.jsx
jmsv6d: So it looks like the Input component forces the refCallback to be a function that will accept the dom node as a parameter and does not allow the new createRef syntax. But this way will be easier to still honor the inputRefCallback prop. You'd create a function like 
```
setInputRef(node) {
    this.inputRef = node;
    if (this.props.inputRefCallback) {
        this.props.inputRefCallback(node);
    }
}
```

Then on the input you would use that for the refCallback prop
`refCallback={this.setInputRef}`
Then you could use `this.inputRef` anywhere you'd need access to the dom node.

jmsv6d: We'd likely want to update Input so that the createRef syntax can be used, but that's probably out of scope of this PR

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -183,36 +223,66 @@ class SearchField extends React.Component {
...
     const inputText = this.context.intl.formatMessage({ id: 'Terra.searchField.search' });
     const buttonText = this.context.intl.formatMessage({ id: 'Terra.searchField.submit-search' });
-    const additionalInputAttributes = Object.assign({ 'aria-label': inputText }, inputAttributes, valueTemp);
+    const clearText = this.context.intl.formatMessage({ id: 'Terra.searchField.clear' });
+    const additionalInputAttributes = Object.assign({ 'aria-label': inputText }, inputAttributes);
+
+    let valueData;
+    if (defaultValue) {
+      valueData = undefined;
+    } else if (value) {
```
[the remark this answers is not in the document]
amichaelparker: Updated: [3c644fc](https://github.com/cerner/terra-core/pull/2317/commits/3c644fc526ca5b970747607ce166f4e43ecc93cf). I changed it to be closer to the Input example, and testing both seems to work on the correct priority.

packages/terra-search-field/src/SearchField.jsx
amichaelparker: Updated: [3c644fc](https://github.com/cerner/terra-core/pull/2317/commits/3c644fc526ca5b970747607ce166f4e43ecc93cf). TY for this! I like this solution much better.

packages/terra-search-field/src/SearchField.jsx
amichaelparker: Finally working: [3c644fc](https://github.com/cerner/terra-core/pull/2317/commits/3c644fc526ca5b970747607ce166f4e43ecc93cf)

packages/terra-search-field/src/SearchField.module.scss
amichaelparker: This should be resolved here: [83eee68](https://github.com/cerner/terra-core/pull/2317/commits/83eee6882b20fa5c793d89e7eea8d429e46ebd52)

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -183,36 +213,64 @@ class SearchField extends React.Component {
       inputAttributes,
       ...customProps
     } = this.props;
+
+    const clearButtonVisible = !this.state.searchText ? 'input-clear' : undefined;
+
+    const clearButtonClassNames = cx([
+      'clear',
+    ]);
```
jmsv6d: Since there is only one class name here, let's just use `className={cx('clear')}` below

packages/terra-search-field/src/SearchField.module.scss
```diff
@@ -16,8 +51,22 @@
...
+        border-color: var(--terra-search-field-input-focus-clear-border-color, #4cb2e9);
+      }
+    }
+
+    .input::-ms-clear,
+    .input::-ms-reveal {
+      display: none;
+    }
+
+    // Focus style when clear button is not present
```
jmsv6d: It looks like this class is applied when the clear button is present

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -183,36 +213,64 @@ class SearchField extends React.Component {
...
 
     return (
       <div {...customProps} className={searchFieldClassNames}>
         <Input
-          className={cx('input')}
+          className={inputClassNames}
           type="search"
           placeholder={placeholder}
           onChange={this.handleTextChange}
+          onFocus={this.handleInputFocus}
```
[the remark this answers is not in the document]
amichaelparker: Very dead! Removed: [6421e1b](https://github.com/cerner/terra-core/pull/2317/commits/6421e1b29105ddc7726dccabaf5e49a2238a34df)

packages/terra-search-field/src/SearchField.jsx
amichaelparker: Updated: [6421e1b](https://github.com/cerner/terra-core/pull/2317/commits/6421e1b29105ddc7726dccabaf5e49a2238a34df)

packages/terra-search-field/src/SearchField.module.scss
amichaelparker: Comment changed! [6421e1b](https://github.com/cerner/terra-core/pull/2317/commits/6421e1b29105ddc7726dccabaf5e49a2238a34df)

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -183,36 +213,59 @@ class SearchField extends React.Component {
...
-          className={cx('input')}
+          className={inputClassNames}
           type="search"
           placeholder={placeholder}
           onChange={this.handleTextChange}
           disabled={isDisabled}
           aria-disabled={isDisabled}
           onKeyDown={this.handleKeyDown}
-          refCallback={inputRefCallback}
+          refCallback={inputRefCallback || this.setInputRef}
```
jmsv6d: You should just be passing in this.setInputRef here, it's already wrapping inputRefCallback. Doing it this way if the consumer passes in an inputRefCallback setInputRef won't get passed in.

amichaelparker: Updated: [6f3c0ac](https://github.com/cerner/terra-core/pull/2317/commits/6f3c0ac8a332bae73c766a02470db83de37aec28). Also updated the jest test associated with the callback!

jmsv6d:
Several of the WDIO screenshots seem off.  On `Auto_Search_Disabled[extended_search].png` and `Auto_Search_Disabled[searched_text].png` this search text has changes. And the width of the search bar appears to be about 16px shorter than it was before.

amichaelparker:
The width wdio differences are from the container changes, and the text differences were from the tests being written to click a 'button', despite there now being 2 buttons. This is fixed, but I'm currently trying to narrow down an oddity that only appears in the wdio screenshots (and not in actual use in any browser): 
Screen Shot 2019-04-12 at 12 02 52 PM
It makes determining visual regression odd.

bjankord:
@neilpfeiffer This is ready for UX review.

mmalaker:
A couple issues noted during functional testing. When using keyboard navigation, the clear button is missing the focus border on the right side. Also, the focus border is not displayed when tabbing to the search icon.

amichaelparker:
Waiting on finishing design touches for integration.

emilyrohrbough:
Something is weird with some of the screenshots:
Screen Shot 2019-05-22 at 10 49 01 AM

neilpfeiffer:
You don't like the crazy colors in the screenshots? Updating with corrected images.

mmalaker:
@neilpfeiffer @amichaelparker one minor issue noted. When placed within the site's search field, the top border of the clear button focus is hidden (see screenshot). Want to make sure this issue doesn't surface when used in the wild. Otherwise, everything looks good and is functioning correctly.

Screen Shot 2019-05-22 at 1 19 07 PM

neilpfeiffer:
Thanks @mmalaker - generally speaking I would assume the screenshot you provided should never be an issue because the searchfield should be used with space around it. The devsite filter was a quick add and didn't get the appropriate designed space around it I would normally expect to see. 

I will add an updated focus styling as a precaution anyway, since we will want to make sure we err on the safe side.

packages/terra-search-field/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ Unreleased
 3.13.0 - (May 14, 2019)
 ------------------
 ### Changed
+* Clear button added to search field for cross-browser compatibility
```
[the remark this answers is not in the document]
amichaelparker: Weird, it [originally](https://github.com/cerner/terra-core/pull/2317/commits/f339b3acc76bdab7878b63e940927c1e80ff2307) was, not sure what happened there. Fixed! [ff6fc2e](https://github.com/cerner/terra-core/pull/2317/commits/ff6fc2e88926d9a719df2757f5a8e59d116aa98d)

amichaelparker:
I pushed a focus fix for the issue @mmalaker raised. Good catch! :)

PR #2321, opened 2019-04-03
[terra-button] - Add HTML title attribute as a documented prop

### LINKED ISSUE #2304: [terra-button] Add feature to show tooltips for icon-only buttons
# Feature Request

## Description
Current Icon-only Buttons do not show the label by design, and usage would be greatly improved by adding browser-native title attribute tooltips that reveal the label to a user when hovering on the icon-only button.

StephenEsser:
### Summary

Added an optional [title](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title) attribute into the list of documented props. 

Buttons declared as `isIconOnly` or variant `utility` require a title attribute. If one is not provided using `title` the Button will fallback to using the required `text` as a title. 

### Additional Details

This is a follow up PR to https://github.com/cerner/terra-core/pull/2319.

Resolves #2304

@cerner/terra


bjankord:
Might need to update some Jest snapshots with this change.

PR #2327, opened 2019-04-05
[terra-form-select] Added Null Checks in Form-Select

amichaelparker:
### Summary
This adds null checks on all 'data-select-active' queries to prevent empty queries from causing errors.

Closes [cerner/terra-framework#562](https://github.com/cerner/terra-framework/issues/592)


mfried1:
Thank you so much for resolving this issue so quickly! 
Do you have any idea when the new version will be available in npm?

bjankord:
@mfried1 Should have this out in our weekly release tomorrow.

PR #2331, opened 2019-04-09
Improve Accessibility of Table Row

JakeLaCombe:
### Summary
Resolves #2285 

This makes the table more accessible. When Using voice over, I had the onClick and onKeyDown events passed on through the table columns so that it works better with voice over. Using VO + Space on a cell selects the entire row. In addition, several corresponding roles were added to various components to improve accessibility of the table.


supreethmr:
it seems to be still not working for tables having selectable rows like (single selectable rows and multi selectable rows). On both voice over and jaws there was no change on key down of row in both single selectable rows and multi selectable rows table, screen readers were still reading just the content of row missing the text 'this is row is selectable'. 

I also noticed that screen readers counting table header for number of rows and reading it has 4 rows when table as only three rows : 
Screen Shot 2019-04-10 at 2 30 44 PM

this is probably due to missing role on [table header](https://github.com/cerner/terra-core/blob/7caae79d542d07e55a27172a567329f508bf15d4/packages/terra-table/src/TableHeader.jsx#L24)  : ` <tr role="rowheader"> ` .

packages/terra-table/src/TableRows.jsx
```diff
@@ -60,12 +72,13 @@ const TableRows = ({
...
   onKeyDown,
+  selectRowHelpTextId,
   ...customProps
 }) => {
   const numberOfCols = getNumberOfColumns(children);
-  const cloneChildren = cloneChildItems(children, onClick, onKeyDown, numberOfCols);
+  const cloneChildren = cloneChildItems(children, onClick, onKeyDown, numberOfCols, selectRowHelpTextId);
   return (
-    <tbody {...customProps}>
+    <tbody {...customProps} rol="rowgroup">
```
supreethmr: nit: 'rol' : 'role'

JakeLaCombe:
I got the row selected text fixed. I didn't have the id passed down to all the components properly.

As for keydown, one thing to note is that if you are focusing on the rows with the screen reader, you hit enter to select the row, and if you are navigating columns with the screen reader, you press the space. Enter seems to be hijacked by the screen reader, so there isn't a lot I can do with that key. I'll see if adding row button anywhere can assist with that.

JakeLaCombe:
I think VO is including the row header as part of the row count. This is something we may not be able to get around honestly, as that's just how the screen reader reads tables. I'll see if there is a way online to restrict the count read to the user.

supreethmr:
> I think VO is including the row header as part of the row count. This is something we may not be able to get around honestly, as that's just how the screen reader reads tables. I'll see if there is a way online to restrict the count read to the user.

Adding <tr role="rowheader"> for table-header solved this issue for me, header was not included as part of row count by VO after this change.

JakeLaCombe:
I added that, and it doesn't include it as part of the table count. However, what I'm seeing is that VO will not allow you to use the keyboard controls to read the header of the column. I'll talk with others to see what the tradeoff is and which solution we would like. One thing that is interesting is that according to this documentation, role of rowheader actually goes to the th element, not the row itself.  (https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Row_Role) I didn't get any accessibility errors with our linter or wdio when I added that role, but it does make me wonder if that has any other accessibility concerns our tests may miss.

packages/terra-table/src/TableRow.jsx
```diff
@@ -18,17 +18,57 @@ const propTypes = {
...
   isSelectable: PropTypes.bool,
+  /**
+   * @private
+   * Function to trigger when the onClick has been called on a row. This passes the onclick
+   * function as an onclick trigger to the row's child columns.
+   */
+  onClick: PropTypes.func,
+  /**
+   * @private
+   * Function to trigger when the onKeyDown has been called on a row. This passes the onclick
```
jmsv6d:
```suggestion
   * Function to trigger when the onKeyDown has been called on a row. This passes the onKeyDown
```

packages/terra-table/src/TableRow.jsx
```diff
@@ -18,17 +18,57 @@ const propTypes = {
...
+   * function as an onclick trigger to the row's child columns.
```
jmsv6d:
```suggestion
   * function as an onKeyDown trigger to the row's child columns.
```

packages/terra-table/src/TableCell.jsx
```diff
@@ -7,17 +7,29 @@ const propTypes = {
...
+  selectRowHelpTextId: undefined,
 };
 
 const TableCell = ({
   content,
+  selectRowHelpTextId,
   ...customProps
 }) => (
-  <td {...customProps} data-terra-table-cell>
+  <td role="gridcell" {...customProps} aria-describedby={customProps.onKeyDown ? selectRowHelpTextId : undefined} data-terra-table-cell>
```
jmsv6d: Why is this checking for onKeyDown callback? I would think this should be determined by if the row has `isSelectable` set to true. Perhaps in TableRow only clone child with selectRowHelpTextId if `isSelectable`

packages/terra-table/translations/en.json
```diff
@@ -0,0 +1,3 @@
+{
+  "Terra.table.selectRow": "This Row is Selectable"
```
jmsv6d: nit: 
```suggestion
  "Terra.table.selectRow": "This row is selectable"
```

packages/terra-table/src/TableCell.jsx
JakeLaCombe: https://github.com/cerner/terra-core/pull/2331/commits/16a40cbff81110b7b731b0f88439ee8a2d79779c

packages/terra-table/translations/en.json
JakeLaCombe: https://github.com/cerner/terra-core/pull/2331/commits/20d278b6fca13465844cc7aa6b62e31d7dd574e2

packages/terra-table/src/TableRow.jsx
JakeLaCombe: https://github.com/cerner/terra-core/pull/2331/commits/44903cd1f8e631e8f41e50b3d47e8fb9d6dc3fcb

packages/terra-table/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Improved Accessibility but adding role of grid and attaching row listeners to columns
```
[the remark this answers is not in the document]
JakeLaCombe: https://github.com/cerner/terra-core/pull/2331/commits/b0a7f71f2b3978aa8efd3273bbf78ce885b4bf58

JakeLaCombe:
So I tried out VO over in other browsers with other websites, and when you are using the VO command, enter does not allow you to trigger any button or link. By design, I believe VO is locking in the space function as your interaction key.

packages/terra-table/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Updated to allow Voice Over to read selectable table rows.
```
bjankord: nit: I believe [VoiceOver is typically concatenated](https://en.wikipedia.org/wiki/VoiceOver) when in written form.

packages/terra-table/src/TableRows.jsx
bjankord: Looks like this should be `role`

packages/terra-table/CHANGELOG.md
JakeLaCombe: https://github.com/cerner/terra-core/pull/2331/commits/a7015c82ddf086876ec90d58d94a8a8aaeb9581d

packages/terra-table/src/TableRows.jsx
JakeLaCombe: Removed altogether. https://github.com/cerner/terra-core/pull/2331/commits/a7015c82ddf086876ec90d58d94a8a8aaeb9581d

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -25,12 +25,27 @@ const propTypes = {
    */
   // eslint-disable-next-line react/forbid-prop-types
   selectedIndexes: PropTypes.array,
+  liveRegion: PropTypes.oneOfType([
```
jmsv6d: Can we add some doc for these props?

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -49,9 +64,35 @@ class SelectableTableRows extends React.Component {
...
+
+        const { liveRegion } = this.props;
+        const { intl } = this.context;
+
+        if (this.props.screenReaderUpdateOnSelected) {
+          this.props.screenReaderUpdateOnSelected(liveRegion, intl);
+        } else {
+          const isSelected = this.props.selectedIndexes.indexOf(index) >= 0;
+
+          // The isSelected value here actually refers to the previous state of the row since
```
jmsv6d: This comment doesn't seem to belong here.

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -49,9 +64,35 @@ class SelectableTableRows extends React.Component {
     const initialOnClick = row.props.onClick;
 
     return (event) => {
-      // The default isSelectable attribute is either undefined or true, unless the consumer specifies the row's isSelectable attribute as false.
       if (row.props.isSelectable !== false) {
         this.handleOnChange(event, index);
+
+        const { liveRegion } = this.props;
```
jmsv6d: Should we go ahead and destructure all the props used in this function?

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -49,9 +64,35 @@ class SelectableTableRows extends React.Component {
...
+          // The isSelected value here actually refers to the previous state of the row since
+          // isSelected is not actually geting updated inside this listener
+          if (liveRegion.current) {
+            liveRegion.current.innerText = '';
+
+            // The isSelected value here actually refers to the previous state of the row since
+            // isSelected is not actually geting updated inside this listener
+            if (!isSelected) {
+              setTimeout(() => {
+                this.props.liveRegion.current.innerText = intl.formatMessage({ id: 'Terra.table.rowSelected' });
```
jmsv6d: Could just use `liveRegion` here.

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -62,12 +103,38 @@ class SelectableTableRows extends React.Component {
...
   wrappedOnKeyDownForRow(row, index) {
     const initialOnKeyDown = row.props.onKeyDown;
-
     return (event) => {
       if (event.nativeEvent.keyCode === KeyCode.KEY_RETURN || event.nativeEvent.keyCode === KeyCode.KEY_SPACE) {
         // The default isSelectable attribute is either undefined or true, unless the consumer specifies the row's isSelectable attribute as false.
         if (row.props.isSelectable !== false) {
           this.handleOnChange(event, index);
+
+          const { liveRegion } = this.props;
```
jmsv6d: Same comments for this function.

packages/terra-table/src/SingleSelectableRows.jsx
```diff
@@ -20,6 +20,17 @@ const defaultProps = {
   onChange: undefined,
 };
 
+const screenReaderUpdateOnSelected = (liveRegion, intl) => {
+  const liveElement = liveRegion.current;
+  liveElement.innerText = '';
+
+  // The isSelected value here actually refers to the previous state of the row since
+  // isSelected is not actually geting updated inside this listener
```
jmsv6d: This comment should be updated, `isSelected` isn't used here.

JakeLaCombe: https://github.com/cerner/terra-core/pull/2331/commits/dad5154e30b000e06bf7aac427a00df44c441bd8

packages/terra-table/src/SelectableTableRows.jsx
JakeLaCombe: https://github.com/cerner/terra-core/pull/2331/commits/3fb9b80ad6657aec97c4c26e072c7f421c0e62fc

packages/terra-table/src/SelectableTableRows.jsx
JakeLaCombe: https://github.com/cerner/terra-core/pull/2331/commits/3fb9b80ad6657aec97c4c26e072c7f421c0e62fc

packages/terra-table/src/SelectableTableRows.jsx
JakeLaCombe: https://github.com/cerner/terra-core/pull/2331/commits/3fb9b80ad6657aec97c4c26e072c7f421c0e62fc

packages/terra-table/src/SelectableTableRows.jsx
JakeLaCombe: https://github.com/cerner/terra-core/pull/2331/commits/3fb9b80ad6657aec97c4c26e072c7f421c0e62fc

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -62,12 +116,41 @@ class SelectableTableRows extends React.Component {
...
+            screenReaderUpdateOnSelected,
+            selectedIndexes,
+          } = this.props;
+
+          const { intl } = this.context;
+
+          if (screenReaderUpdateOnSelected) {
+            screenReaderUpdateOnSelected(liveRegion, intl);
+          } else {
+            const isSelected = selectedIndexes.indexOf(index) >= 0;
```
supreethmr: these will make VO to read unselected row when we click on preselected rows like shown in this : [example](https://terra-core-deployed-pr-2331.herokuapp.com/#/tests/terra-table/table/selectable-table/selectable-table-rows-preselected) 

In the above example table has row [0] and row [2] in selectedIndexes which makes this condition ``` selectedIndexes.indexOf(index) >= 0 ``` to return true on click of row and adds row unselected text to live region even when the row is still selected. 

This works fine when rows are pre-selected using isSelected prop where we can de-select the pre-selected row on click, but Pre-selected rows specified using selectedIndexes cannot be de-selected on click, this will always keep the row in selected state. 

for this scenario we might need to keep a flag which will be true when `this.props.selectedIndexes` has values initially and false when `this.props.selectedIndexes` is not initialized. 

By Adding this flag to below condition we can prevent VO from reading pre-selected rows as unselected rows
`if(!isSelected && isSelectedIndexesIntialized)`

packages/terra-table/src/SelectableTableRows.jsx
```diff
@@ -80,7 +163,6 @@ class SelectableTableRows extends React.Component {
   newPropsForRow(row, index, onClick, onKeyDown) {
     const isSelected = this.props.selectedIndexes.indexOf(index) >= 0;
     const newProps = { };
-
     // Set the isSelected attribute to false for all the rows except the rows whose index is set to state selectedIndex.
     if (isSelected !== row.props.isSelected) {
```
supreethmr: This might be not related to code changes of this PR but I have doubt with this condition 
``` (isSelected !== row.props.isSelected) ```  
with this condition newProps.isSelected will be set to true only when the either of props in above condition is true but this will not set newProps.isSelected to true when both `isSelected` and `row.props.isSelected` are true.  is this is intended to be like this, Because i have seen this scenario in multi-selectable rows table where both `isSelected` and `row.props.isSelected` are true and newProps.isSelected is set to false.

This scenario happens when we have pre-selected rows in multi selectable rows table.  where both `isSelected` and `row.props.isSelected` will become true.  Since selectedIndexes will have index of pre-selected row and props of row too will have true, so in this scenario we might miss-out to update newProps.isSelected to true. 

Changing the above condition to  ``` if(isSelected || row.props.isSelected) ``` will update the newProps.isSelected 
to true on both cases like when either of them are true and when both of them are true.

supreethmr:
Not sure whether we require this  VO to distinguish between highlighted rows in table. just remembered as it was mentioned on issue :  https://github.com/cerner/terra-core/issues/1839#issuecomment-423632145 .  initially 

if we are considering ' VO to distinguish between highlighted rows in table '  then we might need to update live-region in clonedChildItems() as well for pre-selected rows so that VO will be able to distinguish between highlighted rows in table.

packages/terra-table/src/SelectableTableRows.jsx
JakeLaCombe: I'll try again if I get the chance, but I don't think row.props.isSelected would work. If you initialize a table row such as <Table.row isSelected />, that if condition would always be true, even if the row is no longer selected. We will need to solely rely on the selectedIndexes prop passed into the table, which from implementing the details I noticed refers to the previous state of the table.

packages/terra-table/src/SelectableTableRows.jsx
JakeLaCombe: selectedIndexes will actually always be initialized to an empty array. I did try testing it with an empty array, and the screen reader wasn't able to correctly specify whether the row was selected or deselected. For this case, maybe we can tie onto the isSelected prop if that is what is mean to always keep a row selected?

JakeLaCombe: I did an investigation. This is not a perfect solution, and it requires the end user insuring they coded their onChange correctly. If you want the correct selected status to be read when a row is clicked, whether it's preselected or not, you can check for the presence of an onChange. If that does not exist, then we can check the value of the isSelected prop on the row itself.

It's not ideal, but I haven't found other ways that let us determine if the clicked row is selected or not. This component has a lot of moving parts, and it's getting hard to determine the point of change from when a row goes from selected to unselected.

mmalaker:
Functional verification in progress. The changes now make the selectable tables much more accessible in VO. Still seeing issues in JAWS, though. Tabbing into tables in JAWS will read the selectable/selected/unselected states; however, when using the JAWS arrow navigation keys, it ignores those states. Would be best if we could enable JAWS to read the state as well without having to rely on tab navigation.

JakeLaCombe:
I think the only way to get around that is to put visually hidden text inside each individual table cell that identifies the row as selected or unselected.

supreethmr:
> Functional verification in progress. The changes now make the selectable tables much more accessible in VO. Still seeing issues in JAWS, though. Tabbing into tables in JAWS will read the selectable/selected/unselected states; however, when using the JAWS arrow navigation keys, it ignores those states. Would be best if we could enable JAWS to read the state as well without having to rely on tab navigation.

I have modified the code and Tested with Jaws and I was able to navigate cells using up and down arrow keys. @mmalaker can you please test Jaws with recent build to confirm this.

mmalaker:
@supreethmr thanks for your work on this. I just checked the selectable tables again. To clarify, the arrow navigation in JAWS works fine within tables and across page elements. The issue is that JAWS does not read the selected/selectable state when navigating between cells with arrow keys.  It only reads this information when using the Tab key. This means that a user who is navigating a page may not know that the rows in a given table are selectable. 

I think we can shelve that issue for now, however. This PR was originally opened to address VO issues. The changes made in this PR have greatly improved the accessibility of tables overall. I think we can consider this done and can fine-tune the JAWS workflow at a later time.

PR #2339, opened 2019-04-15
Update Field.jsx

emilyrohrbough:
### Summary
I'm am receiving these console warnings, but had to go to the src code to see what prop should be set:

Screen Shot 2019-04-15 at 1 48 23 PM


StephenEsser:
Danger is probably going to fail on the changelog check

PR #2345, opened 2019-04-17
Fix hr margin issue

### LINKED ISSUE #2344: [terra-divider] New margin is not right
# Bug Report

## Description
Divider used to have a margin of 0.  Now it does not.  This gives "padding" in places we were not expecting it.

## Steps to Reproduce
1. Create a page with a terra-action-header.
2. Add a terra-divider immediately after it.

Test with the old version of terra-divider (< 3.6.0).  There is no space between the divider and the content.
Test with the new version (3.6.0).  There is space between the header and the divider.
It is easy to see the space with any control that has its space filled with a color.

We found the issue after Fix #2296 was released.
No Space (old version)
no_space

Space (new version)
space

## Expected Behavior
Given this is a minor update, margin should remain 0 when there is no text.

## Possible Solution
If the margin needs to change, it should only change when there is text.  Or allow the user to opt in.  Or make the change a major version bump, so teams have time to figure out what to do about it.

## Environment

* Component Name and Version: terra-divider 3.6.0
* Browser Name and Version: Chrome Version 72.0.3626.119  (and in our Chrome wdio tests)
* Node/npm Version: 
* Webpack Version:
* Operating System and version (desktop or mobile): desktop

bjankord:
### Summary
Resolves #2344 

I added divs above and below in our wdio test pages so it is more apparent how the margin on the divider is being rendered.


packages/terra-divider/src/terra-dev-site/test/divider/DefaultDivider.test.jsx
```diff
@@ -2,4 +2,10 @@ import React from 'react';
 
 import Divider from '../../../Divider';
 
-export default () => <Divider id="divider" />;
+export default () => (
+  <div>
+    <div style={{ height: '2rem', background: '#40E0D0' }} />
```
yuderekyu: could this cause color contrast issues with theme development? if so, we should create theme variables for the example background colors

bjankord: No text is rendered on these. These divs exist to highlight the box-model of the hr element and the hr elements related margin.

PR #2347, opened 2019-04-19
[terra-button] Use 'aria-label' from custom props.

### LINKED ISSUE #2346: [terra-button] aria-label sent as a custom prop is nullified.
# Bug Report

## Description
`aria-label` that is sent as a custom prop gets nullified due to the line below - 

https://github.com/cerner/terra-core/blob/25696dcf0fb49d22356219252f13ef0ef83b98e7/packages/terra-button/src/Button.jsx#L246

## Additional context
https://github.com/cerner/terra-core/issues/1689#issuecomment-483123764 and https://github.com/cerner/terra-core/issues/1689#issuecomment-483216953.

## Steps to Reproduce

Create a default `terra-button` component with `aria-label` sent as a custom prop.

```
const DefaultButton = () => <Button 
   id="defaultButton"
   text="Button Type is Default"
   type="button"
   aria-label="default button"/>;
```

## Expected Behavior
Button has the `aria-label` set.

## Possible Solution

Default the value of `aria-label` prop in `Button.jsx` to `aria-label` that is present in `customProps`. Use `text` as a fallback for utility buttons and icons. 

## Environment

* Component Name and Version: terra-button@v3.11.0
* Browser Name and Version: All browsers.

## @ Mentions
@neilpfeiffer

nramamurth:
### Summary
Resolves #2346.

@cerner/terra


packages/terra-button/src/Button.jsx
```diff
@@ -233,6 +233,11 @@ class Button extends React.Component {
       </span>
     );
 
+    let ariaLabel = customProps['aria-label'];
+    if (isIconOnly || variant === 'utility') {
```
benbcai: nit: I know this was moved from the lines below..we could use `ButtonVariants.UTILITY` instead of 'utility'.

Issue #968, opened 2017-10-25
Simplify arrange example page


mmalaker:
# Issue Description
The current Arrange page has seven different examples. Several of these allow users to change the alignment of each segment of the arrange component. This is a bit redundant and the content doesn't always flex depending on the selection. For example, the Fit Start image fills the entire fit page, so no matter what alignment value is selected, the example doesn't seem to change.

## Issue Type

- [ ] New Feature
- [x ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
- Reduce number of arrange examples
- Provide content that flexes based on user input and exemplifies the intended usage of the component.

bjankord:
On this issue, we'll want to sync up with @mmalaker to see what guidance he has on this.

PR #2353, opened 2019-04-23
[terra-arrange] Update Doc Examples

amichaelparker:
### Summary
Updates the doc site examples to remove redundancy, fix alignment functionality, and clarify changes in alignment.

Closes [#968](https://github.com/cerner/terra-core/issues/968)


packages/terra-arrange/src/terra-dev-site/doc/arrange/Arrange.doc.jsx
```diff
@@ -24,19 +24,15 @@ const DocPage = () => (
         title: 'Align All Containers',
         example: (
           <div>
-            <ArrangeAlignment alignment="all" fitStart={image} fill={simpleText} />
-            <ArrangeAlignment alignment="all" fill={simpleText} fitEnd={icon} />
-            <ArrangeAlignment alignment="all" fitStart={image} fill={simpleText} fitEnd={icon} />
+            <ArrangeAlignment alignment="all" fitStart={alignExampleDiv} fill={simpleText} fitEnd={alignExampleDiv} />
           </div>
         ),
       },
```
yuderekyu: want to use the `source` attribute to toggle the example code?

packages/terra-arrange/src/terra-dev-site/doc/example/ArrangeAlignment.jsx
```diff
@@ -110,6 +110,7 @@ class ArrangeAlignment extends React.Component {
             {...(this.state.alignFitStart && { alignFitStart: this.state.alignFitStart })}
             {...(this.state.alignFill && { alignFill: this.state.alignFill })}
             {...(this.state.alignFitEnd && { alignFitEnd: this.state.alignFitEnd })}
+            style={{ height: '100%' }}
```
emilyrohrbough: Why are you adding this?

amichaelparker: It may not be necessary any more. Initially, the examples didn't actually work, because the container would scale to whichever was the larger of the icons, so alignment of one icon would work, and the other would not... So this essentially forced the height of the whole arrange container to fill the wrapper.

However, this will likely be gone soon, since I'm overhauling the examples anyway.

packages/terra-arrange/src/terra-dev-site/doc/arrange/Arrange.doc.jsx
amichaelparker: I'm currently in the process of adding this. :) I'm attempting to restructure the examples so the source examples will make sense individually, rather than just being the monolithic one currently being used.

packages/terra-arrange/src/Arrange.jsx
```diff
@@ -25,21 +25,21 @@ const propTypes = {
    */
   fitEnd: PropTypes.element,
   /**
-   * The vertical orientation of all three containers. It will override the aligment of alignFitStart, alignFill and alignFitEnd if given. One of: `center`, `bottom`, `stretch`.
+   * The vertical orientation of all three containers. It will override the aligment of alignFitStart, alignFill and alignFitEnd if given. One of: `center`, `bottom`, `stretch`, `default`.
    */
-  align: PropTypes.oneOf(['center', 'bottom', 'stretch']),
+  align: PropTypes.oneOf(['center', 'bottom', 'stretch', 'default']),
```
emilyrohrbough: Why are we adding a default prop option?

packages/terra-arrange/src/terra-dev-site/doc/example/ArrangeAlignAllContainers.jsx
```diff
@@ -0,0 +1,43 @@
...
+/* eslint-disable import/no-extraneous-dependencies */
+import Arrange from 'terra-arrange/lib/Arrange';
+import { alignExampleDivBlue, textWithBlueBorder } from '../common/examplesetup';
+import styles from './ArrangeExamples.module.scss';
+
+const cx = classNames.bind(styles);
+
+const ArrangeAlignAllContainers = () => (
+  <div>
+    <h4>Align - Default</h4>
```
emilyrohrbough: Is this the correct heading for accessibility of this page?

packages/terra-arrange/src/Arrange.jsx
amichaelparker: Sort of a tester to run by @bjankord. The idea is to have examples for default, center and bottom, and not stretch, but default is stretch.

If the idea is to have top, center, and bottom examples, then you’d probably want default to use something like flex-start, since stretch will fill the height of the container if the consumer doesn’t constrain it.

This is basically just my initial thoughts on making the desired examples, minus stretch.

packages/terra-arrange/src/terra-dev-site/doc/example/ArrangeAlignAllContainers.jsx
amichaelparker: Not sure. Heading might not even be the appropriate way to set it up, so I’ll look into it.

packages/terra-arrange/src/Arrange.jsx
bjankord: I don't recall the reasoning behind the original implementation around the align prop but I believe that in the dropdown on the doc page, selecting "default", the example would not set the align prop at all.
By not specifying the align prop, the fit containers would have the following "default" styles.
```css
align-self: flex-start;
```
The fill container did not have any align-self styles in this "default" state. What this did was align the content in the fit and fill containers vertically to the top.

I could see adding a new value, `top` that is default value for the align prop that would work similar to how we have today. If teams omit the align prop as they are today, it would default the align prop to `top`. For this PR, I think this change is beyond the scope though.

I can sync up with you on this to keep changes to the component source code to a minimum for this PR.

packages/terra-arrange/src/terra-dev-site/doc/example/ArrangeAlignAllContainers.jsx
bjankord: An `h3` element here would better match the document outline.

packages/terra-arrange/src/terra-dev-site/doc/common/examplesetup.jsx
```diff
@@ -1,54 +1,36 @@
...
-  </div>
+const textWithBlueBorder = <div style={{ border: '2px solid #0079BE' }}>{ipsum}</div>;
+
+const alignExampleDiv = (
+  <div
+    style={{
+      backgroundColor: '#000',
+      display: 'block',
+      height: '20px',
+      margin: '1px',
```
bjankord: nit: I'd drop the margin here, I don't want people to think that the arrange is padding components within it.

packages/terra-arrange/src/terra-dev-site/doc/common/examplesetup.jsx
```diff
@@ -1,54 +1,36 @@
...
-  <div style={{ backgroundColor: '#ff5722', height: '100%' }}>
-    {props.children}
-  </div>
+const alignExampleDivBlue = (
+  <div
+    style={{
+      backgroundColor: '#0079BE',
+      display: 'block',
+      height: '20px',
+      margin: '1px',
```
bjankord: nit: I'd drop the margin here, I don't want people to think that the arrange is padding components within it.

packages/terra-arrange/src/terra-dev-site/doc/arrange/Arrange.doc.jsx
amichaelparker: This is implemented in the current deployment.

packages/terra-arrange/src/terra-dev-site/doc/example/ArrangeAlignment.jsx
amichaelparker: This is removed in the current deployment via all the shuffling of examples. :)

packages/terra-arrange/src/Arrange.jsx
amichaelparker: Updated: [68c3a20](https://github.com/cerner/terra-core/pull/2353/commits/68c3a206ca8b941ea1e9f9e7a3e6fad41525c455). I went ahead and removed the default prop and just removed the alignment attributes from all of the examples so they would fallback to the default styling.

packages/terra-arrange/src/terra-dev-site/doc/example/ArrangeAlignAllContainers.jsx
amichaelparker: Updated: [68c3a20](https://github.com/cerner/terra-core/pull/2353/commits/68c3a206ca8b941ea1e9f9e7a3e6fad41525c455).

packages/terra-arrange/src/terra-dev-site/doc/common/examplesetup.jsx
amichaelparker: Updated: [68c3a20](https://github.com/cerner/terra-core/pull/2353/commits/68c3a206ca8b941ea1e9f9e7a3e6fad41525c455).

packages/terra-arrange/src/terra-dev-site/doc/common/examplesetup.jsx
amichaelparker: Updated: [68c3a20](https://github.com/cerner/terra-core/pull/2353/commits/68c3a206ca8b941ea1e9f9e7a3e6fad41525c455).

Issue #2280, opened 2019-03-15
Update react-focus-trap dependency to latest version in terra-overlay


bjankord:
# Issue Description
Currently we use react-focus-trap ^4.0.1
The latest version is 6.0.0

[react-focus-trap changelog](https://github.com/davidtheclark/focus-trap-react/blob/master/CHANGELOG.md)

bjankord:
This could potentially require a major version bump of components that consume react-focus-trap.

bjankord:
I looked into updating this a bit when working in terra-framework. When I updated to v6, I started seeing the following error when popups were opened. 
```
Uncaught Error: You can't have a focus-trap without at least one focusable element.
```

A recommended work around is to use setTimeout in componentDidMount to delay focus trap activation until react renders the DOM: https://github.com/davidtheclark/focus-trap-react/issues/20#issuecomment-365224468

PR #2372, opened 2019-04-30
[terra-overlay] Updated focus-trap-react Version

amichaelparker:
### Summary
This updates the focus-trap-react version in terra-overlay to ^6.0.0.

Closes [#2280](https://github.com/cerner/terra-core/issues/2280)

### Additional Details
Version 5.0.0 of focus-trap-react introduced a [breaking change](https://github.com/davidtheclark/focus-trap-react/blob/master/CHANGELOG.md#500) that requires a single, focusable child element to be present, which caused fullscreen overlays to throw an error in console and break the page. Introducing a single div wrapper around the overlayComponent mitigates this change while maintaining functionality / visual parity.


packages/terra-overlay/src/Overlay.jsx
```diff
@@ -208,7 +208,10 @@ class Overlay extends React.Component {
       <Portal>
         {backgroundScrollContent}
         <FocusTrap>
-          {overlayComponent}
+          {/* div addresses child focus change introduced in focus-trap-react v5.0.0 */}
+          <div>
```
tbiethman: Do any of the styles applied to the overlay need to be applied here as well? https://github.com/cerner/terra-core/pull/2372/files#diff-5c1dfe9c477f29f0ae7f7738c5cb13e9R174

I'm thinking things like the fullscreen properties might be impacted by adding this wrapper.

tbiethman: Turns out the `is-fullscreen` [selector](https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/Overlay.module.scss#L34) that I thought might have issues....isn't actually used anywhere. 🤷‍♂

Issue #2131, opened 2018-12-19
Update Grid to consume terra-breakpoints package


tbiethman:
# Feature Request

## Description
The terra-grid package currently defines its own breakpoints: https://github.com/cerner/terra-core/blob/master/packages/terra-grid/src/Grid.module.scss#L36

We should switch it over to consume the terra-breakpoints package and update its size-based props to match.

bjankord:
Planning to hold off on this change until around the end of Q2/beginning of Q3 to minimize major version bumps in the Terra UI ecosystem.

bjankord:
For this, I'd recommend we create a new `.scss` file in the terra-breakpoints package to store a SCSS variable which holds a map of the breakpoints similar to https://github.com/cerner/terra-core/blob/master/packages/terra-grid/src/_variables.scss#L1-L7

The new map should look something like:
```scss
$terra-breakpoints: (
  tiny: 0px,
  small: 544px,
  medium: 768px,
  large: 992px,
  huge: 1216px,
  enormous: 1440px
);
```

While it may be possible to consume [the JS breakpoint values export](https://github.com/cerner/terra-core/blob/master/packages/terra-breakpoints/src/breakpoints.js#L1-L8) in the SCSS file, I think this would require additional webpack config on our user's part. This is something I want to avoid.

We'd then want to consume this new SCSS partial back in the terra-grid package.

With this change, we'll also need to update the following prop names:
* `col`
* `tiny`
* `small`
* `medium`
* `large`
* `huge`

I'd propose we change these to:
* `col` -> `tiny+`
* `tiny`  -> `small+`
* `small`   -> `medium+`
* `medium`   -> `large+`
* `large` ->  `huge+`
* `huge` -> `enormous+`

or drop the `+` and go with:

* `col` -> `tiny`
* `tiny`  -> `small`
* `small`   -> `medium`
* `medium`   -> `large`
* `large` ->  `huge`
* `huge` -> `enormous`

The names we use for these props should be the same names we go with for the responsive element MVB update.

cc/ @tbiethman @mjhenkes @StephenEsser @ryanthemanuel

PR #2373, opened 2019-04-30
Terra-grid now imports breakpoints from terra-breakpoints

AlexB98:
### Summary
Deleted the existing variables.scss file in terra-grid that was being used for breakpoints. Instead terra-grid now imports the breakpoints from a variables.scss file within terra-breakpoints. Also, changed the naming scheme of the size props.

Resolves [#2131](https://github.com/cerner/terra-core/issues/2131)


packages/terra-breakpoints/src/_variables.scss
```diff
@@ -0,0 +1,8 @@
+$terra-breakpoints: (
+  tiny: 0,
+  small: 544px,
+  medium: 768px,
+  large: 992px,
+  huge: 1216px,
+  enormous: 1440px
+);
```
StephenEsser: Do we still want to keep this file in sass. Are consumers using it in their custom scss?

packages/terra-grid/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Removed variables file, breakpoints now imported from terra-breakpoints
```
StephenEsser: There are breaking changes we should callout here. I don't know if there is a standard for the changelog entry yet, but previously I had listed those changes under a `### Breaking Changes` heading

AlexB98: AH, I wasn't aware of that. Will do!

emilyrohrbough: Here is what I link out for Changelog reference: https://keepachangelog.com/en/1.0.0/

packages/terra-grid/src/GridColumn.jsx
```diff
@@ -27,9 +27,13 @@ const propTypes = {
   /**
    * The number of columns across all breakpoints. From 1-12(inclusively).
    */
-  col: columnRange,
+  tiny: columnRange,
+  /**
+   * The number of columns starts from enormous breakpoint. From 1-12(inclusively).
```
emilyrohrbough: Can we update and change these prop descriptions with this breaking change? I find them to be confusing and always have to guess and check my implementation when trying to use them

packages/terra-grid/src/GridColumn.jsx
```diff
@@ -27,9 +27,13 @@ const propTypes = {
   /**
    * The number of columns across all breakpoints. From 1-12(inclusively).
    */
-  col: columnRange,
+  tiny: columnRange,
```
emilyrohrbough: Should tiny be required since it'll be the default?

packages/terra-grid/CHANGELOG.md
StephenEsser: I think the Breaking Changes section adds more clarity to the change log and helps the changes stand out. For non-breaking changes I think the Changelog reference above is a good standard.

StephenEsser: There are a few other things we'll need to call out here. Including the removal of the `col` prop, the addition of the `enormous` prop,  and how the new breakpoint ranges affect this component. The old API has a different breakpoint mapping now

emilyrohrbough: @StephenEsser I agree we should add a `### Breaking Change` section with the list of changes and possibly link to the upgrade guide

emilyrohrbough:
@AlexB98 We need to add an upgrade guide for these breaking changes

packages/terra-grid/src/GridColumn.jsx
AlexB98: Ya... I wasn't truthfully sure what it meant either. Any clarification as to the significance of the numbers would be appreciated.

packages/terra-grid/CHANGELOG.md
AlexB98: Sure thing, will add a list!

AlexB98:
@emilyrohrbough What does this entail?

AlexB98:
> @AlexB98 We need to add an upgrade guide for these breaking changes

@emilyrohrbough What does this entail?

emilyrohrbough:
I would expect a new section at the top of the existing upgrade guide that explains to consumers what changed and what code changes are required for consumer terra-grid v6.

https://engineering.cerner.com/terra-core/#/components/terra-grid/grid/upgrade-guide

packages/terra-grid/src/GridColumn.jsx
emilyrohrbough: As you resize thew width of your browser, the grid should re-adjust it's layout when a new breakpoint is reach. This layout will reflect column values provided to the props. I'd update an example and test it out a bit to find a description that makes the best sense to you

bjankord: I'd recommend the following for the prop description.
```
Defines the number of columns this container will span in the 12 column grid. It is applied at the `tiny` breakpoint and wider screens.
```

AlexB98:
> Do we have wdio tests for each breakpoint size? I noticed the medium screenshots were updated

Updated the screenshots I saw that used the now nonexistent 'col' prop, and added the 'enormous' prop.

packages/terra-grid/src/GridColumn.jsx
bjankord: This should stay as an optional prop.

AlexB98:
@StephenEsser @emilyrohrbough Any more changes/fixes you would like to see?

StephenEsser:
@AlexB98 

This PR was brought up and discussed in our war-room. It sounded like Brett would sync up with you to discuss any potential changes. 

cc: @bjankord

bjankord:
@StephenEsser I've updated this branch with additions we discussed in our meeting on Friday here: https://github.com/cerner/terra-core/pull/2373/commits/a6dc28763867f417d20553f2f40c6dd761c6dd58

@emilyrohrbough I've created test pages to cover the various column width props here: https://github.com/cerner/terra-core/pull/2373/commits/50e05c062b03d02fb6968505fe8a7586d179994c
@AlexB98 We'll want to get wdio tests in place to capture test evidence of these new pages.

Issue #2376, opened 2019-05-01
[terra-table] Subheaders can clash with striped rows


yuderekyu:
# Bug Report
Stripe table styles combine with subheader styles, if two conditions are met: 
1. subheader is an even row within table. 
2. subheader and striped rows have opaque colors

![Screen Shot 2019-05-01 at 1 46 06 PM](https://user-images.githubusercontent.com/14063284/57035387-83a22500-6c17-11e9-8428-e949ca6267c6.png)

A subheader is defined as a table row (tr). [This](https://github.com/cerner/terra-core/blob/master/packages/terra-table/src/Table.module.scss#L145-L147) selector applies striped row styling to every even table row.

A potential solution is to exclude subheader from that selector.

PR #2377, opened 2019-05-01
[terra-table] Exclude subheader from zebra row styles.

yuderekyu:
### Summary
Closes https://github.com/cerner/terra-core/issues/2376. 

With this change, subheaders will not receive stripe row styles.

### DIFF
--- a/packages/terra-table/CHANGELOG.md
+++ b/packages/terra-table/CHANGELOG.md
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Exclude subheader from zebra row styles.
 
 3.11.0 - (May 1, 2019)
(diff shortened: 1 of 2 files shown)


nramamurth:
Should this UI change be _wdio_ tested?

yuderekyu:
A test already captures this scenario [[1](https://github.com/cerner/terra-core/blob/master/packages/terra-table/tests/wdio/table-spec.js#L98)]. The example sets `isStriped` and has a subheader as an even row [[2](https://github.com/cerner/terra-core/blob/master/packages/terra-table/src/terra-dev-site/test/table/table/TableWithSubheaders.test.jsx#L6)].

This test does not break because we don't have opacity in the open source theme.

Issue #2375, opened 2019-05-01
Update to stylelint-config-terra v3


bjankord:
# Issue Description
Update repo to use stylelint-config-terra-v3

PayalSawant:
I can work on this

bjankord:
Yes, PRs welcome! 👍

bjankord:
Resolved in https://github.com/cerner/terra-core/pull/2379

PR #2379, opened 2019-05-02
stylelint version updated

PayalSawant:
### Summary
stylelint-config-terra version updated #2375.

@cerner/terra

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,7 @@
     "react-router-dom": "^4.0.0",
     "react-test-renderer": "^16.5.2",
     "stylelint": "^9.0.0",
-    "stylelint-config-terra": "^2.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -86,7 +86,7 @@
     "react-router-dom": "^4.0.0",
     "react-test-renderer": "^16.5.2",
     "stylelint": "^9.0.0",
```
bjankord: We'll want to update this to `^10.0.1` to match the peerDependency needed for stylelint-config-terra v3
https://github.com/cerner/stylelint-config-terra/blob/master/package.json#L53

bjankord: Resolved in this commit: https://github.com/cerner/terra-core/pull/2379/commits/e716325a4d0665700769eff3629a03e9d8f5a7cd

Issue #2079, opened 2018-11-30
[terra-demographics-banner] Background image does not fully extend to container


yuderekyu:
# Bug Report
In some cases if a background image is set on `demographics-banner`, it does not extend to all borders. A 1 pixel line is visible on either the bottom or right side, depending on the size of the browser window.

Zoom in on the below screenshot. The `demographics-banner` background image is set to a linear gradient. Notice that the white doc template background bleeds through on the bottom of the `demographics-banner`.
<img width="1392" alt="screen shot 2018-11-30 at 3 16 01 pm" src="https://user-images.githubusercontent.com/14063284/49315366-1301ab80-f4b3-11e8-868c-cb42c47ef778.png">

Could be a browser rounding issue: https://stackoverflow.com/questions/28832046/background-size-leaves-a-one-pixel-line-to-the-right-of-my-browser

## Environment

* Browser Name and Version: Chrome
* Operating System and version (desktop or mobile): Mac High Sierra on desktop

JakeLaCombe:
Can you provide the background style as part of this issue? It will make it easier to debug

yuderekyu:
Ah, I forgot to mention, this is when someone wants both a `background-color` and `background-image`. Add an inline background-image, and set it to some linear gradient.

<img width="1414" alt="screen shot 2019-01-02 at 2 37 08 pm" src="https://user-images.githubusercontent.com/14063284/50611443-0a2e3c00-0e9c-11e9-9149-53780f21e04d.png">

In this case, the background-image does not extend fully to the right. The dark background-color has ~ 1px  where there is no overlap.

Setting `background-clip: padding-box` fixes this case, but not the case in the initial screenshot. Checkout the clinical theme and [ocs-theme-variables branch](https://github.com/cerner/terra-core/compare/ocs-theme-variables#diff-2267f9efeef50e2d1511d46f3a178e10R1) for full styles to debug the initial screenshot

supreethmr:
This Issue seems to be fixed with PR : #2133 . no space / white line is being displayed in the demographics banner for the above specified scenarios.

yuderekyu:
@supreethmr hmm the issue persists within the internal clinical theme -  other styles in combination with `background-image` and  `background-color` may be creating this

yuderekyu:
@supreethmr emailed

supreethmr:
[...]

PR #2380, opened 2019-05-03
[terra-demographics-banner] Set background-clip to border-box

yuderekyu:
### Summary
Closes https://github.com/cerner/terra-core/issues/2079

Chrome has sub pixel rendering issues. Setting `background-clip: border-box` allows the background-image to extend to the border and fully mask the background-color.


bjankord:
Nice find on this @yuderekyu

Issue #2161, opened 2019-01-11
Form-Select speed handling large lists


neilpfeiffer:
# Feature Request

## Description
Certain lists may return 1000+ values which has fairly slow performance if not managed well by application. If possible, would help to create a version of the form-select that either (1) helps better manage with async options, and/or (2) reduce the number of items in the DOM using a virtualized/infinite-list approach, similar project: [react-virtualized-select](https://bvaughn.github.io/react-virtualized-select/).

Since terra-form-select by itself does not have any option to page or manage long lists, it forces all values to load as soon as a user focuses on the field. Some very non-official trial load times for reference, shows the page to be slower as more options are added, as expected. Greater than 1000 items could potential take a few seconds to load on a slower device.

Total Items | Time (ms)
------------ | -------------
10 | 90
100 | 115
1000 | 550

Without consumer doing abundance of service work, could it possible for version/s of a select help to manage extreme lists when a service or business logic is unable to manage effectively on its own.

neilpfeiffer:
cc: @papasmile

dev-cprice:
Did some perf tests in `production` with the current `terra-form-select` https://github.com/dev-cprice/terra-select-perf

Results found: [here](https://github.com/dev-cprice/terra-select-perf/tree/master/results)

While the numbers are smaller than the table above, this is run on a pretty powerful laptop. Obviously, the numbers will be much worse on lower-powered devices, e.g. mobiles.

supreethmr:
Verified this issue with new variant Native-select and it improves performance time to load the options much faster than the time mentioned above. Since Native select has certain limitations compared to existing variants of form-select it may not be resolve the performance issue with existing form-select. 

Possible Cause for the slowness while loading large number of options would be due to the rendering of each options on dropdown open which takes considerably longer time to render when number of options are more like mentioned in above table. 

Since Native-Select variant renders native options the time taken render 1000 options was around 80ms . 

To address this performance issue I have tried reduce the number options that are rendered on dropdown open by displaying only set of options that will be visible to user when it is opened instead of loading all the options at once. 

Since the number of options rendered on Open is reduced performance load time improves considerably compare to existing form-select. I have created Draft PR to test this code changes : 
https://github.com/cerner/terra-core/pull/2938 

In this PR I am trying to calculate the number of options thats fits in at initial render by comparing with Frame height. Rest of the items will be added to list as we scroll down. I am adding same set of options on each scroll so that scrolling behavior aligns with existing behavior.

[...]

PR #2382, opened 2019-05-04
Add visually hidden dropdown to help improve navigation experience on iOS VoiceOver

bjankord:
### Summary
When the dropdown is open, we render it in a portal. This is to allow it to be positioned easily in the DOM and avoid overflow/clipping issues that would arise if the dropdown was positioned relative to the select dropdown triggering element. 

By rendering the dropdown in a React portal, it causes some issues when it comes to navigating the select options when using gestures to navigate while using VoiceOver on iOS. To help with this, we've provided instructional text to guide users to tap on/interact with the toggle (chevron) to shift focus to  the portalled dropdown when interacting with select variants that allow text entry while using VoiceOver on iOS. 

While this works and does allow navigation of the select options, we want to make navigating to the select options more intuitive and allow navigating to the dropdown options via gesture navigation right after the user opens the dropdown when the click on the text input in the select.

To solve for this, we are now rendering a visually hidden dropdown that follows right after the select text input in the DOM. This allows for the virtual indicator to move from the text input to the list of options by double tapping on the text input to open the dropdown, then swiping right to move the virtual indicator into the dropdown option list when using VoiceOver on iOS.

Resolves: https://github.com/cerner/terra-core/issues/2383

### DIFF
--- a/packages/terra-form-select/CHANGELOG.md
+++ b/packages/terra-form-select/CHANGELOG.md
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Added
+* Added visually hidden dropdown to help make gesture navigation while using VoiceOver on iOS more intuitive when using select variants that allow text entry
 
 5.15.0 - (May 1, 2019)
 ------------------
--- a/packages/terra-form-select/src/_Dropdown.jsx
+++ b/packages/terra-form-select/src/_Dropdown.jsx
@@ -98,7 +98,9 @@ const Dropdown = ({
           onMouseDown={preventDefault}
           refCallback={refCallback}
         >
-          {children}
+          <div aria-hidden="true">
+            {children}
+          </div>
         </Hookshot.Content>
       </Hookshot>
     </React.Fragment>
--- a/packages/terra-form-select/src/_Dropdown.jsx
+++ b/packages/terra-form-select/src/_Dropdown.jsx
@@ -79,6 +79,18 @@ const Dropdown = ({
     );
   }
 
+  /**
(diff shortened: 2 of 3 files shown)


dev-cprice:
What are the implications for when the screen reader hits the bottom of the DOM? Will it repeat the visible options?

bjankord:
@dev-cprice I've update the visual dropdown to use `aria-hidden="true"` so only one instance of the dropdown will be interactable / read by screen readers on iOS. Since the visually hidden dropdown is positioned similar to the portal one, tapping on the portal options passes through and triggers the visually hidden dropdown.
* https://github.com/cerner/terra-core/pull/2382/commits/69a61ce114141fa943c4d00fd03560098a7f027b
* https://github.com/cerner/terra-core/pull/2382/commits/f47f9995be0b9501615f6896e44ce8cb555dc1f9

StephenEsser:
Were there any noticeable performance impacts with this change? We have an issue logged now around performance and this change appears to double the amount of options in the DOM.

bjankord:
@StephenEsser I didn't notice any slow downs in rendering on the iOS device, however, the examples only have a few options. I would say there will be a performance hit with this as it does require rendering the list of options in 2 places instead of 1 and like you said, there are already [performance issues](https://github.com/cerner/terra-core/issues/2161) with rendering just 1 set of options.

dev-cprice:
I think there's still some visual sync issues when selecting options 😞😞😞

it works great when no selections are made, but runs into issues when the input box grows to include more selections.

I'll post an example in a sec......

bjankord:
Thanks for the example @dev-cprice. I'm planning to put this PR on hold until I free up some capacity to look into other ways to solve #2383

PR #2384, opened 2019-05-06
[terra-profile-image]Updated Src to be an optional field.

### LINKED ISSUE #2359: [terra-profile-image] Allow consumers to load a default profile image
# Feature Request
Is it possible for src prop of terra-profile-image be made optional. Allowing consumers to load the failed profile image if src is empty/nil?

## Description
Allow consumers to load the default image when src is blank or empty. We as a consumer have the following scenarios for our user list

1. **User exists: Image Might Exist**
     a. Image successfully loads **(Works Today)**
     b. Image fails to load **(Works Today)**

2. **User exists: Image does not exist yet** (we know this before even attempting a image load)
     a. In this scenario we want the exact same Failed Profile image(**Feature Request**)

Today, to get this behavior we are having to pass a dummy url end point, but that increases the network traffic as it keeps trying to make the call to a non existent endpoint and keeps failing. And if we have auto refresh functionality implemented these dummy errors increase.

Also, just want to add that a terra-avatar does not really work for us here as we do not want to draw too much of a end user attention throwing in different kinds of profile images for the above scenarios. 

If further context is needed, please reach out to me on Teams.

pranav300:
### Summary
Made Src be an optional field.
When Src is not given it loads failed image.
Added WDIO and jest.
Resolves #2359


packages/terra-profile-image/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Changelog
 Unreleased
 ----------
 
+### Changed
+* Changed Src to be an optional prop.
```
nramamurth: Let's maybe do something like - 

1. Changed `src` to be an optional prop.
2. Use the avatar image as the default fallback image when `src` is not provided.

packages/terra-profile-image/src/terra-dev-site/test/profile-image/ProfileImageSourceNotProvided.test.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import ProfileImage from '../../../ProfileImage';
+
+const ProfileImageSourceNotProvided = () => (
+  <div>
+    <h2>Profile Image when source is not provided</h2>
+    <ProfileImage src="" alt="could not load profile image" width="75" height="75" />
```
nramamurth: nit: this could simply be 

```
<ProfileImage alt="default" width="75" height="75" />
```

packages/terra-profile-image/tests/jest/ProfileImage.test.jsx
```diff
@@ -6,3 +6,9 @@ it('should render a Terra image with a placeholder prop', () => {
   const wrapper = shallow(<ProfileImage src="profile.jpg" height="75" width="75" alt="A profile image" />);
   expect(wrapper).toMatchSnapshot();
 });
+
+/* Profile Image source is not provided */
+it('should render a Terra image without Source prop', () => {
```
nramamurth: `terra-image` is a component that comes into picture in a success scenario. I'd not mention it here in the test description for a scenario when `src` is not provided.

packages/terra-profile-image/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Changelog
 Unreleased
 ----------
 
+### Changed
+* Changed `src` to be an optional prop.
+* When `src` is not provided it loads avatarn image.
```
bjankord: nit: typo `avatarn` -> `avatar`

packages/terra-profile-image/tests/jest/ProfileImage.test.jsx
```diff
@@ -6,3 +6,9 @@ it('should render a Terra image with a placeholder prop', () => {
...
+  const wrapper = shallow(<ProfileImage src="" height="75" width="75" alt="Source not provided" />);
```
ShettyAkarsh: can we also remove the src attribute here ?

PR #2395, opened 2019-05-10
Remove custom props spread on base

emilyrohrbough:
### Summary
Removed custom props spead on I18nProvider. The I18nProvider does not accept custom props because it renders a React fragment, not a div

Ref: https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/src/I18nProvider.jsx#L22


packages/terra-base/src/Base.jsx
```diff
@@ -74,19 +74,16 @@ class Base extends React.Component {
...
       customMessages,
       strictMode,
-      translationsLoadingPlaceholder,
-      ...customProps
     } = this.props;
 
     const messages = Object.assign({}, this.state.messages, customMessages);
     const renderChildren = strictMode ? (<React.StrictMode>{children}</React.StrictMode>) : children;
 
     if (!this.state.areTranslationsLoaded) return <div>{this.props.translationsLoadingPlaceholder}</div>;
```
StephenEsser: Probably doesn't matter either way, but alternatively this could have used the `translationsLoadingPlaceholder` from the removed props spread above and dropped `this.props`.

PR #2410, opened 2019-05-22
Render action header as a div

bjankord:
### Summary
* Now renders `<ActionHeader />` as a div instead of a header element to avoid duplicate banner landmarks in the DOM

### DIFF
--- a/packages/terra-action-header/src/_ActionHeaderContainer.jsx
+++ b/packages/terra-action-header/src/_ActionHeaderContainer.jsx
@@ -57,14 +57,14 @@ const ActionHeaderContainer = ({
   ) : undefined;
 
   return (
-    <header {...customProps} className={cx(['flex-header', customProps.className])}>
+    <div {...customProps} className={cx(['flex-header', customProps.className])}>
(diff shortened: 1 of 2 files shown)


packages/terra-action-header/src/_ActionHeaderContainer.jsx
```diff
@@ -57,14 +57,14 @@ const ActionHeaderContainer = ({
...
-    <header {...customProps} className={cx(['flex-header', customProps.className])}>
+    <div {...customProps} className={cx(['flex-header', customProps.className])}>
       {startContent && <div className={cx('flex-end')}>{startContent}</div>}
       <div className={cx('flex-fill')}>
         {titleElement}
       </div>
       {content}
       {endContent && <div className={cx('flex-end')}>{endContent}</div>}
-    </header>
+    </div>
```
StephenEsser: I would assume a few snapshots are going to fail with this change

bjankord: I ran jest locally and didn't see any issues in terra-core, there may be some issues downstream in jest tests.

PR #2418, opened 2019-05-24
Dangerfile changelog checking improvements

Blackop778:
### Summary
Enforces that every package that has changes in the `src` folder has a changelog modification. Additionally even if only new files have been added a changelog modification is still needed. All packages that need a changelog modification will be listed as opposed to simply stating that a changelog entry need to be included.

Once this gets merged I will likely mirror the changes to other repositories.

Modification of code by @emilyrohrbough


emilyrohrbough:
With this change, lets bump to the latest danger version

Blackop778:
Updating the danger version requires using babel 7.

dangerfile.js
```diff
@@ -1,27 +1,28 @@
...
 
-const newChangelog = danger.git.created_files.filter((filePath) => {
-  const srcFilePattern = /^packages\/terra-([a-z-])*\/CHANGELOG.md/i;
-  return srcFilePattern.test(filePath);
-});
+const determinePackages = (changedList, path) => {
+  const path1 = path.split('packages/')[1];
+  const packageName = path1.split('/')[0];
+  return changedList.includes(packageName) ? changedList : changedList.concat([packageName]);
+};
```
StephenEsser: This function is a little strange. It is adding items to a parameter. It may be cleaner to remove the adding logic and return the package name, the calling reducer can then add the item.

If we want to keep this function altering params it would be slightly cleaner to make the reducer `changedList` a `Set`. This would remove the lengthy return statement. At the very least we should switch `changedList.concat([packageName])` to `changedList.push(packageName)`. Creating and concatenating an additional array isn't necessary.

StephenEsser: Here is a rough idea of what I meant by removing the reducers. This is just an example, these changes aren't required. 

https://gist.github.com/StephenEsser/5ac86a603d58e0042e40d381c4fcb224

Blackop778: I agree your code made it clearer what was happening. Had to fix an issue with checking files that aren't in any package.

dangerfile.js
```diff
@@ -1,31 +1,37 @@
...
-  fail('Please include a CHANGELOG entry with this PR.');
+// Fail if there are package changes without a CHANGELOG update
+if (missingChangelogs.length > 0) {
+  fail(`Please include a CHANGELOG entry for each changed package this PR. Looks like a CHANGELOG is missing for: \n - ${missingChangelogs.join('\n - ')}`);
 }
 
 // Warn when there is a big PR
-const bigPRThreshold = 1000;
-if (danger.github.pr.additions + danger.github.pr.deletions > bigPRThreshold) {
+if (danger.github.pr.additions + danger.github.pr.deletions > BIG_PR_THRESHOLD) {
```
emilyrohrbough: Lets remove this big threshold. It is 9.9 x / 10 going to complain

PR #2420, opened 2019-05-30
[terra-form-select] Prevent invalid options (options with only spaces) from being added in combobox variant.

### LINKED ISSUE #2414: [terra-form-select] Combobox Variant Renders Options Inconsistently
# Bug Report

## Description
Combo box variant displays an empty object selection when using the space-bar to display the space options. This spacing persists with hover styles when space is removed.

Reason I think it's a bug:
1. This selection 'space' option doesn't render the same height of the other select options.
2. It doesn't show the add option (`+` ) indication to show that it should be added to the list.
3. Steps 7-10 results in an empty option that receives hover styles. The initial open options does not have this spacing.

## Steps to Reproduce
1. Navigate to https://engineering.cerner.com/terra-core/#/raw/tests/terra-form-select/form-select/uncontrolled-combobox

2. Hit `tab` to focus to the select input

3. Hit `space` to open select options
<details><summary>screenshot</summary>
Screen Shot 2019-05-23 at 9 07 11 AM
</details>

4. Click middle of page to remove select focus
<details><summary>screenshot</summary>
Screen Shot 2019-05-23 at 9 07 18 AM
</details>

5. Hit `tab` to move focus to the select input

6. Hit `down arrow` key to open select options. See the empty, select option that is select. Its because it thinks it's a space.

<details><summary>screenshot</summary>
Screen Shot 2019-05-23 at 9 07 29 AM
</details>

7. Hit `backspace` key to remove the space.

<details><summary>screenshot</summary>
Shot 2019-05-23 at 9 13 55 AM
</details> 

8. Click middle of page to remove focus.

9. Hit `tab` to return focus to input.

10. Hit `down arrow` or `space bar` to see an empty option (or gap before options) that receives hover styles.

<details><summary>screenshot</summary>
Shot 2019-05-23 at 9 13 11 AM
</details>

## Expected Behavior
No empty selection option when no option was selected.

## Environment

* Component Name and Version: v5.15.0

nramamurth:
### Summary
Fixes #2414. 

### Additional Details
Updated a couple of Jest tests that should have supposedly tested combobox variant.

### Deployment Link
https://terra-core-deployed-pr-2420.herokuapp.com/#/components/terra-form-select/form-select/select


nramamurth:
@StephenEsser 

Should the workflow listed on #2414  be _wdio_ tested?

packages/terra-form-select/src/_FrameUtil.js
```diff
@@ -100,7 +100,7 @@ class FrameUtil {
     const { hasSearchChanged, searchValue } = state;
 
     if ((variant === Variants.TAG || variant === Variants.COMBOBOX)
-        && (variant !== Variants.TAG || searchValue.trim().length > 0)
```
StephenEsser: We should reach out to @neilpfeiffer to get his thoughts on this change. This line of code was previously allowing a user to clear the combobox by removing all characters within it. With this change, when a user clears the input field and tabs out, the previous value is reverted.

Previous: ( Clears on tab )
ezgif com-video-to-gif (1)

With this change: ( Reverts back to green on tab )
ezgif com-video-to-gif


Steps:

1. Click on a combobox to open the dropdown
2. Select an option
3. Click back into the input field and backspace all character
4. Tab out of the combobox

StephenEsser: The difference here was a true empty string `''` was never added back into the combobox when it was cleared. What we are wanting to avoid in this PR is allowing a series of spaces to be added. `"  "`

StephenEsser: I spoke offline with @neilpfeiffer. We want to keep the previous functionality of allowing the user to clear the combobox by removing the characters, but we still want to prevent blank spaces from being added for the combobox as options.

nramamurth: Thanks for following it up, @StephenEsser and @neilpfeiffer. In that case, could we simply do something like shown below? 

```diff
if ((variant === Variants.TAG || variant === Variants.COMBOBOX)
-  && (variant !== Variants.TAG || searchValue.trim().length > 0)
+  && (searchValue === '' || searchValue.trim().length > 0) // Allow empty string OR legit text.
   && hasSearchChanged
   && onSelect
   && (variant !== Variants.TAG || !FrameUtil.includes(props, searchValue))
   && (variant !== Variants.COMBOBOX || value !== searchValue)) {
      return true;
}
```

StephenEsser: Something similar, we want to make sure the Tag variant doesn't allow empty text to be created as a tag. 

The tricky part is I think we may need to treat blank strings as empty strings. Otherwise the workflow above happens again. 

A user makes a selection ( Green for example ), clicks back on the input, deletes all the characters, enters several spaces, presses tab. This would revert the blank spaces back to Green. I don't think we want that behavior.

bjankord: It may be helpful to document the behavior we intend for the combobox and the tag variants for this scenario clearly so @nramamurth has something actionable to code towards.
e.g. what steps should we test, what edge-cases should we consider.

nramamurth: @bjankord 

That'd be great! In the meanwhile, I am trying to understand how all components within `terra-form-select` come together. I have left a [comment](https://github.com/cerner/terra-core/issues/2414#issuecomment-506196791) on the original issue to take up the discussion further there.

StephenEsser: I left a comment on the issue thread: https://github.com/cerner/terra-core/issues/2414#issuecomment-507322746

packages/terra-form-select/CHANGELOG.md
```diff
@@ -28,6 +28,9 @@ Unreleased
 ### Changed
 * Made required updates to consume terra-toolkit v5 and terra-dev-site v5
 
+### Fixed
```
avinashg1994: This should be moved back to the unreleased section, it might have changed because of subsequent merges.

nramamurth: Good catch! Fixed it - https://github.com/cerner/terra-core/pull/2420/commits/cbf3624e2aae81db6575f6d66cc71975cdcedf24.

mmalaker:
Functional verification completed. The form select combo box example functions as expected and does not accept spaces only as a valid input option.

neilpfeiffer:
Tested: PR #2420
Resolves: Issue #2414
 
**Reviewed:**
- removes ability to add empty strings as accepted value in combobox

**Considerations:**
- future: if value was already accepted and you modify the value or select another option, or press spaces, then select escape or "click-out", the original value should be retained. Will be resolved by issue #2474 or similar.

 
Marking as Verified, ready for Release.
 
**Pass** |
-----------|
UX Review|

PR #2436, opened 2019-06-07
[terra-core] Removed node_modules from .npmignore

### LINKED ISSUE #2415: Investigate removing node_modules from .npmignore
# Feature Request
We might be able to remove `node_modules` from each `.npmignore`, because npm does this by default. Came across this while reading [npm documentation](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package).

> Additionally, everything in node_modules is ignored, except for bundled dependencies. npm automatically handles this for you, so don’t bother adding node_modules to .npmignore.

We're on the latest v8. I'm not 100% sure if v8 defaults to this.

avinashg1994:
### Summary
Removed _node_modules_ from ".npmignore" of each package.

### Additional Details
Verified these changes by running _**npm pack**_ command, before and after the change, locally which generated a tarball in the working directory and by comparing the total files in the tarball both before and after.

resolves #2415 

@cerner/terra


avinashg1994:
Why is the CHANGELOG.md file not being detected by Travis as changed, even though i have update it?

Blackop778:
The danger issue has been fixed in #2431

PR #2437, opened 2019-06-07
[terra-action-header] Updated misleading documentation

### LINKED ISSUE #2422: [terra-action-header] Update misleading 'children' documentation
# Bug Report
Update misleading 'children' documentation

## Description
The 'children' proptype documentation is vague about the acceptable element for the component, as a result consumers mistakenly pass invalid elements. Should be updated to mention that in order to function properly children should only be a single terra-collapsible-menu-view.

https://github.com/cerner/terra-core/blob/master/packages/terra-action-header/src/ActionHeader.jsx#L50

lokesh-0813:
### Summary
Updated terra-action-header misleading `children` documentation.

resolves #2422


packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -47,7 +47,7 @@ const propTypes = {
    */
   onPrevious: PropTypes.func,
   /**
-   * Child element to be displayed on the right end of the header.
+   * Displays a single child element on the right end of the header.
```
StephenEsser: We should double check with @dkasper-was-taken to make sure this is what he had in mind. It's difficult to tell from the issue if he wanted the collapsable item view called out here specifically as the only element intended to be used.

dkasper-was-taken: Yes, this should be clarified that the only intended component for that location is the terra-collapsible-menu-view, as currently consumers are trying to put other elements in their without realizing this.

lokesh-0813: Ok @dkasper-was-taken, I will update it to just single terra-collapsible-menu-view. I assumed it could take any single terra component.

PR #2452, opened 2019-06-12
[terra-responsive-element] - Update breakpoint ranges

### LINKED ISSUE #2048: [terra-responsive-element] Consume terra-breakpoints and update size props

StephenEsser:
### Summary

This PR updates terra-responsive-element to use the breakpoints defined by terra-breakpoints.

- Removed `breakpoints.scss`
- Removed the `defaultElement` prop
- Changed the breakpoint ranges to match those defined in terra-breakpoints
- Added `onChange` callback function that is invoked when a breakpoint range changes
- Added an `onResize` callback function that is invoked when a resize event is triggered
- Added a `children` prop to use in conjunction with `onChange` to create controlled responsive elements

### Additional Details

Resolves #2048

@cerner/terra

[License]: ../LICENSE


StephenEsser:
I have a few more changes I will be pushing soon to update the paginator, demographics banner, and alert to use the new responsive element.

packages/terra-alert/src/Alert.jsx
```diff
@@ -171,8 +171,7 @@ const Alert = ({
 
   return (
     <ResponsiveElement
-      responsiveTo="parent"
```
mjhenkes: did you intend to remove this?

StephenEsser: Yes, parent is the default value of responsiveTo, so it is redundant have in code

Issue #2450, opened 2019-06-12
[terra-status-view] Refactor Buttons


yuderekyu:
# Feature Request
The status view has a `children` prop of type node. The stylesheet will apply spacing styles to any html button passed as a child element (gross). From offline discussions, there are two options.

## Remove Styles
* Remove most spacing styles for the action section and let consumers control spacing.

## Remove Children
* Remove ability to pass any element as a node. Instead, provide a prop `buttonAttrs`, that takes in array of button props. We'll use these to render buttons in the action section and control spacing styles.

Anyone have preferences or feedback?

yuderekyu:
I am mocking out the 2nd option in my development branch `refactor-status-view`.

Issue #2440, opened 2019-06-07
[terra-base] Internet Explorer calculates percentage based font-size differently


yuderekyu:
# Bug Report

## Description
Internet explorer 11 calculates 87.5% font-size as 13.93px. This can cause downstream issues for components that use `rem`.

<img width="960" alt="base1 (1)" src="https://user-images.githubusercontent.com/14063284/59126242-31e16e80-892a-11e9-8126-aef5f32cefea.png">

<img width="960" alt="base2" src="https://user-images.githubusercontent.com/14063284/59126182-0bbbce80-892a-11e9-98ee-aae929ca97cc.png">

## Example 
This causes status view to fail contrast ratios on internet explorer. 

Good ol' status view has a font-size of `1.714rem`. IE automatically rounds rems to the hundredth spot. So the end value is `1.71rem`. Multiply the font size by the base font size `13.93px` to equal  `23.8203px`, which then rounds to `23.8px`. Don't ask me why it rounded to the tenths this time. `23.8px` converts to `17.85pt` font. Under `18pt` fonts follow different color contrast ratios, so this subsequently fails axe tests.

<img width="960" alt="status-view-font1" src="https://user-images.githubusercontent.com/14063284/59130952-81c63280-8936-11e9-93df-0d92039db86d.png">

<img width="960" alt="status-view-font2" src="https://user-images.githubusercontent.com/14063284/59130950-7ffc6f00-8936-11e9-9ae3-11ba6f0ed950.png">

yuderekyu:
Talked offline with @bjankord. Potential solution is to rectify this case by case. For status view, we can set the font-size to `1.73rem` which bumps the font size on internet explorer to `24.13px`. This will affect other browser calculations and is non-passive

PR #2453, opened 2019-06-13
[terra-status-view] Rework component

### LINKED ISSUE #1578: [terra-status-view] Status view does not re-render on resize.
# Issue Description
The status view does not re-render on resize which means the status glyph and message will not appear centered vertically.

## Issue Type

- [ ] New Feature
- [ ] Enhancement
- [x] Bug
- [ ] Other

## Expected Behavior
The status view should re-render during resize, repositioning its message and glyph in the center of its parent.

## Current Behavior
The status view does not re-render during resize so its client height remains fixed.

## Steps to Reproduce

1. Create a status view.
2. Resize it's parent.
3. Observe its position is incorrect.

## Environment

* Component Version: 2.13.0
* Browser Name and Version: Chrome 66.0.3359.181
* Operating System and version (desktop or mobile): macOS 10.12.6

yuderekyu:
### Summary
Rework status-view :D

#### Removed

- Removed children prop. Replaced with buttonAttrs prop
- Removed the default of rendering an error view
- Removed the custom variant
- Removed hide glyph logic based on available container space
- Removed onComponentMount logic to calculate padding
- Removed StatusView.Opts object and export

#### Changed

- Use CSS grid to center component
- Render divider if title and message/buttons exist

### Deployement Link
https://terra-core-deployed-pr-2453.herokuapp.com/#/components/terra-status-view/status-view/status-view

Closes #1578, #2450, #2440


packages/terra-status-view/docs/README.md
```diff
@@ -12,19 +12,32 @@ The status view component provides a customizable icon, message, and buttons in
...
   isGlyphHidden={false}
   variant="error"
   title="A 500 error was thrown from the service."
   message=message="A null pointer exception was thrown."
->
-  <Button text="OK" key="1 />
-  <Button text="Cancel" key="2" />
-</StatusView>
+  buttonAttrs={buttons}
+/>
```
StephenEsser: I think many components have been removing this usage sections. Example code can be found within the doc template examples.

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -11,22 +13,27 @@ const StatusViewVariants = {
...
-   * The glyph that is to be displayed for the `custom` variant.
+   * Display a custom glyph. Overrides a variant's default glyph.
    */
   customGlyph: PropTypes.node,
 
   /**
-   *  Determines if the content should be aligned vertically at the top of the container rather than in the middle.
+   * @private The intl object to be injected for translations. Provided by the injectIntl function.
+   */
+  intl: intlShape.isRequired,
```
StephenEsser: Might need an intl expert's opinion here but I don't think these need to be defined within the props anymore. I think it can be grabbed using the context.

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -36,223 +43,148 @@ const propTypes = {
...
-        <div className={cx('divider')} ref={(element) => { this.dividerNode = element; }}>
-          <Divider />
-        </div>
-      );
-    }
+  let defaultTitle;
+  if (checkVariantExistence(variant)) {
+    defaultTitle = intl.formatMessage({ id: `Terra.status-view.${variant}` });
+  } else {
+    defaultTitle = '';
```
StephenEsser: This won't pass the if statement below because Javascript evaluates empty string as false. This line can be removed, unless the empty string is meant to fall through into the titleSection which might require some logic work below.

packages/terra-status-view/src/StatusView.jsx
yuderekyu: Weren't we moving away from context based on React API instability?

StephenEsser: I'm not entirely sure, we should get one of the intl people to clarify which way we should be implementing this.

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -36,223 +43,148 @@ const propTypes = {
...
   title: PropTypes.string,
 
   /**
-   * Determines the glyph to display, one of the following: `no-data`,
-   * `no-matching-results`, `not-authorized`,
-   * `error`, or `custom`.
+   * Sets the glyph and title using a pre-baked variant. One of the following: `no-data`,
+   * `no-matching-results`, `not-authorized`, or `error`
    */
   variant: PropTypes.oneOf(Object.values(StatusViewVariants)),
```
StephenEsser: Should this be required or can you implement a custom status view by not providing a variant?

yuderekyu: talked offline - `title`, `message`, and/or `customGlyph` props will override the defaults. I thought it was redundant to set the variant to `custom` in order to render those first class overrides. I'll talk to @neilpfeiffer on what use cases exist for custom status views. Maybe we can remove that functionality

packages/terra-status-view/docs/README.md
yuderekyu: updated

packages/terra-status-view/src/StatusView.jsx
yuderekyu: removed

packages/terra-status-view/CHANGELOG.md
```diff
@@ -1,8 +1,19 @@
 Changelog
 =========
 
-Unreleased
+Unreleased - MVB
 ----------
+### Removed
```
mjhenkes: I've seen elsewhere changes like these listed as 'breaking'

yuderekyu: updated

packages/terra-status-view/src/StatusView.jsx
tbiethman: react-intl isn't using the _new_ Context API (yet), but it's still using context. To Stephen's point, both solutions are grabbing the `intl` object from the context. However, in the previous implementation, the StatusView is interfacing with context directly, whereas now `injectIntl` does the context interfacing and forwards `intl` to StatusView as a prop. The new way is future-proofed a bit, as `injectIntl` abstracts the Context API used so StatusView doesn't have to care. If we stick with the old way, StatusView would break when react-intl updates.

I say all this with the caveat that when react-intl does start using the new Context API, we could simplify the implementation here by using the useContext hook. We'd want to make changes eventually to remove that injectIntl wrapper because it's unnecessary at that point. However, we'd have some runway to do so without immediately failing.

TLDR: This is the best way to currently get access to `intl`.

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -36,223 +43,146 @@ const propTypes = {
...
-            <svg className={cx(variant)} />
-          </div>
-        );
-      }
-    }
+  const generatedButtonsArray = [];
+  buttonAttrsArray.forEach(button => generatedButtonsArray.push(
+    <Button {...button} className={cx(['button', button.className])} />,
+  ));
+  return generatedButtonsArray;
```
tbiethman: Wanna use map?
```jsx
return buttonAttrsArray.map(button => <Button {...button} className={cx(['button', button.className])} />);
```

tbiethman: Are key's required here? We should we document that if they are.

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -11,22 +13,27 @@ const StatusViewVariants = {
...
 };
 
 const propTypes = {
   /**
-   * An array of actionable buttons to display.
+   * An array of objects containg terra-button properties. Use these to render buttons in the bottom section.
+   * Example:`[{ text: 'Button 1', key: 1, size: 'medium', variant: 'action', onClick: onClickFunction}]`
    */
-  children: PropTypes.node,
+  buttonAttrs: PropTypes.arrayOf(PropTypes.object),
```
tbiethman: Can we use Button.propType instead of PropTypes.object?

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -36,223 +43,146 @@ const propTypes = {
...
-    if (messageSection || actionSection) {
-      dividerSection = (
-        <div className={cx('divider')} ref={(element) => { this.dividerNode = element; }}>
-          <Divider />
-        </div>
-      );
-    }
+  let defaultTitle;
+  if (checkVariantExistence(variant)) {
+    defaultTitle = intl.formatMessage({ id: `Terra.status-view.${variant}` });
```
tbiethman: All things considered, you could probably use react-intl's `<FormattedMessage />` component here and remove `injectIntl` and the prop. It'll wrap the string in a <span> tag, but I don't think that will impact anything.

packages/terra-status-view/src/terra-dev-site/doc/example/ToggleVariants.jsx
```diff
@@ -0,0 +1,37 @@
...
+            <option value="no-matching-results">No Matching Results</option>
+            <option value="not-authorized">Not Authorized</option>
+          </select>
+        </fieldset>
+      </div>
+    );
+  }
+}
+
+export default ToggleVariants;
```
tbiethman: If you're interested, these example components are prime candidates for the usage of [hooks](https://reactjs.org/docs/hooks-state.html).

packages/terra-status-view/src/StatusView.jsx
yuderekyu: updated and added a comment on the prop comment

packages/terra-status-view/src/StatusView.jsx
yuderekyu: updated - i had to wrap that in a `shape`, otherwise react complains

packages/terra-status-view/src/StatusView.jsx
yuderekyu: updated to FormattedMessage

packages/terra-status-view/src/terra-dev-site/doc/example/ToggleVariants.jsx
yuderekyu: good idea, update to use hooks https://github.com/cerner/terra-core/pull/2453/commits/8ad0a3ff2a728771b709167dc3a0617952b5f864

packages/terra-status-view/docs/README.md
```diff
@@ -1,32 +1,13 @@
 # Terra Status View
 
-The status view component provides a customizable icon, message, and buttons in a view.
+Renders an icon, title, message, and buttons.
```
jmsv6d: This doesn't really tell consumers anything about what this component is or what it should be used for.

packages/terra-status-view/docs/UPGRADEGUIDE.md
```diff
@@ -1,5 +1,37 @@
...
+
+<StatusView buttonAttrs={buttons} />
+```
+
+* **Removed** default behaviour of rendering an error view. Please provide props.
+* **Removed** the `custom` variant. Use first class props `title`, `message`, or `customGlyph` to render respective sections or override a variant's sections.
+
+### Resizing
+* Use `isGlyphHidden` to show or hide the glyph. The `onComponentMount` logic to do this has been removed.
+* Status view defaults to a "centered" alignment, based on available size and remaining space. The top will contain `40%` of space, while the remaining `60%` is delegated to the bottom. Use `setAlignTop` to remove this distribution and render the status view at the top of the container.
```
jmsv6d:
```suggestion
* Status view defaults to a "centered" alignment, based on available size and remaining space. The top will contain `40%` of space, while the remaining `60%` is delegated to the bottom. Use `isAlignedTop` to remove this distribution and render the status view at the top of the container.
```

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -36,223 +40,142 @@ const propTypes = {
...
-    let dividerSection;
-    if (messageSection || actionSection) {
-      dividerSection = (
-        <div className={cx('divider')} ref={(element) => { this.dividerNode = element; }}>
-          <Divider />
-        </div>
-      );
-    }
+  let defaultTitle;
+  if (checkVariantExistence(variant)) {
```
jmsv6d: You should really only need to check that variant is defined, the propType defines which values are allowed, I don't think that we need to guard against improper usage of the API.

yuderekyu: if this check is removed, an erroneous title will render for a non existent variant - `Terra.status-view.{erroneous-title}` - should we let this happen versus not rendering anything? I'm fine either way

packages/terra-status-view/docs/README.md
yuderekyu: updated

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -11,22 +13,24 @@ const StatusViewVariants = {
...
   ERROR: 'error',
-  CUSTOM: 'custom',
 };
 
+/* eslint-disable react/forbid-foreign-prop-types */
 const propTypes = {
   /**
-   * An array of actionable buttons to display.
+   * An array of objects containg terra-button properties. A key attribute is required for each object.
+   * Use this array to render buttons in the bottom section.
```
Blackop778:
```suggestion
   * This array is used to render buttons in the bottom section.
```

packages/terra-status-view/docs/UPGRADEGUIDE.md
```diff
@@ -1,5 +1,37 @@
 # Terra Status View Upgrade Guide
 
+## Changes from version 3 to version 4
+
+### Props
+* **Removed** children prop. This is **replaced** by `buttonAttrs`. This prop accepts an array of objects containing terra-button properties. Example:
```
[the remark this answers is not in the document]
yuderekyu: updated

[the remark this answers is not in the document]
Blackop778: I'd probably call out `key` specifically since the items in the array actually have `id` props too.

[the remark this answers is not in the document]
yuderekyu: oops, I swapped them. I replaced id with key and removed id from the examples

packages/terra-status-view/src/StatusView.jsx
jmsv6d: I think we should. It's an invalid scenario either way and removing the check for a valid variant is less code to upkeep.

yuderekyu: removed

mmalaker:
Functional verification completed. The status view examples display and function as expected with the above changes.

packages/terra-status-view/src/StatusView.jsx
neilpfeiffer: Custom is rare, so ideally most uses should/will choose a prop as their default choice. I think this is a cleaner API to not forcibly need to reset to custom (which is really just an 'un-set' variant). We can update documentation to more prominently encourage variant picking or discouraging custom-use if we need to.

packages/terra-status-view/src/terra-dev-site/doc/example/ToggleVariants.jsx
neilpfeiffer: @yuderekyu +1 on the select and checkbox toggle examples. hooks were a nice touch, really shortened up the implementation.

packages/terra-status-view/src/StatusView.module.scss
```diff
@@ -88,18 +107,15 @@
...
-      margin-top: var(--terra-status-view-actions-button-margin-top, 0);
-      max-width: 100%;
-    }
+  .button {
+    flex: var(--terra-status-view-actions-button-flex, 0 1 auto);
+    flex-shrink: 1;
+    margin-bottom: var(--terra-status-view-actions-button-margin-bottom, 0.71428rem);
+    margin-left: var(--terra-status-view-actions-button-margin-left, 0.357rem);
+    margin-right: var(--terra-status-view-actions-button-margin-right, 0.357rem);
+    margin-top: var(--terra-status-view-actions-button-margin-top, 0);
```
neilpfeiffer: oops - made a mistake. don't need a new theming variable. subtract back https://github.com/cerner/terra-core/pull/2453/commits/20e999e6ff16ee9dc907dc3470a6615482c10492 to remove margin-bottom again.
Then just change `.button { margin-top:` to `0.71428rem` and reduce `.actions { padding-top:` to `0.71428rem` [(line 108)](https://github.com/cerner/terra-core/pull/2453/files#diff-b68e8f9acd496ddbab9eb7f3855aa53eR108) to have the same effect without adding anything new.

neilpfeiffer:
Tested: PR #2453
Resolves: Issues #1578, #2450, #2440

**Reviewed:**
- refactors Status View with minimal to no apparent changes. 

\
\
Verified by @neilpfeiffer, Ready for Release after minor edit: https://github.com/cerner/terra-core/pull/2453#pullrequestreview-255515874.

**Pass** |
-----------|
UX Review|

packages/terra-status-view/src/StatusView.module.scss
ryanthemanuel: https://github.com/cerner/terra-core/pull/2453/commits/807b3881725c3e0cc556ab6f14aa9c057260dce4

ryanthemanuel:
Looks like a breaking change (https://github.com/conorhastings/react-syntax-highlighter/issues/193) . Locking down react-syntax-highlighter until that's fixed.

PR #2455, opened 2019-06-13
[terra-form-select] Added null check in form-select

prmeznar:
### Summary
This adds a null check on an additional 'data-select-active' query to prevent empty queries from causing errors. This is associated to #2327, which is in response to https://github.com/cerner/terra-framework/issues/592

This change is in response to a workflow where on clicking a select field we would call a service to find the options that should be displayed. If, while that service was running, the user enters a value when the select field previously had no value, upon service completion the component would error out due to a null activeOption getting dereferenced on line 499. This would occur when the value was originally empty and when the select box search value was originally populated, but cleared and a value entered before the service completed.


packages/terra-form-select/src/_Menu.jsx
```diff
@@ -490,17 +490,17 @@ class Menu extends React.Component {
    * Options below the dropdown will be scrolled to the bottom of the menu.
    */
   scrollIntoView() {
-    if (!this.state.active) {
+    const activeOption = this.menu.querySelector('[data-select-active]');
+    if (!this.state.active || activeOption === null) {
       return;
     }
```
StephenEsser: Can you provide steps to reproduce the issue this PR solves?

Something might be out of sync, if the state claims there is an active element, it should be possible to query for it. Something suspicious might be going on.

prmeznar: I attempted to outline it above. The short version for us was:
- **Empty** select that calls some service to that returns data that becomes options
- Click into select, before service returns, type in anything
- When service returns, it attempts to render new options, and fails

PR #2465, opened 2019-06-20
[terra-core] Removed inline style for dev-site components Part 2

### LINKED ISSUE #2386: Remove inline styles in dev site components
# Feature Request

## Description

avinashg1994:
### Summary

1. Content Container
2. Demographics Banner
3. Dialog
4. Divider
5. Doc Template
6. Dynamic Grid

PartiallyResolves #2386


packages/terra-dialog/src/terra-dev-site/test/dialog/DefaultDialog.test.scss
```diff
@@ -0,0 +1,6 @@
+:local {
```
emilyrohrbough: Shouldn't these sass files be `name.module.scss`? @yuderekyu @bjankord

bjankord: For consistency sake, yeah they should. We use the `name.module.scss` format to make sure that CSS modules are detected and applied when the scss files are used in tools like create-react-app, rails/webpacker, codesandbox.io. Since it is unlikely people will use these test files in those tools, its not as big of a concern but I think for consistency, we should go with the  `name.module.scss` format for any file that is intended to be used as a css module.

avinashg1994: To maintain consistency i have updated the scss file names to follow the `name.module.scss` format.

PR #2473, opened 2019-06-24
[terra-core] Removed inline style for dev-site components Part 7

### LINKED ISSUE #2386: Remove inline styles in dev site components

avinashg1994:
### Summary

1. Spacer
2. Status View
3. Text

Partially Resolves #2386


packages/terra-status-view/src/terra-dev-site/doc/example/StatusViewDocCommon.module.scss
```diff
@@ -0,0 +1,16 @@
+:local {
+  .status-view-1,
+  .status-view-2,
+  .status-view-3 {
```
emilyrohrbough: Can we rename to status-view-height, stutus-view-width, status-view-overflow?

avinashg1994: Done

PR #2474, opened 2019-06-24
[terra-form-select] Show the display of the active option

### LINKED ISSUE #2257: Select control calls onChange with search value when + Add option is not present
# Bug Report

## Description

In the select component with the 'combobox' variant the value that is given to onChange is not consistent. When the option is selected the option.value is given to onChange. When the + Add option is selected the search value is given to onChange. When the search doesn't match exactly and the user loses focus onChange is called with the current searchValue. Given the last scenario is expected functionality it then becomes an issue when the search results are filtered down to an exact match, the + Add option disappears, then focus is lost, onChange is called with the search value and not the option value of the exact match.

Since + Add disappears the user would expect the exact option value is being applied when in reality a 'free text' value is being applied. In the application we're using the Select control for dose that does impost a significant difference in behavior. In our details pane we also use the 'search' variant where when loses focus does not cause an option to be selected at all. This inconsistent behavior could be confusing. 

## Steps to Reproduce

1. Select controls has options
2. Search till results are filtered to one matching option exactly
3. Control loses focus

Video of the issue here. You can see the option.value is printed in the console when an option is selected, and the search value when it's free text.
https://cernerprod.sharepoint.com/:v:/s/OrdersandPlansDev/Projects/Ef0lMq5VmXhPsqNSJjCYS7gBxS0JQfKGi64kFYJdStwOfw?e=du20s4

## Expected Behavior &  Possible Solution

Possible Resolutions (in order of preference)
# Ideally we would like onChange to be not be called when the user didn't specifically select and option or select the + Add option.
# We would then like the option.value to be sent to onChange when search result have exactly one visible option.
# Always show the + Add option

## Environment

* Component Name and Version: terra-form-select, 4.26.0
* Browser Name and Version: Chrome, latest

## @ Mentions

@StephenEsser

benbcai:
### Summary
When the search string in the select input matches the display of an option, tabbing away would add the search string as a new option. As a result, the menu would have duplicate options with the display being the same but the value would be different. This PR is to show the display of the active option in the select input. When there is a match, the option in the menu would be selected instead of the adding a new option using the search string as the display and value.

### Additional Details
- Closes #2257 
- When navigating the options in the menu using the Up/Down arrow keys, the pre-selected option will be displayed in the select input.
- Move the Add option to be the top option in the menu.

### Deployment Link
https://terra-core-deployed-pr-2474.herokuapp.com/#/components/terra-form-select/form-select/select

Note: This PR currently does not include new tests or fixing existing tests. These changes will come shortly.

@cerner/terra

[License]: ../LICENSE


mmalaker:
I'm seeing some strange keyboard navigation behavior with the form select control. The Enter/Return key should open the menu. Currently only spacebar and down arrow will open the menu. Also, the Dropdown icon should receive tab focus since it is clickable. Finally, on desktop Safari, the menu displays but the items are not highlighted accurately as you arrow through the options
Screen Shot 2019-07-22 at 10 11 31 AM
 (see screenshot)

packages/terra-form-select/src/_Menu.jsx
```diff
@@ -432,11 +452,16 @@ class Menu extends React.Component {
     } else if (keyCode === KeyCode.KEY_END) {
       event.preventDefault();
       this.setState({ active: Util.findLast(children) });
-    } else if (variant === Variants.DEFAULT && keyCode >= 48 && keyCode <= 90) {
+    } else if (/* variant === Variants.DEFAULT && */ keyCode >= 48 && keyCode <= 90) {
```
tbiethman: Still got this comment hiding here

benbcai:
This PR has conflicts with the changes made in #2549. Putting this PR on hold and will refactor once #2549 is merged.

PR #2475, opened 2019-06-25
[terra-core] Removed inline style for dev-site components Part 8

### LINKED ISSUE #2386: Remove inline styles in dev site components

saket2403:
### Summary

1. Scroll 
2. Signature
3. Status 
4. Table 
5. Tag 
6. Toggle Button 

Resolves #2386


packages/terra-scroll/src/terra-dev-site/doc/example/ScrollVertical.module.scss
```diff
@@ -0,0 +1,13 @@
+:local { 
+  .overflow-y-wrapper {
```
avinashg1994: We could combine `ScrollHorizontal.module.scss` and `ScrollVertical.module.scss` into a single scss file say `ScrollCommon.module.scss`. Since "overflow-x-wrapper" and "overflow-y-wrapper" have the same styles.

PR #2479, opened 2019-06-26
[terra-avatar] to allow non square images

### LINKED ISSUE #2259: [terra-avatar] should allow non-square images
# Feature Request

## Description
Current avatar accepts images, but distorts the presentation unless the image provided is a perfect square. https://github.com/cerner/terra-core/pull/2224#issuecomment-469676287

Need to determine if terra-avatar should allow for non-square images to be handled such that they do not distort the image/photo provided. 

screen shot 2019-03-05 at 8 53 56 am

pranav300:
### Summary
Resolves #2259

### Additional Details
Added `object-fit-images` polyfill as `object-fit` property is not supported by IE.


packages/terra-avatar/src/variants/Avatar.jsx
```diff
@@ -58,7 +59,7 @@ const defaultProps = {
 class Avatar extends React.Component {
   constructor(props) {
     super(props);
-
+    objectFitImages();
```
tbiethman: To better encapsulate the polyfill's effects, you could provide an element here to update the individual Avatar instance: https://github.com/bfred-it/object-fit-images#objectfitimagesimages-options

You'd have to do this inside componentDidMount (and probably componentDidUpdate, if changing the avatar props requires execution again).

packages/terra-avatar/src/common/Avatar.module.scss
```diff
@@ -63,7 +63,9 @@
 
   .avatar > .image {
     border-radius: 50%;
+    font-family: 'object-fit: cover;';
```
bjankord: Let's add a comment that this is used to provide settings to the object-fit polyfill

packages/terra-avatar/src/variants/Avatar.jsx
pranav300: ComponentDidMount works fine, as changing the avatar props does not require execution again.

packages/terra-avatar/src/common/Avatar.module.scss
```diff
@@ -63,7 +63,9 @@
 
   .avatar > .image {
     border-radius: 50%;
+    font-family: 'object-fit: cover;'; //This is added to provide settings for object-fit-images polyfill
     height: 100%;
+    object-fit: cover;
```
yuderekyu: if we use object-fit, is it necessary to define 100% height and width?

pranav300: What I have found is height and width 100% is required for non-square images.

PR #2481, opened 2019-06-27
[terra-overlay] Fix for freezing of loading-overlay test pages

### LINKED ISSUE #2476: [terra-overlay] Page freezes on the test pages on terra-dev-site.
# Bug Report

## Description
Page Freezes for  "Loading Overlay" Test Page on terra-dev-site and the user is unable to navigate to anywhere on the page.

## Steps to Reproduce
1. Navigate to : https://engineering.cerner.com/terra-core/#/tests/terra-overlay/overlay
2. Click on "Loading Overlays"
3. Notice that when "Animate Loading Overlay" loads, the page freezes and the user is unable to navigate anywhere on the page.

AnimatedLoadingOverlay
DarkLoadingOverlay

## Expected Behavior
When "Loading Overlay" pages are loaded, the page should not freeze and the user should be able to navigate to anywhere on the page.

## Possible Solution
There are 2 possible solutions to this:

1. Wrapping the "Loading Overlay" in an "OverlayConatiner" solves the issue. 
2. Another possible solution is to provide a Timeout (say, 5 sec) to terminate the Loading Overlay.

## Environment
* Component Name and Version: Terra Overlay
* Browser Name and Version: All Browsers

## @ Mentions
@ShettyAkarsh @ryanthemanuel

avinashg1994:
### Summary
Fixed the freezing of Loading Overlay test pages by wrapping it with an overlay-container.

Fixes #2476


packages/terra-overlay/src/terra-dev-site/test/overlay/LoadingOverlay/CustomMessageLoadingOverlay.test.jsx
```diff
@@ -1,4 +1,24 @@
...
+    super();
+
+    this.state = { show: true };
+  }
+
+  componentDidMount() {
+    setTimeout(() => {
+      this.setState({ show: false });
+    }, 5000);
+  }
```
StephenEsser: We discussed this issue in scrum earlier this week, I believe we want to avoid timers in the test examples. We may want to update this example to use the Container as well and remove the timeout.

cc: @bjankord

avinashg1994: I was made aware that we had to use "Overlay Containers" and not timeout as per discussion on #2476 . I also implemented the same as can be seen in https://github.com/cerner/terra-core/pull/2481/commits/28471e882788407dd7b73e2f8574487a92cec06f commit.
But, there is a test in `loading-overlay-spec.js` to check whether the loading overlay is in "fullscreen mode" or not. This check was added to test the following functionality:
https://github.com/cerner/terra-core/blob/master/packages/terra-overlay/src/Overlay.jsx#L151.
So i changed from using an "Overlay Container" to a "timer" for that test.

Is there any other approach for this while keeping that check?
**'or'**
Do I go ahead with an "Overlay Conainer" by removing that check?

bjankord: I don't have a strong preference on this one way or another. @mjhenkes @ryanthemanuel @tbiethman Thoughts?

bjankord: I'm fine with the timer on the test page.

PR #2487, opened 2019-06-28
Create Terra Dropdown Button

### LINKED ISSUE #2216: Create new Dropdown-Button component
# Feature Request

## Description

Create a new dropdown-style button component with two features:
- one for a split-button (2 targets: regular button plus a dropdown icon-button that reveals the menu) 
- one for a dropdown-button (1 target: clicking button reveals the menu)

Button should also contain a prop for `isBlock` to stretch to fill 100% of container's width.

Additional details posted here: https://zpl.io/a3gPoXy

[Tech design link](https://github.com/cerner/terra-core/issues/2216#issuecomment-501418592)

screenshots:

----
- **Split-Button example:**
screen shot 2019-02-06 at 11 31 18 am

- **Dropdown-Button example:**
screen shot 2019-02-06 at 11 30 43 am

Blackop778:
### Summary
A button that, when pressed, opens a dropdown with more buttons with custom callbacks. Has another variation that also shows a button with a custom callback outside the dropdown.

PR is still waiting for some UX feedback from @neilpfeiffer but code is ready for review.

### Deployment

- [Dropdown Button Documentation](https://terra-core-deployed-pr-2487.herokuapp.com/#/components/terra-dropdown-button/dropdown-button/dropdown-button)
- [Split Button Documentation](https://terra-core-deployed-pr-2487.herokuapp.com/#/components/terra-dropdown-button/dropdown-button/split-button)
- [Test pages](https://terra-core-deployed-pr-2487.herokuapp.com/#/tests/terra-dropdown-button/dropdown-button/block-dropdown-button)

### Expected Interactions

- Enter / Space / Mouse click to trigger primary option / open dropdown / trigger options in dropdown
- Up and down arrows to navigate up / down inside dropdown
- Home to jump to 1st item inside dropdown
- End to jump to last item inside dropdown
- Typing numbers or letters to jump to first option beginning with that character (resets after half a second of no typing)
- Tab / Escape inside dropdown to close dropdown
- Click outside dropdown to close dropdown

### Additional Details
Closes #2216


packages/terra-dropdown-button/.npmignore
```diff
@@ -0,0 +1,4 @@
+*.log
+target
+reports
+tests/**/__snapshots__/
```
emilyrohrbough: missing:
```
aggregate_translations
dev-site-config
screenshots
errorScreenshots
```

packages/terra-dropdown-button/docs/Button.md
```diff
@@ -0,0 +1,3 @@
+# Button
+
+The buttons to be shown in the dropdown. Must have a unique label.
```
emilyrohrbough: This doc seems to be missing info/context.

packages/terra-dropdown-button/docs/SplitButton.md
```diff
@@ -0,0 +1,15 @@
+# Split Button
+
```
emilyrohrbough: Is this two buttons or one? If I understood the design this is technically "one" button for user interactions so describing it as two is confusing

packages/terra-dropdown-button/docs/DropdownButton.md
```diff
@@ -0,0 +1,15 @@
+# Terra Dropdown Button
+
+The Terra Dropdown Button is a button that, when pressed, opens a dropdown with more buttons.
```
emilyrohrbough: Can we describe the use cases for this so we can differentiate from split button?

packages/terra-dropdown-button/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "terra-dropdown-button",
+  "main": "lib/DropdownButton.js",
+  "version": "0.0.0",
+  "description": "A button that, when pressed, opens a dropdown with more buttons",
```
emilyrohrbough: Lets update this to describe this gives access to a dropdown and split button components for selecting a single option (action) form a list of options.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -0,0 +1,133 @@
...
+        {...customProps}
+        buttons={children}
+        isOpen={isOpen}
+        requestClose={this.handleDropdownRequestClose}
+        disabled={disabled}
+        isBlock={isBlock}
+      >
+        <button
+          type="button"
+          className={cx('dropdown-button-type', { 'is-active': isOpen || active }, variant, { 'is-block': isBlock })}
```
emilyrohrbough: nit: can we move this to logic out of the rendering so it's easier to reach the boolean checking?

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -0,0 +1,133 @@
...
+  /**
+   * Determines whether the dropdown should be disabled.
+   */
+  disabled: PropTypes.bool,
+  /**
+   * Determines whether the component should have block styles applied. The dropdown will match the component's width.
+   */
+  isBlock: PropTypes.bool,
+  /**
+   * Sets the text that will be shown on the dropdown button.
```
emilyrohrbough: Should we also be using this for `aria-label` for screen reader accessibility?

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -0,0 +1,133 @@
...
+   */
+  label: PropTypes.string.isRequired,
+  /**
+   * Sets the styles of the component.
+   * Must be one of 'neutral', 'emphasis', or 'ghost'.
```
emilyrohrbough: Can we add `neutral` is default?

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -0,0 +1,133 @@
...
+
+  handleDropdownRequestClose() {
+    this.setState({ isOpen: false });
+  }
+
+  /*
+    In FireFox active styles don't get applied on space
+   */
+  handleKeyDown(event) {
+    if (event.keyCode === KeyCode.KEY_SPACE) {
```
emilyrohrbough: Do we need to toggle focus state for tabbing to and from this component?

emilyrohrbough: Also, do we need to handle `enter` for opening the dropdown as well?

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -0,0 +1,133 @@
...
+      this.setState({ active: true });
+    }
+  }
```
emilyrohrbough: Do we need to carry forward focus styles?

packages/terra-dropdown-button/src/DropdownButton.module.scss
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  .chevron-icon {
+    display: inline-block;
+    // If flex-shrink is set to 1 IE will give the button 0 width
+    flex: 0 0 auto;
+    height: var(--terra-dropdown-button-chevron-height, 14px);
+    margin: auto 0;
+    // position: relative;
+    // top: var(--terra-dropdown-button-chevron-top, 0.1429rem);
```
emilyrohrbough: clean up?

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * Sets the text that will be shown on the primary button which is outside the dropdown.
+   */
+  primaryOptionLabel: PropTypes.string.isRequired,
+  /**
+   * What will be called when the primary button is pressed.
+   */
+  onClick: PropTypes.func.isRequired,
+  /**
+   * Sets the styles of the component. Must be either 'neutral' or 'ghost'.
```
emilyrohrbough: can we add default variant is `neutral`

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -0,0 +1,162 @@
...
+      ...customProps
+    } = this.props;
+
+    const {
+      isOpen,
+      primaryActive,
+      chevronActive
+    } = this.state;
+
+    const finalVariant = variant !== 'emphasis' ? variant : 'neutral';
```
emilyrohrbough: `emphasis` isn't an option? can this be removed?

packages/terra-dropdown-button/src/DropdownButton.jsx
Blackop778: Focus styles seem to appear when tabbed to in Firefox and Chrome.

When pressing `enter` on the buttons containing chevrons `onClick` gets called immediately which opens the dropdown. Active styles are applied to the chevron button while the dropdown is open.

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -0,0 +1,162 @@
...
+
+    return (
+      <DropdownButtonBase
+        {...customProps}
```
emilyrohrbough: Do we want to allow customProp spreading?

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -0,0 +1,162 @@
...
+        buttons={children}
+        isOpen={isOpen}
+        requestClose={this.handleDropdownRequestClose}
+        disabled={disabled}
+        isBlock={isBlock}
+      >
+        <button
+          type="button"
+          className={cx('split-button-primary', { 'is-block': isBlock }, { 'is-active': primaryActive }, finalVariant)}
```
emilyrohrbough: can we break these out like:
```
const primaryButtonClassNames = cx([
  'split-button-primary',
  { 'is-block': isBlock },
  { 'is-active': primaryActive },
  finalVariant
]);
```

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -0,0 +1,162 @@
...
+          aria-disabled={disabled}
+        >
+          {primaryOptionLabel}
+        </button>
+        <button
+          type="button"
+          onClick={this.handleDropdownButtonClick}
+          onKeyDown={this.handleChevronKeyDown}
+          onKeyUp={this.handleChevronKeyUp}
+          className={cx('split-button-chevron', { 'is-active': isOpen || chevronActive }, finalVariant)}
```
emilyrohrbough: again, can we break out

packages/terra-dropdown-button/src/SplitButton.module.scss
```diff
@@ -0,0 +1,72 @@
...
+@import './variables';
+@import './mixins';
+
+:local {
+  .split-button-primary {
+    // stylelint-disable-next-line terra/custom-property-name
+    border-bottom-left-radius: var(--terra-dropdown-button-border-radius, 3px);
+    border-bottom-right-radius: 0;
+    border-style: solid;
+    // stylelint-disable-next-line terra/custom-property-name
```
emilyrohrbough: These probably shouldn't be a disabled stylelint rule

packages/terra-dropdown-button/src/SplitButton.module.scss
```diff
@@ -0,0 +1,72 @@
+@import '~terra-mixins/lib/Mixins';
+
+@import './variables';
+@import './mixins';
+
+:local {
+  .split-button-primary {
+    // stylelint-disable-next-line terra/custom-property-name
```
emilyrohrbough: These probably shouldn't be a disabled stylelint rule

packages/terra-dropdown-button/src/SplitButton.module.scss
```diff
@@ -0,0 +1,72 @@
...
+    }
+  }
+
+  .is-block.split-button-primary {
+    flex-grow: 1;
+  }
+
+  .split-button-chevron {
+    border-bottom-left-radius: 0;
+    // stylelint-disable-next-line terra/custom-property-name
```
emilyrohrbough: These probably shouldn't be a disabled stylelint rule

packages/terra-dropdown-button/src/_Button.jsx
```diff
@@ -0,0 +1,56 @@
...
+   */
+  requestClose: PropTypes.func,
+};
+
+const Option = ({
+  label, onClick, isFocused, isActive, requestClose, ...customProps
+}) => (
+  <li>
+    <div
+      {...customProps}
```
emilyrohrbough: Custom props probably shouldn't cascade this far since it's an internal component

emilyrohrbough:
Should the dropdown options open on key down?
When I tab a dropdown and then open the list of options, should I be able to tab to the next option?

emilyrohrbough:
Seeing a dotted line when focused in firefox when using keyboard nav:
Screen Shot 2019-06-28 at 3 57 50 PM

packages/terra-dropdown-button/tests/wdio/dropdown-button-spec.js
```diff
@@ -0,0 +1,159 @@
...
+      browser.moveToObject('#root', 100, 100);
+    });
+
+    Terra.it.validatesElement('Both closed');
+
+    it('opens the long label dropdown', () => {
+      browser.click('#wide-label');
+      Terra.validates.element('Long label open');
+    });
+
```
emilyrohrbough: lets add a check that it closes on 'escape'

packages/terra-dropdown-button/tests/wdio/dropdown-button-spec.js
```diff
@@ -0,0 +1,159 @@
...
+      browser.moveToObject('[class*=dropdown-button-type]');
+      Terra.validates.element('hover');
+    });
+
+    it('validates the open dropdown', () => {
+      browser.click('[class*=dropdown-button-type]');
+    });
+
+    Terra.it.validatesElement('dropdown open', { selector: '[class*=dropdown-list]' });
+    Terra.it.matchesScreenshot('dropdown button with dropdown open');
```
emilyrohrbough: This can be removed. validateElement takes a screenshot

packages/terra-dropdown-button/tests/wdio/dropdown-button-spec.js
```diff
@@ -0,0 +1,159 @@
...
+  });
+
+  describe('Disabled', () => {
+    before(() => {
+      browser.url('/#/raw/tests/terra-dropdown-button/dropdown-button/disabled-dropdown-button');
+    });
+
+    Terra.it.validatesElement();
```
emilyrohrbough: Because this is composed of multiple components, it'd probably be good to add a test to try to tab to the component and verify it did not receive focus

packages/terra-dropdown-button/tests/wdio/dropdown-button-spec.js
```diff
@@ -0,0 +1,159 @@
...
+  });
+
+  describe('Callback', () => {
+    before(() => {
+      browser.url('/#/raw/tests/terra-dropdown-button/dropdown-button/callback-dropdown-button');
+      // avoid hover styles
+      browser.moveToObject('#root', 100, 100);
+    });
+
+    Terra.it.matchesScreenshot('initial');
```
emilyrohrbough: Lets drop initial and let the default name be used

packages/terra-dropdown-button/tests/wdio/dropdown-button-spec.js
```diff
@@ -0,0 +1,159 @@
...
+      });
+
+      it('closes on tab without running a callback', () => {
+        browser.keys(['Tab']);
+        Terra.validates.screenshot('blue4');
+      });
+    });
+  });
+
+  describe('Width', () => {
```
emilyrohrbough: Can this be a little more descriptive?

packages/terra-dropdown-button/tests/wdio/dropdown-button-spec.js
```diff
@@ -0,0 +1,159 @@
...
+      });
+
+      it('jumps to the first entry', () => {
+        browser.keys(['ArrowDown', 'Home', 'Enter']);
+        Terra.validates.screenshot('red4');
+      });
+
+      it('jumps when typing', () => {
+        browser.keys(['b', 'Enter']);
+        Terra.validates.screenshot('blue3');
```
emilyrohrbough: can we makes these screenshot name more descriptive so they can be understood when reading the full screenshot name

packages/terra-dropdown-button/src/DropdownButton.jsx
Blackop778: What do you mean by that?

packages/terra-dropdown-button/src/DropdownButton.jsx
Blackop778: You're suggesting setting the `aria-label` on the `button`? Wouldn't the screen reader read the text inside the `button`? I suppose it needs to be tested.

packages/terra-dropdown-button/src/SplitButton.jsx
Blackop778: `emphasis` is only an option for the `DropdownButton`. The split button will still show blue `emphasis` styles if given the prop and I find prop types errors rather easy to miss but I'll remove the check.

packages/terra-dropdown-button/src/SplitButton.jsx
Blackop778: Probably? Most components do.

packages/terra-dropdown-button/src/_Button.jsx
Blackop778: Custom props don't cascade down to here, this would only be what the user of the component passes directly to the `Button`. Thinking about it, this really isn't an internal component and probably shouldn't have an underscore prefixed name.

packages/terra-dropdown-button/tests/wdio/dropdown-button-spec.js
Blackop778: The default selector does not capture the dropdown and the dropdown selector captures nothing else. This tests that the button has active styles while the dropdown is open.

Blackop778:
@emilyrohrbough Fixed the Firefox inner focus styles.

According to the UX design document the dropdown should open on enter or space and inside the dropdown the user should navigate with the up arrow, down arrow, home, end, and typing letters.

packages/terra-dropdown-button/docs/DropdownButton.md
bjankord: @jenniferwurth has provided some wording for this on the GitHub issue: https://github.com/cerner/terra-core/issues/2216#issuecomment-501327457 along with wording for the split button.

packages/terra-dropdown-button/docs/SplitButton.md
bjankord: @jenniferwurth has provided some wording for this on the GitHub issue: https://github.com/cerner/terra-core/issues/2216#issuecomment-501327457

packages/terra-dropdown-button/src/Button.module.scss
```diff
@@ -0,0 +1,31 @@
...
+    font-size: var(--terra-dropdown-button-menu-item-font-size, 0.9285714286rem);
+    line-height: var(--terra-dropdown-button-menu-item-line-height, 1.2);
+    outline: 0;
+    padding-bottom: var(--terra-dropdown-button-menu-item-padding-bottom, 4px);
+    padding-left: var(--terra-dropdown-button-menu-item-padding-left, 10px);
+    padding-right: var(--terra-dropdown-button-menu-item-padding-right, 10px);
+    padding-top: var(--terra-dropdown-button-menu-item-padding-top, 4px);
+
+    &.focus {
+      background: var(--terra-dropdown-button-menu-item-focus-background, #ebf6fd);
```
bjankord: Let's change this to background-color.
```suggestion
      background-color: var(--terra-dropdown-button-menu-item-focus-background-color, #ebf6fd);
```

packages/terra-dropdown-button/src/DropdownButton.jsx
bjankord: The screen readers should read the text inside of the button but this would be good to test to confirm.

packages/terra-dropdown-button/src/DropdownButton.module.scss
```diff
@@ -0,0 +1,60 @@
+@import '~terra-mixins/lib/Mixins';
+
+@import './variables';
+@import './mixins';
+
+:local {
+  .dropdown-button-type {
+    border-radius: var(--terra-dropdown-button-border-radius, 3px);
```
bjankord: Update for consistency with other variable names
```suggestion
    border-radius: var(--terra-dropdown-button-dropdown-type-border-radius, 3px);
```

packages/terra-dropdown-button/src/DropdownButton.module.scss
```diff
@@ -0,0 +1,60 @@
...
+
+  .emphasis {
+    @include terra-dropdown-button-variant-color($terra-dropdown-button-emphasis-color-scheme);
+  }
+
+  .chevron-icon {
+    display: inline-block;
+    // If flex-shrink is set to 1 IE will give the button 0 width
+    flex: 0 0 auto;
+    height: var(--terra-dropdown-button-chevron-height, 14px);
```
bjankord: nit: Let's update these to `1rem` for height and width.

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -0,0 +1,172 @@
...
+          onClick={this.handleDropdownButtonClick}
+          onKeyDown={this.handleChevronKeyDown}
+          onKeyUp={this.handleChevronKeyUp}
+          className={chevronClassnames}
+          disabled={disabled}
+          tabIndex={disabled ? '-1' : undefined}
+          aria-disabled={disabled}
+          aria-expanded={isOpen || undefined}
+          aria-haspopup
+          aria-label="More Options"
```
bjankord: We'll want to get translations for this.

packages/terra-dropdown-button/docs/DropdownButton.md
Blackop778: I'm not sure I like her wording for the DropdownButton, it seems to be in a more specific context than I'd say the documentation should be.

packages/terra-dropdown-button/src/SplitButton.module.scss
```diff
@@ -0,0 +1,78 @@
+@import '~terra-mixins/lib/Mixins';
+
+@import './variables';
+@import './mixins';
+
+:local {
+  .split-button-primary {
+    border-bottom-left-radius: var(--terra-dropdown-button-border-bottom-left-radius, 3px);
```
bjankord: We'll want to update the naming to be consistent with other variables under this selector.

```suggestion
    border-bottom-left-radius: var(--terra-dropdown-button-split-type-border-bottom-left-radius, 3px);
```

packages/terra-dropdown-button/src/SplitButton.module.scss
```diff
@@ -0,0 +1,78 @@
...
+    border-bottom-right-radius: 0;
+    border-style: solid;
+    border-top-left-radius: var(--terra-dropdown-button-border-top-left-radius, 3px);
```
bjankord: We'll want to update the naming to be consistent with other variables under this selector.

```suggestion
    border-top-left-radius: var(--terra-dropdown-button-split-type-border-top-left-radius, 3px);
```

packages/terra-dropdown-button/src/SplitButton.module.scss
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  .is-block.split-button-primary {
+    flex-grow: 1;
+  }
+
+  .split-button-chevron {
+    border-bottom-left-radius: 0;
+    border-bottom-right-radius: var(--terra-dropdown-button-border-bottom-right-radius, 3px);
```
bjankord:
```suggestion
    border-bottom-right-radius: var(--terra-dropdown-button-split-type-chevron-border-bottom-right-radius, 3px);
```

packages/terra-dropdown-button/src/SplitButton.module.scss
```diff
@@ -0,0 +1,78 @@
...
+    border-style: solid;
+    border-top-left-radius: 0;
+    border-top-right-radius: var(--terra-dropdown-button-border-top-right-radius, 3px);
```
bjankord:
```suggestion
    border-top-right-radius: var(--terra-dropdown-button-split-type-chevron-border-top-right-radius, 3px);
```

packages/terra-dropdown-button/src/SplitButton.module.scss
```diff
@@ -0,0 +1,78 @@
...
+    // If flex-shrink is set to 1 IE will give the button 0 width
+    flex: 0 0 auto;
+    font-size: var(--terra-dropdown-button-split-type-font-size, 1rem);
```
bjankord:
```suggestion
    font-size: var(--terra-dropdown-button-split-type-chevron-font-size, 1rem);
```

packages/terra-dropdown-button/src/SplitButton.module.scss
```diff
@@ -0,0 +1,78 @@
...
+    line-height: var(--terra-dropdown-button-split-type-line-height, 1.4285714286);
```
bjankord:
```suggestion
    line-height: var(--terra-dropdown-button-split-type-chevron-line-height, 1.4285714286);
```

packages/terra-dropdown-button/src/SplitButton.module.scss
```diff
@@ -0,0 +1,78 @@
...
+
+  .emphasis {
+    @include terra-dropdown-button-variant-color($terra-dropdown-button-emphasis-color-scheme);
+  }
+
+  .chevron-icon {
+    display: inline-block;
+    // If flex-shrink is set to 1 IE will give the button 0 width
+    flex: 0 0 auto;
+    height: var(--terra-dropdown-button-chevron-height, 14px);
```
bjankord:
```suggestion
    height: var(--terra-dropdown-button-split-type-chevron-height, 14px);
```

packages/terra-dropdown-button/src/SplitButton.module.scss
```diff
@@ -0,0 +1,78 @@
...
+    margin: auto 0;
+    position: relative;
+    top: var(--terra-dropdown-button-chevron-top, 0.1429rem);
```
bjankord:
```suggestion
    top: var(--terra-dropdown-button-split-type-chevron-top, 0.1429rem);
```

packages/terra-dropdown-button/src/SplitButton.module.scss
```diff
@@ -0,0 +1,78 @@
...
+    width: var(--terra-dropdown-button-chevron-width, 14px);
```
bjankord:
```suggestion
    width: var(--terra-dropdown-button-split-type-chevron-width, 1rem);
```

packages/terra-dropdown-button/src/_DropdownList.module.scss
```diff
@@ -0,0 +1,14 @@
+:local {
+  .dropdown-list {
+    background: var(--terra-dropdown-button-menu-background, #fff);
```
bjankord: Let's update the variable names like so
```suggestion
    background: var(--terra-dropdown-list-background, #fff);
```

packages/terra-dropdown-button/src/terra-dev-site/doc/DropdownButton/Button.3.doc.jsx
```diff
@@ -0,0 +1,34 @@
...
+import { name } from '../../../../package.json';
+
+// Component Source
+import ButtonSrc from '!raw-loader!../../../../src/Button';
+
+// Example Files
+import ButtonExamples from '../example/Button';
+import ButtonExampleSrc from '!raw-loader!../../../../src/terra-dev-site/doc/example/Button.jsx';
+
+const DocPage = () => (
```
bjankord: Can we drop this page and pull in the button doc to both the dropdown doc page and the split button doc page.

packages/terra-dropdown-button/src/terra-dev-site/doc/example/Button.module.scss
```diff
@@ -0,0 +1,10 @@
+:local {
+  .list {
```
bjankord: Is this used in the examples?

packages/terra-dropdown-button/src/terra-dev-site/doc/example/DefaultDropdownButton.jsx
```diff
@@ -0,0 +1,16 @@
...
+import React from 'react';
+
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import DropdownButton, { Button } from 'terra-dropdown-button/lib/DropdownButton';
+
+export default () => (
+  <DropdownButton
+    label="Dropdown"
+  >
+    <Button label="1st Option" onClick={() => { console.log('hi'); }} />
```
bjankord: Let's update this example to write content to the page instead of a console.log.

packages/terra-dropdown-button/src/terra-dev-site/doc/example/DefaultSplitButton.jsx
```diff
@@ -0,0 +1,17 @@
...
+
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import { SplitButton, Button } from 'terra-dropdown-button/lib/DropdownButton';
+
+export default () => (
+  <SplitButton
+    primaryOptionLabel="Primary option"
+    onClick={() => { console.log('primary option'); }}
+  >
+    <Button label="Abacadabra" onClick={() => { console.log('hi'); }} />
```
bjankord: Let's update this example to write content to the page instead of a console.log.

packages/terra-dropdown-button/src/terra-dev-site/doc/example/DisabledDropdownButton.jsx
```diff
@@ -0,0 +1,17 @@
...
+
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import DropdownButton, { Button } from 'terra-dropdown-button/lib/DropdownButton';
+
+export default () => (
+  <DropdownButton
+    label="Dropdown"
+    disabled
+  >
+    <Button label="1st Option" onClick={() => { console.log('hi'); }} />
```
bjankord: Let's update this example to write content to the page instead of a console.log.

packages/terra-dropdown-button/src/terra-dev-site/doc/example/DisabledSplitButton.jsx
```diff
@@ -0,0 +1,18 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import { SplitButton, Button } from 'terra-dropdown-button/lib/DropdownButton';
+
+export default () => (
+  <SplitButton
+    primaryOptionLabel="Primary option"
+    onClick={() => { console.log('primary option'); }}
+    disabled
+  >
+    <Button label="1st Option" onClick={() => { console.log('hi'); }} />
```
bjankord: Let's update this example to write content to the page instead of a console.log.

packages/terra-dropdown-button/src/terra-dev-site/doc/example/EmphasisDropdownButton.jsx
```diff
@@ -0,0 +1,17 @@
...
+
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import DropdownButton, { Button } from 'terra-dropdown-button/lib/DropdownButton';
+
+export default () => (
+  <DropdownButton
+    label="Dropdown"
+    variant="emphasis"
+  >
+    <Button label="1st Option" onClick={() => { console.log('hi'); }} />
```
bjankord: Let's update this example to write content to the page instead of a console.log.

packages/terra-dropdown-button/src/terra-dev-site/test/dropdown-button/BlockDropdownButton.test.jsx
```diff
@@ -0,0 +1,18 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import DropdownButton, { Button } from '../../../DropdownButton';
+
+export default () => (
+  <DropdownButton
+    label="Primary option"
+    onClick={() => { console.log('primary option'); }}
+    isBlock
+  >
+    <Button label="1st Option" onClick={() => { console.log('hi'); }} />
```
bjankord: Let's update this example to write content to the page instead of a console.log.

packages/terra-dropdown-button/src/terra-dev-site/test/dropdown-button/BlockSplitButton.test.jsx
```diff
@@ -0,0 +1,18 @@
...
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import SplitButton, { Button } from '../../../SplitButton';
+
+export default () => (
+  <SplitButton
+    primaryOptionLabel="Primary option"
+    onClick={() => { console.log('primary option'); }}
+    isBlock
+  >
+    <Button label="1st Option" onClick={() => { console.log('hi'); }} />
```
bjankord: Let's update this example to write content to the page instead of a console.log.

packages/terra-dropdown-button/src/SplitButton.jsx
Blackop778: Submitted a translations request

packages/terra-dropdown-button/src/_DropdownList.module.scss
Blackop778: Stylelint gets upset when the `button-` is removed because then the variable doesn't begin with `--terra-dropdown-button`

packages/terra-dropdown-button/src/terra-dev-site/test/dropdown-button/BlockDropdownButton.test.jsx
Blackop778: I don't know if the test pages (aside from the callback testing page) need to have visible callbacks like this.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -0,0 +1,140 @@
...
+
+const Variants = {
+  NEUTRAL: 'neutral',
+  EMPHASIS: 'emphasis',
+  GHOST: 'ghost',
+};
+
+const propTypes = {
+  /**
+   * The options to display in the dropdown. Should be of type `Button`.
```
nramamurth: This could cause confusion as to whether this is `terra-button` or the subcomponent inside this package.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -0,0 +1,140 @@
...
+      ...customProps
+    } = this.props;
+
+    const {
+      isOpen,
+      active,
+    } = this.state;
+
+    const classnames = cx(
+      'dropdown-button-type',
```
nramamurth: Could we avoid using 'type' as the suffix here? What does it convey?

Blackop778: The type suffix is left over from when `DropdownButtonBase`, `SplitButton`, and `DropdownButton` were all in the same file. The very outermost div (which became `DropdownButtonBase`) had the class `dropdown-button` as the class that encapsulates the component. Dropdown's classname had the `-type` suffix to differentiate itself from that.
It can be removed now though.

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -0,0 +1,191 @@
...
+  intl: intlShape.isRequired,
+};
+
+const defaultProps = {
+  disabled: false,
+  isBlock: false,
+  variant: 'neutral',
+};
+
+const contextTypes = {
```
nramamurth: Do we still need this if you are using `injectIntl` HOC?

packages/terra-dropdown-button/CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+ChangeLog
+=========
+
+Unreleased
+-----------------
+* Initial stable release
```
StephenEsser: ( Commenting here so the conversation can be threaded. Looking at you Github Engineers )

A couple questions regarding the functionality of the UI. 

Clicking the primary Button opens the Dropdown. But clicking the primary Button again does not close it (persistently). Should it? Or does it behave as expected, where clicking while it's open causes it to close and reopen?

image

The next question is whether we should update the cursor for the dropdown button text. I am unable to get a picture, but if you hover the text in the dropdown it will display as the default selection cursor. 
image

packages/terra-dropdown-button/src/Button.jsx
```diff
@@ -0,0 +1,56 @@
...
+  requestClose: PropTypes.func,
+};
+
+const Option = ({
+  label, onClick, isFocused, isActive, requestClose, ...customProps
+}) => (
+  <li>
+    <div
+      {...customProps}
+      onMouseUp={() => { onClick(); requestClose(); }}
```
StephenEsser: This is a little tricky, using the default example as a reference. 

If you open the dropdown then click and hold option 1, drag the mouse down to option 4 ( or any other option ) and then release the mouse, that option will fire a selection. This might lead to some unexpected user actions, it's not the option that you clicked that fires, but whichever one you happen to mouse up on. 

If you were trying to mimic some of the functionality of the Select, the Select records which button was pressed on mouse down and then compares that button against the button that receives the mouse up event. If they don't match it doesn't make a selection.

https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/_Menu.jsx#L446

packages/terra-dropdown-button/src/DropdownButton.jsx
StephenEsser: Do we need to add this twice? Does it already display within the generated props table?

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -0,0 +1,140 @@
...
+    this.handleDropdownButtonClick = this.handleDropdownButtonClick.bind(this);
+    this.handleDropdownRequestClose = this.handleDropdownRequestClose.bind(this);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.handleKeyUp = this.handleKeyUp.bind(this);
+
+    this.state = { isOpen: false, active: false };
+  }
+
+  handleDropdownButtonClick() {
+    this.setState({ isOpen: true });
```
StephenEsser: This is probably a question for the accessibility team, but should the down arrow also open the dropdown when the button is focused?

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -0,0 +1,181 @@
...
+  /**
+   * Sets the text that will be shown on the primary button which is outside the dropdown.
+   */
+  primaryOptionLabel: PropTypes.string.isRequired,
+  /**
+   * What will be called when the primary button is pressed.
+   */
+  onClick: PropTypes.func.isRequired,
+  /**
+   * Sets the styles of the component.  The default is 'neutral'. Must be either 'neutral' or 'ghost'.
```
nramamurth: I am a little confused here. The documentation says that a `SplitButton` could be of `ghost` variant but the [CSS is for `emphasis`](https://github.com/cerner/terra-core/blob/cb4c964627bac4eb94077e0647bd9a998708c394/packages/terra-dropdown-button/src/SplitButton.module.scss#L66-L68)? Also, it'd be nice to have an example on the site.

packages/terra-dropdown-button/src/_Dropdown.jsx
```diff
@@ -0,0 +1,62 @@
...
+  isOpen: PropTypes.bool.isRequired,
+  /**
+   * The Options that should be displayed in the dropdown
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * @private
+   * Width to set the dropdown to. Used when `isBlock` is true. Unset means to autosize.
+   */
+  width: PropTypes.string,
```
nramamurth: This could be a question for the UX. The UX callout shows that the width of the dropdown extends up until the width of the button but the example on the site does not. Is the callout showing the default Dropdown or is it showing how it is when `isBlock` is `true`?

Callout: 
Screen Shot 2019-07-03 at 10 31 09 AM


Site Example: 
Screen Shot 2019-07-03 at 10 41 33 AM

packages/terra-dropdown-button/src/_Dropdown.jsx
```diff
@@ -0,0 +1,62 @@
...
+};
+
+const Dropdown = ({
+  requestClose, isOpen, targetRef, children, width,
+}) => (
+  <Hookshot
+    isOpen={isOpen}
+    isEnabled
+    targetRef={targetRef}
+    attachmentBehavior="flip"
```
nramamurth: nit: could we possibly make use of the enum [here](https://github.com/cerner/terra-framework/blob/master/packages/terra-hookshot/src/Hookshot.jsx#L19) for this?

packages/terra-dropdown-button/src/_DropdownListUtil.js
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
```
nramamurth: nit: There are a few references to `menu` in the comment documentation in this file. They could be cleaned up.

packages/terra-dropdown-button/src/terra-dev-site/doc/example/BlockDropdownButton.jsx
```diff
@@ -0,0 +1,25 @@
...
+const Example = () => {
+  const [message, setMessage] = useState('No option clicked');
+
+  return (
+    <React.Fragment>
+      <DropdownButton
+        label="Dropdown"
+        isBlock
+      >
+        <Button label="1st Option" onClick={() => setMessage('1st option clicked')} />
```
nramamurth: I think it'd be better to use real-world examples (there are a couple pointed out in the [original issue](https://github.com/cerner/terra-core/issues/2216#issuecomment-501327457)) here so that it probably helps consumers better understand the difference between Dropdown Button v/s SplitButton.

packages/terra-dropdown-button/src/SplitButton.jsx
Blackop778: I'm awaiting UX callouts for the ghost Dropdown Button

packages/terra-dropdown-button/CHANGELOG.md
Blackop778: It probably shouldn't reopen like that, looks to be a side effect of changing from `Popup` to `Hookshot`.
The mouse cursor should probably also change.

packages/terra-dropdown-button/src/_Dropdown.jsx
Blackop778: I think that is just a bad example, from the UX callouts:

> The menu is not tied to the size of the button. It may be larger or smaller than the button, except when using Block Style Button.

packages/terra-dropdown-button/src/_Dropdown.jsx
Blackop778: Because those objects are arrays the best I could do is `{Hookshot.attachmentBehaviors[1]}` which is certainly worse than `"flip"`

nramamurth: Oh no, my bad.

packages/terra-dropdown-button/CHANGELOG.md
Blackop778: Addressed in bd46997b4f8140247c8d530b18646f706b1033e5 and f3b73577383099a9bddb5383f6a820868611af97

packages/terra-dropdown-button/src/Button.jsx
Blackop778: Not sure why I was using `onMouseUp`, addressed in f3b73577383099a9bddb5383f6a820868611af97

packages/terra-dropdown-button/src/terra-dev-site/doc/example/BlockDropdownButton.jsx
Blackop778: Done in f3b73577383099a9bddb5383f6a820868611af97

packages/terra-dropdown-button/src/_DropdownListUtil.js
Blackop778: Done in f3b73577383099a9bddb5383f6a820868611af97

packages/terra-dropdown-button/src/_DropdownButtonBase.jsx
```diff
@@ -72,6 +72,7 @@ class DropdownButtonBase extends React.Component {
       { 'is-block': isBlock },
       { disabled },
       customProps.className,
+      'ignore-react-onclickoutside',
```
emilyrohrbough: Did you try directly passing the `disableOnClickOutside ` prop to the Hook.Content component? 
Ref: https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/_Dropdown.jsx#L79

Blackop778: The problem with that is I still want to get notified for clicks outside the `DropdownButton` component.
https://github.com/cerner/terra-core/blob/8c1821643741909627549952b976d1c5f534a9ea/packages/terra-dropdown-button/src/_Dropdown.jsx#L44
`Select` uses `onBlur` to account for this which works if you don't want to know if what got clicked on to move focus was a `button`.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -0,0 +1,140 @@
...
+          onKeyDown={this.handleKeyDown}
+          onKeyUp={this.handleKeyUp}
+          disabled={disabled}
+          tabIndex={disabled ? '-1' : undefined}
+          aria-disabled={disabled}
+          aria-expanded={isOpen || undefined}
+          aria-haspopup
+        >
+          <span className={cx('dropdown-button-text')}>{label}</span>
+          <span className={cx('chevron-icon')} />
```
emilyrohrbough: Looking at the call outs, it looks like this should be using the caret down icon, not the chevron

emilyrohrbough:
I'm noticing a gap between the button and dropdown. Is this intended? 
Screen Shot 2019-07-09 at 10 14 03 AM

emilyrohrbough:
I'm clicking through the doc examples and test examples. When I click on a dropdown option, the selected dropdown option is not updated when I re-open the dropdown.

Blackop778:
@emilyrohrbough what do you mean "is not updated when I re-open the dropdown"? Is it retaining focus or active styles? The callback not getting called?

emilyrohrbough:
I'm not sure other's opinions on this, but I would like to rename the Button component to something else. When I see this, I am expecting to see terra-button, not a dropdown option. It could be we use like:
```
<DropdownButton>
    <DropdownButton.Button>
</DropdownButton>
```

packages/terra-dropdown-button/docs/DropdownButton.md
```diff
@@ -0,0 +1,18 @@
...
+It is intended to allow the user to choose from a list of options with no special emphasis on any option.
+
+Children must be `Button` subcomponent for proper functionality and appearance.
```
StephenEsser: This seems a little ambiguous between the terra-button and the DropdownButton's Button.

packages/terra-dropdown-button/src/Button.jsx
```diff
@@ -0,0 +1,58 @@
...
+
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+// Keyboard events are handled in _DropdownButton.jsx
+const Option = ({
+  label, onClick, isFocused, isActive, requestClose, ...customProps
+}) => (
+  <li>
+    <div
+      {...customProps}
+      onClick={() => { onClick(); requestClose(); }}
```
StephenEsser: This order of these functions might have the potential to cause issues if the onClick causes a focus change. Assuming these events fire synchronously if the action of the Button changes focus, what happens at the completion of requestClose? When the popup closes, does it attempt to return focus to the button?

StephenEsser: There's something strange about firing two functions in an anonymous block, but if the focus doesn't turn out to be an issue this should be fine.

packages/terra-dropdown-button/src/Button.jsx
```diff
@@ -0,0 +1,58 @@
+import React from 'react';
```
tbiethman: Should this file be Option.jsx? Or should the class here be renamed to Button?

packages/terra-dropdown-button/src/Button.jsx
StephenEsser: Although, the onClick is being called without any parameters. This might be worth talking about in scrum.

packages/terra-dropdown-button/src/Button.jsx
Blackop778: It used to be `Option.jsx` but it was decided it should be renamed from `Option` to `Button`. Guess the function name didn't get changed.

packages/terra-dropdown-button/src/Button.jsx
Blackop778: A minor gotcha about passing the `event` parameter would be that sometimes the event is a `onClick` and sometimes a `onKeyDown`. The `event` parameter hadn't been excluded for any real reason, I just hadn't thought to include that.

StephenEsser: I'm second guessing the argument I made for calling this onClick, but I can't think of a name that feels more appropriate. I still think callback is pretty generic, but because this fires on key down too I'm a little unsure. How does Button handle this? Does it fire on click onKeyDown (enter / space)?

emilyrohrbough:
Should there be icon prop options available for both the primary button and dropdown button options?

packages/terra-dropdown-button/src/Button.jsx
Blackop778: terra-button allows the consumer to pass a `onClick` and/or `onKeyDown` listener. Being a HTML button (most of the time) `onClick` would get called when enter or space are pressed.

Another thing about the event is that true focus is actually on the list and not on the item itself, so `target` and possible others would have to be changed before the event is passed on.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -0,0 +1,140 @@
...
+
+const propTypes = {
+  /**
+   * The options to display in the dropdown. Should be comprised of the subcomponent `Button`.
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * Determines whether the dropdown should be disabled.
+   */
+  disabled: PropTypes.bool,
```
tbiethman: This should be `isDisabled` to match the other boolean props throughout. You'll just have to map isDisabled->disabled for the button element.

Blackop778: Funny you mention that, it was originally like that. Stephen suggested to have it changed the other way https://github.com/cerner/terra-core/issues/2216#issuecomment-502746195

tbiethman: @StephenEsser
> I think isDisabled should be disabled. For html attributes we generally match naming conventions unless there is a chance they can conflict. But in this context they share the same meaning.

Terra's standalone Button component uses `isDisabled`: https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L50

But the various Form components use `disabled`: https://github.com/cerner/terra-core/blob/master/packages/terra-form-textarea/src/Textarea.jsx#L39

I personally find the difference between the two naming schemes jarring, but I'm really just looking for some consistency in this pattern. Are we expecting to update <Button /> to remove `isDisabled` at some point?

StephenEsser: I'll try to find the previous thread, it might be on the locked war-room notes. We had discussed in early terra development that we would match boolean properties that map directly to the html property. 

For example:
- disabled
- required
- readonly

cc: @bjankord - Do you remember this discussion / have opinions on this?

packages/terra-dropdown-button/src/Button.jsx
tbiethman: If we boil it down, this component is just a collection of actions that a user can select. All this component needs to do is render the actions and notify the consumer that an action was selected. The manner of that selection shouldn't factor into the API.

Can we change the prop from 'onClick' to 'onSelect' and universally execute the callback after the dropdown closes (due to keyup, clicks, etc.)? It'd be migrating away from a directly semantic `<button>` API, but given the complexities of this component, I feel like that is appropriate.

packages/terra-dropdown-button/src/DropdownButton.jsx
StephenEsser: I have returned from searching through the email graveyard. 

First I stumbled upon this:
https://github.com/cerner/terra-core/pull/305#discussion_r115284789

Which lead me to this: 

https://engineering.cerner.com/terra-ui/#/contributing/terra-ui/conventions

Which contains this:

image

tbiethman: Stephen bringing the receipts. I guess the sacred texts have spoken. 

(It's still weird)

packages/terra-dropdown-button/src/DropdownButton.jsx
Blackop778: Fixed in aca4376fcf55905535d77b2168118fd85dfb4a15

packages/terra-dropdown-button/src/Button.jsx
Blackop778: Fixed in fa6e738959360e067e0c7ca807365bedf22725a8

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -0,0 +1,140 @@
...
+class DropdownButton extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.handleDropdownButtonClick = this.handleDropdownButtonClick.bind(this);
+    this.handleDropdownRequestClose = this.handleDropdownRequestClose.bind(this);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.handleKeyUp = this.handleKeyUp.bind(this);
+
+    this.state = { isOpen: false, active: false };
```
yuderekyu: @cerner/terra Do we have a stance on when to use hooks versus class components?

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -0,0 +1,140 @@
...
+      children,
+      disabled,
+      isBlock,
+      label,
+      variant,
+      ...customProps
+    } = this.props;
+
+    const {
+      isOpen,
```
yuderekyu: nit, want to make this destructuring one line haha

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -0,0 +1,181 @@
...
+    super(props);
+
+    this.handleDropdownButtonClick = this.handleDropdownButtonClick.bind(this);
+    this.handleDropdownRequestClose = this.handleDropdownRequestClose.bind(this);
+    this.handlePrimaryKeyDown = this.handlePrimaryKeyDown.bind(this);
+    this.handlePrimaryKeyUp = this.handlePrimaryKeyUp.bind(this);
+    this.handleCaretKeyDown = this.handleCaretKeyDown.bind(this);
+    this.handleCaretKeyUp = this.handleCaretKeyUp.bind(this);
+
+    this.state = { isOpen: false, caretActive: false, primaryActive: false };
```
yuderekyu: nit, for clarity, can we add 'is' to these state attributes
```suggestion
    this.state = { isOpen: false, caretIsActive: false, primaryIsActive: false };
```

packages/terra-dropdown-button/src/_Dropdown.jsx
```diff
@@ -0,0 +1,62 @@
...
+    contentAttachment={{ vertical: 'top', horizontal: 'start' }}
+    targetAttachment={{ vertical: 'bottom', horizontal: 'start' }}
+  >
+    <Hookshot.Content
+      onEsc={requestClose}
+      onOutsideClick={requestClose}
+    >
+      <FocusTrap focusTrapOptions={{ returnFocusOnDeactivate: true, clickOutsideDeactivates: true }}>
+        <div>
```
yuderekyu: is this div necessary?

packages/terra-dropdown-button/src/Button.jsx
Blackop778: > I still think callback is pretty generic, but because this fires on key down too I'm a little unsure. How does Button (terra-button) handle this? Does it fire onClick for onKeyDown (enter / space)?

terra-button calls onClick when enter or space is pressed, but that is just a feature of how native buttons work. The items inside the dropdown are only fancy `div`s as currently implemented.

StephenEsser: I am in favor of using `onSelect` to alleviate the confusion between the actions and abstract away the implementation details of how the action is invoked. I think we should avoid multiple callback functions if possible.

That being said, we should consider the implications of how this may affect future components with a similar behavior and navigate towards some level of consistency. The Menu component for example uses `onClick` to notify the user a selection has been made, regardless of if the selection was made from keyboard interaction.

https://engineering.cerner.com/terra-ui/#/components/terra-menu/menu/menu

packages/terra-dropdown-button/src/Button.module.scss
```diff
@@ -0,0 +1,32 @@
...
+      &.active {
+        background-color: var(--terra-dropdown-button-menu-item-focus-active-background-color, #ebf6fd);
+        box-shadow: var(--terra-dropdown-button-menu-item-focus-active-box-shadow, none);
+        color: var(--terra-dropdown-button-menu-item-focus-active-color, #1c1f21);
+      }
+    }
+
+    &:hover {
+      background: var(--terra-dropdown-button-menu-item-hover-background, #ebf6fd);
+      outline: 0;
```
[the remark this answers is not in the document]
Blackop778: That does not seem to be the case, my best guess is that it got copied over from the focus style.

packages/terra-dropdown-button/src/Button.jsx
```diff
@@ -0,0 +1,52 @@
...
+      onClick={() => { onClick(); requestClose(); }}
+      role="button"
+      tabIndex="0"
+      className={cx([
+        'button',
+        { active: isActive },
+      ])}
+    >
+      {label}
+    </div>
```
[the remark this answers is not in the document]
Blackop778: The `jsx-a11y` eslint plugin complains about a non-interactive element having an interactive role, a mouse event listener and tab index if this change is made. Also prevents the accessibility API from seeing it as a list item containing a button.

packages/terra-dropdown-button/src/_Dropdown.jsx
Blackop778: It used to be when focus was on the `DropdownList` but doesn't seem to be needed anymore.

packages/terra-dropdown-button/src/Button.jsx
Blackop778: > The order of these functions might have the potential to cause issues if the onClick causes a focus change. Assuming these events fire synchronously if the action of the Button changes focus, what happens at the completion of requestClose? When the popup closes, does it attempt to return focus to the button?

I've confirmed that (regardless of order) if `onClick` tries to focus something that thing will get focused and then blurred as the dropdown unmounts. Even if I try to change the focus trap's props when the item gets clicked it unmounts before the props update. Fixing this would require changes in `focus-trap-react`.

Blackop778: Focus fixed in b12e0ce243a3101ed65658c1c4314cf10002fa7d, renaming to `onSelect` soon

packages/terra-dropdown-button/src/Item.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import styles from './Item.module.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
```
mjhenkes: @neilpfeiffer, should the drop down items have a 'selected' option that adds a checkmark for toggled items? I only ask because thats what we ended up doing in terra-dev-site for the toolbar buttons.

packages/terra-dropdown-button/src/_DropdownButtonBase.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+const propTypes = {
+  /**
+   * The button or buttons that are the primary option and caret
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * The options to display in the dropdown. Should be of type `Button`.
+   */
+  buttons: PropTypes.node.isRequired,
```
mjhenkes: Change to items to match the file name changes.

packages/terra-dropdown-button/src/_DropdownList.jsx
```diff
@@ -0,0 +1,169 @@
+import React from 'react';
```
mjhenkes: I assume we investigated using terra-menu but couldn't?

Blackop778: I hadn't actually, although it looks to be different in keyboard interaction (menu uses tab instead of up/down arrow), some style variances, it appears to prefer to open left instead of right, and uses `Popup` which we moved away from to avoid issues with mobile screens due to not being able to scroll away from the dropdown.

packages/terra-dropdown-button/src/DropdownButton.jsx
bjankord: We talked some in are meeting on Friday. Both are available to use. If you find that you are needing nearly all the lifecycle methods and features provided with class components, it may be best to stick with a class. For new stuff, the react team recommends using hooks.

packages/terra-dropdown-button/src/_DropdownButtonBase.jsx
Blackop778: Fixed in 7b2111474afa8a329a9bc4035196cbf6f1264d25

mmalaker:
Functional verification underway. Everything looks good except for JAWS on desktop Chrome. When the dropdown menus are open, the focus indication doesn't always display. JAWS will read the dropdown items correctly, but the focus styling is missing. Another issue is that when using the arrow keys to move focus around the page, when JAWS reaches the dropdown, focus is trapped there. It seems that the page scrolling feature of the browser takes over on arrow up/down. You can tab out of the control, however, then continue with JAWS navigation. Still, this doesn't seem like the best user experience.

Blackop778:
@mmalaker those issues should be resolved now. Unfortunately VoiceOver in Safari does not announce how many items are in the menu or the index of the item the user is on ("Item 3 of 4") anymore.

mmalaker:
Functional verification completed. The dropdown button examples now display and function as expected on all supported platforms. The JAWS focus trap issue appears to have been unique to a single testing device instance.

Blackop778:
As a status update I believe the PR is good to go code wise but needs some style updates and review from UX.

neilpfeiffer:
Tested: PR #2487 
Resolves: Issue #2216
 
**Reviewed:**
- Follows pattern considerations documented in the current Cerner Design Standards: [Dropdown-Buttons & Split-Buttons](http://url)
- Follows UX defined design implementation for Terra-Dropdown-Button
- Follows defined theme designs: terra-opensource-theme
 
 
**Considerations for Future Enhancements:**
- update applied Focus state styling to detect for mouse/keyboard interaction
- update Focus state to new Terra outline style (black dashed line)
- add "Emphasis" as an available variant for Split-Button
 
\
\
Marking as Verified by @neilpfeiffer, ready for release.
 
**Pass** |
-----------|
UX Review|

packages/terra-dropdown-button/src/Item.jsx
neilpfeiffer: @mjhenkes No - the list of items in the Dropdown and Splitbuttons are straight buttons, not toggles, so "selecting" or "pressing" them immediately would do an action. There is no persistent "selected" state for these buttons or the items in the list. That would be a menu or other component. This is "pick and execute".

packages/terra-dropdown-button/src/DropdownButton.module.scss
```diff
@@ -0,0 +1,74 @@
...
+
+:local {
+  .dropdown-button {
+    border-radius: inherit;
+    border-style: solid;
+    border-width: 1px;
+    cursor: pointer;
+    display: inline-flex;
+    flex: 1 1 auto; // Needed for IE 10 to linebreak things properly
+    //flex-direction: row;
```
bjankord: Let's remove this commented out code.

packages/terra-dropdown-button/src/_variables.scss
```diff
@@ -0,0 +1,92 @@
...
+  active-color: var(--terra-dropdown-button-active-color-emphasis, #fff),
+  active-dropdown-font-weight: var(--terra-dropdown-button-dropdown-type-active-font-weight-emphasis, 400),
+  //active-split-font-weight: var(--terra-dropdown-button-split-type-active-font-weight-emphasis, 400),
+  //active-split-caret-box-shadow: var(--terra-dropdown-button-split-type-caret-active-box-shadow-emphasis, none),
+  background-color: var(--terra-dropdown-button-background-color-emphasis, #0079be),
+  border-color: var(--terra-dropdown-button-border-color-emphasis, #0065a3),
+  color: var(--terra-dropdown-button-color-emphasis, #fff),
+  dropdown-box-shadow: var(--terra-dropdown-button-dropdown-type-box-shadow-emphasis, none),
+  dropdown-font-weight: var(--terra-dropdown-button-dropdown-type-font-weight-emphasis, 400),
+  //split-box-shadow: var(--terra-dropdown-button-split-type-box-shadow-emphasis, none),
```
bjankord: Let's remove this commented out code.

packages/terra-dropdown-button/src/_variables.scss
```diff
@@ -0,0 +1,92 @@
...
+  //split-font-weight: var(--terra-dropdown-button-split-type-font-weight-emphasis, 400),
+  //split-caret-box-shadow: var(--terra-dropdown-button-split-type-caret-box-shadow-emphasis, inset 1px 0 0 0 rgba(255, 255, 255, 0.2)),
+  disabled-color: var(--terra-dropdown-button-disabled-color-emphasis, #fff),
+  focus-background-color: var(--terra-dropdown-button-focus-background-color-emphasis, #0079be),
+  focus-border-color: var(--terra-dropdown-button-focus-border-color-emphasis, #0065a3),
+  focus-dropdown-box-shadow: var(--terra-dropdown-button-dropdown-type-focus-box-shadow-emphasis, 0 0 0 1px rgba(76, 178, 233, 0.7), 0 0 0 3px rgba(76, 178, 233, 0.35)),
+  //focus-split-box-shadow: var(--terra-dropdown-button-split-type-focus-box-shadow-emphasis, 0 0 0 1px rgba(76, 178, 233, 0.7), 0 0 0 3px rgba(76, 178, 233, 0.35)),
```
bjankord: Let's remove this commented out code.

packages/terra-dropdown-button/src/_variables.scss
```diff
@@ -0,0 +1,92 @@
...
+  //focus-split-caret-box-shadow: var(--terra-dropdown-button-split-type-caret-focus-box-shadow-emphasis, 0 0 0 1px rgba(76, 178, 233, 0.7), 0 0 0 3px rgba(76, 178, 233, 0.35)),
+  hover-background-color: var(--terra-dropdown-button-hover-background-color-emphasis, #0065a3),
+  hover-color: var(--terra-dropdown-button-hover-color-emphasis, #fff),
+  //hover-split-caret-box-shadow: var(--terra-dropdown-button-split-type-caret-hover-box-shadow-emphasis, none),
```
bjankord: Let's remove this commented out code.

dkasper-was-taken: Updated.

packages/terra-dropdown-button/src/_variables.scss
dkasper-was-taken: Updated.

packages/terra-dropdown-button/src/_variables.scss
dkasper-was-taken: Updated.

packages/terra-dropdown-button/src/DropdownButton.module.scss
dkasper-was-taken: Updated.

neilpfeiffer: +1 missed it when I moved it up a level

packages/terra-dropdown-button/src/_variables.scss
neilpfeiffer: these are the emphasis variant enhancement for split button that is ready to go for a quick-add after UX gets a chance to review the enhancement request. Have it stashed so fine to remove.

PR #2493, opened 2019-07-01
[terra-form-select] Focus Incorrectly Returning to Input When Navigating Away

### LINKED ISSUE #242: Update z-index scale
We discussed updating the z-index scale to multiples of 1000.

Update this file:
https://github.com/cerner/terra-core/blob/master/packages/terra-legacy-theme/src/_zindex.scss

To this:
```
$terra-z-index-1: 1000 !default;
$terra-z-index-2: 2000 !default;
$terra-z-index-3: 3000 !default;
$terra-z-index-4: 4000 !default;
$terra-z-index-5: 5000 !default;
$terra-z-index-6: 6000 !default;
$terra-z-index-7: 7000 !default;
$terra-z-index-8: 8000 !default;
$terra-z-index-9: 9000 !default;
$terra-z-index-10: 10000 !default;
$terra-z-index-11: 11000 !default;
```

Do we want these variables to still go to 11?

### LINKED ISSUE #2421: [form-select] Focus Incorrectly Returning to Input When Navigating Away
# Bug Report

## Description
for `multiple` and `tag` variants, when clicking outside the select options menu, focus is incorrectly be returned to the select input (although the focus styles are not applied). This is also broken when hitting the escape key. I have to click two times to remove the focus from select.

This does not appear to be an issue when using the mouse to make selections.

This does not appear to be an issue with the other variants.

## Steps to Reproduce

Manual test the options tests in terra-form-select spec for `multiple` or `tag` variants and then click the middle of the page to view the issue. This was revealed in toolkit v5 uplift.

### Example
1. Navigate to https://engineering.cerner.com/terra-core/#/raw/tests/terra-form-select/form-select/uncontrolled-multiple
2. Enter the keys: `Tab`, `Space`, `Enter` to open dropdown and select the first item
3. Click the middle of the page. Focus is trapped in input without focus styles.
[Recording of behavior](https://cl.ly/962a02e20317)

### Example
1. Navigate to https://engineering.cerner.com/terra-core/#/raw/tests/terra-form-select/form-select/uncontrolled-tag
2. Enter the keys: `Tab`, `ArrowDown `, `ArrowDown `, `Enter` to open dropdown and select the second item
3. Enter the keys: `ESCAPE` to cancel menu options. Focus is trapped in input without focus styles.

## Expected Behavior
Using the keyboard selection of options should removes focus from the form select input as if using mouse selection.

## Environment

* Component Name and Version: `terra-form-select v5.15.0`
* Affected Variants: multiple & tag

## @ Mentions
@StephenEsser  @bjankord

jmsv6d:
### Summary
Fixes #2421
On selection with the `enter` key the assumption was being made that the menu would be closed afterwards and the focus needed stay in the input. I added a check to see if it is a variant that does not allow multiple selections before setting `closedViaKeyEvent` state to `true`


packages/terra-form-select/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ Unreleased
 
 5.18.0 - (June 28, 2019)
 ------------------
+### Fixed
```
emilyrohrbough: oops. move to under unreleased section

packages/terra-form-select/tests/wdio/select-spec.js
```diff
@@ -983,9 +983,6 @@ describe('Select', () => {
       after(() => {
         browser.keys(['Backspace']); // remove selected option
         browser.click('#root');
-
-        // talk with Brett. Seems like a bug. even with escape
-        browser.click('#root');
```
StephenEsser: Hi @bjankord.

PR #2495, opened 2019-07-03
Add refCallback prop to visually hidden text component

bjankord:
### Summary
Adding a refCallback to the visually hidden component. This allows users the ability to write text into the visually-hidden-text container via `innerText` which seems to be one of the few ways we can get IE 11 + JAWS to accurately read aria-live content.


packages/terra-visually-hidden-text/src/VisuallyHiddenText.jsx
```diff
@@ -6,6 +6,10 @@ import styles from './VisuallyHiddenText.module.scss';
 const cx = classNames.bind(styles);
 
 const propTypes = {
+  /**
+   * Callback ref to pass into the dom element.
+   */
```
yuderekyu: want to add how this is useful for aria-live - similar to the changelog entry

bjankord: Updated in this commit: https://github.com/cerner/terra-core/pull/2495/commits/63d75ab828f340f68ef3f820950f28e568fe5b6a

PR #2498, opened 2019-07-05
Change to use Terra.describeViewports part 1

Blackop778:
### Summary
Part 1 of 8 (or more!) of changing tests to align with cerner/terra-framework#729


packages/terra-alert/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Update tests to use `Terra.describeViewports` for setting viewports
```
StephenEsser: What caused the screenshot diffs?

Blackop778: I believe the old code https://github.com/cerner/terra-core/pull/2498/files?short_path=debb75d#diff-9234219fdf2a8f3279af2bcab788b0d6L34 wasn't properly structured so the before hook in the above line wasn't being called and the viewport never got changed to tiny. The screenshots were newly generated.

PR #2500, opened 2019-07-08
Add node.contains polyfill for IE 11 and IE 10

bjankord:
### Summary
In one of our internal tools, we've run into issues with code that uses Node.contains. While the code seems to work in IE 10 and IE 11 in our testing, teams have reported issues with it not working as expected. To help mitigate this concern, we're adding a `Node.contains` polyfill to terra-overlay.


packages/terra-overlay/src/_contains-polyfill.js
```diff
@@ -0,0 +1,19 @@
+// Source: https://github.com/Financial-Times/polyfill-service/pull/183/files
+// Needed for IE11 and IE10
+if (!document.contains) {
```
StephenEsser: This might not be an issue, but would we want to check if `contains` already exists on node before adding it? I'm not sure what would happen if another package in the future needed this polyfill, would they just overwrite each other?

bjankord: Updated in this commit: https://github.com/cerner/terra-core/pull/2500/commits/89bfcaa561120db6f21d5c0db11f527933d1a33a

PR #2501, opened 2019-07-08
Change to use Terra.describeViewports part 3

Blackop778:
### Summary

I made a few misc updates to the DocTemplate spec file. The test had become broken and no longer opened the code for the example. I also changed all the tests to run under the `large` viewport. Those resulted in the screenshot changes.

In Form Fieldset I changed it so both screenshots are taken at both viewports so a new screenshot was generated.


packages/terra-doc-template/tests/wdio/doc-template-spec.js
```diff
@@ -1,39 +1,29 @@
-describe('DocTemplate', () => {
-  before(() => browser.setViewportSize(Terra.viewports('huge')[0]));
-
+Terra.describeViewports('DocTemplate', ['large'], () => {
```
emilyrohrbough: Why did this get updated to large when it previously tests huge?

packages/terra-form-checkbox/tests/wdio/form-checkbox-field-spec.js
```diff
@@ -50,24 +65,6 @@ describe('CheckboxField', () => {
       browser.click('[for="ux-dept"]');
     });
 
-    Terra.it.matchesScreenshot({ viewports: [viewports[0]] });
-  });
-
-  describe('CheckboxField Optional', () => {
```
emilyrohrbough: two test cases were dropped. Was this intentional?

packages/terra-form-checkbox/tests/wdio/form-checkbox-spec.js
```diff
@@ -36,53 +31,46 @@ describe('Checkbox', () => {
...
       });
 
-      Terra.it.matchesScreenshot();
-      Terra.it.isAccessible();
+      Terra.it.validatesElement();
     });
   });
 
   describe('Disabled', () => {
     beforeEach(() => browser.url('/#/raw/tests/terra-form-checkbox/form-checkbox/checkbox/disabled-checkbox'));
```
emilyrohrbough: Is it possible to remove these beforeEach hooks? These can cause unexpected behaviors and slow down test execution time

packages/terra-form-checkbox/tests/wdio/form-checkbox-field-spec.js
emilyrohrbough: nvm I see they were added above.

packages/terra-doc-template/tests/wdio/doc-template-spec.js
Blackop778: The PropsTable screenshot requires a custom taller viewport, which was being saved to `large`. It was also causing the screenshots taken after it to be `large` because they are no longer being passed a `viewports` option. I decided to make them all large.

packages/terra-form-checkbox/tests/wdio/form-checkbox-spec.js
Blackop778: Fixed in cbac7ee370e2e95511fa85f71a127ab4a258ebec

PR #2503, opened 2019-07-08
Change to use Terra.describeViewports part 5

Blackop778:
### Summary


packages/terra-markdown/tests/wdio/markdown-spec.js
```diff
@@ -1,9 +1,9 @@
-describe('Markdown', () => {
-  before(() => browser.setViewportSize({ width: 1000, height: 10000, name: 'large' }));
+Terra.describeViewports('Markdown', ['large'], () => {
   describe('Default', () => {
     before(() => browser.url('/#/raw/tests/terra-markdown/markdown/markdown'));
 
-    Terra.it.validatesElement({ selector: '#root' });
+    Terra.it.isAccessible();
+    Terra.it.matchesScreenshot({ selector: '#root', viewports: [{ width: 1000, height: 10000, name: 'large' }] });
```
emilyrohrbough: We don't want to do this. We really should be using a terra-defined viewport. What is being captured here?

Blackop778: https://github.com/cerner/terra-core/blob/227b8ea4e851b21312e87fb2d292e86d35c15b10/packages/terra-markdown/tests/wdio/__snapshots__/reference/en/chrome_large/markdown-spec/Default%5Bdefault%5D.png
is what is being captured

mjhenkes: I did this 😬. I wanted to capture the whole rendered markdown document in one image.

Blackop778: I removed this screenshot and made the `scss` screenshot smaller so it would fit in a regular viewport

PR #2511, opened 2019-07-10
Change to use Terra.describeViewports part 7

Blackop778:
### Summary

In single select table the `beforeEach` hook was causing the 2nd and 3rd rows to not get selected. For some reason changing this caused yellow focus glow to appear.


packages/terra-table/tests/wdio/single-select-table-spec.js
```diff
@@ -1,73 +1,80 @@
...
 
-      Terra.it.matchesScreenshot('third-row', { viewports });
+      Terra.it.matchesScreenshot('third-row');
     });
 
     describe('Single-Select Table - Keyboard ENTER key', () => {
+      // reset focusable point for keyboard navigation
+      before(() => {
+        browser.click('[data-terra-table-header-cell]:nth-child(1)');
+        browser.pause(50);
```
Blackop778: For some reason removing the `beforeEach` hook caused the blue line that for some reason appears at the bottom of the 1st and 2nd row selected screenshots to sometimes not appear, this causes it to always appear.

packages/terra-table/tests/wdio/single-select-table-spec.js
```diff
@@ -1,73 +1,80 @@
...
 
-      Terra.it.matchesScreenshot('third-row', { viewports });
+      Terra.it.matchesScreenshot('third-row');
     });
 
     describe('Single-Select Table - Keyboard SPACE key', () => {
+      // reset focusable point for keyboard navigation
+      before(() => {
+        browser.click('[data-terra-table-header-cell]:nth-child(1)');
+        browser.pause(50);
```
Blackop778: For some reason removing the `beforeEach` hook caused the blue line that for some reason appears at the bottom of the 1st and 2nd row selected screenshots to sometimes not appear, this causes it to always appear.

packages/terra-table/tests/wdio/single-select-table-spec.js
```diff
@@ -1,73 +1,80 @@
-const viewports = Terra.viewports('medium');
-
-describe('Single-Select Table', () => {
+Terra.describeViewports('Single-Select Table', ['medium'], () => {
```
emilyrohrbough: How come the selections changed in these screenshots?

Blackop778: The `beforeEach` hook was causing the 2nd and 3rd rows to not get selected like the test names suggested they were. Not sure why the focus glow became yellow.

emilyrohrbough: The yellow focus glow is related to the Chrome version we are testing

packages/terra-table/tests/wdio/multi-select-table-spec.js
```diff
@@ -1,29 +1,27 @@
-const viewports = Terra.viewports('medium');
```
mjhenkes: why did the above screenshots change?

Blackop778: There was a `beforeEach` hook  that brought focus to the top of the table between every test. One of the tests was supposed to focus each row but was not able to because of the `beforeEach` hook. I think some of the screenshots hadn't been updated since the focus glow became yellow.

mjhenkes: I guess it's the table screenshots from the single select table

mjhenkes: and i see your response below

PR #2515, opened 2019-07-11
[terra-overlay] Disabled document scroll if Overlay is not relative to the triggering container.

### LINKED ISSUE #2514: Overlay disables scrolling
# Bug Report

## Description
We have a test page where the view is in a OverlayContainer with an Overlay and children.  Prior to version 3.18.0, after the overlay was turned off, the view would scroll.  At the 3.18.0 version, the view would no longer scroll.

## Steps to Reproduce

Different steps than described above, but is an example of a symptom.
1. Go to terra-ui.com>Components>terra-overlay
2. Inspect the DOM, noting the top-most element's styles.
3. Click a "Trigger Container Overlay" button.
4. Notice the element changes to have a style of overflow:hidden.  It will not go away.

In the Fix #2483, changes were made to the disableContainerChildrenFocus function such that the document's style.overflow is ALWAYS set to 'hidden'.  Previously, the document's style was only set if the overlay is NOT isRelativeToContainer.
Note: The mirror image code in enableContainerChildrenFocus only changes the style in the case the overlay is NOT isRelativeToContainer.  Meaning if the Overlay isRelativeToContainer the document is permanently left with an overflow style of hidden.

## Expected Behavior
After an overlay is stopped, the scroll behavior should return to what it was before.

## Possible Solution
Move the line to set overflow = hidden up one line, such that it is again in the ELSE block. i.e. it only executes if the overlay is NOT isRelativeToContainer.

## Environment

* Component Name and Version: terra-overlay 3.18.0
* Browser Name and Version: I was testing in chrome, but I don't think it would matter.
* Webpack Version:
* Operating System and version (desktop or mobile): desktop Mac

## @ Mentions
@bjankord

nramamurth:
### Summary

Fixes #2514.

### DIFF
--- a/packages/terra-overlay/CHANGELOG.md
+++ b/packages/terra-overlay/CHANGELOG.md
@@ -4,6 +4,9 @@ Changelog
 Unreleased
 ----------
 
+### Fixed
+* Disabled document scroll if Overlay is not relative to the triggering container.
+
 3.18.0 - (July 9, 2019)
 ------------------
 ### Fixed
--- a/packages/terra-overlay/src/Overlay.jsx
+++ b/packages/terra-overlay/src/Overlay.jsx
@@ -115,8 +115,8 @@ class Overlay extends React.Component {
         document.querySelector(selector).setAttribute('data-overlay-count', `${inert + 1}`);
         document.querySelector(selector).setAttribute('inert', '');
       }
+      document.documentElement.style.overflow = 'hidden';
     }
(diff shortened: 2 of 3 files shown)


nramamurth:
@bjankord, done: https://github.com/cerner/terra-core/pull/2515/commits/33ca9a68ff08b9ddf03c4c9e5d9b45e113ad475c

PR #2517, opened 2019-07-11
Docs Updated around requirement of terra-base and translations.

### LINKED ISSUE #2391: Improve docs around the requirement of terra-base and aggregate-translations
# Feature Request

## Description
Add documentation on all component pages to make it clear that using Terra UI components require the use of terra-base to be wrapped at the root of your mounting React app as we'll as requiring the use of aggregate-translations.

Consumers are running into some confusion when using Terra UI components in regards to the requirements of using terra-base. This is something we can better document and provide guidance on to eliminate some of the learning curve when getting started with Terra UI components.

saket2403:
### Summary
Added documentation on all components with translations to make it clear that using Terra UI components require the use of terra-base to be wrapped at the root of the mounting React app as well as requiring the use of aggregate-translations.
Resolves #2391

### DIFF
--- a/packages/terra-action-header/docs/README.md
+++ b/packages/terra-action-header/docs/README.md
@@ -8,7 +8,7 @@ The terra-action-header component is a header bar containing a title and optiona
   - `npm install terra-action-header`
 
 ## Implementation Notes:
-The Action-Header component must be composed inside the [Base][1] component with a locale in order for it to load the correct translation strings. These translation strings are used for the default Action-Header title when the application does not specify a title for the Action-Header.
+The Action-Header component must be composed inside the [Base][1] component with a locale in order for it to load the correct translation strings.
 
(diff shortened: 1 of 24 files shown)


packages/terra-action-header/docs/README.md
```diff
@@ -7,6 +7,11 @@ The terra-action-header component is a header bar containing a title and optiona
 - Install with [npmjs](https://www.npmjs.com):
   - `npm install terra-action-header`
 
+## Implementation Notes:
+The Action-Header component must be composed inside the [Base][1] component with a locale in order for it to load the correct translation strings. These translation strings are used for the default Action-Header title when the application does not specify a title for the Action-Header.
```
nramamurth: I think it adds maintainability to mention, "These translation strings are used for the default Action-Header title when the application does not specify a title for the Action-Header." If it is removed or if more text is internationalized in the future, then the developer has to come and update it here.

PR #2522, opened 2019-07-15
[terra-action-header] Document that maximize/minimize will not be shown with back button

### LINKED ISSUE #2313: ActionHeader does not render Maximize/Minimize buttons if Back button is present
# Bug Report

## Description
The ActionHeader has explicit logic that will prevent the Maximize and Minimize buttons from rendering if a Back button is also being rendered. This doesn't seem appropriate for most use cases. Is this a bug or an explicit design decision?

## Expected Behavior
The ActionHeader can render Back and Maximize/Minimize buttons at the same time.

## Possible Solution
We should just need to remove the backButton presence check here: https://github.com/cerner/terra-core/blame/master/packages/terra-action-header/src/ActionHeader.jsx#L101

## Environment

* Component Name and Version: terra-action-header/latest

Blackop778:
### Summary
Updates the documentation for `onMinimize` and `onMaximize` to state that the maximize/minimize buttons will not appear when the back button is shown and adds an example of using a custom implementation to add a maximize/minimize button while showing the back button.

### Additional Details
Closes #2313


packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -30,12 +30,18 @@ const propTypes = {
    * Callback function for when the expand button is clicked.
    * The expand button will not display if this is not set or on small viewports.
    * Only the expand button will be rendered if onMaximize and onMinimize are set.
+   *
+   * *Note: If `onBack` is set the maximize button will not appear and must be provided as a child inside
```
bjankord:
```suggestion
   * *Note: If `onBack` is set, the maximize button will not appear and must be provided as a child inside
```

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -30,12 +30,18 @@ const propTypes = {
...
+   * a `Collapsible Menu View`. See examples for an example.*
    */
   onMaximize: PropTypes.func,
   /**
    * Callback function for when the minimize button is clicked.
    * The minimize button will not display if this is not set or on small viewports.
    * Only the expand button will be rendered if both onMaximize and onMinimize are set.
+   *
+   * *Note: If `onBack` is set the minimize button will not appear and must be provided as a child inside
```
bjankord:
```suggestion
   * *Note: If `onBack` is set, the minimize button will not appear and must be provided as a child inside
```

mlh758:
Would it be possible for you to also update the docs for the `children` prop to clarify that you need to pass in a menu view and not just menu items? The current documentation is a little ambiguous to me.

Blackop778:
@mlh758 Done in 46a2bec4d1fc4952ddd0aeb25246df808efa6b46

mlh758:
Thank you!

packages/terra-action-header/package.json
```diff
@@ -29,6 +29,7 @@
     "prop-types": "^15.5.8",
     "terra-button": "^3.17.0",
     "terra-button-group": "^3.17.0",
+    "terra-collapsible-menu-view": "^6.5.0",
```
mjhenkes: I'm conflicted about adding a dependency for just an example.

bjankord: Yeah, the odd part about it is with terra-collapsible-menu-view being it terra-framework, it will make major version bumps weird when releasing terra-collapsible-menu-view.

I'm fine with it being a dependency if we move terra-action-header to terra-framework.

ryanthemanuel: Let's just document rather than adding this dependency just to have an example

Blackop778: Done in 380ceaf77a0f7641c18c8b17327a3849634461a9

PR #2524, opened 2019-07-16
Fix list linting errors

Blackop778:
### Summary
I've seen a few of my builds fail to linting errors here, but not consistently. Not sure why.


packages/terra-list/src/terra-dev-site/doc/guides/SectionList.jsx
```diff
@@ -1,9 +1,11 @@
 import React from 'react';
+/* eslint-disable import/no-extraneous-dependencies */
 import List, {
   Item,
   Section,
   Utils,
-} from 'terra-list/lib/index'; // eslint-disable-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+} from 'terra-list/lib/index';
+/* eslint-enable */
```
StephenEsser: These should probably enable the rule that was disabled 

```
/* eslint-enable import/no-extraneous-dependencies */
```

Blackop778: This was changed in e9a5d37a7c9efac19c5fa440f1777cc6d0f54f75

PR #2531, opened 2019-07-17
[terra-button] Show focus styles when programmatically focused

### LINKED ISSUE #2525: [terra-button] Update visual behavior to handle programmatic focus
# Feature Request
Update visual style for the programmatic setting of focus.

## Description
When setting focus on a terra-button through javascript default to keyboard style focus to ensure when navigating between elements with other keys than tab.

This relates back to the implementation of menus within the terra-application-navigation, where keyboard navigation uses arrow keys as well as the tab key.

## @ Mentions
@scottwilmarth

Blackop778:
### Summary
`Button` will now show focus styles when programmatically focused. I'm not really sure why it uses a focus state and `is-focused` class but I'm leaving that alone.

### Additional Details
Closes #2525


packages/terra-button/src/_mixins.scss
```diff
@@ -8,7 +8,8 @@
   color: map-get($color-map, color);
   font-weight: map-get($color-map, font-weight);
 
-  &.is-focused {
+  &.is-focused,
+  &:focus {
```
dkasper-was-taken: Solving the problem this way will force mouse selection to also get focus, which we don't want, so we'll need to try solving this another way.

packages/terra-button/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changelog
 
 Unreleased
 ----------
+### Fixed
+* Fixed Button not showing focus styles when programmatically focused
```
emilyrohrbough: What do you mean by programmatically focused?

Blackop778: Programmatic focus would be getting a reference to the DOM element and calling the `focus()` method.

emilyrohrbough:
>  I'm not really sure why it uses a focus state and is-focused class but I'm leaving that alone.

This is to allow keyboard focus styles to be applied to the button consistently across browsers: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Button#Clicking_and_focus

I wish I had the original docs I found on this, but that was a while ago

Blackop778:
> > I'm not really sure why it uses a focus state and is-focused class but I'm leaving that alone.
> 
> This is to allow keyboard focus styles to be applied to the button consistently across browsers: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Button#Clicking_and_focus
> 
> I wish I had the original docs I found on this, but that was a while ago

I guess it makes more sense if we don't want the button to display focus styles when it gets clicked. Don't think I've seen that documented anywhere though.

packages/terra-button/tests/wdio/button-spec.js
```diff
@@ -238,4 +238,18 @@ Terra.describeViewports('Button', ['tiny'], () => {
...
     Terra.it.validatesElement({ selector: '#compactButton' });
   });
+
+  describe('Programmatic Focus', () => {
+    before(() => {
+      browser.url('/#/raw/tests/terra-button/button/programmatic-focus');
+      browser.click('#button1');
+    });
+
+    Terra.it.matchesScreenshot('Button 2 focused programmatically');
```
emilyrohrbough: Shouldn't this be button 1?

Blackop778: button 1's callback focuses button 2 and vice versa. Button 1 was just clicked so button 2 is now focused.

packages/terra-button/src/_mixins.scss
Blackop778: Fixed in 06a47fb35b2e4d69e5204e17572a246f8b1357a6

PR #2534, opened 2019-07-18
[terra-doc-template] Only render example syntax when expanded

tbiethman:
### Summary

Recently, I found that the DataGrid examples rendered through the ExampleTemplate were a lot less performant than the similarly sized test examples. The markup generated by the ExampleTemplate includes the highlighted markup syntax even when the code is not expanded open. 

We can hide it until presentation to be more performant.


packages/terra-doc-template/src/ExampleTemplate.jsx
```diff
@@ -113,10 +115,12 @@ class ExampleTemplate extends React.Component {
                 <span className={cx('chevron-right')} />
               </button>
             </div>
-            <div className={cx('code', { 'is-expanded': this.state.isExpanded })} aria-hidden={!this.state.isExpanded}>
-              <SyntaxHighlighter language="jsx" style={theme} customStyle={{ margin: '0', borderRadius: '0' }}>
-                {exampleSrc}
-              </SyntaxHighlighter>
+            <div className={cx('code', { 'is-expanded': isExpanded })} aria-hidden={!isExpanded}>
+              {isExpanded ? (
```
StephenEsser: It probably doesn't make too much of a difference, is there a benefit to the full ternary syntax over the && syntax?

```jsx
isExpanded && <SyntaxHighlighter />
```

```jsx
isExpanded ? <SyntaxHighlighter /> : undefined
```

tbiethman: The first one will result in rendering `{false}` instead of `{undefined}`. That's the extent of the difference, I believe.

PR #2535, opened 2019-07-18
[terra-core] Update lerna to v3

### LINKED ISSUE #1792: Update to lerna v3
# Feature Request

## Description
[Lerna v3.0.0 was released on 2018-08-10](https://github.com/lerna/lerna/blob/master/CHANGELOG.md). The latest stable release is 3.1.4. This issue is to determine changes needed to update to v3 and implement said changes in terra-core.

Looking through the changelog from the version we use to v3 briefly, I didn't really see any breaking changes for our usage. One thing of note is that the key `ignore` in the lerna.json file changed to `ignoreChanges`, though it sounds like `ignore` will still work, it was just deprecated. However, a deeper dive into the changes will be required to update.

pranav300:
### Summary
Updated Lerna to latest version `^3.15.0` from version `^2.1.2`.
Resolves #1792


lerna.json
```diff
@@ -1,9 +1,9 @@
 {
-  "lerna": "2.1.2",
+  "lerna": "3.15.0",
```
bjankord: The lerna key looks like it can be removed, it is now longer listed in [the lerna.json documentation](https://github.com/lerna/lerna#lernajson).

lerna.json
```diff
@@ -9,5 +7,11 @@
...
     }
   },
-  "packages": ["packages/*"]
-  }
+  "packages": [
+    "packages/*"
+  ],
+  "version": [
+    "independent",
+    "3.15.0"
```
bjankord: I don't think this `"3.15.0"` line needs to be here. The docs state that version key is used to specify the current version of the repo, since we're using independent versioning, we should use:

```
"version": [
  "independent"
]
```

pranav300: updated here https://github.com/cerner/terra-core/pull/2535/commits/74606e6605036404696ff870060dc8d4b1229dee.

package.json
```diff
@@ -85,7 +85,7 @@
     "glob": "^7.1.2",
     "identity-obj-proxy": "^3.0.0",
     "jest": "^24.8.0",
-    "lerna": "^2.1.2",
+    "lerna": "^3.15.0",
```
bjankord: 3.16.1 was released recently, lets set that as the lower bound.

pranav300: Fixed here https://github.com/cerner/terra-core/pull/2535/commits/653d751d7ad2331e1e96bdd3a131237b83b4e096.

PR #2536, opened 2019-07-18
[search-field] Autofocus feature added

### LINKED ISSUE #1091: Missing Autofocus feature in SearchField
# Issue Description
Currently there is no autofocus prop for the SearchField. User has to click inside the SearchField to bring the focus.

## Issue Type

- [X] New Feature
- [ ] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
As soon SearchField is displayed the focus should be in the input box.

## Current Behavior

N/A

saket2403:
### Summary
Resolves #1091

### DIFF
--- a/packages/terra-search-field/src/SearchField.jsx
+++ b/packages/terra-search-field/src/SearchField.jsx
@@ -77,6 +77,11 @@ const propTypes = {
    */
   // eslint-disable-next-line react/forbid-prop-types
   inputAttributes: PropTypes.object,
+
+  /**
+   * When true, will enable the auto-focus.
+   */
+  enableAutoFocus: PropTypes.bool,
 };
 
 const defaultProps = {
@@ -89,6 +94,7 @@ const defaultProps = {
   searchDelay: 250,
   value: undefined,
   inputAttributes: undefined,
+  enableAutoFocus: true,
 };
 
 const contextTypes = {
@@ -211,6 +217,7 @@ class SearchField extends React.Component {
       value,
       inputRefCallback,
       inputAttributes,
+      enableAutoFocus,
       ...customProps
     } = this.props;
 
@@ -230,6 +237,10 @@ class SearchField extends React.Component {
       additionalInputAttributes.value = value;
     } else {
       additionalInputAttributes.defaultValue = defaultValue;
+    }
+    /* If block to set autoFocus based on prop value. */
+    if (enableAutoFocus === true) {
+      additionalInputAttributes.autoFocus = true;
     }
 
     const clearButton = this.state.searchText && !isDisabled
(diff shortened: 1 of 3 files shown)


packages/terra-search-field/src/SearchField.jsx
```diff
@@ -89,6 +94,7 @@ const defaultProps = {
   searchDelay: 250,
   value: undefined,
   inputAttributes: undefined,
+  enableAutoFocus: true,
```
nramamurth: Should this be `false` by default?

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -231,6 +238,10 @@ class SearchField extends React.Component {
     } else {
       additionalInputAttributes.defaultValue = defaultValue;
     }
+    /* If block to set autoFocus based on prop value. */
+    if (enableAutoFocus === true) {
```
nramamurth: I'd simply do  ``` additionalInputAttributes.autoFocus = enableAutoFocus; ``` and remove the `if` block.

StephenEsser:
We may need to reach out to UX and other associates that work closely with accessibility to determine if this is a feature we want to support. Auto focus comes with a few challenges. 

1.  In general, is autofocus accessible when run on page load?
2. What happens if two search fields are displayed on the same page and both request autofocus?
3. Should all inputs support an autofocus feature or is autofocus special to SearchField?

packages/terra-search-field/src/SearchField.jsx
StephenEsser: Booleans should default to false whenever possible. Booleans that default to true introduce a discouraged API. 

```jsx
<SeachField autoFocus={false} />
```

Vs

Auto focus on:
```jsx
<SeachField autoFocus />
```
Auto focus off:
```jsx
<SeachField />
```

yuderekyu:
Yeah, there's a11y issues. Quoting MDN:
> Warning: Automatically focusing a form control can confuse visually-impaired people who using screen-reading technology. When autofocus is assigned, screen-readers "teleport" their user to the form control without warning them beforehand.

packages/terra-search-field/src/SearchField.jsx
supreethmr: I am assuming this condition is added to make sure auto-focus attribute gets added to search-input only when it is required i.e. when enableFocus is true. Because adding attribute autofocus to search-input by default will make input to be auto-focused even for enableFocus = false.

nramamurth: > Because adding attribute autofocus to search-input by default will make input to be auto-focused even for enableFocus = false.

No, it does not. The condition is redundant imo.

saket2403: @nramamurth . Removing if works perfectly fine. But I guess we'll have to wait for UX review to find out whether this enhancement is required.

bjankord:
From talking with @scottwilmarth, it sounds like this is not something that we want to enable for all consumers due to accessibility concerns.

PR #2510, opened 2019-07-10
[terra-form-select] Added aria-label for every input field

pranav300:
### Summary
Resolves #2505

PR #2549, opened 2019-07-25
terra-form-select.filter(simplicity).reduce(complexity)

dev-cprice:
### Summary

Slice n' dice up `terra-form-select` into specialized components..

#### Rationale

We want to flatten out the select code and remove any premature abstractions that introduce complexity. This means that the code will be less DRY (more wet? 💦), but more maintainable.

In the future, we can iterate on any abstractions we feel add to the clarity of the code (read as: extract out any variant independent code, e.g. `<Tag />`, `<Option />`, `FooUtil.someMethodHere()`, etc).

#### Current plan:

- add variant specific directories
- move variant dependent code into each variant's directory, e.g.:

```
combobox/
L Combobox.jsx
L Frame.jsx
L Menu.jsx
default/
L Default.jsx
L Frame.jsx
L Menu.jsx
```

- move variant independent code into a shared directory, e.g.

```
shared/
L SelectUtil.js
L MenuUtil.js
L FrameUtil.js
L SharedUtil.js
L _AddOption.jsx
L _ClearOption.jsx
L _ClearOption.module.sccs
L ...
```

- new Select component which acts as a switch router to the new variant specific select implementations, seen [here](https://github.com/cerner/terra-core/blob/split-select-up/packages/terra-form-select/src/Select.jsx#L135-L151)
- fix any plumbing for tests & docs 🔧

@cerner/terra

[License]: ../LICENSE


StephenEsser:
Theming will likely be impacted by these changes. I wanted to comment to make sure it's in mind while doing this refactor. 

If every component has its own unique markup, can we continue to share the theme variables or will each variant now need to implement its own complete set of theme variables?

dev-cprice:
@StephenEsser That's a good concern/question. So, right now they're all sharing the same css as prior to the refactor. So, they all share the same theming vars as well. I left it this way since it's most passive, but it may make sense to split out theming vars into variants eventually.

So far, I've just duplicated the Frame, Menu, and Select code for each of the variants and removed any logic that was dependent on the variant prop. For something like [this](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/_Frame.jsx#L688), the variant was just hardcoded in, like for combobox:

```diff
const selectClasses = cx([
  'select',
-  variant,
+ 'combobox',
// ...
```

StephenEsser:
Is there an intention to flatten out the Frame from each component? The additional level of state can result in duplicate renders. I don't see much value in preserving the Frame for each component.

dev-cprice:
I initially planned on doing that, but wanted to keep the code as passive as possible. I think that the distinction between a frame and select is pointless now and can cause more harm than good.

---

I still see value in flattening, but wanted to do this iteratively so that each variant's specific behavior & functionality was maintained without regression. I've been tied up lately, but once I update the tag variant's Menu, then I'll be able to flatten if we all agree it's best.

StephenEsser:
+1 to flatten out the Frame.

The other tricky problem this refactor presents is the Field component. Will each variant have it's own Field wrapper (ComboboxField, MultiSelectField, TagField)? The issue with a single Wrapper will be code splitting. Any use of a SelectField as it is implemented now will result in pulling in every single component into the bundle.

dev-cprice:
Tbh, I'm not a fan of the single select component export for the same reason. It's a switch with some prop filtering per variants since we can't just do full on prop pass through (else we spread useless things onto the dom elements via customProps).

I would much rather see exporting each variant individually and having individual field wrappers as well.. that's a breaking change/version bump tho. Would be nice for docs too, since it's would allow us to have individual entries for each variant and be more explicit which props are applied.

StephenEsser:
Adding individual Field wrappers would be passive as long the original Field wrapper still accepted the variant prop and used the Select switch to return the correct component. Is your passivity concern around how they get exported / imported?

bjankord:
Talked about skipping the flattening of frame and handling that in a separate PR. We also talked about exporting Combobox, ComboboxField etc. along with Select and SelectField for passivity.

dev-cprice:
Thoughts on naming? All of the following are shallow lib imports, e.g:
`import ComboboxSelect from 'terra-form-select/lib/ComboboxSelect';`

- ComboboxSelect
- ComboboxSelectField
- DefaultSelect
- DefaultSelectField
- MultipleSelect
- MultipleSelectField
- SearchSelect
- SearchSelectField
- Select (legacy compat, uses variant prop)
- SelectField (legacy compat, uses variant prop)
- TagSelect
- TagSelectField

dev-cprice:
Also on my radar:
I still need to add docs on the use/import of the new variant selects.

dev-cprice:
Will be addressing conflicts introduced from https://github.com/cerner/terra-core/pull/2510

packages/terra-form-select/CHANGELOG.md
```diff
@@ -4,8 +4,10 @@ ChangeLog
 Unreleased
 ----------
 ### Changed
-* updated package.json test scripts
+* WIP: splitting up terra-form-select
```
bjankord: Lets get this updated to be accurate about the changes. Let's also group all the change bullet points under the `Unreleased` section under 1 `### Changed` heading.

packages/terra-form-select/src/terra-dev-site/doc/form-select/Select.1.doc.jsx
```diff
@@ -1,11 +1,12 @@
 import React from 'react';
 import DocTemplate from 'terra-doc-template';
 import ReadMe from '../../../../docs/README.md';
+import { name } from '../../../../package.json';
 
 /* eslint-disable import/no-webpack-loader-syntax, import/first, import/extensions, import/no-unresolved, import/no-duplicates */
 // Component Source
-import OptionSrc from '!raw-loader!../../../../src/_Option';
-import OptGroupSrc from '!raw-loader!../../../../src/_OptGroup';
+import OptionSrc from '!!raw-loader!../../../../src/shared/_Option';
```
bjankord: Curious, what's the extra `!` for in front of the raw-loader?

bjankord:
@dev-cprice [Naming](https://github.com/cerner/terra-core/pull/2549#issuecomment-519273468) sounds good to me. Here some other options for `DefaultSelect` that I'm curious what others think:
* Listbox: This ties to the idea of [aria-role="listbox"](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/listbox_role)
* Dropdown: Another term people often use for this type of control

> Any thoughts on how we want to document this? I could add individual pages for the variants in the doc site?

Let's go with that option, I like the idea of individual pages for the variants, that will help with searching for those components in the doc site.

packages/terra-form-select/src/terra-dev-site/doc/form-select/Select.1.doc.jsx
dev-cprice: interesting.. not sure where that came from, will fix 👍

packages/terra-form-select/CHANGELOG.md
```diff
@@ -4,10 +4,16 @@ ChangeLog
 Unreleased
 ----------
 ### Added
+* Added Combobox and ComboboxField exports and doc site entries
+* Added DropdownSelect and DropdownSelectField exports and doc site entries
```
StephenEsser: What is the DropdownSelect? How is this different from the Default Select?

dev-cprice: DropdownSelect is the DefaultSelect, just renamed. See: https://github.com/cerner/terra-core/pull/2549#issuecomment-519890741

image

dev-cprice: We decided against naming DefaultSelect ➡️Select due to passivity (in case people are pulling in Select via `import Select from 'terra-form-select/lib/Select';`, which is common enough). We want the default export to remain `lib/Select.js`, which now points to the switch selector between the actual individual variant implementations.

packages/terra-form-select/CHANGELOG.md
```diff
@@ -4,10 +4,16 @@ ChangeLog
...
+* Added MultipleSelect and MultipleSelectField exports and doc site entries
```
StephenEsser: A few of the other front-end libraries refer to this as the Multiselect, something to consider, but probably not a huge deal either way.

dev-cprice: I'm fine with either approach. I went with `MultipleSelect` since the variant was `'multiple'` instead of `'multi'`. I'm open to change if we all agree.

packages/terra-form-select/CHANGELOG.md
```diff
@@ -4,10 +4,16 @@ ChangeLog
...
+* Added SearchSelect and SearchSelectField exports and doc site entries
+* Added TagSelect and TagSelectField exports and doc site entries
```
StephenEsser: Similar to Combobox, what are peoples thoughts on dropping the Select suffix. `Search` and `Tag` respectively? Search may have a little confusion with the existing SearchField, but `SearchSelect` and `TagSelect` feels really strange, especially with Combobox not following the same pattern.

packages/terra-form-select/CHANGELOG.md
StephenEsser: Hmm, that's tricky. I would expect the default export to be `Select`. A `SingleSelect` might be another option to consider if it absolutely must have a different name than `Select`.

packages/terra-form-select/CHANGELOG.md
dev-cprice: Naming is hard.

Just like the confusion with `SearchField`, `Tag` sounds like it's just the `_Tag` component (pill, chip, tag, whatever it's called) which multi-selected values are displayed as.

packages/terra-form-select/CHANGELOG.md
dev-cprice: It's definitely not ideal, but passivity is a virtue/feature.

I can vibe with `SingleSelect`.

packages/terra-form-select/src/Combobox.jsx
```diff
@@ -0,0 +1,231 @@
...
+    if (allowClear) {
+      if (selectPlaceholder.length === 0) {
+        clearOptionDisplay = defaultPlaceholder;
+      } else {
+        clearOptionDisplay = selectPlaceholder;
+      }
+    }
+
+    return (
+      <Frame
```
StephenEsser: Do we have an issue logged to flatten out the Frame?

dev-cprice: Not logged yet. Was going to log one immediately after this merges, but I can go ahead and just ref this PR

dev-cprice: Logged here: https://github.com/cerner/terra-core/issues/2582

bjankord:
Talked in scrum and landed on:
* Select (stays as is)
* Single Select (Used to be Dropdown Select / Default Select )
* Combobox
* Multi Select (Used to be Multiple Select)
* Tag Select
* Search Select

packages/terra-form-select/CHANGELOG.md
StephenEsser: Spoke offline and the consensus was to keep these as SearchSelect and TagSelect

Issue #2542, opened 2019-07-19
Update SearchField examples so they are more helpful


bjankord:
# Feature Request

## Description
Currently, the search field code samples display this as the code:

```
import React from 'react';
// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
import SearchFieldExampleTemplate from 'terra-search-field/lib/terra-dev-site/doc/example/SearchFieldExampleTemplate';

const SearchFieldExample = () => (
  <SearchFieldExampleTemplate />
);

export default SearchFieldExample;
```

This doesn't offer much insight to how to actually use the SearchField component. It would be better if the code examples were more concrete so people could copy/paste them and have them be functional.

ShettyAkarsh:
closed with #2550

PR #2550, opened 2019-07-26
[terra-search-field] Updates Terra SearchField examples to show usage.

lokesh-0813:
### Summary
Updates the terra-search-field examples code to show detail implementation of existing features.

Solves : #2542


bjankord:
This looks good to me. I noticed the `Search Field that has Autocomplete, Autocorrect, Autocapitalize, and Spellcheck disabled` example on the deployed site doesn't render any code when you toggle the code example. It looks like it was this way before this change as well. Let's get that updated as a part of this PR.

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldBasic.jsx
```diff
@@ -1,9 +1,38 @@
...
-import SearchFieldExampleTemplate from 'terra-search-field/lib/terra-dev-site/doc/example/SearchFieldExampleTemplate';
+import SearchField from 'terra-search-field/lib/SearchField';
 
-const SearchFieldExample = () => (
-  <SearchFieldExampleTemplate />
-);
+class SearchFieldBasic extends React.Component {
+  constructor(props) {
+    super(props);
+    this.invalidSearchMessage = 'The defualt minimum search length is 2.';
```
StephenEsser: It is a little strange to declare this string as a class property. A couple other options would be to declare a constant above or inline the message as a fallback.

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldBasic.jsx
```diff
@@ -1,9 +1,38 @@
 import React from 'react';
 // eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
-import SearchFieldExampleTemplate from 'terra-search-field/lib/terra-dev-site/doc/example/SearchFieldExampleTemplate';
+import SearchField from 'terra-search-field/lib/SearchField';
 
-const SearchFieldExample = () => (
-  <SearchFieldExampleTemplate />
-);
+class SearchFieldBasic extends React.Component {
```
StephenEsser: This example is simple enough I think it could be written with hooks: https://reactjs.org/docs/hooks-intro.html

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldBasic.jsx
```diff
@@ -1,9 +1,38 @@
...
+    super(props);
+    this.invalidSearchMessage = 'The defualt minimum search length is 2.';
+    this.state = { searchText: '', message: this.invalidSearchMessage };
+    this.handleSearch = this.handleSearch.bind(this);
+    this.handleInvalidSearch = this.handleInvalidSearch.bind(this);
+  }
 
-export default SearchFieldExample;
+  handleSearch(searchText) {
+    this.setState({ searchText, message: 'Search Text: ' });
```
StephenEsser: I think message as a state variable can be removed entirely. Here it is just being set to the same string on every state update. We can add a little logic into the render to display the different messages based on the searchText length.

length > 2 message otherwise display the invalid message

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldBasic.jsx
StephenEsser: Here's an example of how it could be switched to use hooks:

```jsx
// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
import SearchField from 'terra-search-field/lib/SearchField';

const INVALID_MESSAGE = 'The default minimum search length is 2.';

const SearchFieldExample = () => {
  const [searchText, setSearchText] = useState('');

  const message = searchText.length > 2 ? `Search text: ${searchText}` : INVALID_MESSAGE;

  return (
    <div>
      <p>{message}</p>
      <SearchField onSearch={setSearchText} onInvalidSearch={setSearchText} />
    </div>
  );
};

export default SearchFieldExample;
```

lokesh-0813:
@StephenEsser  - Updated example code to use `react hooks`.
@bjankord - `search field with inputattributes` example codes are seen in the docs now.

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldDisabled.jsx
```diff
@@ -1,11 +1,20 @@
...
+
+const SearchFieldDisabled = () => {
+  const [searchText, setSearchText] = useState('');
+
+  const message = searchText.length >= 2 ? `Search text: ${searchText}` : INVALID_MESSAGE;
+
+  return (
+    <div>
+      <p>{message}</p>
+      <SearchField isDisabled onSearch={setSearchText} onInvalidSearch={setSearchText} />
```
ShettyAkarsh: I dont think so we require the code changes for this example. as it is to show the disabled control.

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldDisableAutoSearch.jsx
```diff
@@ -1,11 +1,20 @@
...
+
+const SearchFieldDisableAutoSearch = () => {
+  const [searchText, setSearchText] = useState('');
+
+  const message = searchText.length >= 2 ? `Search text: ${searchText}` : INVALID_MESSAGE;
+
+  return (
+    <div>
+      <p>{message}</p>
+      <SearchField disableAutoSearch onSearch={setSearchText} onInvalidSearch={setSearchText} />
```
ShettyAkarsh: Even for this example <SearchField DisableAutoSearch /> will be fine.

packages/terra-search-field/src/terra-dev-site/doc/search-field/SearchField.1.doc.jsx
```diff
@@ -21,7 +21,7 @@ import SearchFieldDefaultValueSrc from '!raw-loader!../../../../src/terra-dev-si
 import SearchFieldDisabled from '../example/SearchFieldDisabled';
 import SearchFieldDisabledSrc from '!raw-loader!../../../../src/terra-dev-site/doc/example/SearchFieldDisabled';
 import SearchFieldDisabledAutoFormatting from '../example/SearchFieldDisableAutoFormatAttrs';
-import SearchFieldDisabledAutoFormattingSrc from '../example/SearchFieldDisableAutoFormatAttrs';
+import SearchFieldDisabledAutoFormattingSrc from '!raw-loader!../../../../src/terra-dev-site/doc/example/SearchFieldDisableAutoFormatAttrs';
```
ShettyAkarsh: do we need this change ?

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldDisabled.jsx
lokesh-0813: Earlier this example was using a template file which was not descriptive. Hence moved it to a new file so consumers can get a clear idea.

packages/terra-search-field/src/terra-dev-site/doc/search-field/SearchField.1.doc.jsx
lokesh-0813: Yes. Currently the official website does not show the sample code for `Disabled Auto Formatting`. This change fixes that.

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldDisabled.jsx
lokesh-0813: That makes sense.

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldDisableAutoSearch.jsx
lokesh-0813: Updated this as well.

PR #2554, opened 2019-07-26
Update markdown dependencies

bjankord:
### Summary
* Updated marked to `0.7.x`
* Updated primsjs to `~1.17.1`

[marked 0.7.0](https://github.com/markedjs/marked/releases/tag/v0.7.0) has breaking changes in it but I could not find anything in our use that would break with this update. Planning to consider this a minor update.

Updating primsjs to prevent duplicate package issue with dependency build in with [react-syntax-highlighter](https://github.com/conorhastings/react-syntax-highlighter/blob/master/package.json#L12)

Screen Shot 2019-07-26 at 2 50 07 PM

### DIFF
--- a/packages/terra-markdown/CHANGELOG.md
+++ b/packages/terra-markdown/CHANGELOG.md
@@ -3,6 +3,9 @@ Changelog
 
 Unreleased
 ----------
+### Changed
+* Updated marked to 0.7.x
+* Updated primsjs to ~1.17.1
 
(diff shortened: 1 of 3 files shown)


bjankord:
Looks like there is a difference in the diff syntax highlighting in markdown content.

Old version:
Screen Shot 2019-07-26 at 3 53 55 PM

New version:
Screen Shot 2019-07-26 at 3 54 04 PM

PR #2564, opened 2019-08-05
[content-contaner] Example and docs added for scrollRefCallback prop usage.

### LINKED ISSUE #2556: Add documentation around usage of scrollRefCallback prop to content-container examples/docs
# Feature Request

## Description
Add documentation around usage of scrollRefCallback prop to content-container examples/docs. If you are familiar with refCallbacks, the prop pattern is apparent, but if this is the first time you've seen this pattern, it may be confusing.

saket2403:
### Summary
Fixes #2556


packages/terra-content-container/src/terra-dev-site/doc/example/ContentContainerScrollRef.jsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import ContentContainer from 'terra-content-container/lib/ContentContainer';
+import classNames from 'classnames/bind';
+
+import styles from './ContentContainerDocCommon.module.scss';
+
+const cx = classNames.bind(styles);
+
+const containerRef = React.createRef();
```
ShettyAkarsh: This would throw some warnings in the console as scrollRefCallback expects function to be passed. as this will pass an object.

ShettyAkarsh: With this example we cannot differentiate with fill container example we have currently. we will have to draft this example to show what actually we can do using scrollRefCallback prop for the end user.

bjankord: Yeah, maybe display the scroll position values on the example page. These can be referenced from the ref we return in the `scrollRefCallback` prop.

saket2403: Resolved in 52bf2a9

Issue #1813, opened 2018-08-30
Parse package name and latest version from artifactory 


m-samkc:
# Feature Request

## Description
Badge in [terra-doc-template](https://github.com/cerner/terra-core/blob/master/packages/terra-doc-template/src/DocTemplate.jsx#L94) looks for package in npm repository, However we have package published in internal repo and would like to reference. 

## Additional Context / Screenshots
We could have new prop called `internalNpmPackage` in `DocTemplate`. If we have user defined value for the prop then we build `href ` and `img src` with internal repository url otherwise always fallback to normal npm lookup. 

![screen shot 2018-08-30 at 3 06 08 pm](https://user-images.githubusercontent.com/38563089/44876296-48092280-ac66-11e8-8378-fa3c92a3ebed.png)

## Cosiderations  
We have npm package published to internal [artifactory](https://cernerrepos.net/) and would like to fetch package name and latest version to display.

bjankord:
Discussed updating this to make the badge be more customizable to handle internal repos.

bjankord:
Ran into an issue that made me think about this issue. The doc site has a stale deployment where the npm badge pulls the latest version from the released npm package, but the docs show an older version. This causes confusion as the user will think they are looking at the latest released docs.

This can happen if npm publish is successful but the doc site deploy fails.

This made me think that this badge should be pulling the version directly from the package.json file instead of pulling from npm service. I think this could mitigate all the concerns noted on this issue.

emilyrohrbough:
@mjhenkes would that require compiling two versions of the site and then toggling them to change which documentation you are referencing?

bjankord:
Talked with @mmalaker  and @amichaelparker about this today. VoiceOver in Safari seems to get trapped in the SVG DOM loaded as a background image [in this img](https://github.com/cerner/terra-core/blob/master/packages/terra-doc-template/src/DocTemplate.jsx#L97).

Planning to update this so it is no longer a dynamically generated SVG via badgegen.

### Tech Design
Replace [this rendering](https://github.com/cerner/terra-core/blob/master/packages/terra-doc-template/src/DocTemplate.jsx#L97) with the following which would use a new prop `packageVersion`.

```jsx
// do some cool styling in here via the .badge class to make this look nice.
 const badge = (
    <a className={cx('badge')} href={`https://www.npmjs.org/package/${packageName}`}>
      <span>Version:</span><span>{`${packageVersion}</span>
    </a>
  );
```

[...]

PR #2571, opened 2019-08-07
Render Replica npm-style Badge in doc-template

amichaelparker:
### Summary
Ref: [This comment](https://github.com/cerner/terra-core/issues/1813#issuecomment-518840423). This renders a replica npm badge to replace the dynamically generated one from badgen. This allows projects that don't have their packages hosted on npm to not have a 404 badge on their doc site.

Closes [#1813](https://github.com/cerner/terra-core/issues/1813).

### Additional Details
This adds packageVersion and packageUrl props, and a badge will be rendered based on what is provided. If neither are, it will fall back to the current badgen badge. It also addresses an issue with Voiceover being able to jump out of focus trapped modals to the badge in some circumstances.


packages/terra-doc-template/src/DocTemplate.jsx
```diff
@@ -92,11 +102,24 @@ const DocTemplate = ({
...
   }
 
-  const badge = (
-    <a href={`https://www.npmjs.org/package/${packageName}`}>
-      <img src={`https://badgen.net/npm/v/${packageName}`} alt="NPM version" />
-    </a>
-  );
+  const badge = packageVersion
+    ? (
+      <div className={cx('badge-container')}>
```
bjankord: Let's turn this into a component within the terra-doc-template package. It should make testing easier this way. I'm thinking a new file, `packages/terra-doc-template/src/NpmBadge.jsx` that we can import here.

amichaelparker: Changed! [af52c67](https://github.com/cerner/terra-core/pull/2571/commits/af52c67081f807649628e8e8f07b4e9957635768)

packages/terra-doc-template/src/NpmBadge.module.scss
```diff
@@ -0,0 +1,28 @@
+:local {
+  .badge {
+    color: #fff;
+    font-family: Verdana,DejaVu Sans,sans-serif;
+    font-size: 11px;
```
bjankord: let's update this to `0.7857rem`

packages/terra-doc-template/src/NpmBadge.module.scss
```diff
@@ -0,0 +1,28 @@
...
+    text-decoration: none;
+  }
+
+  .badge-container {
+    margin-bottom: 0.4em;
+  }
+
+  .badge-name,
+  .badge-version {
+    padding: 3.25px 5px;
```
bjankord: let's update this to `0.232rem 0.357rem`

packages/terra-doc-template/src/terra-dev-site/doc/example/BadgeExamples.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies, import/no-unresolved, import/extensions
+import NpmBadge from '../../../NpmBadge';
+import { name } from '../../../../package.json';
+
+const packageUrl = 'https://engineering.cerner.com/terra-ui/components/terra-doc-template/doc-template/doc-template';
+const packageVersion = '2.13.0';
```
bjankord: For this file, let's actually pull the version from the package.json file.

```jsx
import { name, version } from '../../../../package.json';
```
and then pass `version` to the `packageVersion` prop. For the tests files, we can keep the version static.

packages/terra-doc-template/src/NpmBadge.module.scss
```diff
@@ -0,0 +1,28 @@
+:local {
+  .badge {
+    color: #fff;
+    font-family: Verdana,DejaVu Sans,sans-serif;
+    font-size: 0.7857rem;
```
emilyrohrbough: Should we have been overriding font-size and font-family ?

packages/terra-doc-template/src/DocTemplate.jsx
```diff
@@ -17,6 +18,14 @@ const propTypes = {
    * The given component's npm package name.
    */
   packageName: PropTypes.string,
+  /**
+   * The url for the provided package if not in npm. This will be ignored if packageVersion is not set.
+   */
+  packageUrl: PropTypes.string,
```
emilyrohrbough: What url do we expect people to provide internally? Will this differ from the component source code URL we already supply on the page? 

What if someone want so to render the badge with a version but doesn't care to supply a link?

emilyrohrbough:
>  It also addresses an issue with Voiceover being able to jump out of focus trapped modals to the badge in some circumstances.

Was voice over functionally verified for these changes?

packages/terra-doc-template/src/NpmBadge.module.scss
amichaelparker: The normal npm badge also uses its own font size and family. This allows the non-badgen version of the badge to maintain consistency with the dynamic version.

packages/terra-doc-template/src/DocTemplate.jsx
amichaelparker: They might want to point it to an internal source, or they may just want it to loop right back to the same page the badge is found on to simply not have a 404 badge.

If they supply a version without a url, it falls back to passing the version to the normal npm link, pointed at that version of the package. If that doesn't exist, the url prop gives them the option to point it wherever they desire.

PR #2574, opened 2019-08-08
Remove react and react-dom as dependencies from terra-section-header

dv297:
### Summary
When consuming terra-section-header, a warning appears stating that "Multiple versions of react were found". I looked around and didn't see any other components declaring React as a dependency, only as a peer dependency, which makes sense. 

### Additional Details
Error
```
ERROR in react
  Multiple versions of react found:
    16.8.6 ./~/react
    16.9.0 ./~/terra-section-header/~/react
```


bjankord:
Thanks @dv297 for the contribution, looks like this slipped in by accident and we didn't notice it.

PR #2577, opened 2019-08-09
Update code owners

bjankord:
### Summary
Update code owners


bjankord:
Was curious if adding code owners in this PR would prompt them on this review or if this needs to be merged before they get auto added to PRs to review for packages they are code owners on. Looks like it may be the later.

Issue #2294, opened 2019-03-21
Update label to be string prop-type on terra-form-field


neilpfeiffer:
# Feature Request

## Description
Form Fields currently allow the label to be a `node`, which potentially breaks the display when injecting other non-inline-only text elements (`terra-text`) or has the potential to completely break accessibility if something non-text based would be passed in.

https://github.com/cerner/terra-core/blob/a367ca5fd76861a8c251f1a1bfd20ae542fdbddf/packages/terra-form-field/src/Field.jsx#L50

\
\
Suggested fix is to remove ability for labels to be nodes, and switch to string only, similar to [form-fieldset](https://github.com/cerner/terra-core/blob/a367ca5fd76861a8c251f1a1bfd20ae542fdbddf/packages/terra-form-fieldset/src/Fieldset.jsx#L23). 

`Major Version Bump` -- Will break any consumers who have injected other components and elements into form-field labels as non-standard usages.

## Components affected
form-field
terra-form-checkbox/lib/CheckboxField
terra-form-input/lib/InputField
terra-form-radio/lib/RadioField
terra-form-select/lib/SelectField
terra-form-textarea/lib/TextareaField

JakeLaCombe:
Is there a way we can communicate the design standard to other developers? I've had requests come in wanting to decorate text such as in the option node. If we have a doc that we can link them to that indicates text should always be styled like this, it would help with the migration.

bjankord:
Planning to hold off on this change until around the end of Q2/beginning of Q3 to minimize major version bumps in the Terra UI ecosystem.

PR #2583, opened 2019-08-13
[form-textarea] Added label prop Part-1

### LINKED ISSUE #2570: [axe-core-warning] Ensures every form element has a label
# Bug Report
Accessibility is violated on a few TextArea test pages and had a consumer tell me that they ran into this issue while running their _wdio_ tests as well.

<details>
 <summary>WDIO Axe Error Report</summary>

```
"description": "Ensures every form element has a label",
  "help": "Form elements must have labels",
  "helpUrl": "https://dequeuniversity.com/rules/axe/3.2/label?application=axeAPI",
  "id": "label",
  "impact": "critical",
  "nodes": [
    {
      "all": [],
      "any": [
        {
          "data": null,
          "id": "aria-label",
          "impact": "serious",
          "message": "aria-label attribute does not exist or is empty",
          "relatedNodes": []
        },
        {
          "data": null,
          "id": "aria-labelledby",
          "impact": "serious",
          "message": "aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty",
          "relatedNodes": []
        },
        {
          "data": null,
          "id": "implicit-label",
          "impact": "critical",
          "message": "Form element does not have an implicit (wrapped) <label>",
          "relatedNodes": []
        },
        {
          "data": null,
          "id": "explicit-label",
          "impact": "critical",
          "message": "Form element does not have an explicit <label>",
          "relatedNodes": []
        },
        {
          "data": null,
          "id": "non-empty-title",
          "impact": "serious",
          "message": "Element has no title attribute or the title attribute is empty",
          "relatedNodes": []
        }
      ],
      "failureSummary": "Fix any of the following:\n  aria-label attribute does not exist or is empty\n  aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty\n  Form element does not have an implicit (wrapped) <label>\n  Form element does not have an explicit <label>\n  Element has no title attribute or the title attribute is empty",
      "html": "<textarea id=\"auto-resizable\" cols=\"2\" rows=\"2\" class=\"Textarea-module__textarea___2LAQp\" style=\"min-height: 48px; caret-color: transparent;\">Default Value</textarea>",
      "impact": "critical",
      "none": [],
      "target": [
        "#auto-resizable"
      ]
    }
  ],
  "tags": [
    "cat.forms",
    "wcag2a",
    "wcag332",
    "wcag131",
    "section508",
    "section508.22.n"
  ]

```
</details>

## Steps to Reproduce

1. Go to https://engineering.cerner.com/terra-core/tests/terra-form-textarea/form-textarea/auto-resizable-textarea
2. Run the axe test.

### Additional Information
It looks like the rule is being ignored in the _wdio_ tests [1] for TextArea. Is this for a reason?

[1] https://github.com/cerner/terra-core/blob/21630dbf90a05bb93567aeba3a687843ad142246/packages/terra-form-textarea/tests/wdio/form-textarea-spec.js#L1-L3

pranav300:
### Summary
Added `ariaLabel` prop to remove accessibility violations.
Partially resolves #2570


packages/terra-form-textarea/src/Textarea.jsx
```diff
@@ -79,6 +79,10 @@ const propTypes = {
    * functionality such as managing focus, selection, or animations.
    */
   refCallback: PropTypes.func,
+  /**
+   * Value to set for the aria-label.
+   */
+  label: PropTypes.string.isRequired,
```
StephenEsser: Adding a required prop will be a breaking change, this will require a major version bump

packages/terra-form-textarea/src/TextareaField.jsx
```diff
@@ -168,6 +168,7 @@ const TextareaField = (props) => {
         id={inputId}
         onChange={onChange}
         value={value}
+        label={label}
```
StephenEsser: The label for the TextareaField should be handled by the Field wrapper which creates a native `<label>`. Do we always want to supply both? Making the label required within the base Textarea may force our hand.

packages/terra-form-textarea/src/Textarea.jsx
StephenEsser: Thoughts on making this prop `ariaLabel`?

packages/terra-form-textarea/src/TextareaField.jsx
bjankord: We want the aria-label on the input to be non-required. In the field composition, we want the label to be required.

packages/terra-form-textarea/src/Textarea.jsx
bjankord: Let's make this non-required and change the prop to ariaLabel. If people use the TextareaField, we require a label for that component and map it to a label element so the aria-label on the textarea wouldn't be needed. Its only if you render the `<Textarea />` without our `<Field />` composition that the textarea needs to be labelled.

packages/terra-form-textarea/src/TextareaField.jsx
pranav300: Fixed here https://github.com/cerner/terra-core/pull/2583/commits/cba49d48d985f385a4e210fc459814560d02c3b2.

packages/terra-form-textarea/src/Textarea.jsx
pranav300: fixed here https://github.com/cerner/terra-core/pull/2583/commits/cba49d48d985f385a4e210fc459814560d02c3b2.

packages/terra-form-textarea/tests/jest/Textarea.test.jsx
```diff
@@ -122,29 +123,29 @@ it('should not set the textarea to resizable when viewed on browser with msMaxTo
...
   const wrapper = render(textarea);
   expect(wrapper).toMatchSnapshot();
 });
 
 it('should set the ref when refCallback is passed into the component', () => {
   const refCallback = jest.fn();
 
-  const textarea = <Textarea refCallback={refCallback} />;
+  const textarea = <Textarea ariaLabel="label" refCallback={refCallback} />;
   const wrapper = mount(textarea);
```
ShettyAkarsh: No need to update all the tests with this new prop i guess . as this is not the required one here.

packages/terra-form-textarea/src/terra-dev-site/doc/example/Resizable.jsx
```diff
@@ -9,6 +9,7 @@ const ResizableExample = () => (
       size="small"
       defaultValue="I automatically resize as you type in more information (except in mobile)."
       id="resizable"
+      ariaLabel="text-area-label"
```
ShettyAkarsh: Even the docs i would rather update only the some examples with this change. let say small, large and medium size examples.

packages/terra-form-textarea/src/Textarea.jsx
nramamurth: @bjankord 

> Its only if you render the <Textarea /> without our <Field /> composition that the textarea needs to be labelled.

If accessibility is violated without it, then shouldn't it be a required prop?

packages/terra-form-textarea/src/terra-dev-site/doc/example/Resizable.jsx
pranav300: Fixed here https://github.com/cerner/terra-core/pull/2583/commits/0aa2ae3fc943878309d490d58b2707b61e888845.

packages/terra-form-textarea/tests/jest/Textarea.test.jsx
pranav300: Fixed here 0aa2ae3.

packages/terra-form-textarea/src/Textarea.jsx
bjankord: @nramamurth You are correct, when using `<Textarea />` without a label would be an accessibility violation. In a future release, we can make it a required prop, at this time though, we don't want to introduce a breaking change to this package.  

One piece of complexity to work out with making `ariaLabel` a required prop is around a type compatibility issue with the `<TextareaField />` `label` prop. The TexareaField component has [a required `label` prop, however it is set to a node type.](https://github.com/cerner/terra-core/blob/master/packages/terra-form-textarea/src/TextareaField.jsx#L15), where as for `ariaLabel` on the Textarea, we want/need to use a string type. 

If we make `<Texarea />` have a required `ariaLabel`, we need to find some way to get the `label` from `<TextareaField label={<span>Test label</span>}>` down to the `<Textarea />` `ariaLabel` prop. The solution for this is likely [introduce a breaking change](https://github.com/cerner/terra-core/issues/2294) for `<TextareaField />` to take a string for the `label` prop instead of a node so we can cleanly pass that string down to the `<Textarea />` `ariaLabel` prop. 

We'll also want to test with screen readers and see how it reads `<TextareaField label="Test label" />` when that component has both a `<label />` element mapping to the `textarea` and an `ariaLabel` on the `textarea` element.

PR #2589, opened 2019-08-14
Example/Tests file clean up

bjankord:
### Summary
Update config to turn off the following rules in terra-dev-site directory:

```
"import/no-extraneous-dependencies"
"import/no-unresolved"
"import/extensions"
```

Removed all comments disabling these rules.

Update import paths from 
```
import ActionFooter from 'terra-action-footer/lib/ActionFooter';
```
to
```
import ActionFooter from 'terra-action-footer';
```


bjankord:
> The world is a slightly better place now

4fd

bjankord:
@mjhenkes Nah, I just did find and replace in each package.

PR #2592, opened 2019-08-15
[terra-form-select] Add AT announcement for option onClick and unselection via Backspace key for multiple variants

### LINKED ISSUE #2590: [terra-form-select] MultiSelect and TagSelect onClick Option and Backspace Don't Handle AT Announcements Properly
# Bug Report

## Description

Both multiselect variants of terra-form-select (Multi and Tag) do not announce on assistive technology that an option has been unselected using clicks, taps, or backspace, while unselecting an option with the `Enter` key does.

## Steps to Reproduce

For both [MultiSelect](https://engineering.cerner.com/terra-ui/components/terra-form-select/form-select/multi-select) and [TagSelect](https://engineering.cerner.com/terra-ui/components/terra-form-select/form-select/tag-select),

1. navigate to the component's page  and scroll to the examples (the first one is sufficient to demo)
2. turn on AT, e.g JAWS, mac VoiceOver, or ios VoiceOver (most notable on phones since there's no workflow for `ArrowUp`/`ArrowDown` and `Enter`)
3. Click (or tap) on an option to select, it'll say `{option.display} Selected`
4. Click or tap on on the same option, or press backspace, it will just say `{option.display}` or nothing

## Expected Behavior

When clicking on a selected option, it should AT announce `{option.display} Unselected` just like `ArrowUp`/`ArrowDown` and `Enter` workflow.

## Possible Solution

We just need to update the `visuallyHiddenComponent`'s `innerHTML`, identical to [here](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/multiple/Menu.jsx#L373), but in the [MultiSelect's Menu.handleOptionClick](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/multiple/Menu.jsx#L392) and [TagSelect's Menu.handleOptionClick](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/tag/Menu.jsx#L337)

## Environment

* Component Name and Version: terra-form-select, MultiSelect and TagSelect
* Using Assistive Technology, eg. JAWS, mac VoiceOVer, or iOS VoiceOver

## @ Mentions

@nickpith @bjankord

dev-cprice:
### Summary
closes #2590

Add assistive technology announcement in `MultiSelect` and `TagSelect` for option `onClick` (selection & unselection) and unselection via the `Backspace` key. 

[deployment](https://terra-core-deployed-pr-2592.herokuapp.com/)

@cerner/terra

[License]: ../LICENSE


dev-cprice:
WDIO tests passing locally, failing on travis
image

🙃🙃🙃

packages/terra-form-select/src/multiple/Frame.jsx
```diff
@@ -332,7 +335,18 @@ class Frame extends React.Component {
       event.preventDefault();
       this.openDropdown(event);
     } else if (keyCode === KeyCode.KEY_BACK_SPACE && !this.state.searchValue && value.length > 0) {
-      this.props.onDeselect(value[value.length - 1]);
+      const unselectedTxt = intl.formatMessage({ id: 'Terra.form.select.unselected' });
+      const lastOptionValue = value[value.length - 1];
+      const lastOption = React.Children.toArray(children).find(child => child.props.value === lastOptionValue);
```
StephenEsser: I wasn't able to test this on the deployed examples because the values and displays are identical. But I think we may need to use the findByValue utility to search for the appropriate child, specifically for the opt groups. This will do a flat search through the opt group but it won't search through the options of each group.

dev-cprice: Ah, good catch. Forgot about the OptGroups. Will fix that

packages/terra-form-select/src/tag/Menu.jsx
```diff
@@ -340,13 +340,27 @@ class Menu extends React.Component {
...
-    if (MenuUtil.includes(value, option.props.value)) {
-      onDeselect(option.props.value, option);
+
+    const selectedTxt = intl.formatMessage({ id: 'Terra.form.select.selected' });
+    const unselectedTxt = intl.formatMessage({ id: 'Terra.form.select.unselected' });
+    const shouldUnselectOption = MenuUtil.includes(value, option.props.value);
+    const optionATClickText = shouldUnselectOption ? unselectedTxt : selectedTxt;
+
+    if (visuallyHiddenComponent && visuallyHiddenComponent.current) {
+      visuallyHiddenComponent.current.innerHTML = `${option.props.display} ${optionATClickText}`;
```
StephenEsser: This is probably a question for the translations team or someone more familiar with translations, but the grammar in different languages may not always be in this order. 

`Selected` for example has a parameterized version: https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/translations/en-US.json#L8

dev-cprice: I agree. When I copied this from https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/multiple/Menu.jsx#L375, I thought it was odd that we weren't translating like:

```js
const unselectedText = intl.formatMessage({ id: 'Terra.form.select.unselected',  { text: option.props.display } });
```

dev-cprice: Translations added and fix made

packages/terra-form-select/src/multiple/Menu.jsx
```diff
@@ -341,40 +338,25 @@ class Menu extends React.Component {
...
-      /*
-        Detecting if browser is not Safari before updating aria-live as there is some odd behaivor
-        with VoiceOver on desktop, that causes the selected option to be read twice when this is
-        is added to aria-live container.
-        When we shift focus back to select, VoiceOver automatically reads the display text.
-        Using aria-hidden on the display does not prevent VO from reading the display text and so it
-        results in reading the display text followed by reading the aria-live message which is
-        the display text + 'selected'
-        */
-      if (SharedUtil.isSafari()) {
```
dev-cprice: removed this check.. leftover from the refactor (missed it then)

doesn't appear to make a difference, especially since both blocks run the same thing

packages/terra-form-select/src/multiple/Menu.jsx
```diff
@@ -341,40 +338,25 @@ class Menu extends React.Component {
...
-        this.props.visuallyHiddenComponent.current.innerText = `${option.props.display} ${unselectedTxt}`;
-      } else {
-        this.props.visuallyHiddenComponent.current.innerText = `${option.props.display} ${unselectedTxt}`;
+      if (visuallyHiddenComponent && visuallyHiddenComponent.current) {
+        visuallyHiddenComponent.current.innerHTML = intl.formatMessage({ id: 'Terra.form.select.unselectedText' }, { text: option.props.display });
+      }
+
+      if (onDeselect) {
```
StephenEsser: Now that the frames are variant specific these callback functions (onSelect / onDeselect) could probably be marked as required. This would remove a lot of the unnecessary guard checks. The Menu should be a private implementation but if you're concerned about passivity this could be a change for another time. But the callback functions should be considered guaranteed here, even if they are not marked as required.

dev-cprice: True, they are guaranteed to be here, given that we always pass `handleX` from above

It ain't public, so I'm fine with making them `.isRequired`

dev-cprice: Actually, this isn't entirely true..

https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/multiple/Frame.jsx#L582-L594

`onDeselect` is passed from `props`, which has a defaultValue of `undefined` [here](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/multiple/Frame.jsx#L106)

So, in order to be confident that they are present, I say we either:

- [ ] add defaultProps of `() => {}` for all callbacks
- [ ] keep truthy checks for callbacks
- [ ] add bound wrapper methods in `Frame` (later, the individual variant implementations)

The last option is similar to the first, but would end up basically adding more noise for the same effect, with maybe a lil more safety e.g.:

```jsx
handleOnDeselect = (value) => {
  const { onDeselect } = this.props;

  if (onDeselect) {
    onDeselect(value);
  }
}
```

bjankord: > We can tackle it in another issue. 
:+1: I'd leave the code as-is. We can talk about it in our weekly sync but I don't want to derail this PR.

StephenEsser: These functions are guaranteed to exist from the base Multiple component.

https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/MultiSelect.jsx#L203-L204

dev-cprice: This can/will be address when the Frame and Selects are flattened then.

StephenEsser: When the frame is flattened this will become more obvious

mmalaker:
Functional verification completed. The multiselect and tag select examples now function as expected with screen readers. One issue noted is that tapping on tags within the text field on mobile does not delete them. That will be addressed as a separate issue.

PR #2601, opened 2019-08-20
[terra-icon] Fix BiDi.

### LINKED ISSUE #2599: [terra-icon] isBidi no longer functioning
# Bug Report
The IconBase prop of isBidi is not longer funcitoning.

## Description
The property of isBidi either passed or defaulted isn't being honored within icons.

## Steps to Reproduce
1. Navigate to terra-ui site, go to themable icons page, switch to rtl mode.

## Expected Behavior
Icon with isBidi set should flip along the vertical axis.

## Possible Solution
[dir='rtl'] .icon.is-bidi  -> [dir=rtl] .icon.is-bidi

nramamurth:
### Summary
Closes #2599.

Note: double-quotes fixes it too but Linter throws up 🤔


packages/terra-icon/src/Icon.module.scss
```diff
@@ -16,7 +16,7 @@
 
   // We need dir='rtl'
```
mjhenkes: nit: update this comment

PR #2602, opened 2019-08-20
Added Tooltip for IconConsultInstructionsForUse 

### LINKED ISSUE #2575: [terra-icon] Add Tooltip text to the Icon IconConsultInstructionsForUse
# Feature Request

## Description

Tooltip Mentioning "Electronic Information for Use Icon." needs to be added for the icon 
IconConsultInstructionsForUse. as per the UX documentation this tooltip needs to be added for to support accessibility

## @ Mentions
@ryanthemanuel @neilpfeiffer

supreethmr:
### Summary
Resolves #2575 

### Additional Details
Added `span` around `svg` to display tooltip for Icon.


packages/terra-icon/src/icon/IconConsultInstructionsForUse.jsx
```diff
@@ -1,14 +1,21 @@
 /* eslint-disable */
 import React from 'react';
```
bjankord: This file is auto-generated, any manual edits made to it will be overwritten next time [the generate-icon npm script is run](https://github.com/cerner/terra-core/blob/master/packages/terra-icon/package.json#L45). Another way to do this is to detect the ConsultInstructionsForUse icon while generating icons and use a custom template for generating that icon instead of [the default template](https://github.com/cerner/terra-core/blob/d064b2fc0856d5315ab573e68fcf0d228f1ff890/packages/terra-icon/scripts/src/generate-icon/template.txt).

packages/terra-icon/src/icon/IconConsultInstructionsForUse.jsx
```diff
@@ -1,14 +1,21 @@
...
 import IconBase from '../IconBase';
+import { FormattedMessage } from 'react-intl';
```
StephenEsser: react-intl will need to be added to the Icon package. I believe the current strategy is to add react-intl as a peerDependency

packages/terra-icon/src/icon/IconConsultInstructionsForUse.jsx
ShettyAkarsh: @bjankord Just curious, like in what scenario we will be regenerating the existing icons ? like is it during the creation of new icons ?

supreethmr: Added new Template for Icons having tooltip and made that template generic for all the theme icons : dc294a9ec3f4f8825ecd45093a01510045db3e4a

packages/terra-icon/src/icon/IconConsultInstructionsForUse.jsx
supreethmr: updated here : https://github.com/cerner/terra-core/blob/dc294a9ec3f4f8825ecd45093a01510045db3e4a/packages/terra-icon/package.json#L32

packages/terra-icon/src/icon/IconConsultInstructionsForUse.jsx
bjankord: @ShettyAkarsh When the design team releases a new version of https://github.com/cerner/one-cerner-style-icons, we run a script that pulls in the SVGs and a CSV mapping aliases to the SVGs and use that to generate the icons. 

In that process, we clean out the existing `packages/terra-icon/src/icon` directory and create a new empty `packages/terra-icon/src/icon` directory before writing all of the in all the generated icon files to it.

* https://github.com/cerner/terra-core/blob/master/packages/terra-icon/scripts/src/generate-icon/removeMakeDirectories.js
* https://github.com/cerner/terra-core/blob/master/packages/terra-icon/scripts/src/generate-icon/index.js#L11

packages/terra-icon/scripts/src/generate-icon/renderJsx.js
```diff
@@ -19,25 +22,21 @@ const staticIcons = iconData
...
   } else {
-    fs.readFile(themeIconTemplatePath, 'utf-8', (error, text) => {
-      if (error) {
-        reject(error);
-      } else {
-        const compiled = template(text);
-        resolve(new ReactIcon(icon.name, compiled({ icon })));
-      }
-    });
+    templatePath = (iconsWithTooltip.includes(icon.name)) ? themeIconWithTooltipTemplatePath : themeIconTemplatePath;
```
nramamurth: This logic is a bit confusing to me, does this mean that the static icons cannot/should not support tooltip?

supreethmr: eIfu icon is themable icon and uses [theme template](https://github.com/cerner/terra-core/blob/dc294a9ec3f4f8825ecd45093a01510045db3e4a/packages/terra-icon/scripts/src/generate-icon/template.txt) so created tooltip template using theme template. If any requirement comes for static icons in future then we might need to create one more tooltip template for static Icons too.

bjankord: Let's rename `iconsWithTooltip` to `themeIconsWithTooltip` for clarity,  and then we can add logic for static icons if any requirement for static icons comes up in the future.

supreethmr: renamed : 4fb01f556396ec875479704ccff8be8bf3dbea43

supreethmr:
Waiting for Translations for Icon Tooltip text : https://jira.cerner.com/browse/TRANS-6969

PR #2621, opened 2019-09-03
[terra-avatar] Changed fallback to initials from user icon for avatar

### LINKED ISSUE #2559: [terra-avatar] Allow Fallback to Initials Display
# Feature Request

## Description
Currently, the image fallback logic of the avatar displays a generic icon. However, this fallback does not differentiate between different 'Avatars' if multiple images fail to load.
If the image fails to load and initials are provided - the avatar should fallback to the initials display instead (or at least be optional).

supreethmr:
### Summary
Resolves #2559 

### Additional Details
generateImage has been modified to fallback to initials when provided for invalid image.

### Deployment Links
https://terra-core-deployed-pr-2621.herokuapp.com/components/terra-avatar/avatar/avatar
https://terra-core-deployed-pr-2621.herokuapp.com/raw/tests/terra-avatar/avatar/avatar/invalid-image-avatar-with-initials


packages/terra-avatar/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed 
+* Changed generateImage function to Fallback to Initials if Provided.
```
nramamurth: Maybe saying 'Changed the load behavior such that if the image fails to load and initials are provided, the avatar falls back to the initials display.' is more clear to a consumer.

packages/terra-avatar/src/variants/Avatar.jsx
```diff
@@ -86,8 +86,8 @@ class Avatar extends React.Component {
     let avatarContent;
 
     if (image) {
-      avatarContent = generateImage(image, alt, isAriaHidden, AVATAR_VARIANTS.USER, this.handleFallback);
-    } else if (initials && (initials.length === 1 || initials.length === 2)) {
+      avatarContent = generateImage(image, alt, isAriaHidden, AVATAR_VARIANTS.USER, this.handleFallback, initials);
+    } else if (initials && initials.length <= 2) {
```
nramamurth: Just curious, could these redundant checks be probably avoided if the default value of `initials` is set to an empty string instead of `undefined` and if the length of the `initials` prop is custom-validated during datatype validation?

nramamurth:
I think that the docs need to be updated. For instance, 

[1] https://github.com/cerner/terra-core/blob/master/packages/terra-avatar/docs/avatar.md
[2] https://github.com/cerner/terra-core/blob/master/packages/terra-avatar/src/terra-dev-site/doc/avatar/Avatar.b.doc.jsx#L33 - is it still the default fallback icon?

packages/terra-avatar/src/common/AvatarUtils.jsx
```diff
@@ -60,17 +60,21 @@ const getColorVariant = (hashValue) => {
...
  */
-const generateImagePlaceholder = (alt, isAriaHidden, variant) => {
-  const avatarIconClassNames = cx(['icon', variant]);
+const generateImagePlaceholder = (alt, isAriaHidden, variant, initials) => {
+  if (initials && initials.length <= 2) {
+    const avatarTextClassNames = cx('initials');
+    return <span className={avatarTextClassNames} alt={alt} aria-label={alt} aria-hidden={isAriaHidden}>{initials.toUpperCase()}</span>;
+  }
 
+  const avatarIconClassNames = cx(['icon', variant]);
```
yuderekyu: Based on new UX requirements, we should remove this icon logic and default the placeholder to be initials.

packages/terra-avatar/src/variants/Avatar.jsx
```diff
@@ -86,8 +86,8 @@ class Avatar extends React.Component {
...
       const avatarTextClassNames = cx('initials');
       avatarContent = <span className={avatarTextClassNames} alt={alt} aria-label={alt} aria-hidden={isAriaHidden}>{initials.toUpperCase()}</span>;
     } else {
```
yuderekyu: we can get rid of the else block now

packages/terra-avatar/src/common/AvatarUtils.jsx
```diff
@@ -60,17 +60,21 @@ const getColorVariant = (hashValue) => {
 /**
  * Render placeholder.
  */
-const generateImagePlaceholder = (alt, isAriaHidden, variant) => {
-  const avatarIconClassNames = cx(['icon', variant]);
+const generateImagePlaceholder = (alt, isAriaHidden, variant, initials) => {
```
yuderekyu: want to update this to accept an object containing these properties? something i learned recently :)
```suggestion
const generateImagePlaceholder = ( 
  { alt, isAriaHidden, variant, initials }
) => {
```

packages/terra-avatar/src/common/AvatarUtils.jsx
```diff
@@ -60,17 +60,21 @@ const getColorVariant = (hashValue) => {
...
+  const avatarIconClassNames = cx(['icon', variant]);
   return <span className={avatarIconClassNames} role="img" aria-label={alt} alt={alt} aria-hidden={isAriaHidden} />;
 };
 
 /**
  * Render image with placeholder.
  */
-const generateImage = (image, alt, isAriaHidden, variant, handleFallback) => {
-  const icon = generateImagePlaceholder(alt, isAriaHidden, variant);
+const generateImage = (image, alt, isAriaHidden, variant, handleFallback, initials) => {
```
yuderekyu: same object update here

packages/terra-avatar/src/variants/Avatar.jsx
```diff
@@ -86,8 +86,8 @@ class Avatar extends React.Component {
     let avatarContent;
 
     if (image) {
-      avatarContent = generateImage(image, alt, isAriaHidden, AVATAR_VARIANTS.USER, this.handleFallback);
-    } else if (initials && (initials.length === 1 || initials.length === 2)) {
+      avatarContent = generateImage(image, alt, isAriaHidden, AVATAR_VARIANTS.USER, this.handleFallback, initials);
+    } else if (initials && initials.length <= 2) {
       const avatarTextClassNames = cx('initials');
       avatarContent = <span className={avatarTextClassNames} alt={alt} aria-label={alt} aria-hidden={isAriaHidden}>{initials.toUpperCase()}</span>;
```
yuderekyu: can we call `generateImagePlaceholder` here instead?

packages/terra-avatar/src/common/AvatarUtils.jsx
supreethmr: I was bit confused with new Ux requirements for removing fallback icon from avatar. since there   was no response on the comment  I made in [Issue page](https://github.com/cerner/terra-core/issues/2559#issuecomment-526664922) I Proceeded with logic which adds fallback to initials without changing the existing default Icon fallback logic.

packages/terra-avatar/src/common/AvatarUtils.jsx
supreethmr: Do I need to create object containing these properties [here](https://github.com/cerner/terra-core/blob/9b6a5d4ec8ec4ced52e8cacbcb62be86f2a35395/packages/terra-avatar/src/variants/Avatar.jsx#L87) like below and to pass that object to generateImagePlaceholder() function..? 

```javascript
const avatarAttributes = {
   alt : 
   isAriaHidden :
   variant : 
   initials : 
}
```

packages/terra-avatar/src/common/AvatarUtils.jsx
yuderekyu: Let's sync with UX on this

cc @neilpfeiffer @jeremyfuksa

packages/terra-avatar/src/common/AvatarUtils.jsx
yuderekyu: yup!

mmalaker:
During functional verification the initials fallback is working fine, but on JAWS on Chrome and on VO in iOS, the screen reader reads initials only and not the name, which should be available to the screen reader. VO on Safari desktop reads the avatar name correctly.

packages/terra-avatar/CHANGELOG.md
supreethmr: Updated [here](https://github.com/cerner/terra-core/blob/565f1417aaeb24a8bcc372b80c65ba24ebe360ab/packages/terra-avatar/CHANGELOG.md)

packages/terra-avatar/src/variants/Avatar.jsx
supreethmr: Redundant checks have been removed and validating this only at generateImagePlaceholder function : [here](https://github.com/cerner/terra-core/pull/2621/commits/565f1417aaeb24a8bcc372b80c65ba24ebe360ab)

supreethmr:
> I think that the docs need to be updated. For instance,
> 
> [1] https://github.com/cerner/terra-core/blob/master/packages/terra-avatar/docs/avatar.md
> [2] https://github.com/cerner/terra-core/blob/master/packages/terra-avatar/src/terra-dev-site/doc/avatar/Avatar.b.doc.jsx#L33 - is it still the default fallback icon?

Updated [Doc files](https://github.com/cerner/terra-core/blob/565f1417aaeb24a8bcc372b80c65ba24ebe360ab/packages/terra-avatar/docs/avatar.md)

packages/terra-avatar/src/common/AvatarUtils.jsx
supreethmr: Got approval from @jeremyfuksa for having fallback only when initials are provided : [conversation-link](https://github.com/cerner/terra-core/issues/2559#issuecomment-529600379)

packages/terra-avatar/src/variants/Avatar.jsx
supreethmr: else has been removed since this is now being handled in generateImagePlaceholder() function

packages/terra-avatar/src/common/AvatarUtils.jsx
supreethmr: changed to object params : [here](https://github.com/cerner/terra-core/blob/565f1417aaeb24a8bcc372b80c65ba24ebe360ab/packages/terra-avatar/src/variants/Avatar.jsx#L88-L101)

packages/terra-avatar/src/common/AvatarUtils.jsx
supreethmr: updated [here](https://github.com/cerner/terra-core/blob/565f1417aaeb24a8bcc372b80c65ba24ebe360ab/packages/terra-avatar/src/variants/Avatar.jsx#L88-L101)

packages/terra-avatar/src/variants/Avatar.jsx
supreethmr: yes now [generateImagePlaceholder](https://github.com/cerner/terra-core/blob/565f1417aaeb24a8bcc372b80c65ba24ebe360ab/packages/terra-avatar/src/variants/Avatar.jsx#L100) function can return initials too.

packages/terra-avatar/src/common/AvatarUtils.jsx
```diff
@@ -60,18 +60,22 @@ const getColorVariant = (hashValue) => {
 /**
  * Render placeholder.
  */
-const generateImagePlaceholder = (alt, isAriaHidden, variant) => {
-  const avatarIconClassNames = cx(['icon', variant]);
+const generateImagePlaceholder = (avatarParams) => {
+  if (avatarParams.initials && avatarParams.initials.length <= 2) {
```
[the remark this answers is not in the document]
supreethmr: would like to know why :) does de-structuring objects have any advantages..?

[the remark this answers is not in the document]
yuderekyu: That's a good reason to update - readability contributes to maintainability. This code base receives contributions from many people, not just one. Not opting to write accessible code whenever possible is a disservice to others who may or will have to understand this logic.

We should also strive to store frequently used object members into local variables. Take `avatarParams.initials`. We reference that three separate times. With destructuring, we just reference the local variable `initials`. This is terse, readable, and (probably) slightly more performant.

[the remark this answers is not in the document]
supreethmr: Thank you @yuderekyu  for the explanation : updated [here](https://github.com/cerner/terra-core/blob/706b5915a650187e747e00b562faf81a2525c414/packages/terra-avatar/src/common/AvatarUtils.jsx#L64-L88)

mmalaker:
Regarding my previous [comment](https://github.com/cerner/terra-core/pull/2621#issuecomment-529643813) I checked the screen reader behavior again. Now none of the screen readers read the person name. They only read the initials. Even though visually we are falling back to the initials, the important information is the person name. The person name should be available even if an avatar image is not. Person name should always be what a screen reader reads back when focus is on the avatar.

supreethmr:
> Regarding my previous [comment](https://github.com/cerner/terra-core/pull/2621#issuecomment-529643813) I checked the screen reader behavior again. Now none of the screen readers read the person name. They only read the initials. Even though visually we are falling back to the initials, the important information is the person name. The person name should be available even if an avatar image is not. Person name should always be what a screen reader reads back when focus is on the avatar.

are you referring to the value in `alt` prop as person name here ..?
It reads text passed into ```alt``` prop when initials are not provided. but when Initials were provided I think even in default implementation it still reads Initials not the text provided in `alt` prop.
So would like to know the steps to re-create the difference you mentioned in above comment.

mmalaker:
@supreethmr correct. I'm referring to the alt text, which in an avatar would be a person or patient name. Here is some [guidance](https://webaim.org/techniques/alttext/) on alt text. The trick is that we have a mix of text and non-text content being presented in the fallback avatar. That probably confuses screen readers. I think @bjankord may have some ideas for how we can get this to work better.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
@mmalker I have tested with screen reader it reads both initial and the person name in VO. can you test it again.

supreethmr:
> @supreethmr correct. I'm referring to the alt text, which in an avatar would be a person or patient name. Here is some [guidance](https://webaim.org/techniques/alttext/) on alt text. The trick is that we have a mix of text and non-text content being presented in the fallback avatar. That probably confuses screen readers. I think @bjankord may have some ideas for how we can get this to work better.

@mmalaker I have made changes to avatar to make person name to be accessible for Screen Readers.

mmalaker:
During functional verification JAWS is working well, but both VoiceOver on Safari and VO on iOS now skip over the avatar completely. This is going to need further investigation.

yuderekyu:
@mmalaker how are you navigating to component? Tab seems to skip over, but if I go into the content area with `shift-ctrl-opt-down-arrow`, I'm able to focus and announce the avatars. I'm under the impression tab is for links or form controls, is that wrong?

packages/terra-avatar/docs/avatar.md
```diff
@@ -1,6 +1,6 @@
 # Terra Avatar
 
-The `Avatar` variant represents a person - it displays an image or initials in a circular frame. If neither are provided, a fallback user icon displays. This is the default export of the `terra-avatar` package.
+The `Avatar` variant represents a person - it displays an image or initials in a circular frame. If valid image is not provided and initials is provided, then fallback initial displays. If neither are provided, a fallback user icon displays. This is the default export of the `terra-avatar` package.
```
mjhenkes: Based on this description, is this still a breaking change? Initials are not required and if not supplied the component still has the default behaviour?

supreethmr: Yes but if initials are supplied fallback will happen to initials instead of user Icon which was not the behaviour  before these code changes. So I Think it is a breaking change.

mmalaker:
Functional verification completed. The avatar fallback is working as designed. Confirmed that using the VO commands: control + option + R/L arrow, will enable VoiceOver on desktop Safari to read out the name of the avatar image. Some accessibility issues remain, however. VoiceOver on iOS navigation still skips over the avatar. Also, we need to determine user expectations regarding when the VO commands are needed. Most page elements are accessible by using the arrow keys only. These accessibility issues will be addressed separately.

packages/terra-avatar/docs/avatar.md
mjhenkes: I don't see that as breaking at all. The api does not change and consumers would get the behavior we expect.

mjhenkes: I see you are setting initials to be required. This paragraph needs to be updated.

```suggestion
The `Avatar` variant represents a person - it displays an image or initials in a circular frame. If valid image is not provided then the avatar falls back to displaying initials. This is the default export of the `terra-avatar` package.
```

supreethmr: > I don't see that as breaking at all. The api does not change and consumers would get the behavior we expect.

In scenario : user would be expecting avatar to display user-icon when image fails to load but with this code changes it will display initials when image fails to load instead of displaying user-Icon. (this was something which was not happening before) so will these not be considered as breaking change..?

PR #2628, opened 2019-09-05
Update wcig-inert path to import directly the ES5 version of the package

bjankord:
### Summary
[wcig-inert recently added a `module` key to their package.json](https://github.com/WICG/inert/commit/c27413c1a06f61a5d798522588238ca02d9f639a).
With our webpack setup, webpack will prefer resolving `module` over `main` keys defined in the package.json. This change makes the wcig-inert import explicitly import the ES5 version of wcig-inert.

### DIFF
--- a/packages/terra-overlay/CHANGELOG.md
+++ b/packages/terra-overlay/CHANGELOG.md
@@ -6,6 +6,8 @@ Unreleased
 ### Added
 * Props table markdown file import added to doc site page
 
+### Changed
+* Update wcig-inert path to import directly the ES5 version of the package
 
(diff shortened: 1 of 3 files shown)


emilyrohrbough:
I thought we were going to make the changes in the webpack configuration?

mjhenkes:
With the change to webpack we don't need this right?

mjhenkes:
Though i suppose, this change would help people not consuming our webpack config.

bjankord:
@mjhenkes Yes, this should resolve the issue for anyone not using terra-toolkit that is using this package.

PR #2635, opened 2019-09-09
[terra-form-select] SearchSelect: Move Input into Dropdown Menu

### LINKED ISSUE #2593: [terra-form-select] [a11y] SearchSelect: Move Input into DropdownMenu as a Sibling to Options for Enhanced iOS VoiceOver Navigation
# Feature Request

## Description

Sub-issue of #2383

Move the `SearchSelect` `input` into the dropdown menu as a sibling of the options so that iOS VoiceOver swipe navigation behaves as expected: focus moves from input to the options list.

## Constraints

- Must be visually identical to current implementation.
- Must be functionally identical to current implementation.
- Should appropriately handle when the menu `isAbove` or not (input is at the top when menu is below, input is at the bottom when  menu is above)
  - Since VoiceOver operates on the DOM order (builds the a11y tree based on DOM order), the DOM order should not change when toggling where the input is rendered
- If enough options are present to show a scroll bar, the input shouldn't scroll with the options

## Extra Information

Exploration of how to solve this:
- [code](https://github.com/dev-cprice/lets-get-flexy)
- [demo](https://dev-cprice.github.io/lets-get-flexy/)

## @ Mentions

@nickpith @bjankord

dev-cprice:
### Summary
closes #2593 

### Deployment Link
https://terra-core-deployed-pr-2635.herokuapp.com/components/terra-form-select/form-select/search-select

@cerner/terra

[License]: ../LICENSE


packages/terra-form-select/src/search/Dropdown.module.scss
```diff
@@ -0,0 +1,22 @@
...
+  .dropdown {
+    background-color: var(--terra-form-select-dropdown-background-color, #fff);
+    border: var(--terra-form-select-dropdown-border, 1px solid #dedfe0);
+    border-bottom-left-radius: var(--terra-form-select-dropdown-border-bottom-left-radius, 3px);
+    border-bottom-right-radius: var(--terra-form-select-dropdown-border-bottom-right-radius, 3px);
+    border-top: 0;
+    border-top-left-radius: 0;
+    border-top-right-radius: 0;
+    box-shadow: var(--terra-form-select-dropdown-box-shadow, 0 3px 12px -5px rgba(0, 0, 0, 0.25));
+  }
```
StephenEsser: Things start to get a little dicey with theme-able variables. Will each variant redefine the dropdown scss or is this an intermediate step while uplifting all variants?

dev-cprice: This is just an intermediate step.

dev-cprice: The only change here from the one in `src/shared/` is that this removes the `overflow: auto;`

packages/terra-form-select/src/search/Frame.jsx
```diff
@@ -173,18 +189,19 @@ class Frame extends React.Component {
...
+      ref: this.selectInputRef,
       onChange: this.handleSearch,
       onFocus: this.handleInputFocus,
       onBlur: this.handleInputBlur,
-      onMouseDown: this.handleInputMouseDown,
       'aria-label': this.ariaLabel(),
       'aria-describedby': ariaDescribedBy,
       'aria-disabled': disabled,
-      'aria-owns': this.state.isOpen ? 'terra-select-menu' : undefined,
+      'aria-owns': this.state.isOpen ? menuId : undefined, // TODO: this may need to be removed..
```
bjankord: Let's determine if this needs to be removed or left in.

packages/terra-form-select/src/search/Frame.jsx
```diff
@@ -199,14 +216,30 @@ class Frame extends React.Component {
...
    * Closes the dropdown.
    */
   closeDropdown() {
+    const { current: selectInput } = this.selectInputRef;
+    const { current: select } = this.selectRef;
+
     this.setState({
       isAbove: false,
-      isFocused: document.activeElement === this.input || document.activeElement === this.select,
+      isFocused: document.activeElement === selectInput || document.activeElement === select, // TODO: fixme
```
bjankord: What needs fixed here?

packages/terra-form-select/src/search/Frame.jsx
```diff
@@ -264,7 +279,10 @@ class Frame extends React.Component {
     }
 
     const { dropdownAttrs, maxHeight } = this.props;
-    const { select, dropdown } = this;
+    // const { dropdown } = this;
```
bjankord: Let's remove commented out code if it isn't needed

packages/terra-form-select/src/search/Frame.jsx
```diff
@@ -273,16 +291,23 @@ class Frame extends React.Component {
    */
   handleBlur(event) {
     // The check for dropdown.contains(activeElement) is necessary to prevent IE11 from closing dropdown on click of scrollbar in certain contexts.
-    if (this.dropdown && (this.dropdown === document.activeElement && this.dropdown.contains(document.activeElement))) {
+    const { isOpen } = this.state;
+    const { onBlur } = this.props;
+
+    // TODO: fixme
```
bjankord: Let's get this cleaned up

packages/terra-form-select/src/search/Frame.jsx
```diff
@@ -326,7 +347,7 @@ class Frame extends React.Component {
   handleMouseDown(event) {
     // Preventing default events stops the search input from losing focus.
     // The default variant has no search input therefore the mouse down gives the component focus.
-    event.preventDefault();
+    // event.preventDefault();
```
bjankord: Let's remove this code if it is no longer needed.

packages/terra-form-select/src/search/Frame.jsx
```diff
@@ -335,7 +356,7 @@ class Frame extends React.Component {
    * @param {event} event - The mouse down event.
    */
   handleInputMouseDown(event) {
-    event.stopPropagation();
+    // event.stopPropagation();
```
bjankord: Let's remove this code if it is no longer needed.

packages/terra-form-select/src/search/Frame.jsx
```diff
@@ -366,11 +387,13 @@ class Frame extends React.Component {
    * Handles the toggle button mouse down events.
    */
   handleToggleButtonMouseDown() {
+    // const { current: selectInput } = this.selectInputRef;
```
bjankord: Let's get this cleaned up.

packages/terra-form-select/src/shared/useScrollToActiveOption.js
```diff
@@ -0,0 +1,26 @@
...
+
+    if (activeOption === null) {
+      return;
+    }
+
+    const dropdownRect = menu.parentNode.getBoundingClientRect();
+    const optionRect = activeOption.getBoundingClientRect();
+
+    if (optionRect.top < dropdownRect.top) {
+      activeOption.scrollIntoView();
```
bjankord: `scrollIntoView` doesn't look to be supported in Safari or IE: https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#Browser_compatibility

We'll need to either polyfill it or use a different method for scrolling to the active option.

packages/terra-form-select/src/search/Frame.jsx
dev-cprice: This is tied to focus management/blur.. working through the logic to determine when the select is focused is complicated since `document.activeElement` seems to consistently be `body` when the menu is closed

debugging this right now

packages/terra-form-select/src/search/Frame.jsx
dev-cprice: working through the blur logic right now

packages/terra-form-select/src/shared/useScrollToActiveOption.js
dev-cprice: This is the same code we used before https://github.com/cerner/terra-core/blob/terra-form-select%405.29.0/packages/terra-form-select/src/search/Menu.jsx#L376-L380

IE and safari don't support `scrollIntoViewOptionsExperimental`, which is the `scrollIntoView({ ... })` API (options hash), but they do support `scrollIntoView()` with/without the boolean `alignToTop` arg

https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView#Parameters

bjankord:
Some behavior seems to be missing when testing in Safari w/ VoiceOver enabled.
Pressing up/down arrow keys highlights the options visually but the highlighted option is not being announced via the screen reader.

Another issue I noticed in Safari when using VoiceOver is that if I focus on the input, then press down (the menu opens), and then press tab, my focus is shifted out of the select but then it is shifted back to the input. I'd expect it to shift out of the select to the next focusable element in this case and not shift focus back into the select.

dev-cprice:
@bjankord yeah, I'll be pushing up some changes soon for the AT announcements

As for the tab event, that _was_ working as you expect, but I'll revisit that. I recently made a change that might've reverted that behavior

dev-cprice:
@bjankord 

So, after looking into the tab behavior with voiceover on, the only way I can reliably get this to work as expected is to do the following in the search `Frame`:

I'm using [tabbable](https://www.npmjs.com/package/tabbable) ([1.7kB minified / 861B gzipped + minified](https://bundlephobia.com/result?p=tabbable@4.0.0)) to achieve manual navigation as relying on the default behavior of the `tab` keypress doesn't work with VoiceOver on, unfortunately... Using tabbable kinda seems like overkill _and_ a code smell since it'll bloat the _*too-big-to-fail*_ select bundle even more (even if by 861B at best, 1.7kB at worst).

Initially, I focused on the `selectInput` and then delegated to the `default behavior` of the tab keypress, but VoiceOver navigates the virtual indicator back to the `selectInput`, even though the next tabbable element has focus. Mutating the `handleMenuRequestClose`'s `event` object's `target` to point to the `selectInput` (as opposed to the `menuInput` and then delegating to default behavior doesn't work either, unfortunately :(

working code in chrome, safari, firefox, and safari + voiceover:
```jsx
  /**
   * Handle the menu onRequestClose event
   */
  handleMenuRequestClose(event) {
    const { keyCode, shiftKey } = event;
    const { current: selectInput } = this.selectInputRef;

    /* on tab or shift+tab, manual navigation */
    if (keyCode === KeyCode.KEY_TAB) {
      this.closeDropdown();

      // find the selectInput among the all of the document's tabbable elements
      const allTabbableElements = tabbable(document);
      const selectInputIdx = allTabbableElements.findIndex(el => el === selectInput);

      const previousElement = allTabbableElements[selectInputIdx - 1];
      const nextElement = allTabbableElements[selectInputIdx + 1];

      // check for shiftKey so we know to navigate to the previous element,
      // if it exists
      if (shiftKey && previousElement) {
        // prevent tab press from prematurely focusing on the wrong element
        event.preventDefault();

        // manually focus on the previous element
        previousElement.focus();

      // focus on the next element, if it exists
      } else if (nextElement) {
        // prevent tab press from prematurely focusing on the wrong element
        event.preventDefault();

        // manually focus on the next element
        nextElement.focus();
      }

    /* on escape, just close the dropdown and focus on the select input */
    } else if (keyCode === KeyCode.KEY_ESCAPE) {
      this.closeDropdown();

      // focus back on the select input
      if (selectInput) {
        selectInput.focus();
      }
    }
  }
```

bjankord:
@dev-cprice Let's leave that code on the side for now, I'm fine with the current behavior and moving forward with considering the tab shifting the virtual indicator back to the input on blur a VoiceOver bug. If users find issue with this when using VoiceOver, we can re-evaluate and look at implementing the code sample you noted.

dev-cprice:
looks like the select hookshot content shifts over 1px to the right and down when it opens...

packages/terra-form-select/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* a11y: Moved input into the Menu, make menu options adjacent to the input
```
bjankord: Let's add the following:
```
### Added
* Added new CSS custom property: `--terra-form-select-focus-border-bottom-left-radius`
* Added new CSS custom property: `--terra-form-select-focus-border-bottom-right-radius`

File: packages/terra-form-select/src/search/Menu.jsx [hunk 11]
```diff
@@ -1,416 +1,390 @@
...
-       * attributes, you'll need to manually verify that the virtual indicator on iOS is still shifted
-       * to the dropdown / the selected item in the dropdown if an item is selected when the dropdown
-       * is opened.
-       */
+  });
+
+  const menuClasses = cx([
+    'menu',
+    { 'is-above': isAbove },
+    { 'is-focused': true }, // TODO: fixme
```
bjankord: Is this todo still being worked on or this fixed now?

dev-cprice: I guess this can be removed now. There's not a moment where the menu is open and it's not focused.

packages/terra-form-select/src/search/Menu.jsx
```diff
@@ -1,416 +1,390 @@
...
+    { 'is-above': isAbove },
+    { 'is-focused': true },
+    { 'is-invalid': isInvalid },
+  ]);
+
+  /* eslint-disable react/forbid-dom-props */
+  return (
+    <div ref={menuRef} id={menuId} className={menuClasses} aria-describedby={ariaDescribedBy}>
+      <div className={cx('menu-container')}>
+        <div className={cx('display')} aria-label={ariaLabel}>
```
bjankord: I believe we'll need to add `role="application"` here. JAWS doesn't seem to honor up/down arrow key events in its default "browse" mode. Switching this over to "application" mode will allow JAWS to pass through the up/down arrow key events like we expect.

packages/terra-form-select/translations/en-US.json
```diff
@@ -1,5 +1,6 @@
 {
   "Terra.form.select.add": "Add \"{text}\"",
+  "Terra.form.select.close": "Close",
```
benbcai: A new en-CA translation file just got merged today: https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/translations/en-CA.json
Can you also add this string translation to the en-CA file?

dev-cprice: This PR might end up being scrapped. Since we've hit a lot of regressions with existing functionality, we're exploring new options going forward with UX.

If we continue down this path, I'll be sure to add it though!

dev-cprice:
Closing this in favor of the simpler fix in #2658

PR #2644, opened 2019-09-13
Pin webpack version to 4.39.3 until bugs are fixed in newer webpack r…

bjankord:
### Summary
Temporary resolution to fix issue related to buggy behavior with webpack v4.40.0

Related:
* https://github.com/webpack/webpack/issues/9692
* https://github.com/webpack/webpack/issues/9693


bjankord:
Webpack seems to have resolved the issues we were seeing with 4.40.0 in the 4.40.1 release.

PR #2646, opened 2019-09-16
[terra-form-select] Restore passivity with prop-types

### LINKED ISSUE #2622: [terra-form-select] Array of numbers for MultiSelect produces proptype warnings
# Bug Report

## Description

Previously on terra-form-select 5.25.0, given a multiple variant (MultiSelect or TagSelect), we could pass a list of numbers for the value prop:

[Select.jsx:100](https://github.com/cerner/terra-core/blob/a3e7707420cd6d09b20d15c7eecb78b11843384d/packages/terra-form-select/src/Select.jsx#L100)
[SelectField.jsx:102](https://github.com/cerner/terra-core/blob/a3e7707420cd6d09b20d15c7eecb78b11843384d/packages/terra-form-select/src/SelectField.jsx#L102)

```
<Select
  variant={'multiple'}
  value={[12, 34, 56, 78]}
>
  {...}
</Select>
```

When the large refactor happened for terra-form-select 5.26.0, the proptype validation requires value to be an array of strings. [MultiSelect.jsx:93](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/MultiSelect.jsx#L93)

## Steps to Reproduce
Create a Select component with a 'multiple' variant
Set the value to an array of numbers
Receive proptype validation warnings

image

## Expected Behavior
This code would not produce a proptype error.

## Possible Solution

Update [MultiSelect.jsx:93](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/MultiSelect.jsx#L93) to accept an array of numbers or strings

## Environment
* Component Name and Version: terra-form-select@5.26.0+
* Browser Name and Version: Latest Chrome
* Node/npm Version: Node 8, npm 5
* Operating System and version (desktop or mobile): Any

## @ Mentions
@dev-cprice

dev-cprice:
### Summary
closes #2622 

@cerner/terra

[License]: ../LICENSE

### DIFF
--- a/packages/terra-form-select/src/Combobox.jsx
+++ b/packages/terra-form-select/src/Combobox.jsx
@@ -16,7 +16,7 @@ const propTypes = {
    */
   children: PropTypes.node,
   /**
-   * The default selected value.
+   * The default selected value. Can be a string, number, or array of strings/numbers.
    */
   defaultValue: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.array]),
   /**
@@ -86,7 +86,7 @@ const propTypes = {
    */
   required: PropTypes.bool,
   /**
-   * The selected value.
+   * The selected value. Can be a string, number, or array of strings/numbers.
    */
   value: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.array]),
 };
--- a/packages/terra-form-select/src/Combobox.jsx
+++ b/packages/terra-form-select/src/Combobox.jsx
@@ -18,7 +18,7 @@ const propTypes = {
   /**
    * The default selected value.
    */
-  defaultValue: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  defaultValue: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.array]),
   /**
    * Whether the select is disabled.
    */
@@ -88,7 +88,7 @@ const propTypes = {
   /**
    * The selected value.
    */
-  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.array]),
 };
 
 const defaultProps = {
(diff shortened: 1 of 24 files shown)


packages/terra-form-select/src/Combobox.jsx
```diff
@@ -88,7 +88,7 @@ const propTypes = {
   /**
    * The selected value.
    */
-  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+  value: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.array]),
```
nramamurth: nit: the props table generated on the site does not show this information well (it says that the type is a ‘union’). I’d explicitly call this piece out in the inline comment above so that the consumers are aware of what types this prop can be when they look up documentation.

dev-cprice: fixed: https://github.com/cerner/terra-core/pull/2646/commits/734f0b26bb3a546fcbbdbbd1babc37bd9461b17c

packages/terra-form-select/src/Combobox.jsx
```diff
@@ -110,20 +110,14 @@ const defaultProps = {
...
-
   constructor(props) {
     super(props);
 
+    const { value, defaultValue } = props;
+
     this.state = {
       tags: [],
-      value: Combobox.defaultValue(props),
+      value: SelectUtil.defaultValue({ defaultValue, value }),
```
StephenEsser: I'm wondering if it would have been possible to scope this change only within the Select component.

Rough Pseudo code
```
const { variant,  value, defaultValue, ...otherProps } = props;

const passiveDefaultValue = SelectUtil.defaultValue({ defaultValue, value }); 

switch (variant) {
  case: Variants.Variant:
    return <Variant defaultValue={passiveDefaultValue} ...otherProps />;
}
```

Mainly because the variants themselves are new, so it is completely passive to keep these strict prop-types if a user was to import them directly. `import Combobox from 'terra-form-select/lib/Combobox'`. It is only non-passive when using the base Select as the entry point.

StephenEsser: It gets tricky for sure, I think the lists are a single value. `[value]`. I'm not really sure what would happen if a list of two or more items were provided to a single select variant. `[value1, value2]`. Chaos, I assume. I think your current solution satisfies the issue, it's just unfortunate we have to expose new prop types to specialized components.

We could probably transform the value in a very similar way to the defaultValue, but it's probably not worth the risk. Controlled values should probably just pass through.

dev-cprice: Yeah, it's just such an odd use case... And it doesn't really work, see below:

image

Same behavior whether we use `value` or `defaultValue` here.

I thought about logging an issue to circle back and remove the weird list prop-types for the single selects for the next breaking change

StephenEsser: +1 to circling back around on the next major version

dev-cprice: Issue logged: https://github.com/cerner/terra-core/issues/2652

PR #2654, opened 2019-09-17
Update cross-env dev dependency version

bjankord:
### Summary
Update cross-env dev dependency version.

Question, do we need cross-env still now that Windows development has been migrated to docker?
Should we just get rid of this dependency.


bjankord:
The only place I see it used is in the root wdio command: 
https://github.com/cerner/terra-core/blob/09102a476642bbef59a73567cf83f6591ed48d93/package.json#L89

PR #2658, opened 2019-09-20
[terra-form-select] Select mobile a11y

dev-cprice:
### Summary
In our infinite quest to make the select mobile accessible, we have introduced a new prop that is opt-in for making the dropdown menu render `position: absolute` in the normal DOM flow: `isTouchAccessible`.

**[DEPLOYMENT LINK](https://terra-core-deployed-pr-2658.herokuapp.com/components/terra-form-select/form-select/select)**

### NOTE

This fix only applies for select variants with an input in them (all except `default` / `<SingleSelect />`

### Description

As explored and discovered in #2635, portaling the `<input />` introduces a lot more complexity than we're comfortable with ***and*** it introduces regressions on nearly all other targets (browsers, screen readers, and all combinations thereof). In an effort to combat that complexity and maintain passivity and functionality for all targets, we are introducing a way to remove the portaled menu.

This is the same thing that the UK Gov did [here](https://github.com/alphagov/accessible-autocomplete), with working deployed examples [here](https://alphagov.github.io/accessible-autocomplete/examples/), however (with our own UX guidance) we are also making the dropdown `position: absolute;`

@cerner/terra

[License]: ../LICENSE

cc @nickpith


packages/terra-form-select/src/shared/isMobile.jsx
```diff
@@ -0,0 +1,16 @@
...
+ * On the userAgent check against "Mobi", quoting MDN:
+ * > In summary, we recommend looking for the string “Mobi” anywhere in the
+ * > User Agent to detect a mobile device
+ * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent#Mobile_Tablet_or_Desktop
+ *
+ * We also need to ensure that touch events are present in the window, since our
+ * assumption is that mobile uses touch as the primary interface
+ */
+export default function isMobile() {
+  return /Mobi/.test(navigator.userAgent) && 'ontouchstart' in window;
```
dev-cprice: before iOS 13, this worked on all mobile devices

however, iPads will no longer say Mobile and will instead use the Desktop experience that mac os provides

surprisingly, Apple did add support for [visual viewport](https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API), so we can fall back to this

this is an interesting problem to solve because we don't want to simply say "if the device ~quacks like a duck, it must be a duck~ supports touch, then it's mobile" This is because more and more desktop PCs are supporting touch and we really only want to target devices that use a virtual on-screen keyboard.

reworking this since it affects how we handle blurring the input on keyboard overlay devices (phones & tablets)

dev-cprice: turns out we don't necessarily need this for the blur logic... it should be able to be simplified by attaching `onTouchStart` to the select's `<input>` and setting internal state for something like `inputFocusedByTouch`

since the `<input>` `onBlur` fires when the onscreen keyboard is dismissed, we can now better handle this in the frame `onBlur` by looking at `inputFocusedByTouch`

still need to explore what happens when the user just taps on a new input though... 🧐

however, we should keep the above comment in mind when thinking about our mobile voiceover specific logic

dev-cprice: so, I can determine when a touch event focuses the select, even with voice over on

using that, I can determine if I should blur the select or not (need to look at the `event.relatedTarget`, which will be `null` if the keyboard was dismissed, else populated by another target

the current behavior of blurring the select closes the menu and clears the input search value

however, by not blurring, the select menu will be visible until interacted with or manually closed...

packages/terra-doc-template/src/ExampleTemplate.module.scss
```diff
@@ -3,7 +3,6 @@
     border: 1px solid 	#e8e8e8;
     border-radius: 3px;
     margin-bottom: 0.9375rem;
-    overflow: hidden;
```
dev-cprice: necessary for `useSemanticDropdown` examples to break out

packages/terra-form-select/src/search/Frame.jsx
```diff
@@ -374,6 +388,13 @@ class Frame extends React.Component {
     }
   }
 
+  /**
+   * TODO: docme
```
dev-cprice: reminder to myself to doc this (and others)...

dev-cprice:
I think some minor styling issues are introduced to the single select variant, working on those now. Then I'll add in tests. For now, removing WIP status on this PR.

packages/terra-form-select/src/shared/_Dropdown.module.scss
```diff
@@ -20,4 +20,10 @@
     border-top-right-radius: var(--terra-form-select-dropdown-border-top-right-radius, 3px);
     box-shadow: var(--terra-form-select-dropdown-above-box-shadow, 0 -3px 12px -5px rgba(0, 0, 0, 0.25));
   }
+
+  .semantic {
+    position: absolute;
+    top: 100%;
```
bjankord: I'd recommend going with these styles:
```scss
    // 0 needs px unit for calc to work correctly
    /* stylelint-disable length-zero-no-unit */
    left: calc(0px - var(--terra-form-select-dropdown-left-offset-width, 1px));
    /* stylelint-enable length-zero-no-unit */
    top: calc(100% + var(--terra-form-select-dropdown-top-offset-width, 1px));
```

The `--terra-form-select-dropdown-left-offset-width` and `--terra-form-select-dropdown-top-offset-width` should be equal to the width of the border used in `--terra-form-select-dropdown-border`.  We'll also want to test this in RTL, we may need to define a right offset for that directionality.

packages/terra-form-select/src/shared/isMobile.jsx
bjankord: What's this used to solve again?

packages/terra-form-select/src/shared/_Dropdown.module.scss
dev-cprice: Earlier this morning when I initially added those, it shifted everything by 1px. Lemme see again, just to make sure..

bjankord: Talked offline, planning to go with the styling as-is.

bjankord: Added left: 0; to account for IE needing explicit style defined for positioning.

mmalaker:
Functional verification completed. The form select examples display and function as expected on mobile. 
Two separate issues found that relate only to desktop: #2661 and #2663.

packages/terra-doc-template/src/ExampleTemplate.module.scss
StephenEsser: This is a little concerning, but I'm not seeing any issues browsing through the other doc examples. It might be worth double checking the dev site in IE 10 just to make absolutely certain this overflow isn't doing anything.

Also, the crazy space on the `border` property above cannot be unseen.

packages/terra-form-select/src/Combobox.jsx
```diff
@@ -85,6 +85,10 @@ const propTypes = {
    * Whether the field is required.
    */
   required: PropTypes.bool,
+  /**
+   * Render dropdown menu in normal DOM flow with position absolute. Renders in a portal by default.
+   */
+  useSemanticDropdown: PropTypes.bool,
```
StephenEsser: Nominating a name change here. Names are challenging, but here are a few rough ideas.

- `disablePortal`
- `isPositionAbsolute`
- `isAbsolutePositioned`
- `isInline`
- `isDropdownInline`
- `isInlineDropdown`

`isPortaled` would be nice, but we want a variable that defaults to false. I don't particularly like any of these names, but maybe some combination of them will work.

dev-cprice: Yeah... Naming. What a constant battle. We started with isDropdownInline, but that was before we had position absolute'd the menu. So it made more sense... I chose useSemanticDropdown after we talked with UX on Monday since it is using semantic markup now and with the hope that the name would lead to teams wanting to use semantic markup...

Even thought about something like preferAccessibility to try to guilt people into opting in 😂

nickpith: I'm not sure most people will understand the semantic markup terminology. Whatever we choose I think accessibility needs to be called out in the code comments and prop documentation for clarity and reinforcement. 

Although it goes against the idea that all components should be accessible by default, I like `preferAccessibility` or `isAccessible` because it clearly says why I would care about this and then later when it is `true` it will be a code smell when people are using `isAccessible=false`. Much like how `dangerouslySetInnerHtml` is a cue that I'm doing something abnormal.

dev-cprice: Yeah, I do agree that "use semantic" will be lost on most consumers. I'm split on which one I like more: `preferAccessibility` or `isAccessible`

dev-cprice: @bjankord Crushing it. Love it. Sounds the most reasonable.

nickpith: +1

neilpfeiffer:
Tested: PR #2658
 
**Reviewed:**
- Reviewed new opt-in prop for initial enhancement for touch accessibility (iOS VoiceOver specifically). May become standard behavior in future major version and removed as optional prop.

 
Marking as Verified, ready for Release.
 
**Pass** |
-----------|
UX Review|

PR #2676, opened 2019-10-01
[terra-form-input] - Add placeholder as a property

### LINKED ISSUE #2675: [terra-form-input] - Add placeholder as a supported property
# Feature Request

Add the html placeholder attribute as a declared property for the Input and InputField so it displays within the generated props table on the terra-ui site.

## Description

There are several [site examples](https://engineering.cerner.com/terra-ui/components/terra-form-input/form-input/input) using the placeholder attribute, however, the placeholder is not officially declared within the Input props table and is instead being applied as a custom property.

StephenEsser:
### Summary

This PR adds the `placeholder` attribute as a declared propType within the `<Input />` and `<InputField />` components. 

### Additional Details

Several [site examples](https://engineering.cerner.com/terra-ui/components/terra-form-input/form-input/input) are displaying placeholder text. The property should be shown within generated props table.

Resolves #2675

### Testing

Coincidentally there were already several tests for the placeholder attribute even though it was not a declared propType.

- [Test 1](https://github.com/cerner/terra-core/blob/master/packages/terra-form-input/tests/wdio/form-input-spec.js#L36)
- [Test 2](https://github.com/cerner/terra-core/blob/master/packages/terra-form-input/tests/wdio/form-input-spec.js#L71)
- [Test 3](https://github.com/cerner/terra-core/blob/master/packages/terra-form-input/tests/jest/Input.test.jsx#L25)

I added a couple additional placeholder tests for the InputField.

@cerner/terra

[License]: ../LICENSE

### DIFF
--- a/packages/terra-form-input/src/InputField.jsx
+++ b/packages/terra-form-input/src/InputField.jsx
@@ -72,6 +72,10 @@ const propTypes = {
    */
   onChange: PropTypes.func,
   /**
+   * Placeholder text.
+   */
+  placeholder: PropTypes.string,
+  /**
    * Ref callback to pass into the ref attribute of the html input element.
    */
   refCallback: PropTypes.func,
@@ -109,6 +113,7 @@ const defaultProps = {
   isLabelHidden: false,
   labelAttrs: {},
   onChange: undefined,
+  placeholder: undefined,
   maxWidth: undefined,
   refCallback: undefined,
   required: false,
@@ -134,6 +139,7 @@ const InputField = (props) => {
     labelAttrs,
     maxWidth,
     onChange,
+    placeholder,
     refCallback,
     required,
     showOptional,
@@ -157,6 +163,7 @@ const InputField = (props) => {
   }
 
   const inputType = type || inputAttrs.type;
+
   return (
     <Field
       label={label}
@@ -180,6 +187,7 @@ const InputField = (props) => {
         id={inputId}
         type={inputType}
         onChange={onChange}
+        placeholder={inputAttrs.placeholder || placeholder}
         value={value}
         defaultValue={defaultValue}
         refCallback={refCallback}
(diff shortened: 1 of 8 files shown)


packages/terra-form-input/tests/jest/InputField.test.jsx
```diff
@@ -2,145 +2,159 @@ import React from 'react';
 import IconHelp from 'terra-icon/lib/icon/IconHelp';
 import InputField from '../../src/InputField';
 
-it('should render a default InputField component', () => {
-  const textarea = <InputField inputId="test-input" label="Label" />;
-  const wrapper = shallow(textarea);
-  expect(wrapper).toMatchSnapshot();
-});
+describe('InputField', () => {
```
StephenEsser: This file looks like it had a lot of changes, but the largest change was wrapping the it statements within this describe block. These tests were flat/unscoped previously.

packages/terra-form-input/src/InputField.jsx
```diff
@@ -180,6 +187,7 @@ const InputField = (props) => {
         id={inputId}
         type={inputType}
         onChange={onChange}
+        placeholder={inputAttrs.placeholder || placeholder}
```
emilyrohrbough: I would expect placeholder to be valued first if both were provided since it's a first class prop

StephenEsser: For passivity the inputAttrs are valued first. The order probably isn't as important because it is only an issue if both attributes are provided which consumers can easily reconcile, but I think it makes sense to prioritize the existing API when a change like this is made that can override the old API. This is how disabled is being used now in all of the form fields, for better or for worse.

StephenEsser: I went ahead and flipped these: https://github.com/cerner/terra-core/pull/2676/commits/a9c6e70ac392f5a2e6019746106660d581c435a9

nramamurth:
dat label tho 🔥

Issue #2678, opened 2019-10-02
[Terra-Overlay] throws removeAttribute undefined error when being displayed


po016255:
# Bug Report

## Description
It appears we get an error in the terra-overlay when displaying the overlay by setting isOpen=true.  (ie debugger kicks off a data load that shows the overlay and the error occurs before the data returns so it most likely when the overlay becomes visible, not being hidden)
When clicking yes on the script error to dismiss it, the underlying summary successfully finishes loading.

[Post on teams](https://teams.microsoft.com/l/message/19:0adf7161dd244f87aa6415f7c5577836@thread.skype/1570037453018?tenantId=fbc493a8-0d24-4454-a815-f4ca58e8c09d&groupId=45f47bfa-38a9-444d-8933-9c20266ae38c&parentMessageId=1570037453018&teamName=Terra&channelName=Component%20Support&createdTime=1570037453018)
Issue cannot be recreated in standalone chrome or ie11/10.
It only happens when running the component in an mpage.

The bug appears to be related to showing the overlay when setting isOpen=true.  [Line of code in plan-js.](https://github.cerner.com/ipoc/goals-js/blob/180edf86bf3e5c9c4ae96527f9cb3bdbc40043d3/src/goals/components/views/BaseView.jsx#L349)
[This appears to be the offending line of code in terra-overlay that throws an exception.](https://github.com/cerner/terra-core/blob/b1562abd458cb6c999fcf59e45ba4dab6eb43403/packages/terra-overlay/src/Overlay.jsx#L132)

## Steps to Reproduce
1. Load OnePlan inside an mpage. (plan-js)
2. Click tab in mpage to view goals and activity summary. 
2. Click an activity to view details. (opens slide panel with detail view)
3. Click button to modify an activity.  (view in panel changes to modify view)
3. Click apply. (triggers reload of goals content and closes slide panel)
4. Overlay displays while data is being loaded and error is thrown. (isOpen is set to true on overlay)

## Additional Context / Screenshots
![image](https://user-images.githubusercontent.com/12768193/66068218-d0659100-e512-11e9-97d3-7df29bfaa72f.png)

![image](https://user-images.githubusercontent.com/12768193/66068277-ea06d880-e512-11e9-9ada-c60f9ea5234e.png)

Gif of error in solgm domain.  
![overlay_error](https://user-images.githubusercontent.com/12768193/66068843-f2abde80-e513-11e9-80fe-19f69569e9d1.gif)

## Expected Behavior
[...]

PR #2679, opened 2019-10-02
Update how imperative DOM manipulations are run in terra-overlay

bjankord:
### Summary

Related: #2678 

Updated imperative macrotasks to encapsulate all imperative DOM conditionals and manipulation. This should resolve some issues we're seeing with one of internal rendering engines.


packages/terra-overlay/src/Overlay.jsx
```diff
@@ -126,27 +126,30 @@ class Overlay extends React.Component {
 
   enableContainerChildrenFocus() {
```
tbiethman: Can we execute this logic in componentDidUpdate instead? There's probably a quirk with the inert implementation I'm missing, but from what I'm looking at here, it seems like the microtasks (TIL the technical term, thank you) would be unnecessary if we're doing the DOM manipulation in the expected lifecycle functions rather than render.

bjankord: Good call, updated in this commit: https://github.com/cerner/terra-core/pull/2679/commits/cd17e89e5a8b2b135b21c3569fa566ea147bcc6c

packages/terra-overlay/src/Overlay.jsx
```diff
@@ -87,9 +86,17 @@ class Overlay extends React.Component {
     document.addEventListener('keydown', this.shouldHandleESCKeydown);
   }
 
+  componentDidUpdate() {
+    if (this.props.isOpen) {
+      this.disableContainerChildrenFocus();
```
tbiethman: We probably need to call  disableContainerChildrenFocus in componentDidMount as well (if isOpen is set)

bjankord: Updated in this commit: https://github.com/cerner/terra-core/pull/2679/commits/c8f96a27a024cf20ad45da9a569d8d713a676128 

I've also updated some of the test examples to verify this behavior where we set the overlay to isOpen="true" on initial mount.

PR #2680, opened 2019-10-03
Fix IE 10 intl loading issue

emilyrohrbough:
### Summary
The intl polyfill was always included in the terra-i18n setup. We then tried to use the `global.Intl` definition to determine if we would need to load intl data, however the `intl` polyfill directly adds itself as the global object if it is undefined. This means we never loaded the intl data when it was needed. This change reflects the[ setup recommendation outline in the `intl` docs for using this polyfill with webpack](
https://github.com/andyearnshaw/Intl.js/tree/4bd382c5df206f328b35ca23579904f603b84f0b#intljs-and-browserifywebpack).

Resolves https://github.com/cerner/terra-core/issues/2681


packages/terra-i18n/tests/jest/i18nLoader.test.js
```diff
@@ -16,25 +21,42 @@ describe('i18nLoader', () => {
...
+  describe('when intl is provided by browser', () => {
+    beforeAll(() => {
+      global.Intl = {};
+      i18nLoader('en', jest.fn());
     });
+    afterAll(() => { global.Intl = undefined; });
 
-    afterEach(() => {
-      delete process.env.NODE_ENV;
+    it('it does not adds the intl polyfill ', () => {
```
tbiethman: Nitpicking here, but can we loose the 'it's within the test descriptions?

packages/terra-i18n/tests/jest/i18nLoader.test.js
```diff
@@ -16,25 +21,42 @@ describe('i18nLoader', () => {
...
+    beforeAll(() => {
+      global.Intl = undefined;
+      i18nLoader('en', jest.fn());
     });
+    afterAll(() => { global.Intl = undefined; });
 
-    it('still throws error when callback is not function', () => {
-      const invalidCallback = () => i18nLoader('en');
+    it('it adds the intl polyfill ', () => {
+      // expect('intl').toHaveBeenCalled();
```
tbiethman: nix this feller

bjankord:
Tested in IE 10 with the code that was triggering an error and it works as expected now:
Screen Shot 2019-10-04 at 9 14 40 AM

Issue #1516, opened 2018-05-15
[terra-paginator] update focus+active styles


neilpfeiffer:
# Issue Description
Update :focus and :active styles, shifting from pseudo selectors to applied classes (`is-focused` and `is-active`) updated with state, same pattern as terra-button. Theming variables will remain unchanged.

## Issue Type

- [ ] New Feature
- [x] Enhancement
- [ ] Bug
- [ ] Other

## Expected Behavior
Will remove the focus-glow that remains after you click "< Previous" and "Next >" with a mouse/touch press.

PR #2686, opened 2019-10-07
[terra-paginator] Update Paginator Focus and Active Style Application

amichaelparker:
### Summary
This updates terra-paginator to apply focus and active styling via classes, rather than pseudo selectors.

Closes [#1516](https://github.com/cerner/terra-core/issues/1516)

### Additional Details
The new button previously used state to apply styles, but now uses applied data attributes to accomplish the same. See [this comment](https://github.com/cerner/terra-core/pull/2686#discussion_r337067094).


packages/terra-paginator/src/_PaginatorButton.jsx
```diff
@@ -0,0 +1,100 @@
...
+
+  handleOnKeyDown(event) {
+    // Add active state to FF browsers
+    if (event.nativeEvent.keyCode === KeyCode.KEY_SPACE) {
+      this.setState({ active: true });
+    }
+
+    // Add focus styles for keyboard navigation
+    if (event.nativeEvent.keyCode === KeyCode.KEY_SPACE || event.nativeEvent.keyCode === KeyCode.KEY_RETURN) {
+      this.setState({ focused: true });
```
neilpfeiffer: Nice work @amichaelparker 
Let's update the approach from internal state to using a data attribute instead, similar to terra-application-navigation and terra-list:

Example here of it as a utility:
https://github.com/cerner/terra-framework/blob/c28c37b8a7030143cccfd4970d8d75b8885ed677/packages/terra-application-navigation/src/utils/helpers.js#L9-21

Example here of it in use (using onBlur and onMouseDown):
https://github.com/cerner/terra-framework/blob/c28c37b8a7030143cccfd4970d8d75b8885ed677/packages/terra-application-navigation/src/utility-menu/_UtilityMenuHeaderButton.jsx#L65-L66


And corresponding CSS to match:
https://github.com/cerner/terra-framework/blob/c28c37b8a7030143cccfd4970d8d75b8885ed677/packages/terra-application-navigation/src/utility-menu/UtilityMenuHeaderButton.module.scss#L29-34

amichaelparker: I moved everything over, and it now uses the data attributes to apply the styles. The button doesn't strictly need to be a separate private component at this point, but I think it also keeps some of the clutter out of the main pagination component.

packages/terra-paginator/src/_PaginatorButton.jsx
```diff
@@ -0,0 +1,63 @@
...
+    this.shouldShowFocus = false;
+    // Wait until after onFocus has been triggered on browsers where it will get triggered for click
+    setTimeout(() => { this.shouldShowFocus = true; });
+  }
+
+  render() {
+    const {
+      children,
+      onClick,
+      ...customProps
```
jmsv6d: This is a private component, there's really no need to allow for custom props.

packages/terra-paginator/src/_PaginatorButton.jsx
```diff
@@ -0,0 +1,63 @@
...
+    } = this.props;
+
+    return (
+      <button
+        className={cx(customProps.className)}
```
jmsv6d: You can just define className as a prop. Or just have a variant prop and apply the styles here.

packages/terra-paginator/src/ControlledPaginator.jsx
```diff
@@ -102,7 +103,7 @@ class Paginator extends React.Component {
         return;
       }
       pageButtons.push((
-        <button
+        <PaginatorButton
           aria-label={intl.formatMessage({ id: 'Terra.paginator.pageIndex' }, { pageNumber: val })}
```
jmsv6d: These aria attributes aren't going to get applied because you don't spread `customProps` in `PaginatorButton`. I would suggest just making defining props for them instead of using customProps anyway.

packages/terra-paginator/src/_PaginatorButton.jsx
```diff
@@ -0,0 +1,63 @@
...
+        onBlur={(e) => disableFocusOnBlur(e)}
+        onClick={onClick}
+        onFocus={this.handleFocus}
+        onKeyDown={(e) => enableFocusActiveStyles(e)}
+        onKeyUp={(e) => disableFocusActiveStyles(e)}
+        onMouseDown={this.handleMouseDown}
+        type="button"
```
jmsv6d: Since you define the `type` here you can remove it in all the usages.

packages/terra-paginator/src/_PaginatorButton.jsx
amichaelparker: Changed: [243ad46](https://github.com/cerner/terra-core/pull/2686/commits/243ad46953a80fa9e1f6fa19c0924da3ab009538)

packages/terra-paginator/src/_PaginatorButton.jsx
amichaelparker: Prop added: [243ad46](https://github.com/cerner/terra-core/pull/2686/commits/243ad46953a80fa9e1f6fa19c0924da3ab009538)

packages/terra-paginator/src/ControlledPaginator.jsx
amichaelparker: Good catch! I added props for the various attributes: [243ad46](https://github.com/cerner/terra-core/pull/2686/commits/243ad46953a80fa9e1f6fa19c0924da3ab009538)

packages/terra-paginator/src/_PaginatorButton.jsx
```diff
@@ -0,0 +1,91 @@
...
+  onClick: PropTypes.func,
+};
+
+class PaginatorButton extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleFocus = this.handleFocus.bind(this);
+    this.handleMouseDown = this.handleMouseDown.bind(this);
+
+    this.shouldShowFocus = true;
```
jmsv6d: You could utilize `useRef` for this and make this a functional component.

amichaelparker: Migrated the component over to the functional variety: [30998be](https://github.com/cerner/terra-core/pull/2686/commits/30998bee8fb4b9c77312b0b144243e42d0dc274f). The effect functions just using a local variable for `shouldShowFocus` since it is just being used (in this case) as a dumb flag, but if there are other issues with it I have missed, let me know!

amichaelparker:
@neilpfeiffer Are the changes in this PR still needed?

packages/terra-paginator/src/ProgressivePaginator.jsx
```diff
@@ -154,57 +151,53 @@ class ProgressivePaginator extends React.Component {
...
-            aria-disabled={selectedPage === 1}
-            aria-label={intl.formatMessage({ id: 'Terra.paginator.previous' })}
+          </PaginatorButton>
+          <PaginatorButton
+            ariaDisabled={selectedPage === 1}
+            ariaLabel={intl.formatMessage({ id: 'Terra.paginator.previous' })}
             className={cx(['nav-link', 'previous', 'icon-only', selectedPage === 1 ? 'is-disabled' : null])}
             tabIndex={selectedPage === 1 ? null : '0'}
             onClick={this.handlePageChange(previousPageIndex)}
             onKeyDown={this.handleOnKeyDown(previousPageIndex)}
```
emilyrohrbough: It doesn't look like these `onKeyDown` props are being used and executed by the `PaginatorButton`. Am I missing it?

packages/terra-paginator/src/ProgressivePaginator.jsx
```diff
@@ -154,57 +151,53 @@ class ProgressivePaginator extends React.Component {
...
-          </button>
-          <button
-            aria-disabled={selectedPage === totalPages}
-            aria-label={intl.formatMessage({ id: 'Terra.paginator.last' })}
+          </PaginatorButton>
+          <PaginatorButton
+            ariaDisabled={selectedPage === totalPages}
+            ariaLabel={intl.formatMessage({ id: 'Terra.paginator.last' })}
             className={cx(['nav-link', selectedPage === totalPages ? 'is-disabled' : null])}
             tabIndex={selectedPage === totalPages ? null : '0'}
```
emilyrohrbough: It doesn't look like this `tabIndex ` prop is being used and honored by the `PaginatorButton`. Am I missing it?

packages/terra-paginator/src/_PaginatorButton.jsx
```diff
@@ -0,0 +1,83 @@
...
+import { disableFocusActiveStyles, disableFocusOnBlur, enableFocusActiveStyles } from './_paginationUtils';
+import styles from './Paginator.module.scss';
+
+const cx = classNames.bind(styles);
+
+const propTypes = {
+  /**
+   * Sets aria-current attribute.
+   */
+  ariaCurrent: PropTypes.oneOfType([
```
emilyrohrbough: I might be missing it, but it seems this prop is only used 2x in the changes and both are passing a boolean value. When would a string value be passed?

packages/terra-paginator/src/_PaginatorButton.jsx
```diff
@@ -0,0 +1,83 @@
...
+    className,
+    onClick,
+  } = props;
+
+  return (
+    <button
+      aria-current={ariaCurrent}
+      aria-disabled={ariaDisabled}
+      aria-label={ariaLabel}
+      className={cx(className)}
```
emilyrohrbough: Should we be binding the PaginatorButton's class styles to the already bound styles passed ? Looking at some of these changes, it looks like this prop is being used like:
```
<PaginatorButton
    className={cx(['nav-link', selectedPage === totalPages ? 'is-disabled' : null])}
/>
```

packages/terra-paginator/src/_PaginatorButton.jsx
```diff
@@ -0,0 +1,83 @@
...
+      onBlur={(e) => disableFocusOnBlur(e)}
+      onClick={onClick}
```
emilyrohrbough: When comparing to terra-button, do we need to implement similar focus logic on the Paginator Button when an onClick is triggered? https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L138

packages/terra-paginator/src/_PaginatorButton.jsx
```diff
@@ -0,0 +1,83 @@
...
+
+  const handleFocus = (event) => {
+    if (shouldShowFocus) event.currentTarget.setAttribute('data-focus-styles-enabled', 'true');
+  };
+
+  const handleMouseDown = () => {
+    // Prevent button from showing focus styles when clicked
+    shouldShowFocus = false;
+    // Wait until after onFocus has been triggered on browsers where it will get triggered for click
+    setTimeout(() => { shouldShowFocus = true; });
```
emilyrohrbough: Should this timeout value have a time value as well, not just the function to execute when the timeout is reached

When comparing to terra-button, should we be preventing default on the event? https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L212

packages/terra-paginator/src/_PaginatorButton.jsx
```diff
@@ -0,0 +1,83 @@
...
+   * Callback function triggered when clicked.
+   */
+  onClick: PropTypes.func,
+};
+
+function PaginatorButton(props) {
+  let shouldShowFocus = true;
+
+  const handleFocus = (event) => {
+    if (shouldShowFocus) event.currentTarget.setAttribute('data-focus-styles-enabled', 'true');
```
emilyrohrbough: Can we add this to the Pagination Utils with the other helpers that that set/remove focus/active styles?

also, can we format like this if statement like the rest of the code base?
```
if (shouldShowFocus) {
   // dos tuff
}
```

packages/terra-paginator/src/ProgressivePaginator.jsx
amichaelparker: Looks like this was leftover from moving things around. Removed! [3c4f04a](https://github.com/cerner/terra-core/pull/2686/commits/3c4f04a82bd80d77dc4fc4f35ba36c5d85deb003)

packages/terra-paginator/src/ProgressivePaginator.jsx
amichaelparker: Added this to the button - [3c4f04a](https://github.com/cerner/terra-core/pull/2686/commits/3c4f04a82bd80d77dc4fc4f35ba36c5d85deb003)

packages/terra-paginator/src/_PaginatorButton.jsx
amichaelparker: I changed the paginator button to just use the passed down props = [3c4f04a](https://github.com/cerner/terra-core/pull/2686/commits/3c4f04a82bd80d77dc4fc4f35ba36c5d85deb003)

packages/terra-paginator/src/_PaginatorButton.jsx
amichaelparker: Added, this rectified the focus issue in Firefox and other browsers - [3c4f04a](https://github.com/cerner/terra-core/pull/2686/commits/3c4f04a82bd80d77dc4fc4f35ba36c5d85deb003)

packages/terra-paginator/src/_PaginatorButton.jsx
amichaelparker: The timeout doesn't appear to be necessary because of the way the events are handled now, and styling works as expected without. Prevent default added to event - [3c4f04a](https://github.com/cerner/terra-core/pull/2686/commits/3c4f04a82bd80d77dc4fc4f35ba36c5d85deb003)

packages/terra-paginator/src/_PaginatorButton.jsx
amichaelparker: Changed! [3c4f04a](https://github.com/cerner/terra-core/pull/2686/commits/3c4f04a82bd80d77dc4fc4f35ba36c5d85deb003)

packages/terra-paginator/src/_PaginatorButton.jsx
amichaelparker: Node evaluates `1 === 1 && 'page'` as a string but, as best as I can tell, there isn't a reason to pass in the string 'page' to ariaCurrent and it just makes the logic confused, so I took it out and fixed the prop type, and it still functions as expected -  [3c4f04a](https://github.com/cerner/terra-core/pull/2686/commits/3c4f04a82bd80d77dc4fc4f35ba36c5d85deb003)

packages/terra-paginator/src/ControlledPaginator.jsx
```diff
@@ -64,24 +63,6 @@ class Paginator extends React.Component {
     };
   }
 
-  handleOnKeyDown(index) {
```
emilyrohrbough: How is the is logic being handled now? It doesn't seem like this state update got pulled forward unless I'm missing it

packages/terra-paginator/src/_paginationUtils.jsx
```diff
@@ -1,3 +1,61 @@
+import * as KeyCode from 'keycode-js';
+
+/**
+ * Enables focus styles for the target of the given event. Typically used as an onBlur callback on selectable elements.
+ * See: terra-application-navigation/src/utils/helpers.js
+ */
+const enableFocusActiveStyles = (event) => {
```
emilyrohrbough: We should add new jest tests for these helpers

packages/terra-paginator/src/_paginationUtils.jsx
```diff
@@ -1,3 +1,61 @@
...
+};
+
+/**
+ * Disable focus style on Blur event.
+ */
+const disableFocusOnBlur = (event) => {
+  event.currentTarget.setAttribute('data-focus-styles-enabled', 'false');
+};
+
+let shouldShowFocus = true;
```
emilyrohrbough: Why is this being defined here? Should this be managed in the component's state?

packages/terra-paginator/src/_PaginatorButton.jsx
```diff
@@ -0,0 +1,75 @@
...
+
+  return (
+    <button
+      aria-current={ariaCurrent}
+      aria-disabled={ariaDisabled}
+      aria-label={ariaLabel}
+      className={className}
+      onBlur={(e) => disableFocusOnBlur(e)}
+      onClick={onClick}
+      onFocus={(e) => handleFocus(e)}
```
emilyrohrbough: Can we define these like:
```
onFocus={handleFocus}
onKeyDown={enableFocusActiveStyles}
onKeyUp={disableFocusActiveStyles=}
onMouseDown={handleMouseDown}
```

packages/terra-paginator/src/ControlledPaginator.jsx
amichaelparker: The key down styles are being handled [directly in the button](https://github.com/cerner/terra-core/blob/2b3154ad8c49b58953711badfaa65d391fe0e718/packages/terra-paginator/src/_PaginatorButton.jsx#L62).

packages/terra-paginator/src/_PaginatorButton.jsx
amichaelparker: Did most of these, but had to compromise on a couple to pass the state change down for the previous comment change.

packages/terra-paginator/src/_paginationUtils.jsx
amichaelparker: I went ahead and added this to the button's state, and passed the setter to the utility functions.

packages/terra-paginator/src/ControlledPaginator.jsx
amichaelparker: I added onKeyDown for the selection to point to the page change function (since they were nearly identical) and added a check for key presses.

jeremyfuksa:
Looks good so far, but will do official UX review after the new Jest tests are complete.

PaulSolDev:
Should tab focus highlight disabled buttons? This appears to suggest the object is selectable:
image

jeremyfuksa:
> Should tab focus highlight disabled buttons? This appears to suggest the object is selectable:
Aside from this problem, it seems to be looking and working like it needs to.

amichaelparker:
> Should tab focus highlight disabled buttons? This appears to suggest the object is selectable:
> Aside from this problem, it seems to be looking and working like it needs to.

I've got this fixed currently, just working on finishing up the Jest tests.

PaulSolDev:

PR #2689, opened 2019-10-09
[terra-form-select] Fixed IE issue that required two clicks to open the dropdown menu after initial selection.

### LINKED ISSUE #2648: [terra-form-select] Drop-down does not open on IE 11 after initial selection
# Bug Report

## Description

Select field drop down required two clicks to open after selecting an option.

## Steps to Reproduce

1. Click on a terra select field on IE 11 to open the drop down
2. Select a drop down option
3. Click on the select field again
4. Drop down will not open unless clicked twice

## Expected Behavior

Expected the drop down to open on first click.

## Environment

* Component Name and Version: terra-select-field (this is able to be reproduced on the terra site)
* Browser Name and Version: IE 11

nramamurth:
### Summary
Fixes #2648


packages/terra-form-select/src/single/Frame.jsx
```diff
@@ -256,21 +267,15 @@ class Frame extends React.Component {
...
-
-    this.closeDropdown();
 
-    if (this.props.onBlur) {
-      this.props.onBlur(event);
+      setTimeout(() => {
+        if (document.querySelector(this.selectMenu) !== document.activeElement) {
+          _onBlur();
+        }
+      }, 10);
```
emilyrohrbough: Why is this timeout needed?

nramamurth: There is [this strange behavior](https://github.com/cerner/terra-core/issues/2648#issuecomment-539909574) I noticed w.r.t. the blur event on IE that seemed to align with our [existing inline code comment](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/single/Frame.jsx#L258). When I read these comments [[1]](https://github.com/facebook/react/issues/3751#issuecomment-415863466) [[2]](https://github.com/catamphetamine/react-responsive-ui/blob/77b7ebc2e43b35595b06cda3ac697b75505e2174/source/utility/focus.js#L30-L45) about blur behaves on IE in React, that seemed to make sense and their trick worked. 

Why 10? I honestly randomly guessed it.

nramamurth:
@emilyrohrbough 
> Can you describe the tests to test these changes? 

If you're talking about the steps to functionally verify on the deployed page, then following the steps mentioned [here](https://github.com/cerner/terra-core/issues/2648#issue-494222772) on IE should do. 

> Which of our current test are validating this behavior is being tested?

I did not see any but [I have added one](https://github.com/cerner/terra-core/pull/2689/commits/2ebe816dc834b681da699f7e8f35249227397563). However, this test might come handy when IE tests are run (probably in the future).

packages/terra-form-select/src/single/Frame.jsx
emilyrohrbough: You are right. This unusual. If this is the work-around, we should add an inline comment with links to these issues so those less familiar with this code will understand why this timeout is here.

packages/terra-form-select/src/single/Frame.jsx
```diff
@@ -248,6 +248,17 @@ class Frame extends React.Component {
...
+    // eslint-disable-next-line no-underscore-dangle
+    const _onBlur = () => {
+      this.setState({ isFocused: false });
+
+      this.closeDropdown();
+
+      if (this.props.onBlur) {
+        this.props.onBlur(event);
+      }
+    };
```
StephenEsser: Is this only an issue with the single select variant?

nramamurth: Yes.

StephenEsser: Can you explain why only this variants needs the special onBlur logic?

nramamurth: Sure. The existing inline code comment explains it - I noticed that unlike the default variant, the blur event does not get triggered when one's trying to open the dropdown menu on the other variants. This special logic takes care of suppressing the rest of the blur logic by not closing the dropdown menu immediately upon Frame click in the default variant.

https://github.com/cerner/terra-core/blob/22e6cb5e503a02e8242061730274ff3bd1c61eb4/packages/terra-form-select/src/single/Frame.jsx#L251-L255

packages/terra-form-select/src/single/Frame.jsx
nramamurth: Done.

PR #2691, opened 2019-10-10
[terra-form-select] Searchable fields' ID moved to input element

### LINKED ISSUE #2674: [terra-form-select] Searchable select fields should be properly labeled via htmlFor
# Feature Request

## Description

Searchable select field variants (e.g. all except `<SelectField variant="default" />`/`<SingleSelectField />`) should render the htmlFor on the select input, rather than the select div.

image

Quoting testing partner:

> Label's for attribute should point to an input element, not a div
>
```html
<label for="single-search-select-recipients" class="Field_label__2C3qa">
  <div class="Field_required__33ntZ">*</div>
  To (single-select):
</label>
```
>
> Doing this would also allow you to remove all the aria-labels scattered throughout.

saket2403:
### Summary
Fixes #2674


nramamurth:
Is this change non-passive?

packages/terra-form-select/src/multiple/Frame.jsx
```diff
@@ -661,8 +664,8 @@ class Frame extends React.Component {
           <span id={labelId}>{this.ariaLabel()}</span>
           <span id={descriptionId}>{this.renderDescriptionText()}</span>
         </div>
-        <div className={cx('display')} aria-label={this.ariaLabel()}>
-          {this.getDisplay(displayId, ariaDescribedBy)}
+        <div className={cx('display')}>
+          {this.getDisplay(displayId, ariaDescribedBy, id)}
```
nramamurth: nit: maybe it's worth considering changing the API to accept an object and then de-structure it to extract whatever that is needed inside `getDisplay`.

packages/terra-form-select/src/tag/Frame.jsx
```diff
@@ -642,6 +643,8 @@ class Frame extends React.Component {
     const descriptionId = `terra-select-screen-reader-description-${uniqueid()}`;
     const customAriaDescribedbyIds = customProps['aria-describedby'];
     const ariaDescribedBy = customAriaDescribedbyIds ? `${descriptionId} ${customAriaDescribedbyIds}` : descriptionId;
+    const { id } = customProps;
+    delete customProps.id;
```
StephenEsser: This can be deconstructed in the props spread above so it doesn't need to be deleted from the customProps. You'll have to add id as a PropType for the linter, but I think that's reasonable since this component does use the id passed in.

```
const propTypes = {
  id: PropTypes.string, 
}
```

    const {
      id,
      ...customProps
    } = this.props;

packages/terra-form-select/src/tag/Frame.jsx
```diff
@@ -187,7 +187,7 @@ class Frame extends React.Component {
     this.input = input;
   }
 
-  getDisplay(displayId, ariaDescribedBy) {
+  getDisplay(displayId, ariaDescribedBy, id) {
```
StephenEsser: The id can probably be grabbed from the props so it doesn't need to be passed around as a parameter

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -208,6 +208,7 @@ class Frame extends React.Component {
       className: cx('search-input'),
```
StephenEsser: > Doing this would also allow you to remove all the aria-labels scattered throughout.

There's a note on the issue that mentions this change would allow us to remove the aria-labels. This will need to be tested with a screen reader to ensure it still works as expected, but we may want to remove those labels with this change.

saket2403: Earlier `aria-label` was read twice by the screen because it was rendering for both `select div` and the `input field`. I have already removed the aria-label for the select div. Removing the label for input field gives axe violations.I don't think that is what we want.
@StephenEsser

packages/terra-form-select/src/multiple/Frame.jsx
saket2403: Added id as prop in 5bb136d

packages/terra-form-select/src/tag/Frame.jsx
saket2403: Added id as prop in 5bb136d

packages/terra-form-select/src/tag/Frame.jsx
saket2403: Added id as prop in 5bb136d

packages/terra-form-select/src/combobox/Frame.jsx
StephenEsser: We do not want any axe violations if the aria-label is required on the input, I see that you were able to remove the aria-labels from the div and perhaps that is the only one we'll be able to remove

packages/terra-form-select/tests/wdio/select-field-spec.js
```diff
@@ -52,7 +52,7 @@ Terra.describeViewports('Select Field', ['tiny'], () => {
     Terra.it.validatesElement();
 
     it('should open the dropdown by clicking the toggle', () => {
-      browser.click('#select-field');
+      browser.click('[class*="tag"]');
```
StephenEsser: These wildcard selectors can be flaky if we add different classes down the road that include the tag suffix. There are data attributes we may be able to use on the frame and toggle buttons.

saket2403: Resolved in 85bc519

StephenEsser:
This change will likely break existing test cases for consumers, but because this is not an exposed API and the change is being made to fix an accessibility concern I lean towards this being considered passive change. We should get a few maintainers to weigh in before making the final decision.

cc: @mjhenkes @ryanthemanuel @tbiethman @dkasper-was-taken

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -37,6 +37,10 @@ const propTypes = {
    */
   // eslint-disable-next-line react/forbid-prop-types
   dropdownAttrs: PropTypes.object,
+  /**
+   * The id of the input field.
+   */
+  id: PropTypes.string,
```
emilyrohrbough: Should this prop name be scoped with `inputId`? Currently if I was using select, I could pass an `id` to the component via `customProps`. I'd be a little surprised if this `id` value was suddenly being spread on the input.

saket2403: Changed in 22b163c

packages/terra-form-select/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ChangeLog
 =========
+### Changed
+* Id moved to input element instead of select div for searchable select fields.
```
emilyrohrbough: Can we update this changelog entry to be:
```
### Added
* Added `inputId` prop to allow an id to be passed to the input element for searchable select fields.
```

emilyrohrbough:
@StephenEsser  I think this is now passive with updating the prop name from `id` to `inputId`

packages/terra-form-select/CHANGELOG.md
ryanthemanuel:
```suggestion
* Added `inputId` prop to allow an id to be passed to the input element for searchable select fields.
```

packages/terra-form-select/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ChangeLog
 =========
+### Changed
+* Added `inputId` prop to allow an id to be passed to the input element for searchable select fields.
```
avinashg1994: Should be moved under the "Unreleased" section.

saket2403:
htmlFor mapped to `inputId` and prop made private in 48d7bda 
@StephenEsser

packages/terra-form-select/src/ComboboxField.jsx
```diff
@@ -99,6 +99,11 @@ const propTypes = {
    * The value of the select. Can be a string, number, or array of strings/numbers.
    */
   value: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.array]),
+  /**
+   * @private
+   * The id of the input field.
+   */
+  inputId: PropTypes.string,
```
StephenEsser: `inputId` shouldn't be a prop on the Field components, only on the corresponding variants. The id of the Field components should map into the inputId of the corresponding variants.

packages/terra-form-select/src/ComboboxField.jsx
```diff
@@ -176,7 +183,7 @@ const ComboboxField = ({
       isInvalid={isInvalid}
       isInline={isInline}
       isLabelHidden={isLabelHidden}
-      htmlFor={selectId}
+      htmlFor={inputId}
```
StephenEsser: This should stay as 

```
htmlFor={selectId}
```

packages/terra-form-select/src/ComboboxField.jsx
```diff
@@ -186,6 +193,7 @@ const ComboboxField = ({
         aria-describedby={ariaDescriptionIds}
         disabled={selectAttrs.disabled || disabled}
         id={selectId}
+        inputId={inputId}
```
StephenEsser: This should map to the selectId.

```
inputId={selectId}
```

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -618,6 +625,7 @@ class Frame extends React.Component {
       value,
       ...customProps
     } = this.props;
+    delete customProps.inputId;
```
StephenEsser: In general I think `delete` should be used as a last resort. What issues were encountered with deconstructing the `inputId`? I see that delete was introduced in [this commit](https://github.com/cerner/terra-core/pull/2691/commits/48d7bdac314670d0ec95978991e53ebd453a1ebc#diff-b3ca3cbbd16d91fb22a942728a8132aaL607-L629).

```jsx
const {
  inputId,
  ...customProps
}
```

saket2403: `delete` is used to ensure that the `id` is rendered on the input element only and not on the select div

packages/terra-form-select/src/ComboboxField.jsx
saket2403: Resolved in f1d8d8f

packages/terra-form-select/src/ComboboxField.jsx
saket2403: Resolved in f1d8d8f

packages/terra-form-select/src/ComboboxField.jsx
saket2403: Resolved in f1d8d8f

packages/terra-form-select/src/combobox/Frame.jsx
saket2403: Changed in aa361eb

ShettyAkarsh:
@saket2403  you should add the jest test cases for all the searchable select fields variants here.

jeremyfuksa:
JAWS with Windows 10 is not reading the value that was selected via keyboard.

PaulSolDev:

PR #2696, opened 2019-10-11
[terra-button] Add disabled theme styling for Button

### LINKED ISSUE #2183: [terra-button] Add disabled theme styling
## Enhancement Request

For better themeing control, need to add additional style blocks and themeable variables for button styles:
- main one to add is `color` for the disabled `Utility` variant
- need to evaluate with OCS plus other themes if the other variants need styles as well (background-color/border/box-shadow/color)

supreethmr:
### Summary
Resolves #2183 

@cerner/terra


benbcai:
Now that we have theme variables for disabled buttons, do we still want to keep the [opacity of 0.3](https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.module.scss#L74) for disabled buttons? Keeping this would continue to add the opacity on top of what ever background color is used.

cc @neilpfeiffer @jeremyfuksa

claytonbeese:
I'm a bit fuzzy on the interaction between the opacity, the font color, and the background color when this gets themed.  Are we able to control all those attributes independently?

benbcai:
The opacity for disabled and the background color and text color are all themeable and are independent of each other. Since the opacity for disabled is also themeable, perhaps we shouldn't get rid of it.

hboege:
Just to clarify on the variables available for theming disabled buttons.
- There is 1 variable for all variants for disabled-opacity.
- There is 1 variable for each variant for disabled-background-color
- There is 1 variable for each variant for font-color (not disabled-font-color).

jeremyfuksa:
Can someone redeploy this so I can do UX verification? I had hoped fixing the conflict in the changelog would force a rebuild, but it didn't.

packages/terra-button/src/_mixins.scss
```diff
@@ -27,6 +27,11 @@
     color: map-get($color-map, active-color);
   }
 
+  &:disabled,
+  &.is-disabled {
+    background-color: map-get($color-map, disabled-background-color);
```
neilpfeiffer: Missing the original issue of providing themability to the text **color**, mainly in utility style button, not exclusively background-colors. After reviewing recently updated other theme designs, this may no longer be needed. Confirming before merging.

packages/terra-button/src/_variables.scss
```diff
@@ -7,6 +7,7 @@ $terra-button-neutral-color-scheme: (
   active-border-color: var(--terra-button-active-border-color-neutral, #52585b),
   active-color: var(--terra-button-active-color-neutral, #fff),
   background-color: var(--terra-button-background-color-neutral, #dedfe0),
+  disabled-background-color: var(--terra-button-disabled-background-color-neutral, rgb(222, 223, 224)),
```
neilpfeiffer:
```suggestion
  disabled-background-color: var(--terra-button-disabled-background-color-neutral, #dedfe0),
```

packages/terra-button/src/_variables.scss
```diff
@@ -26,6 +27,7 @@ $terra-button-emphasis-color-scheme: (
   active-border-color: var(--terra-button-active-border-color-emphasis, #0065a3),
   active-color: var(--terra-button-active-color-emphasis, #fff),
   background-color: var(--terra-button-background-color-emphasis, #0079be),
+  disabled-background-color: var(--terra-button-disabled-background-color-emphasis, rgb(0, 121, 190)),
```
neilpfeiffer:
```suggestion
  disabled-background-color: var(--terra-button-disabled-background-color-emphasis, #0079be),
```

packages/terra-button/src/_variables.scss
```diff
@@ -83,6 +87,7 @@ $terra-button-action-color-scheme: (
   active-border-color: var(--terra-button-active-border-color-action, transparent),
   active-color: var(--terra-button-active-color-action, #fff),
   background-color: var(--terra-button-background-color-action, #0079be),
+  disabled-background-color: var(--terra-button-disabled-background-color-action, rgb(0, 121, 190)),
```
neilpfeiffer:
```suggestion
  disabled-background-color: var(--terra-button-disabled-background-color-action, #0079be),
```

packages/terra-button/src/_mixins.scss
neilpfeiffer: Confirmed that in addition to background-color, for full control of disabled styling, we should add to the color map:
```
disabled-border-color
disabled-box-shadow
disabled-color
```
For default-theme, match the disabled values to the same current normal values.

supreethmr: Above color maps have been added.

packages/terra-button/src/_variables.scss
supreethmr: changed : dc4ad87808ed40be28c911ec6a214fbde13d07a5

packages/terra-button/src/_variables.scss
supreethmr: changed : dc4ad87808ed40be28c911ec6a214fbde13d07a5

packages/terra-button/src/_variables.scss
supreethmr: changed : dc4ad87808ed40be28c911ec6a214fbde13d07a5

packages/terra-button/src/_mixins.scss
neilpfeiffer: Marking Resolved.

neilpfeiffer:
Will need themes (specifically lowlight) ready before merging.

hboege:
It looks like in the end, 4 new variables were added for the disabled state for each variant of button.
- background-color
- color
- border-color
- box-shadow

However, it does not look like any changes were made to the opacity style in the Button.module.scss.  Will teams be able to properly style with this combination of variables?  
For example, in lowlight, the terra-button-disabled-opacity variable has already been set to 0.3 to cover previous versions of the button. (We can't change that, or we would break old versions of the button.)  If the changes above are released, whatever colors we set the background-color to will also get a 0.3 opacity.  I'm not sure if it is possible to set a color to > 100% opacity so when we take 30% of it, we get the right color.

I might be completely wrong in this, but wanted to be sure the addition of variables and the maintenance of the old opacity styling will allow teams to style the components/controls correctly.  
@neilpfeiffer @claytonbeese what are your thoughts on this?

claytonbeese:
I do have some concerns if the new release will cause the buttons to show BOTH the new colors and the 0.3 opacity.  The colors were intended to match the desired colors and contrasts for the button.  Applying an 0.3 opacity on top of that will cause the disabled buttons to practically disappear against the background.

When the initial colors were requested, I was under the understanding that we would ONLY see the new colors OR the opacity, not both.

neilpfeiffer:
@hboege and @claytonbeese 

The new variables will give _other_ themes who have the ability to make different color choices the latitude to do so. Generally for all existing themes (default-opensource and lowlight) the pattern will be to continue to use the same solid colors (no transparency) as the existing color, with the `terra-button-disabled-opacity` variable as the general master control to indicate dithering/disabled. For other themes, the intent might be to set the `terra-button-disabled-opacity` to `1`  -- but then pass the opacity responsibility to the individual elements with RGBA colors instead to create a different effect. 

This generally will not apply to lowlight theme nor cause concern, since all new color values needed for these variables will match the base versions at full opacity, no changes necessary and will not break old versions.

neilpfeiffer:
Outstanding issues needing to be resolved for Lowlight theme have been resolved and needed PRs have been provided and approved. 

Dismissing Review for @mjhenkes Requested Changes here https://github.com/cerner/terra-core/pull/2696#pullrequestreview-320907427 since now is resolved.

neilpfeiffer:
Reapplying `Do Not Merge` label for @ryanthemanuel to clear. 

Otherwise cleared to Merge when Ready.

supreethmr:
Since we have fusion theme ported, can I update these variables in fusion theme with this PR or do I need to open new PR for adding these variables into fusion theme files. 
@ryanthemanuel

neilpfeiffer:
@supreethmr you can go ahead and include the Fusion theme values with this PR.

supreethmr:
> @supreethmr you can go ahead and include the Fusion theme values with this PR now that it has been ported, I will re-review for UX after you have added them.

Added disabled button theme variables in Fusion-Theme : 4843bbed88a5433dd5811981190fd1d70a5b42d2

hboege:
The lowlight theme has been updated for the new variables.

neilpfeiffer:
UX +1 to additional disabled button theme variables in Fusion-Theme: https://github.com/cerner/terra-core/commit/4843bbed88a5433dd5811981190fd1d70a5b42d2

packages/terra-button/themes/orion-fusion-theme/orion-fusion-theme.scss
```diff
@@ -38,6 +38,11 @@
   --terra-button-font-weight-neutral: 400;
   --terra-button-hover-background-color-neutral: #b2b5b6;
   --terra-button-hover-color-neutral: #000;
+  --terra-button-disabled-background-color-neutral: #dedfe0;
+  --terra-button-disabled-color-neutral: #000;
+  --terra-button-disabled-border-color-neutral: #6f7477;
+  --terra-button-disabled-box-shadow-neutral: inset 0 3rem 1.5rem -2rem #f4f4f4;
+
```
emilyrohrbough: nit: extra line

PR #2698, opened 2019-10-14
Label prop type changed to string

### LINKED ISSUE #2294: Update label to be string prop-type on terra-form-field

saket2403:
### Summary
Fixes #2294


packages/terra-form-checkbox/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
```
StephenEsser: We're a little inconsistent about this, but I think it helps to label this title as `### Breaking Changes` to help standout from passive changes.

saket2403: Resolved in 76b8d32

packages/terra-form-checkbox/docs/UPGRADEGUIDE.md
```diff
@@ -1,5 +1,13 @@
 # Terra Form Checkbox Upgrade Guide
 
+## Changes from version 3 to version 4
+
+### Props
+
+#### Updated
+* `labelText` prop now only accepts string.
+* `legend` prop now only accepts string.
+
```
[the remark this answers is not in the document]
saket2403: Resolved in 76b8d32

packages/terra-form-checkbox/docs/UPGRADEGUIDE.md
```diff
@@ -1,5 +1,12 @@
 # Terra Form Checkbox Upgrade Guide
 
+## Changes from version 3 to version 4
+
+### Props
+
+#### Updated
```
StephenEsser: The Props and Updated sections might need to be reworked to make the description below make a little more sense. I think you had a section before that listed the changes. I think it helps to list the changes and then providing description/guidance for upgrading. 

"What happened?"

"Here's how to upgrade"

saket2403: Updated in b447243

packages/terra-form-field/docs/UPGRADEGUIDE.md
```diff
@@ -1,5 +1,15 @@
...
 
+## Changes from version 3 to version 4
+
+### Props
+
+#### Updated
+* `label` prop's type changed from node to string.
+ 
+##### Steps to Upgrade
+To upgrade from 3.0 to 4.0 each existing Field must be reviewed and updated to ensure only a string is being passed to the `labelText` prop. Solutions passing node data types need to be reworked to provide only a string.
```
avinashg1994: Should be **`label`** since form-field does not have `labelText`.

packages/terra-form-textarea/docs/UPGRADEGUIDE.md
```diff
@@ -1,5 +1,15 @@
...
 
+## Changes from version 3 to version 4
+
+### Props
+
+#### Updated
+* `label` prop's type changed from node to string.
+
+##### Steps to Upgrade
+To upgrade from 3.0 to 4.0 each existing Textarea must be reviewed and updated to ensure only a string is being passed to the `labelText` props. Solutions passing node data types need to be reworked to provide only a string.
```
avinashg1994: Should be **`label`** since form-field does not have `labelText`.

packages/terra-form-field/docs/UPGRADEGUIDE.md
saket2403: Changed 83b428d

packages/terra-form-textarea/docs/UPGRADEGUIDE.md
saket2403: Changed 83b428d

packages/terra-form-checkbox/docs/UPGRADEGUIDE.md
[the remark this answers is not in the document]
supreethmr: I think we can provide code change example from v3 to v4 like [this](https://github.com/cerner/terra-core/blob/master/packages/terra-avatar/docs/upgrade-guide.md#v1-code)

packages/terra-form-checkbox/docs/UPGRADEGUIDE.md
```diff
@@ -1,5 +1,16 @@
...
+* `labelText` prop's type changed from node to string.
+* `legend` prop's type changed from node to string.
+
+##### Steps to Update
```
emilyrohrbough: Can we add some context as to why this change was make?

saket2403: Added in 6db571a

PR #2704, opened 2019-10-15
[terra-overlay] Updating tests to cover update lifecycle changes

tbiethman:
### Summary
Thought of a decent way to test the previous changes to the update lifecycle handling. I verified that the pre-fix code failed these tests.


packages/terra-overlay/tests/wdio/overlay-spec.js
```diff
@@ -69,6 +69,38 @@ Terra.describeViewports('Overlay', ['huge'], () => {
...
+      });
+
+      Terra.it.validatesElement('open', { selector: '#terra-Overlay--fullscreen' });
+
+      it('does not increment overlay count during update', () => {
+        browser.execute('const updateEvent = new CustomEvent("overlay.forceUpdateForTest"); document.dispatchEvent(updateEvent);');
+
+        expect(browser.getAttribute('#root', 'inert')).to.equal('true');
+        expect(browser.getAttribute('#root', 'aria-hidden')).to.equal('true');
+        expect(browser.getAttribute('#root', 'data-overlay-count')).to.equal('1');
```
tbiethman: This is the value that was previously incrementing during update and causing the problems

PR #2706, opened 2019-10-16
[terra-list] Accessibility Issue fixed to handle selectable and non-selectable items correctly

### LINKED ISSUE #2664: [terra-list] Handle selectable & disabled list items
# Feature Request

## Description

Update terra list to handle selectable and disabled list items.

While I was upgrading my project to terra-toolkit v5, I ran into the following accessibility violation:
image. 

I encountered this issue with the following [list](https://github.cerner.com/orders-plans/orion-orders-js/blob/master/src/bundles/order-details/components/views/diagnoses/diagnoses-list/AssociateDiagnosesView.jsx#L90) and [items](https://github.cerner.com/orders-plans/orion-orders-js/blob/master/src/bundles/order-details/components/views/diagnoses/diagnoses-list/DiagnosesListItem.jsx#L36). I reached out to @dkasper-was-taken on [teams](https://teams.microsoft.com/l/message/19:ec7057f0024d451194d19071dc151f86@thread.skype/1569423448435?tenantId=fbc493a8-0d24-4454-a815-f4ca58e8c09d&groupId=45f47bfa-38a9-444d-8933-9c20266ae38c&parentMessageId=1569423448435&teamName=Terra&channelName=Testing%20Support&createdTime=1569423448435) and he came up with a  workaround that solved my problem using custom props. Following his suggestion, I logged an issue here for a potential enhancement and/or further discussion.

## @ Mentions

@dkasper-was-taken

saket2403:
### Summary
Fixes #2664


packages/terra-list/src/List.jsx
```diff
@@ -58,10 +58,17 @@ const List = ({
   if (role && role.length > 0 && role !== 'none') {
     attrSpread.role = role;
   }
+  const allItems = children ? [] : undefined;
 
+  React.Children.forEach(children, (child) => {
```
dkasper-was-taken: We do not want to clone each child item here and it will degrade performance and cause the list items to always re-render on update.

packages/terra-list/src/ListItem.jsx
```diff
@@ -94,6 +94,10 @@ const ListItem = ({
     attrSpread['data-item-show-focus'] = 'true';
     attrSpread.onBlur = ListUtils.wrappedEventCallback(onBlur, event => event.currentTarget.setAttribute('data-item-show-focus', 'true'));
     attrSpread.onMouseDown = ListUtils.wrappedEventCallback(onMouseDown, event => event.currentTarget.setAttribute('data-item-show-focus', 'false'));
+  } else if (customProps.listrole === 'listbox') {
```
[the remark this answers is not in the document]
saket2403: There is no `isDisabled` prop in ListItem , so what do you mean by a disabled item?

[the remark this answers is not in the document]
dkasper-was-taken: I've referring to the aria states. aria-selected="true" and aria-disabled="true" can be valid at the same time. So we cannot hard code the selected state to false.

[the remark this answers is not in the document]
saket2403: But shouldn't aria-selected be false for a non-selectable item?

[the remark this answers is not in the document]
dkasper-was-taken: There are a few workflow concepts that need to be provided for involving an interactable item that cannot have its state altered. An option that is either defaulted/selected in conjunction with another option, a disabled list with currently selected values, and an item within an intermediate state with updates pending either from a service or action being carried out.

packages/terra-list/src/List.jsx
saket2403: Added condition in c967e3a

packages/terra-list/src/ListItem.jsx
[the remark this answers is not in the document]
saket2403: Resolved in c967e3a

packages/terra-list/src/List.jsx
```diff
@@ -58,10 +58,20 @@ const List = ({
   if (role && role.length > 0 && role !== 'none') {
     attrSpread.role = role;
   }
-
+  let allItems = children ? [] : undefined;
+  if (role === 'listbox') {
+    React.Children.forEach(children, (child) => {
```
dkasper-was-taken: We don't want to loop through children and clone elements here. The goal is to have to consuming HOC set all relevant data on the list and list items. This reduces the impact of large lists. If a list item needs a prop set it should only the first time by the consumer.

saket2403: Added Prop in df874d5

packages/terra-list/src/List.jsx
saket2403: Added Prop in df874d5

packages/terra-list/src/ListItem.jsx
```diff
@@ -54,13 +54,18 @@ const propTypes = {
    * @private Callback function not intended for use with this API, but if set pass it through to the element regardless.
    */
   onMouseDown: PropTypes.func,
+  /**
+   * Set to true for non-selectable items in a list having role as listbox.
+   */
+  isRoleListbox: PropTypes.bool,
```
nramamurth: This is confusing to me. 

- Why do we need this prop at all?
- Don't we recommend using 'listbox' as the role for all lists for accessibility? If not, why?
- The description 'Set to true for non-selectable items in a list having role as listbox.' doesn't sound right to me.

packages/terra-list/src/ListItem.jsx
```diff
@@ -94,6 +100,10 @@ const ListItem = ({
     attrSpread['data-item-show-focus'] = 'true';
     attrSpread.onBlur = ListUtils.wrappedEventCallback(onBlur, event => event.currentTarget.setAttribute('data-item-show-focus', 'true'));
     attrSpread.onMouseDown = ListUtils.wrappedEventCallback(onMouseDown, event => event.currentTarget.setAttribute('data-item-show-focus', 'false'));
+  } else if (isRoleListbox) {
```
nramamurth: What am I missing if we simply do ⬇️? I am assuming that the list item is "disabled" if it is not selectable. 

```

const attrSpread = {
   role: 'option',
   'aria-disabled': isSelectable,  
};

if (isSelectable) {
    attrSpread.onClick = ListUtils.wrappedOnClickForItem(onClick, onSelect, metaData);
    attrSpread.onKeyDown = ListUtils.wrappedOnKeyDownForItem(onKeyDown, onSelect, metaData);
    attrSpread.tabIndex = '0';
    attrSpread.role = 'option';
    attrSpread['aria-selected'] = isSelected;
    attrSpread['data-item-show-focus'] = 'true';
    attrSpread.onBlur = ListUtils.wrappedEventCallback(onBlur, event => event.currentTarget.setAttribute('data-item-show-focus', 'true'));
    attrSpread.onMouseDown = ListUtils.wrappedEventCallback(onMouseDown, event => event.currentTarget.setAttribute('data-item-show-focus', 'false'));
}
```

packages/terra-list/src/ListItem.jsx
saket2403: Listbox is set as role for list only if it has items that are selectable as described [here](https://github.com/cerner/terra-core/blob/01cba497294d6b055d4482275a1689b85cef52ca/packages/terra-list/src/List.jsx#L28)

packages/terra-list/src/ListItem.jsx
saket2403: That would mean setting role : `option` for all items which will give axe violations for all other cases as pointed out  by Dave in this [comment](https://github.com/cerner/terra-core/issues/2664#issuecomment-541790941)

nramamurth: Fair enough. Then add the role iff `isSelectable` is true but I am trying to understand why we cannot leverage the value of `isSelectable` to set `aria-disabled`.

packages/terra-list/src/ListItem.jsx
```diff
@@ -94,6 +100,10 @@ const ListItem = ({
...
+    attrSpread.role = 'option';
+    attrSpread['aria-selected'] = isSelected;
+    attrSpread['aria-disabled'] = 'true';
```
emilyrohrbough: Why would we assign aria-disabled if an item is selected? I would expect disabled to be applied if the item was disabled from receiving a click event

packages/terra-list/src/ListItem.jsx
emilyrohrbough: I think @dkasper-was-taken 's suggestion on this issue was was to give consumers a way to indicate if a an item in a selectable list was disabled or not via an `isDisabled` prop. I'm not sure consumers would know when to use `isRoleListbox`:

```jsx
<List  role="listbox"/>
   // this item needs aria-disabled
   <Item isSelectable isDisabled />
   // this item needs aria-disabled
   <Item isSelectable={false}  isDisabled />
   <Item isSelectable isSelected />
</ List>
```

This would look something like:
```diff
const attrSpread = {};
  if (isSelectable) {
    attrSpread.onClick = ListUtils.wrappedOnClickForItem(onClick, onSelect, metaData);
    attrSpread.onKeyDown = ListUtils.wrappedOnKeyDownForItem(onKeyDown, onSelect, metaData);
    attrSpread.tabIndex = '0';
    attrSpread.role = 'option';
+   attrSpread['aria-disabled'] = isDisabled
    attrSpread['aria-selected'] = isSelected;
    attrSpread['data-item-show-focus'] = 'true';
    attrSpread.onBlur = ListUtils.wrappedEventCallback(onBlur, event => event.currentTarget.setAttribute('data-item-show-focus', 'true'));
    attrSpread.onMouseDown = ListUtils.wrappedEventCallback(onMouseDown, event => event.currentTarget.setAttribute('data-item-show-focus', 'false'));
  }
```

packages/terra-list/src/ListItem.jsx
emilyrohrbough: @neilpfeiffer Should we be applying disabled styles for this concept or is the non-selectable styles enough to differentiate this non-clickable item behavior?

dkasper-was-taken:
After discussion with UX we don't want to derive disabled styles using the role. We shouldn't leverage isSelectable and a newly created isDisabled prop. The role of the ListItem should be an override, not present in the API, but instead extracted from the customProps. The customProps role would override the default value of null/listitem and the selectability role of option. The customProps value of role, if present, would be unpacked similar to how we allow for the addition of className to our classes.

We also only want to add the disabled aria entry if both isSelectable and isDisabled, otherwise we don't want it decorated on the DOM as the browser already infers the positive setting.

I would expect it to look something like this.

```diff
  const attrSpread = {};
  if (isSelectable) {
+    if (isDisabled) {
+     attrSpread['aria-disabled'] = true;
+    } else {
      attrSpread.onClick = ListUtils.wrappedOnClickForItem(onClick, onSelect, metaData);
      attrSpread.onKeyDown = ListUtils.wrappedOnKeyDownForItem(onKeyDown, onSelect, metaData);
      attrSpread.tabIndex = '0';
-     attrSpread.role = 'option';
+     attrSpread.role = customProps.role ? customProps.role : 'option';
-     attrSpread['aria-selected'] = isSelected;
      attrSpread['data-item-show-focus'] = 'true';
      attrSpread.onBlur = ListUtils.wrappedEventCallback(onBlur, event => event.currentTarget.setAttribute('data-item-show-focus', 'true'));
      attrSpread.onMouseDown = ListUtils.wrappedEventCallback(onMouseDown, event => event.currentTarget.setAttribute('data-item-show-focus', 'false'));
    }
+   attrSpread['aria-selected'] = isSelected;
  }


UX has yet to determine if there is to be a visual change to accompany the work of adding disabled to list & table.

**Comment by saket2403** [comment]
>I would expect it to look something like this.

According to this change, `role='option'` is not getting set for items having `isSelectable = false` and are present in a list having `role='listbox'`

File: packages/terra-list/src/ListItem.jsx (hunk 5 above)
Comment (nramamurth) [reply packages/terra-list/src/ListItem.jsx]: That's the same line of thought that I had too and also simply assumed that if an item is not selectable (`isSelectable` is truthy), then `aria-disabled` should automatically be true. This way, we needn't really add a new prop.

**Comment by ShettyAkarsh** [comment]
> > I would expect it to look something like this.
> 
> Do we want to set `isSelectable` and `isDisabled` for a non-selectable item?
> If so here: [afe54e1](https://github.com/cerner/terra-core/commit/afe54e1fd12a4b07e315b47f0f1a2ab7fd99b081)


In this case user might set  isdisabled prop when the  purpose of the item is to be non-selectable. i tried the same with this new code and the same error occurs when we run the axe-core. as we are setting aria-disabled only when isselectable and isdisabled prop is set.  in this case we might have to add one more else for the main if condition to set

```
if (isSelectable) {
}
else {
    if (isDisabled) {
      attrSpread['aria-disabled'] = true;
    }
    attrSpread.role = 'option';
  }

```
cc. @dkasper-was-taken

File: packages/terra-list/src/ListItem.jsx (hunk 6 above)
Comment (ShettyAkarsh) [reply packages/terra-list/src/ListItem.jsx]: @neilpfeiffer any updates on this disabled styles ?

**Comment by saket2403** [comment]
`role='option'` should only be set for items with `isSelectable= true`. This change would fail all other cases where role of `list` is not `listbox`. @ShettyAkarsh .

**Comment by dkasper-was-taken** [comment]
@ShettyAkarsh
Role `option` and `aria-selected` needs to be placed on every child in a list with the role of `listbox`.  `aria-disabled` is only a valid on an interactable list item. So in this case any use of `isDisabled` outside of a `isSelectable` listitem isn't valid, and should be suppressed. 

```
<ul role="listbox" aria-label="<required-label>">
  <li aria-selected="false" role="option" />
  <li aria-selected="true" role="option" />
  <li aria-selected="false" role="option" aria-disabled="true" />
  <li aria-selected="true" role="option" aria-disabled="true" />
</ul>
```
or 

```
<ul> // defaulted to 'role="list"'
  <li /> // defaulted to 'role="listitem"'
  <li />
  <li />
  <li />
</ul>
```

We also want to avoid placing aria-disabled unnecessarily onto the DOM as all modern browsers use `aria-diabled="false"` as the implicit value, and can cause issues in some obscure cases.

**Comment by ShettyAkarsh** [comment]
> @ShettyAkarsh
> Role `option` and `aria-selected` needs to be placed on every child in a list with the role of `listbox`. `aria-disabled` is only a valid on an interactable list item. So in this case any use of `isDisabled` outside of a `isSelectable` listitem isn't valid, and should be suppressed.
> 
> ```
> <ul role="listbox" aria-label="<required-label>">
>   <li aria-selected="false" role="option" />
>   <li aria-selected="true" role="option" />
>   <li aria-selected="false" role="option" aria-disabled="true" />
>   <li aria-selected="true" role="option" aria-disabled="true" />
> </ul>
> ```
> 
> or
> 
> ```
> <ul> // defaulted to 'role="list"'
>   <li /> // defaulted to 'role="listitem"'
>   <li />
>   <li />
>   <li />
> </ul>
> ```
> 
> We also want to avoid placing aria-disabled unnecessarily onto the DOM as all modern browsers use `aria-diabled="false"` as the implicit value, and can cause issues in some obscure cases.

Oh,  so in that we might have to update the doc for the isDisabled Prop here.

`  /**
   * Whether or not the list item is disabled.
   */
  isDisabled: PropTypes.bool,`

 something like valid when isSelectable prop is set.

**Comment by saket2403** [comment]
> Oh, so in that case we might have to update the doc for the isDisabled Prop here.

Updates docs in 8cec3d1

File: packages/terra-list/docs/ListItem.md [hunk 7]
```diff
@@ -1,6 +1,6 @@
 # List Item
 
-The list item provides a pattern for selection/selectability for li surrounded child content. Whenever, the list item has a prop `isSelectable=true` then `aria-label` must be provided to List for accessibility.  The onSelect event is the primary means for interaction, as it yields the associated metaData when valid selection occurs, though individual onClick and onKeyDown events can be applied through the use of customProps if greater granularity is required.
+The list item provides a pattern for selection/selectability for li surrounded child content. Whenever, the list item has a prop `isSelectable=true` then `aria-label` must be provided to List for accessibility.  The onSelect event is the primary means for interaction, as it yields the associated metaData when valid selection occurs, though individual onClick and onKeyDown events can be applied through the use of customProps if greater granularity is required. Both `isSelectable=true` and `isDisabled=true` should be set for a non-selectable item in a list containing a combination of both selectable and non-selectable items, to prevent any accessibility issues.
```
emilyrohrbough: > Both `isSelectable=true` and `isDisabled=true` should be set for a non-selectable item in a list containing a combination of both selectable and non-selectable items, to prevent any accessibility issues.

This seems weird. I would expect to only need to supply `isDisabled=true`

emilyrohrbough: I take it back, this makes sense because you should only disabled a selectable item

packages/terra-list/src/ListItem.jsx
```diff
@@ -54,13 +54,18 @@ const propTypes = {
    * @private Callback function not intended for use with this API, but if set pass it through to the element regardless.
    */
   onMouseDown: PropTypes.func,
+  /**
+   * Whether or not the list item is disabled.
```
emilyrohrbough: Based on my confusion above, lets enhance this doc to mention this will only be applied to selectable list items/ when `isSelectable` is defined.

saket2403: Updated in bc302c1

saket2403:
> Looks like we are missing a wdio test with a disabled item that validates this accessibility issue is resolved

Updated wdio in bc302c1

packages/terra-list/src/terra-dev-site/doc/example/ListItem.jsx
```diff
@@ -10,12 +10,16 @@ const ListSectionExample = () => (
   <List role="listbox" aria-label="example-label">
```
supreethmr: I think we might need to consider adding aria-label as prop for List similar to code changes that are made to fix axe violation in [Infinite-List] (https://github.com/cerner/terra-framework/pull/999).

saket2403: In list aria-label can be passed by custom-prop to the `ul` element so no need of an extra prop.

packages/terra-list/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* `isDisabled` prop added to fix accessibility issue to handle selectable and non-selectable items correctly.
```
emilyrohrbough: Should there be disabled styles applied to the list item when isDisabled = true?

packages/terra-list/src/ListItem.jsx
```diff
@@ -75,6 +81,7 @@ const ListItem = ({
   onMouseDown,
   onSelect,
   refCallback,
+  isDisabled,
   ...customProps
 }) => {
   const listItemClassNames = cx([
```
dkasper-was-taken: We'll want to ensure we don't display a disabled element as having interaction from hover styles as well.
```diff
- { 'is-selectable': isSelectable },
+ { 'is-selectable': !isDisabled && isSelectable },
```

saket2403: Changed 1c0de15

saket2403:
> We may want to incorporate ListSections into this before we merge it in. Teams are having a similar issue with SectionHeaders.
> 
> The role="header" logic would then need to be moved to a newly created internal div.

Added here 36c7e1a

packages/terra-list/src/ListSection.jsx
```diff
@@ -37,29 +37,40 @@ const propTypes = {
...
   }
+  const attrSpread = {};
+  if (role && role.length > 0 && role !== 'none') {
+    attrSpread.role = role;
+  }
 
   return (
     <React.Fragment>
-      <SectionHeader {...customProps} isCollapsible={isCollapsible} isCollapsed={isCollapsed} />
+      <SectionHeader {...attrSpread} {...customProps} isCollapsible={isCollapsible} isCollapsed={isCollapsed} />
```
emilyrohrbough: Could we directly pass in role instead of adding it to an attrSpread object?

saket2403: Yes, removed attrSpread here d403d32

packages/terra-list/docs/ListItem.md
nramamurth: @saket2403 I still think that the phrase should be reworded. Why would we want to set `isSelectable=true` for an item that is non-selectable?

Did you mean to say something like this? ⬇️

> Both `isSelectable=true` and `isDisabled=true` should be set to represent the disabled state of a selectable item in a list containing a combination of both selectable and non-selectable items, to prevent any accessibility issues.

packages/terra-list/docs/ListSection.md
```diff
@@ -1,6 +1,6 @@
 # List Section
 
-The list section component creates a section header and layouts it out with the passed child list items. The section provides the ability to display and collapse a section within a logical group. The list section will render as a flat list.
+The list section component creates a section header and layouts it out with the passed child list items. The section provides the ability to display and collapse a section within a logical group. The list section will render as a flat list.Both `isSelectable=true` and `isDisabled=true` should be set for a non-selectable item in a list containing a combination of both selectable and non-selectable items to prevent any accessibility issues. Use `role=listbox` if section has selectable items and no parent `List` element. Use `role=option` if section has selectable items and parent `List` element with `role=listbox`.
```
nramamurth: Let's have bullet points for the last three sentences. I think it is more readable and looks separated out.

saket2403: Added bullets b154202

packages/terra-list/src/ListSubsection.jsx
```diff
@@ -59,7 +66,7 @@ const ListSubsection = ({
 
   return (
     <React.Fragment>
-      <SubsectionHeader {...customProps} isCollapsible={isCollapsible} isCollapsed={isCollapsed} />
+      <SubsectionHeader {...customProps} role={role} isCollapsible={isCollapsible} isCollapsed={isCollapsed} />
```
dkasper-was-taken: This isn't needed, as customProps will already spread the value.

packages/terra-list/src/ListSectionHeader.jsx
```diff
@@ -107,6 +114,9 @@ const ListSectionHeader = ({
     attrSpread.onBlur = ListUtils.wrappedEventCallback(onBlur, event => event.currentTarget.setAttribute('data-item-show-focus', 'true'));
     attrSpread.onMouseDown = ListUtils.wrappedEventCallback(onMouseDown, event => event.currentTarget.setAttribute('data-item-show-focus', 'false'));
   }
+  if (role === 'option' || role === 'listbox') {
```
dkasper-was-taken: 'listbox' isn't a invalid role. Likely here we want to just do presence check, and if a role exists, override the role value.

We'll also want to add the isDisabled behavior/props to this as well.

packages/terra-list/src/ListSubsection.jsx
saket2403: Changed 15166a3

packages/terra-list/src/ListSectionHeader.jsx
saket2403: Added

packages/terra-list/src/ListSectionHeader.jsx
```diff
@@ -107,6 +121,12 @@ const ListSectionHeader = ({
     attrSpread.onBlur = ListUtils.wrappedEventCallback(onBlur, event => event.currentTarget.setAttribute('data-item-show-focus', 'true'));
     attrSpread.onMouseDown = ListUtils.wrappedEventCallback(onMouseDown, event => event.currentTarget.setAttribute('data-item-show-focus', 'false'));
   }
+  if (role && role.length > 0 && role !== 'none') {
```
dkasper-was-taken: This can be cleaned up if we don't default a role. A simple
```
if (role) {
```
would suffice.

packages/terra-list/src/ListSectionHeader.jsx
```diff
@@ -73,6 +85,8 @@ const ListSectionHeader = ({
   onSelect,
   refCallback,
   title,
+  role,
+  isDisabled,
   ...customProps
 }) => {
   const sectionHeaderClassNames = cx([
```
dkasper-was-taken: Sections/Groups within lists can get complicated. With some further research we’ll need

The role of "heading" needs to move to a div or h# tag internal to the li.

Next, I believe there are 4 scenarios we need to validate for, so tests for each might be appropriate. I’ve listed some considerations for each.

Scenario 1: Default List with Static Section header
  - li with role of "presentation" or "none
  - A child h# tag containing the title

Scenario 2: Default List with Collapsible Section Header
  - li with role of "presentation"
  - A child interactive element, perhaps a button, containing a child h# tag containing the title
  - A child ul containing the relation options? Or simple the remaining grouped li’s with role option contained within the presentation?

Scenario 3: Interactive List with Static Section Header
  - li with role of "presentation"
  - A child h# tag containing the title

Scenario 4: Interactive List with Collapsible
  - li with role of "presentation" or a role of “option”? Needs further testing.
  - A child interactive element, perhaps a button, containing a child h# tag containing the title
  - A child ul containing the relation options? Or simple the remaining grouped li’s with role option contained within the presentation?

With the pending deprecation of the terra-infinite we no longer need to worry about flat list of list items. We may need to evaluate the inclusion of nested ul’s again, I know it had some issues within iOS at the time. Here is a good reference:
https://w3c.github.io/aria-practices/examples/listbox/listbox-grouped.html

packages/terra-list/src/ListSectionHeader.jsx
saket2403: Changed cf4a187

packages/terra-list/src/ListSectionHeader.jsx
saket2403: Moved `role="heading"` to title element. cf4a187
Role= "presentation" is giving accessibility issue with parent having role="listbox", which is required if any or all other items have role as option.

dkasper-was-taken: Do you have an example of the error? There are other changes, like moving the tab-index, aria-expanded, etc all moving to internal elements. Also potentially moving all child lis inside another ul presentations as well.

dkasper-was-taken: My guess is the error of having no options at all, when all options are collapse or only the section header is used in the example. There may need to be more testing to figure out the correct roles, but the goal for this is that the section header shouldn't be read as/or totalled in the listbox counts.

saket2403: In the list-section example when we have selectable items, having section role as `option` means that the Voice Over does not read it as heading, but if we have heading as role then there are accessibility violations saying that the role should be option as the list's role is listbox.

dkasper-was-taken: We definitely don't want the role of 'heading' on the item, and more than likely 'option' isn't the correct role either. You will need to spike on what the proper fix is. Like I mentioned above, combing a presentation with a interior button that has the header inside may be correct for the section header itself, but in the case of all sections collapsed we will no longer have an option present within the 'listbox', so we will either need to revise that strategy to merely hide the list items, or find a way to get the screen reader to still read how many items are within the listbox even with them no longer on the DOM.

dkasper-was-taken:
Closing as a result of discussion with UX:
https://github.com/cerner/terra-core/issues/2664#issuecomment-641667078

PR #2707, opened 2019-10-17
[terra-form-select]Wrong options selected in mobile view

### LINKED ISSUE #2684: [terra-form-select] Variants with an input (all except single select) are causing rerenders on scroll which calls scrollIntoView
# Bug Report

## Description

Select variants with an input (all except single select) are causing rerenders on scroll which calls `Menu.scrollIntoView`. This causes scroll jumping and also prevents you from being able to properly select all items in large lists. The list will scroll immediately to the active option before an option is selected, causing the wrong option to be selected. This behavior appears after the big internal refactor (e.g. v5.25.0). While this issue is hard to replicate on desktop (generally, the active option is set via mouse hover over an option when scrolling and we don't scroll in this case), this is doubly bad on mobile. See example video [here](https://youtu.be/0-kbggVR-mg)

## Expected Behavior

No scroll jump when scrolling.

## Possible Solution

We probably don't need to call `this.scrollIntoView()` on each `componentDidUpdate()` (see [here](https://github.com/cerner/terra-core/blob/terra-form-select%405.32.0/packages/terra-form-select/src/multiple/Menu.jsx#L197)). This was likely done to guarantee that all state changes were flushed before trying to scroll to the active option.

Instead, we should just call `this.scrollIntoView()` on changes to the `active` state var (using `this.setState`'s callback arg), e.g. on key up:

```jsx
handleKeyDown(event) {
  if (event.keyCode === KeyCode.KEY_UP) {
    this.setState(
      { active: MenuUtil.findPrevious(this.state.children, this.state.active) },
      () => this.scrollIntoView()
    );
  }
  // ...
}
```

## Environment

* Component Name and Version: terra-form-select v5.32.0

lokesh-0813:
### Summary
Fix for form-select wrong options being selected when user scrolls through the menu and selects an option in mobile view

Fixes #2684


packages/terra-form-select/src/combobox/Menu.jsx
```diff
@@ -296,12 +296,16 @@ class Menu extends React.Component {
...
-      this.setState({ active: MenuUtil.findPrevious(children, active) });
+      this.setState({ active: MenuUtil.findPrevious(children, active) }, () => {
+        this.scrollIntoView();
+      });
       this.updateCurrentActiveScreenReader();
     } else if (keyCode === KeyCode.KEY_DOWN) {
       this.clearScrollTimeout();
       this.scrollTimeout = setTimeout(this.clearScrollTimeout, 500);
-      this.setState({ active: MenuUtil.findNext(children, active) });
+      this.setState({ active: MenuUtil.findNext(children, active) }, () => {
```
StephenEsser: The callback function can be passed without an anonymous call. 

```jsx
this.setState({}, this.scrollIntoView)
```

packages/terra-form-select/src/combobox/Menu.jsx
```diff
@@ -153,7 +154,6 @@ class Menu extends React.Component {
 
   componentDidUpdate() {
     this.updateNoResultsScreenReader();
-    this.scrollIntoView();
```
ShettyAkarsh: so we do not need this call during searching of an item and then selecting it ? is it still working fine as before ?

lokesh-0813: Yes the functionality remains the same.

packages/terra-form-select/src/combobox/Menu.jsx
lokesh-0813: Updated here c5d5a69bb0c00408c9a02f2abaed7a5b5050c938

StephenEsser:
We should work on creating consistent commit messages. This is probably a larger discussion, but there are ~3 different commit message formats open full pull request right now. 

- [terra-form-select] - (Preserving the terra prefix)
- [form-select] - (excluding the terra prefix)
- [List] - (excluding the terra prefix but capitalizes the component name)

nramamurth:
@StephenEsser +1, and there are a few that do not have a component name at all.

yuderekyu:
> We should work on creating consistent commit messages. This is probably a larger discussion, but there are ~3 different commit message formats open full pull request right now.
> 
> * [terra-form-select] - (Preserving the terra prefix)
> * [form-select] - (excluding the terra prefix)
> * [List] - (excluding the terra prefix but capitalizes the component name)

I lean towards the first option to be 1-1 with our components.

packages/terra-form-select/CHANGELOG.md
```diff
@@ -3,7 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
-
+### Fixed
+* Fixed re-renders of scroll after every componentDidUpdate call.
 5.32.0 - (October 15, 2019)
```
ryanthemanuel: `* Fixed re-renders of scroll after every componentDidUpdate call.

5.32.0 - (October 15, 2019)`

PR #2715, opened 2019-10-21
[terra-overlay] Add aria screenreader a11y to Loading Overlay.

### LINKED ISSUE #2540: [terra-overlay] Add aria screenreader a11y to loading overlay
# Bug Report

## Description
Loading Overlay currently can have text to support a visual loading indicator, but text is not required. Need to uplift loading-overlay to have proper aria-live, description, etc. for screenreader AT when present on screen, such that all users have awareness/understanding of the loading state.

nramamurth:
### Summary
Fixes #2540.


packages/terra-overlay/src/LoadingOverlay.jsx
```diff
@@ -59,7 +59,7 @@ const LoadingOverlay = ({
   delete customProps.onRequestClose;
 
   return (
-    <Overlay {...customProps} className={cx('loading-overlay', customProps.className)} rootSelector={rootSelector}>
+    <Overlay {...customProps} className={cx('loading-overlay', customProps.className)} rootSelector={rootSelector} aria-live="polite">
```
yuderekyu: should we toggle `aria-busy` based on `isOpen`?

nramamurth: Good point. If we need to do that, then shouldn't it be set on the parent node or the container that is _literally_ loading / busy?

yuderekyu: Yeah, we probably need documentation that points this out for containers

nramamurth: Right. 

@tbiethman @mmalaker What are your thoughts on this?

tbiethman: Would we gain anything by marking it with `aria-busy`? Based on what I'm reading [here](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-busy), it seems as though we might be getting the same functional result by inerting the content behind the overlay (like we do already).

nramamurth: I tried adding it to the container and it didn't make a difference to me.

tbiethman: @scottwilmarth do you know if aria-busy is something we could/should generically set on content underneath our overlays? Even if that content is already made inert and unaccessible?

scottwilmarth:
@tbiethman and I chatted about this issue and whether or not to [add aria-busy](https://www.w3.org/TR/wai-aria-1.0/states_and_properties#aria-busy) to the overlay. I can't say whether this is appropriate to use or not. Is there risk in not adding it for now? If not, my thought is we hold off and re-evaluate later. Whatever is done, it must be tested to ensure the user's expectation is met. If there is harm in not adding something, aria-busy may be the option to try first.

nramamurth:
@scottwilmarth 

> Is there risk in not adding it for now?

I did not find a difference in the screen-reader behavior with/without it. I'd appreciate it if you could test the functionality out in the deployed site and let me know if the current change (sans `aria-busy`) looks good.

ryanthemanuel:
@nramamurth I'd say let's just go ahead and hold off on adding it for now.

PaulSolDev:

PR #2722, opened 2019-10-24
Fixed axe violation for keyboard accessibility of scrollable content

### LINKED ISSUE #2716: Axe Violation for keyboard accessibility of scrollable content
## Description
Axe Core fails in certain components giving violation for keyboard accessibility.

## Error Report
```
Elements that have scrollable content should be accessible by keyboard 
```
## Components

- Content   Container
- Props Table
- Scroll
- Dialog

saket2403:
### Summary
Fixes #2716 

## Additional Details
Changes in terra-scroll also fixes the issue for terra-dialog and terra-content-container


packages/terra-scroll/src/Scroll.jsx
```diff
@@ -21,7 +21,8 @@ const Scroll = ({
   refCallback,
   ...customProps
 }) => (
-  <div {...customProps} className={cx(['scroll', customProps.className])} ref={refCallback}>
+  // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+  <div {...customProps} tabIndex="0" className={cx(['scroll', customProps.className])} ref={refCallback}>
```
dkasper-was-taken: TabIndex needs to be selectively added. ContentContainers are used in nested fashion and we don't want all containers to supply a tabIndex. This also affects how an app and main elements are structured. The behavior will need to be evaluated when we the accessibility pass is implemented.

benbcai: @saket2403 Adding tabIndex here is causing the these [button-group](https://github.com/cerner/terra-core/blob/master/packages/terra-button-group/tests/wdio/button-group-spec.js#L41-L67) tests to fail because the first tab would put focus on the terra-scroll and the second tab would then put focus on the first button in the button group.

saket2403: Changed Implementation in 0823ae5
cc @benbcai  @dkasper-was-taken

benbcai:
@saket2403 There are some test examples in terra-image and terra-responsive-element that are also having this violation when using axe-core 3.4.0. These are the [test examples](https://github.com/cerner/terra-core/compare/keyborad-a11y...axe-violations) I see that needs updated.

packages/terra-content-container/src/ContentContainer.jsx
```diff
@@ -54,7 +54,7 @@ const ContentContainer = ({
     <div {...customProps} className={contentLayoutClassNames}>
       {header && <div className={cx('header')}>{header}</div>}
       <div className={cx('main')}>
-        <Scroll className={cx('normalizer')} refCallback={scrollRefCallback}>
+        <Scroll tabIndex={fill ? 0 : -1} className={cx('normalizer')} refCallback={scrollRefCallback}>
```
jeremyfuksa: When clicking on this object in the standard container, the outline shows up.  `Scroll` needs to have `outline: none` assigned to it.

dkasper-was-taken: We aren't going to be able to do this broadly. Multiple nested content container will have major issues with tabindexes and break accessibility. This needs evaluation at the framework level. For the time being we may need to find a way to alter the config to turn off this axe rule until such time as we can circle back to this.

benbcai:
After discussing the implications of the changes on this PR for addressing [this rule](https://dequeuniversity.com/rules/axe/3.4/scrollable-region-focusable) introduced in axe-core-3.3.0, we decided to hold off on these changes and disable this new rule for now in [terra-toolikit](https://github.com/cerner/terra-toolkit/issues/343). We will revisit this rule once we have a better plan on how to address it. I will close this PR after I disable the rule to ensure the violations no longer appear.

benbcai:
Closing this and we will visit how to properly handle accessibility with scrollable content when we tackle #2787.

PR #2714, opened 2019-10-21
[terra-action-header] Set buttons to utility variant.

nramamurth:
### Summary
Closes #1835.

PR #2739, opened 2019-11-07
Added New Theme Variable to set padding of action buttons

### LINKED ISSUE #2738: Action Buttons in Action header are not properly aligned with fusion theme
# Bug Report
Action Buttons in Action header are misaligned in fusion theme.

## Description
Right and Left action buttons are not center aligned like title in header. this could be because of missing variable for setting padding of action buttons in action header unlike [title](https://github.com/cerner/terra-core/blob/fa60de1861435c35d1bff8d0de75cd36721b07d1/packages/terra-action-header/src/ActionHeaderContainer.module.scss#L52).

Screen Shot 2019-11-07 at 11 25 24 PM

## Steps to Reproduce
1. Open  Action Header with Fusion Theme

## Expected Behavior
Action Buttons should be properly aligned in Action Header.

## Possible Solution
By adding Theme variable to set padding value for action buttons( left and right buttons ) .

supreethmr:
### Summary
Resolves #2738 

### Additional Details
New theme variable added to set padding value for action buttons( left and right buttons ).

@cerner/terra


packages/terra-action-header/src/ActionHeader.module.scss
```diff
@@ -9,6 +9,11 @@
     margin-left: var(--terra-action-header-margin-left, 1.429rem);
   }
 
+  .left-buttons,
+  .right-buttons {
+    padding-top: var(--terra-action-header-button-padding-top);
```
StephenEsser: If feels a little strange to use padding to calculate an offset to center content. If this component is using flexbox then using something like align-items center may have a similar centering effect without needing to calculate these offsets. There's a few ways to center content without padding, flexbox is just an example.

supreethmr: I agree using align-items is more appropriate solution to align buttons to center. since action-header had already align-items set to flex-start changing this to center will create minor discrepancy in existing action-header. So I went with using padding value to align action buttons to center like [Title](https://github.com/cerner/terra-core/blob/1061a1a9d6e60ab314a451a4465d30fdfa57020f/packages/terra-action-header/src/ActionHeaderContainer.module.scss#L52).

neilpfeiffer: Reviewing UX needs for theme and best approach.

neilpfeiffer:
Closing https://github.com/cerner/terra-core/pull/2739 as not needed, any other adjustment items or new variables needed for theme adjustments will be added in https://github.com/cerner/terra-core/pull/2714 and resolved there.

Issue #1969, opened 2018-10-12
Investigate Button programmatic focus styling


dkasper-was-taken:
# Bug Report

## Description
Buttons use custom focus styling, and when they are programmatically focused the default implementation leaves them focused, but without focus highlight.  I'm unsure what the expectation should be, this may need UX input.

## Steps to Reproduce
1. Use node.focus()
2. Observe button has no focus highlight.

ravikiran6565:
Hi Dave,

Actually we debugged component in different browsers(chrome, safari, firefox and IE 10 & 11) and we checked for the tabIndex value, once modal-manager closed we expect focus should retain to triggered button but expect behaviour is working fine in chrome and all other browsers are losing it's focus.

Analysis what we have done.
Ex: In chrome after closing modal-manager tabIndex value is  > 0( which means focus was internally get selected to triggered button and browser will highlight the button).

Ex: Other browsers(safari, firefox, IE 10 and 11) after closing modal-manager tabIndex value is -1(which means focus was not selected to any elements in component and browser fail to highlight button)

We have done this analysis from our end but we are not sure what is internal functionality of modal-manager. With our analysis we gave this inputs if we are wrong please correct us.

Thanks,
Ravikiran V

dkasper-was-taken:
To move some of the offline discussion here:
Chrome buttons behave differently, it leaves focus on the button when it's clicked.  Modal uses a focus-trap, which uses tabbable.  These save the currently active target, and restore it on release. Navigating by keyboard behaves consistently, as all browser leave focus on the button, but clicking will result in odd behavior.

A possible solution would be the programmatically set focus to the button in the onClick before an action occurs.  I don't know if this a change that can be provided broadly, as it may result in unintended behavior with other controls, but it could possibly work in a limited capacity.

bjankord:
Right now, [we set `outline: none`](https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.module.scss#L54-L58) on `:focus` and `:active`, then [programmatically apply custom focus styles on keydown](https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L139-L142), and then we [remove our custom focus styles on blur](https://github.com/cerner/terra-core/blob/master/packages/terra-button/src/Button.jsx#L118-L120). 

I've been curious about inversing that logic. Instead, we apply our custom focus styles with `:focus` and in our onClick handler, we check the `event.type` to filter if the custom focus styles get nuked.

[...]

PR #2740, opened 2019-11-08
Button onfocus

### LINKED ISSUE #2688: [terra-button] DST Clarification Button does not take focus in Safari/Firefox
# Bug Report

## Description
In Safari and Firefox, clicking the time clarification button from a DST workflow for the terra-date-time-picker and then selecting a timezone does not return focus to the clarification button, nor does it invoke the consumer provided onBlur prop. This is particularly troublesome when that onBlur prop how the updates to the date time value get applied.

## Steps to Reproduce
1. Navigate to https://engineering.cerner.com/terra-ui/components/terra-date-time-picker/date-time-picker/date-time-picker in firefox or safari
2. Enter the ambiguous hour (e.g., 11/03/2019, 1:22) - modal appears, clarify your time
3. Modal disappears, click clarification button, modal reappears. Select opposite time, notice no button highlighting (terra-button 3.20.0 fixed how the button would be highlighted in chrome in this workflow, in firefox/safari it's not just that it's not highlighted, it does not even have focus)
4. If you like, run the workflow in chrome and observe the difference

I think this is an issue with the terra-button and the fixes around programmatic focus being browser dependent, which is why I logged it here. If my assumption is wrong I apologize and this may be a terra-date-time-picker issue.
https://jira2.cerner.com/browse/ORDPLANDEV-29226 has video

## Expected Behavior
Either for the button to have focus after the modal closes (after being launched from the button), or for the provided onBlur method to be invoked.

## Possible Solution
Right now I know the date-time-picker attempts to put focus back on the button as that is where it was before the modal was launched. If that behavior isn't necessary, just invoking the onBlur method of the consumer upon the modal being dismissed would work for us... but the correct fix is probably getting focus set on the button.

## @ Mentions
@benbcai @dev-cprice  @mjhenkes

benbcai:
### Summary
Fixes #2688, #1969 

### Additional Details
Clicking on a button does not get focused on Firefox, Safari, and IE. Hence, clicking away from it does not trigger onBlur, causing the issue on #2688. This PR sets the focus on the button when clicked.


packages/terra-button/src/Button.jsx
```diff
@@ -133,6 +134,19 @@ class Button extends React.Component {
     }
   }
 
+  handleClick(event) {
```
supreethmr: I assume this issue  of focus not shifting back to button in Safari and FireFox was due to button not being set as [previouslyFocusedElement](https://github.com/davidtheclark/focus-trap-react/blob/e67d26c5ff512228007bd6eb55aea9841eb7e09d/src/focus-trap-react.js#L10) in focus-trap and instead `main` element was being updated as previouslyFocusedElement. 

So Curious to know to keep button as active element and so that it gets updated as [previouslyFocusedElement](https://github.com/davidtheclark/focus-trap-react/blob/e67d26c5ff512228007bd6eb55aea9841eb7e09d/src/focus-trap-react.js#L10) when popup opens. do we need handleClick() event OR can we just set the focus to button on handleMouseDown() event like below.
```javascript
  handleMouseDown(event) {
    this.shouldShowFocus = false;
    setTimeout(() => { this.shouldShowFocus = true; });

    if (this.props.onMouseDown) {
      this.props.onMouseDown(event);
    }

    event.preventDefault();
    event.currentTarget.focus();
  }
```

benbcai: @supreethmr The issue is that in FF, Safari, and IE when you click on the button, it does not trigger the onFocus event like it does in Chrome. Since it didn't become focused after clicked, clicked outside would also not trigger the onBlur event, which causes the issue reported on #2688. Clicking and focus on a button varies by browser and OS. See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Button#Clicking_and_focus. 

This change is to set the focus on the button when clicked. The mouseDown event is triggered on both left and right mouse click but the focus should only be set on the button being clicked. Hence, I need to do that in handleClick.

supreethmr: Got it Thank you @benbcai  for the info.

packages/terra-button/src/Button.jsx
```diff
@@ -186,11 +200,16 @@ class Button extends React.Component {
     // Prevent button from showing focus styles when clicked
     this.shouldShowFocus = false;
     // Wait until after onFocus has been triggered on browsers where it will get triggered for click
-    setTimeout(() => { this.shouldShowFocus = true; });
+    setTimeout(() => { this.shouldShowFocus = true; }, 300);
```
yuderekyu: just curious, does this still work without setting  `300` inside the timeout?

benbcai: handkeKeyDown is called before handleClick above to put focus on the button. Since we don't want focus glow when clicking we need to wait until onclick is triggered before setting shouldShowFocus back to true.

PR #2741, opened 2019-11-12
Added Prop to disable search when required

### LINKED ISSUE #2305: [terra-form-select] - Optional search functionality
# Feature Request

## Description
For the variant="multiple",  the Select react component should be an option to turn off the default search functionality. The dropdowns in my app are small and do not need the search functionality.  In addition, on the android platform the search forces the popping up of the keyboard which often covers the dropdown.

supreethmr:
### Summary
Resolves #2305 

@cerner/terra

[License]: ../LICENSE


nramamurth:
I'd like to see an example added to the site.

packages/terra-form-select/src/multiple/Frame.jsx
```diff
@@ -223,7 +228,7 @@ class Frame extends React.Component {
             </li>
           ) : null}
         <li className={cx('search-wrapper')}>
-          <input {...inputAttrs} value={searchValue} />
+          {(this.props.isSearchDisabled) ? <div className={cx('placeholder')}>{placeholder || '\xa0'}</div> : <input {...inputAttrs} value={searchValue} /> }
```
nramamurth: Keyboard navigation seems broken.

multiple-select

nramamurth:
This is prolly a UX question - does it make sense for 'I' bar to be the mouse pointer when search is disabled? 

multiple-select

packages/terra-form-select/src/multiple/Frame.jsx
nramamurth: In a case when a value is already selected up-front, is this by design that we'd want to show the placeholder _only_ when search is disabled?

Before: 

Screen Shot 2019-11-12 at 2 49 46 PM

After: 
Screen Shot 2019-11-12 at 2 50 00 PM

packages/terra-form-select/src/multiple/Frame.jsx
```diff
@@ -223,7 +228,7 @@ class Frame extends React.Component {
             </li>
           ) : null}
         <li className={cx('search-wrapper')}>
-          <input {...inputAttrs} value={searchValue} />
+          {(isSearchDisabled) ? <div className={cx('placeholder')}>{placeholder || '\xa0'}</div> : <input {...inputAttrs} value={searchValue} /> }
```
nramamurth: `placeholder` is _always_ going to be present ([either default or custom](https://github.com/cerner/terra-core/blob/1061a1a9d6e60ab314a451a4465d30fdfa57020f/packages/terra-form-select/src/MultiSelect.jsx#L199-L200)). What is the need for this condition? ⬇️

```
placeholder || '\xa0'
```

nramamurth:
I am curious, when search is disabled, do we still need to be able to remove the options by pressing 'backspace' key? 

cc / @StephenEsser

supreethmr:
> I'd like to see an example added to the site.

Added [Example](https://github.com/cerner/terra-core/blob/148aaf18db9acf04ced1b29d35d98c3ada852ee4/packages/terra-form-select/src/terra-dev-site/doc/form-select/MultiSelect.07.doc.jsx#L70-L72) for Search Disabled.

packages/terra-form-select/src/multiple/Frame.jsx
supreethmr: it was due to placeholder div fixed it : 148aaf18db9acf04ced1b29d35d98c3ada852ee4

supreethmr:
> This is prolly a UX question - does it make sense for 'I' bar to be the mouse pointer when search is disabled?
> 
> multiple-select

Added class [is-search-disabled ](https://github.com/cerner/terra-core/blob/fe092d4a4ea3c249c5e62b897518016c00b777b3/packages/terra-form-select/src/shared/_Frame.module.scss#L391-L393)  to change cursor to pointer when search is disabled.

packages/terra-form-select/src/multiple/Frame.jsx
supreethmr: I missed that...! this condition is not required we can remove this from other files too.

packages/terra-form-select/src/multiple/Frame.jsx
supreethmr: replacing div with input fixed this : 148aaf18db9acf04ced1b29d35d98c3ada852ee4

supreethmr:
> I am curious, when search is disabled, do we still need to be able to remove the options by pressing 'backspace' key?
> 
> cc / @StephenEsser

I feel it should be there since it is multi select removing options with backspace provides better accessibility for users. 

would like to know your thoughts @StephenEsser

StephenEsser:
We'll want to get UX guidance on this change. There are a handful of concerns around accessibility, interactions, and visual appearance. 

The first question for UX would be if we want to proceed with making this change.

A multi-select without an input field to search should probably read differently on a screen reader and it may not render an input at all (rather than rendering a read-only input).

The focus management, keyboard interactions, and placeholder would also likely behave differently.

> I am curious, when search is disabled, do we still need to be able to remove the options by pressing 'backspace' key?

If there is not a focusable input I would not expect backspace to remove selected options, but we'd want to get feedback from our design team to appropriately answer this question.

cc: @neilpfeiffer

neilpfeiffer:
Closing as needs UX design guidance before solving for issue #2305.

\
\
Linking to comment added to issue as well: https://github.com/cerner/terra-core/issues/2305#issuecomment-552991164
> This technically would fall as a new standalone variant -- most likely would be shown similar to the single-select, except be able to select multiple values -- and would require it's own rollup design pattern (not use pills/tags) -- similar to an [example like this](https://3.bp.blogspot.com/-mg8pbAHtdYg/XDOaDK813ZI/AAAAAAAAANo/kNBjFWd_Ds4U0MH5Wqll36ph08DR93FOQCLcBGAs/s1600/bootstrap-jquery-muli-select-drop-down-list-with-the-checkbox-in-asp-dot-net.png).
> 
> UX needs to validate if this a type of multiselect Terra will support and what the rollup pattern needs to be. Further design is required before implementing -- hold on working until approach is determined and design pattern provided. 
>

PR #2747, opened 2019-11-15
[terra-alert] Updated uncontrolled responsive-element to controlled responsive-element

### LINKED ISSUE #2744: [terra-alert] Uplift to the controlled responsive element
# Feature Request

Investigate uplifting the terra-alert onto the controlled responsive element or alternative. 

## Description

The uncontrolled responsive element has been [flagged for deprecation](https://github.com/cerner/terra-core/pull/2729). The terra-alert should either be uplifting onto the controlled responsive element implementation or alternatively the responsive element should be removed from the component.

pranav300:
### Summary
Resolves #2744


packages/terra-alert/src/Alert.jsx
```diff
@@ -162,26 +149,20 @@ const Alert = ({
   );
 
   return (
-    <ResponsiveElement
-      tiny={(
-        <div {...attributes} className={narrowAlertClassNames}>
-          <div className={bodyClassNameForNarrowParent}>
+    <ResponsiveElement onChange={value => setBreakpoint(value)}>
+      {
+        breakpoint // Check for a breakpoint before rendering the alert
```
StephenEsser: This PR may remove the need to perform this check: https://github.com/cerner/terra-core/pull/2761. When 2761 merges can you do some unscripted testing to verify if the flicker is still an issue if the breakpoint check is removed?

pranav300: I tested the changes in #2761 with the changes in this branch, after removing the breakpoint check, I can see flicker happening when render time is greater than 20ms for the component. Below which it's not visible.

StephenEsser: After implemented the responsive element changes on this branch was the project recompiled? The update to the responsive element should remove the need for the breakpoint check prior to rendering.

packages/terra-alert/src/Alert.jsx
```diff
@@ -145,7 +132,7 @@ const Alert = ({
     );
   }
   if (onDismiss || action) {
-    bodyClassNameForNarrowParent = cx(['body', 'body-narrow']);
+    bodyClassNameForParent = cx(['body', `alert-parent-${breakpoint}`, 'narrow']);
```
yuderekyu: probably out of scope, but I wonder if the classNames would be easier to maintain if we inline conditional checks. This way we have assign className values once and eliminate some code duplication.

```jsx
const alertSectionClassName = cx([
  'section', 
  { 'section-custom': type === AlertTypes.CUSTOM },
]);

const bodyClassNameForParent = cx([
  'body', 
  `alert-parent-${breakpoint}`,
  { 'narrow': onDismiss || action },
]);

Comment (pranav300) [reply packages/terra-alert/src/Alert.jsx]: Updated here https://github.com/cerner/terra-core/pull/2747/commits/095c1958ce84835a6d936b112675fa9bddf36951.

File: packages/terra-alert/src/Alert.jsx [hunk 3]
```diff
@@ -162,26 +144,20 @@ const Alert = ({
...
-            {getAlertIcon(type, customIcon)}
-            {alertMessageContent}
-          </div>
-          {actionsSection}
-        </div>
-      )}
-      small={(
-        <div {...attributes} className={wideAlertClassNames}>
-          <div className={cx(['body', 'body-std'])}>
+    <ResponsiveElement onChange={value => setBreakpoint(value)}>
```
emilyrohrbough: Looking at the styles in the style sheet, it seems we only care about flipping between a narrow and wide Alert.

Can we add a check in the onChange callback we provide to only update the state when the breakpoint flips from narrow to large? 

If we do this, we shouldn't need to  add new styles, but use the existing ones.

```jsx
const [isNarrow, setIsNarrow] = useState();
const alertClassNames = cx([
    'alert-base',
    type,
    { 'narrow': isNarrow }
    { 'wide': !isNarrow }
    attributes.className,
    { [`${customColorClass}`]: type === AlertTypes.CUSTOM },
  ]);

const bodyClassNameForParent = cx([
    'body',
    'body-std'
   { 'body-std': !isNarrow || (isNarrow && !dismiss && !action}
   { 'body-narrow': isNarrow && (dismiss || action) }
  ]);

<ResponsiveElement
    onChange={(value) => {
         const showNarrowAlert = value === 'tiny';
         if (showNarrowAlert !== isNarrow ) {
           setBreakpoint(showNarrowAlert);
         }
     }
>
      <div {...attributes} className={alertClassNames}>
          <div className={bodyClassNameForNarrowParent}>
            {getAlertIcon(type, customIcon)}
            {alertMessageContent}
          </div>
          {actionsSection}
        </div>
</ ResponsiveElement>
```

packages/terra-alert/src/Alert.module.scss
```diff
@@ -40,11 +44,16 @@
...
 
-  .body-std {
+  .alert-parent-tiny,
+  .alert-parent-small,
+  .alert-parent-medium,
+  .alert-parent-large,
+  .alert-parent-huge,
+  .alert-parent-enormous {
     margin-bottom: var(--terra-alert-std-margin-bottom, 0.7143rem);
     margin-right: var(--terra-alert-std-margin-right, 0.6667rem);
```
emilyrohrbough: As is, we shouldn't be applying margin-right to the narrow alert that has an actin or dismiss:
https://github.com/cerner/terra-core/pull/2747/files#diff-e3a10ba555be3590e95544911b047ebcL148

emilyrohrbough: By changing the classes that use these variables, does that mean we should be introducing a new variables to reflect the new classes in the name?

PR #2752, opened 2019-11-19
[terra-dropdown-button] Removed hookshot to fix accessibility issues.

### LINKED ISSUE #2632: [terra-dropdown-button] - Mouse / Keyboard Accessibility Interactions
# Bug Report

## Description

UX has requested that mouse and keyboard interactions behave slightly different. Currently using a mouse to click the toggle of the SplitButton/DropdownButton will open the menu and focus/highlight the first option in the menu. 

The request is that when the menu is opened via a mouse click the first option is not focused/highlighted by default.

However, when navigating to the  SplitButton/DropdownButton via the keyboard, opening the menu with the keyboard should focus/highlight the first option in the menu.

## Steps to Reproduce

1. Navigate to the [example site](https://engineering.cerner.com/terra-core/components/terra-dropdown-button/dropdown-button/dropdown-button).
2. Click on any of the examples.
3. Observe the first menu action item is auto-focused/highlighted.

image

## Expected Behavior

Clicking a DropdownButton/SplitButton should open the menu but not auto-highlight the first option.

Opening the DropdownButton/SplitButton using the keyboard should auto-highlight the first option.

saket2403:
### Summary
Fixes #2632


packages/terra-dropdown-button/CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ChangeLog
 =========
+### Changed
+* Removed Hookshot to fix accessibility issues in dropdown menu.
```
StephenEsser: Can you move this under the unreleased section right below?

packages/terra-dropdown-button/package.json
```diff
@@ -42,6 +42,6 @@
     "precompile": "rm -rf lib",
     "test": "npm run test:jest && npm run test:wdio",
     "test:jest": "jest --config ../../jestconfig.js",
-    "test:wdio": "../../node_modules/.bin/wdio ../../node_modules/terra-dev-site/config/wdio/wdio.conf.js"
+    "test:wdio": "cd ../.. && npx wdio ./packages/$npm_package_name/wdio.conf.js"
```
StephenEsser: I haven't seen this before, are other components doing this?

StephenEsser:
I don't think hookshot can be removed entirely as a passive change. The positioning changes can cause consuming applications to experience some unpredictable clipping and height distortions. 

For example, with these changes a dropdown button opening within a header will either expand the height of the header or clip the menu if the overflow of the header is set to hidden. To stay passive a prop to flex the component will need to be added.

If we keep decide to move forward with these changes I think the dropdown button will need to be major version bumped and consuming applications will need to carefully redesign certain workflows to account for the changes.

@ryanthemanuel

packages/terra-dropdown-button/CHANGELOG.md
saket2403: Changed 447252c

packages/terra-dropdown-button/package.json
saket2403: Yes other components are doing this, as in [here](https://github.com/cerner/terra-core/blob/6e84583b6841055f49bd4a52ebeb23a378a43384/packages/terra-button/package.json#L31).

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -2,10 +2,13 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import * as KeyCode from 'keycode-js';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import onClickOutside from 'react-onclickoutside';
```
supreethmr: Adding this as dependency will fix this import lint error.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -48,6 +51,7 @@ const defaultProps = {
   isDisabled: false,
   variant: 'neutral',
 };
+const WrappedButtonBase = onClickOutside(DropdownButtonBase);
```
supreethmr: I think we can do this wrapping in _DropdownButtonBase instead of here like below 
`export default reactOnClickOutside(DropdownButtonBase);`

packages/terra-dropdown-button/src/DropdownButton.jsx
saket2403: Added in f8e6d07

packages/terra-dropdown-button/src/_Dropdown.module.scss
```diff
@@ -0,0 +1,28 @@
+:local {
```
supreethmr: Are these styles similar to hookshot.? because it looks like dropdown-right class in split-button pushes dropdown to behind the side menu in dev-site where as it was displayed upfront when hookshot was present like shown in below screenshots.

Without hookshot  : 
Screen Shot 2019-11-20 at 3 32 37 PM

With hookshot :
Screen Shot 2019-11-20 at 3 33 52 PM

saket2403: Ooops! Changed here bcf170a

PR #2753, opened 2019-11-20
[terra-form-select] Upon selection, keep one dropdown open and close the others (if already open) 

### LINKED ISSUE #2750: [terra-form-select] Opening one dropdown in a table will not close another opened dropdown
# Bug Report

## Description
When a terra-table contains a terra-form-select dropdown on each row, opening one select will not automatically close the previously opened select.

This works as expected on Chrome and Edge; only having the issue in IE.

## Steps to Reproduce
1. Render a terra-table with a terra-form-select on each row.
2. Open a select on one row.
3. Open a select on a different row - the previous select remains open.
4. Open a select on another row - the previous two selects remain open.

image

## Expected Behavior
When one select is opened, close other selects in the same table.

## Environment
* Component Name and Version: "terra-form-select": "^5.36.0", "terra-table": "^3.26.0",
* Browser Name and Version: IE 10
* Node/npm Version: [e.g. Node 8/npm 5]  node 10.16 / npm 6.9
* Webpack Version: "webpack": "^4.30.0",

nramamurth:
### Summary
Fixes #2750


packages/terra-form-select/src/single/Frame.jsx
```diff
@@ -468,8 +471,10 @@ class Frame extends React.Component {
     const descriptionId = `terra-select-screen-reader-description-${uniqueid()}`;
     const customAriaDescribedbyIds = customProps['aria-describedby'];
     const ariaDescribedBy = customAriaDescribedbyIds ? `${descriptionId} ${customAriaDescribedbyIds}` : descriptionId;
+    const selectMenuId = `terra-select-menu-${uniqueid()}`;
```
StephenEsser: Creating a new id every time the component re-renders doesn't seem ideal, but I see the other values above are doing the same thing. Before merging can you verify this doesn't do anything strange to screen readers?

packages/terra-form-select/tests/wdio/select-spec.js
```diff
@@ -344,7 +358,7 @@ Terra.describeViewports('Select', ['tiny'], () => {
       });
 
       it('dropdown should be focused', async () => {
-        (await browser.hasFocus('#terra-select-menu')).should.be.true;
+        (await browser.hasFocus('[class*=menu]')).should.be.true;
```
StephenEsser: Wildcard selectors can be flaky if anything else on the page has a similar class name. I think this is probably safe in this isolated test, if we wanted to be safe we could potentially introduce a data attribute similar to the frame. (Which I just noticed says combobox, but I think they all do)

https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/single/Frame.jsx#L489

PaulSolDev:

PR #2762, opened 2019-11-25
[terra-image] Leverage react fragment to prevent double requests from…

### LINKED ISSUE #2757: [terra-image] Two img requests made when placeholder prop is provided
# Bug Report

## Description
Two requests for the image src are made by terra-image when the placeholder prop is provided instead of one.

## Steps to Reproduce
1. Launch our single page application (PowerChart Touch) in DEVENG in chrome
2. Open chrome network tools and navigate to any given patient's chart.
3. Notice how two requests for the patient thumbnail are sent instead of one.

## Expected Behavior
Only one request to be made.

## Possible Solution
The issue is occurring as the img tag for the placeholder state is recreated by react once it has loaded. The solution I found was to wrap the img tag in the render function of terra-image in a div such that the structure is similar to what is rendered during the during the placeholder state lifecycle.

## Environment

* Component Name and Version:
* Node/npm Version: [8.16.2/6.4.1]
* Webpack Version: 4.41.2
* Operating System and version (desktop or mobile): Mac OSX

jeffmerten:
… occuring

### Summary
Use React Fragment to wrap placeholder and image in order to prevent the img request from happening twice.

### Additional Details
Previously the img element would be created twice as it was in a div structure that changed when the state changed to on the initial image load. The re-render would cause a new img element to be created. 

The fix is to wrap the img element in react fragment so it doesn't get recreated by react.

The change has been tested by myself in one of our single page applications for our demographics thumbnail image.

Closes #2757

### DIFF
--- a/packages/terra-image/src/Image.jsx
+++ b/packages/terra-image/src/Image.jsx
@@ -143,13 +143,20 @@ class Image extends React.Component {
       src, variant, isFluid, alt, placeholder, height, width, onLoad, onError, fit, ...customProps
     } = this.props;
 
-    const imageClasses = cx([
+    const imageClassAttributes = [
       'image',
       fit,
       variant,
       customProps.className,
       { fluid: isFluid },
-    ]);
+    ];
+
+    if (placeholder && this.state.isLoading) {
+      imageClassAttributes.push('hidden');
+    }
+
+    const imageClasses = cx(imageClassAttributes);
+
     delete customProps.className;
     if (!this.state.isLoading) {
       objectFitImages(this.ImageRef.current);
@@ -157,14 +164,14 @@ class Image extends React.Component {
     if (placeholder) {
       if (this.state.isLoading) {
         return (
-          <div>
-            <div className={cx('hidden')}>{this.createImage(customProps, imageClasses)}</div>
-            <div>{placeholder}</div>
-          </div>
+          <React.Fragment>
+            {this.createImage(customProps, imageClasses)}
+            {placeholder}
+          </React.Fragment>
         );
       }
 
-      return this.state.isError ? placeholder : this.createImage(customProps, imageClasses);
+      return this.state.isError ? placeholder : <React.Fragment>{this.createImage(customProps, imageClasses)}</React.Fragment>;
     }
 
     return this.createImage(customProps, imageClasses);
--- a/packages/terra-image/src/Image.jsx
+++ b/packages/terra-image/src/Image.jsx
@@ -171,7 +171,7 @@ class Image extends React.Component {
         );
       }
 
-      return this.state.isError ? placeholder : <>{this.createImage(customProps, imageClasses)}</>;
+      return this.state.isError ? placeholder : this.createImage(customProps, imageClasses);
     }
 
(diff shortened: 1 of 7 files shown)


packages/terra-image/src/Image.jsx
```diff
@@ -143,28 +143,35 @@ class Image extends React.Component {
...
       objectFitImages(this.ImageRef.current);
     }
     if (placeholder) {
       if (this.state.isLoading) {
         return (
-          <div>
-            <div className={cx('hidden')}>{this.createImage(customProps, imageClasses)}</div>
-            <div>{placeholder}</div>
-          </div>
+          <React.Fragment>
```
yuderekyu: nitpick, can we use shorthand instead? `<>`

jeffmerten: I kind of prefer React.Fragment since it's easier on my old man eyes, but I'll go with the shorthand since you guys are gonna maintain terra-image more than I going forward.

packages/terra-image/src/Image.jsx
```diff
@@ -143,28 +143,35 @@ class Image extends React.Component {
...
-          </div>
+          <>
+            {this.createImage(customProps, imageClasses)}
+            {placeholder}
+          </>
         );
       }
 
-      return this.state.isError ? placeholder : this.createImage(customProps, imageClasses);
+      return this.state.isError ? placeholder : <>{this.createImage(customProps, imageClasses)}</>;
```
tbiethman: Is the fragment here necessary?

jeffmerten: Nope, was just being safe with my initial fix. Just tested without it and works.

packages/terra-image/src/Image.jsx
```diff
@@ -143,24 +143,31 @@ class Image extends React.Component {
...
     } = this.props;
 
-    const imageClasses = cx([
+    const imageClassAttributes = [
       'image',
       fit,
       variant,
       customProps.className,
       { fluid: isFluid },
-    ]);
```
emilyrohrbough: Can you add like:

```diff
   const imageClasses = cx([
      'image',
      fit,
      variant,
      customProps.className,
      { fluid: isFluid },
+     { hidden: placeholder && this.state.isLoading }
    ]);

```

jeffmerten: I had something like this initially but decided against it as I didn't want to collide with the custom attributes a consumer could provide after the image had loaded.

jeffmerten: Let me know if you think this is a valid concern or not. From the way our team uses terra-image/profile-image it wouldn't affect us at all, and I'm sure anyone applying a hidden attribute to terra-image for some odd use case could find a work around.

jeffmerten: Nevermind, didn't know classNames removes falsy values.

emilyrohrbough: yeah it removes falsey values! It's great 😄

mjhenkes:
Thanks @jeffmerten :)

PR #2764, opened 2019-11-27
[terra-form-textarea]Disallowed Resize when size or rows are provided.

### LINKED ISSUE #2543: [terra-form-textarea] Resize API
# Feature Request
`terra-form-textarea` examples show that resizability is turned off when using the `size` prop. For this functionality, we are actually applying a custom stylesheet to examples: https://github.com/cerner/terra-core/blob/master/packages/terra-form-textarea/src/terra-dev-site/doc/example/FormTextareaDocCommon.module.scss

`terra-form-textarea` will always have resizability set to true, unless it is `disabled` or `size = full`. 

If we want this behavior, we should make this explicit for examples by converting this stylesheet into an inline style again. 

If we do not want this behavior, we should update the API to set `resize: none` when using the `size` or `row` props.

saket2403:
### Summary
Fixes #2543


packages/terra-form-textarea/src/Textarea.jsx
```diff
@@ -104,7 +103,7 @@ const defaultProps = {
   placeholder: undefined,
   required: false,
   rows: null,
-  size: 'small',
+  size: undefined,
```
StephenEsser: Changing the default props is a non-passive change. This will require a major version bump of the component if we want to continue with this solution.

emilyrohrbough:
I have always thought the `size` and `rows` props were used to set the default rendered size and then it gave consumers the opportunity to resize if they wanted a bigger or smaller area.

It feels like we should add a new prop called `disableResize` or something similar to give consumers a way to opt-in to this functionality.

packages/terra-form-textarea/src/Textarea.jsx
saket2403: Added prop in fe94213

saket2403:
> It feels like we should add a new prop called `disableResize` or something similar to give consumers a way to opt-in to this functionality.

Added the prop in fe94213

PaulSolDev:
Functional review complete. No issue identified.

PR #2766, opened 2019-12-03
[terra-dropdown-button] Fixed Focus issue for mouse and keyboard interactions.

### LINKED ISSUE #2632: [terra-dropdown-button] - Mouse / Keyboard Accessibility Interactions

saket2403:
### Summary
Fixes #2632 

we are using `initialFocus` prop of Focus-Trap to prevent focus on first option when dropdown opened via mouse click. 
```javascript
<FocusTrap focusTrapOptions={{ returnFocusOnDeactivate: true, initialFocus: openedViaKeyboard ? '' : buttonRef, clickOutsideDeactivates: true }}>
```
`initialFocus` prop will be set to buttonRef which will be holding ref of [dropdownbutton](https://github.com/cerner/terra-core/blob/a3314900b693a8cf0ca8c78a7a61baf5b0069b35/packages/terra-dropdown-button/src/DropdownButton.jsx#L164) when opened via mouse click. `openedViaKeyboard` boolean is used to differentiate between keyboard event  and mouse event.

`initialFocus` will be set to empty to set focus on first option when dropdown is opened via keyboard ( `openedViaKeyboard` is true ). Focus trap will set to first accessible element when `initialFocus` prop is set to '' empty.

This introduced new issue by making dropdown list inaccessible in VO/JAWS mode when opened via mouse click. this was due to holding focus back on dropdown button on open ( when opened via mouse click ) instead of moving focus to first list item. 

Since dropdown list is rendered in portal dom order will be broken as it renders outside of root div this makes focus to shift from button to other elements in page instead of list items on navigation up and down key press.

To fix this issue I used keydown event handler of DropdownButton and SplitButton. Added check for up and down arrow key so that we can set focus to first list item on down arrow key and on last list item on up arrow key press.

```javascript
if (event.keyCode === KeyCode.KEY_DOWN && this.state.isOpen && !this.state.openedViaKeyboard) {
      // puts focus on first list element on down arrow key press when dropdown is opened by mouse click
      const firstOption = this.dropdownList.querySelector('[data-terra-dropdown-first-list-item = "true"]');
      firstOption.focus();
} else if (event.keyCode === KeyCode.KEY_UP && this.state.isOpen && !this.state.openedViaKeyboard) {
      // puts focus on last list element on up arrow key press when dropdown is opened by mouse click
      const lastOption = this.dropdownList.querySelector('[data-terra-dropdown-last-list-item = "true"]');
      lastOption.focus();
}
```
`openedViaKeyboard` boolean is used to ensure that focus sets to list items on navigation key press only when dropdown is opened via mouse click. 

setting focus on list option in keydown event resolved VO and JAWS issue of accessing list items through key board.

#### Before Code Changes 
mouse click on dropdown-button -> opens dropdown with first option in list being highlighted
Screen Shot 2020-01-09 at 6 27 25 PM

#### After Code Changes 
mouse click on dropdown-button -> opens dropdown by holding focus on dropdown-button
Screen Shot 2020-01-09 at 10 49 57 PM

#### New Functionalities added 
when dropdown is opened via mouse click : 
On keydown press : 
Screen Shot 2020-01-09 at 11 09 44 PM

On keyup press : 
Screen Shot 2020-01-09 at 11 09 58 PM


packages/terra-dropdown-button/tests/wdio/dropdown-button-spec.js
```diff
@@ -59,7 +59,7 @@ Terra.describeViewports('Dropdown Button', ['medium'], () => {
     });
 
     it('closes dropdown on escape without running a callback', () => {
-      browser.keys(['Escape']);
+      browser.keys(['Tab', 'Escape']);
```
supreethmr: do we need 'tab' key here..? Dropdown still closes on escape key press.

packages/terra-dropdown-button/tests/wdio/dropdown-button-spec.js
```diff
@@ -69,7 +69,7 @@ Terra.describeViewports('Dropdown Button', ['medium'], () => {
     });
 
     it('opens the dropdown with space', () => {
-      browser.keys(['Escape', 'Space']);
+      browser.keys(['Tab', 'Escape', 'Space']);
```
supreethmr: here also 'Tab' is not required to close dropdown. Please check for other test cases also.

supreethmr:
Focus issue of dropdown (focus not going into dropdown on key down press) has been addressed with commit : e6707e13790226d459fbd937a35b0d1744a4eff6 
Made changes for keyboard navigation when drop-down is opened by mouse click with reference to : https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-2/menubar-2.html

StephenEsser:
Would you mind adding the specific scenarios these code changes are intended to alter to the PR summary? Such as the expected outcome from clicking the dropdown or opening it with the keyboard and focus interaction states. Occasionally issues have lengthly threads on the originally logged issue to track down the expected changes. A brief recap is really helpful to provide with the PR.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -57,29 +57,57 @@ class DropdownButton extends React.Component {
...
   handleDropdownButtonClick() {
+    if (this.state.isOpen) {
+      this.setState({ isKeyboardEvent: false });
+    }
     this.setState(prevState => ({ isOpen: !prevState.isOpen }));
   }
 
   handleDropdownRequestClose(callback) {
     this.setState({ isOpen: false }, typeof callback === 'function' ? callback : undefined);
+    this.setState({ isKeyboardEvent: false, isActive: false });
```
StephenEsser: These set state calls can probably be combined. React should batch state updates, but I am not familiar with what happens when some of the batched calls have a callback function. The callback can probably be moved into a variable to shorten the declaration.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -57,29 +57,57 @@ class DropdownButton extends React.Component {
...
    */
   handleKeyDown(event) {
-    if (event.keyCode === KeyCode.KEY_SPACE) {
+    if (event.keyCode === KeyCode.KEY_SPACE || event.keyCode === KeyCode.KEY_RETURN) {
       this.setState({ isActive: true });
+      this.setState({ isKeyboardEvent: true });
+    } else if (event.keyCode === KeyCode.KEY_DOWN && this.state.isOpen && !this.state.isKeyboardEvent) {
+      // puts focus on first list element on down arrow key press when dropdown is opened by mouse click
+      // the div inside the li is what is actually focusable so need to go 2 layers down
+      this.dropdownList.childNodes[0].childNodes[0].focus();
```
StephenEsser: This will break if the semantic structure of the list ever changes. It might be better to query the dropdownList for a specific data attribute on the focusable element. Or query for list items and focus the first item in the first list item.

StephenEsser: Does this introduce or resolve any of the voice over concerns? Specifically swipe gesture on iOS. Do they register as keyboard or mouse interactions? How is focus shifted using voice over on desktop and on mobile?

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -57,29 +57,57 @@ class DropdownButton extends React.Component {
     this.handleDropdownRequestClose = this.handleDropdownRequestClose.bind(this);
     this.handleKeyDown = this.handleKeyDown.bind(this);
     this.handleKeyUp = this.handleKeyUp.bind(this);
+    this.setButtonNode = this.setButtonNode.bind(this);
+    this.getButtonNode = this.getButtonNode.bind(this);
+    this.setListNode = this.setListNode.bind(this);
+    this.state = { isOpen: false, isActive: false, isKeyboardEvent: false };
```
nramamurth: I don't understand what `isKeyboardEvent` signifies in context of this component. What does it mean? Could it be something like `keyboardAccessed`, `keyPressed` or `openedViaKeyboard` (quite long)?

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -49,6 +49,17 @@ const propTypes = {
...
+   * @private
+   * Callback to tell the parent it should close the dropdown
+   */
+  requestClose: PropTypes.func,
+  /**
+   *  @private
+   * The associated metaData to be provided in the onSelect callback.
+   */
+  // eslint-disable-next-line react/forbid-prop-types
+  metaData: PropTypes.object,
```
nramamurth: Doesn't each item have its associated `metaData` already? Is this specific to just the primary item? I'd not expect `SplitButton` to have metaData per se.

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -63,21 +74,50 @@ class SplitButton extends React.Component {
...
     this.handlePrimaryKeyDown = this.handlePrimaryKeyDown.bind(this);
     this.handlePrimaryKeyUp = this.handlePrimaryKeyUp.bind(this);
     this.handleCaretKeyDown = this.handleCaretKeyDown.bind(this);
     this.handleCaretKeyUp = this.handleCaretKeyUp.bind(this);
+    this.setButtonNode = this.setButtonNode.bind(this);
+    this.getButtonNode = this.getButtonNode.bind(this);
+    this.setListNode = this.setListNode.bind(this);
+
+    this.state = {
+      isOpen: false, caretIsActive: false, primaryIsActive: false, isKeyboardEvent: false,
```
nramamurth: Same comment as above. Please change `isKeyboardEvent` to what it actually signifies w.r.t. the component.

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -63,21 +74,50 @@ class SplitButton extends React.Component {
...
   handleDropdownButtonClick() {
+    if (this.state.isOpen) {
+      this.setState({ isKeyboardEvent: false });
+    }
     this.setState(prevState => ({ isOpen: !prevState.isOpen }));
   }
 
+  handlePrimaryButtonClick(event) {
+    event.stopPropagation();
+    event.currentTarget.focus();
```
nramamurth: This works but should we instead set `isPrimaryActive` to true in state? That seems to automatically take care of styles and that would remain consistent with how click keydown event handles the focus.

packages/terra-dropdown-button/tests/wdio/split-button-spec.js
```diff
@@ -207,7 +207,7 @@ Terra.describeViewports('Split Button', ['medium'], () => {
     });
 
     it('opens the long option dropdown', () => {
-      browser.keys(['Escape']);
+      browser.refresh();
```
nramamurth: Why is this needed?

supreethmr:
> Would you mind adding the specific scenarios these code changes are intended to alter to the PR summary? Such as the expected outcome from clicking the dropdown or opening it with the keyboard and focus interaction states. Occasionally issues have lengthly threads on the original logged issue to track down the expected changes. A brief recap is really helpful to provide with the PR.

Sure, I have updated the PR description with details of code changes that I have made let me know if I have missed something.

packages/terra-dropdown-button/src/DropdownButton.jsx
supreethmr: Changed it to [data-att ](https://github.com/cerner/terra-core/blob/cf4a1d2744dbd30b29ba636528a1ebd6c739c3a0/packages/terra-dropdown-button/src/DropdownButton.jsx#L99).

Yes this was added to resolve navigation issue that was found while testing with VO and JAWS. Due to holding focus with button on open was making dropdown list inaccessible for VO and JAWS hence handled this in keydown event.

this issue was happening only with keyboard and not with touch / swipe gestures so this has no impact on them.

I have tested for swipe gesture on IOS it works as expected.

packages/terra-dropdown-button/src/DropdownButton.jsx
supreethmr: changed to : openedViaKeyboard

packages/terra-dropdown-button/src/SplitButton.jsx
supreethmr: yes it is just specific to primary-item because even primary-button is considered as item in SplitButton.

packages/terra-dropdown-button/src/SplitButton.jsx
supreethmr: changed to openedViaKeyboard

packages/terra-dropdown-button/src/SplitButton.jsx
supreethmr: we need this because setting primaryIsActive state will not set focus styles in safari and firefox so we need to set focus manually. 
reference : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Button#Clicking_and_focus

packages/terra-dropdown-button/src/SplitButton.jsx
nramamurth: Then why do you need a prop here when each item already has metaData prop?

supreethmr: yes each item has metadata but we don't have one for primary-button. so we need this prop.

packages/terra-dropdown-button/src/DropdownButton.jsx
supreethmr: changed  : [here](https://github.com/cerner/terra-core/blob/cf4a1d2744dbd30b29ba636528a1ebd6c739c3a0/packages/terra-dropdown-button/src/DropdownButton.jsx#L86-L87)

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -57,29 +57,56 @@ class DropdownButton extends React.Component {
...
   handleDropdownButtonClick() {
+    if (this.state.isOpen) {
+      this.setState({ openedViaKeyboard: false });
+    }
     this.setState(prevState => ({ isOpen: !prevState.isOpen }));
   }
 
   handleDropdownRequestClose(callback) {
-    this.setState({ isOpen: false }, typeof callback === 'function' ? callback : undefined);
+    const callbackOnRequestClose = typeof callback === 'function' ? callback : undefined;
```
nramamurth: Could this be simplified to - 

`this.setState({ isOpen: false, openedViaKeyboard: false, isActive: false }, callback);`

Is there a scenario wherein the argument that is passed in is not a function?

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -49,6 +49,17 @@ const propTypes = {
    * The intl object to be injected for translations.
```
nramamurth: https://github.com/cerner/terra-core/pull/2766#discussion_r360082208 needs to be addressed in this file too.

packages/terra-dropdown-button/src/SplitButton.jsx
nramamurth: In my understanding, primary button is not a separate item - it is built off of one of the items that are passed in as children - and metaData is part of that item. 

Also, 

- if this is the face-up component that the consumers will use to create a SplitButton, then how does this work because it is marked as private? 
- is there an example or a test page that prints out the metaData of the primary button when selected?

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -96,13 +137,22 @@ class SplitButton extends React.Component {
...
       this.setState({ caretIsActive: true });
+      this.setState({ openedViaKeyboard: true });
+    } else if (event.keyCode === KeyCode.KEY_DOWN && this.state.isOpen && !this.state.openedViaKeyboard) {
+      // puts focus on first list element on down arrow key press when dropdown is opened by mouse click
+      const firstOption = this.dropdownList.querySelector('[data-terra-dropdown-first-list-item = "true"]');
+      firstOption.focus();
+    } else if (event.keyCode === KeyCode.KEY_UP && this.state.isOpen && !this.state.openedViaKeyboard) {
+      // puts focus on last list element on up arrow key press when dropdown is opened by mouse click
+      const lastOption = this.dropdownList.querySelector('[data-terra-dropdown-last-list-item = "true"]');
+      lastOption.focus();
```
nramamurth: What happens when there is only one item (assuming that it is a valid use-case)? Which data attribute will that item have? I'd think that this blows up.

packages/terra-dropdown-button/src/SplitButton.jsx
nramamurth: Why is `stopPropagation` necessary? I don't see this logic in `terra-button`. Also, I'd leave the MDN link in the code comment for code readers to understand in the future. 

https://github.com/cerner/terra-core/blob/ce4c1c129427a28161c0c079c8c9647e27c9d98f/packages/terra-button/src/Button.jsx#L137-L148

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -96,13 +137,22 @@ class SplitButton extends React.Component {
   }
 
   handleCaretKeyDown(event) {
-    if (event.keyCode === KeyCode.KEY_SPACE) {
+    if (event.keyCode === KeyCode.KEY_SPACE || event.keyCode === KeyCode.KEY_RETURN) {
       this.setState({ caretIsActive: true });
+      this.setState({ openedViaKeyboard: true });
+    } else if (event.keyCode === KeyCode.KEY_DOWN && this.state.isOpen && !this.state.openedViaKeyboard) {
+      // puts focus on first list element on down arrow key press when dropdown is opened by mouse click
+      const firstOption = this.dropdownList.querySelector('[data-terra-dropdown-first-list-item = "true"]');
```
nramamurth: This could be simplified to ⬇️

`this.dropdownList.querySelector('[data-terra-dropdown-first-list-item]');`

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -96,13 +137,22 @@ class SplitButton extends React.Component {
   }
 
   handleCaretKeyDown(event) {
-    if (event.keyCode === KeyCode.KEY_SPACE) {
+    if (event.keyCode === KeyCode.KEY_SPACE || event.keyCode === KeyCode.KEY_RETURN) {
       this.setState({ caretIsActive: true });
+      this.setState({ openedViaKeyboard: true });
+    } else if (event.keyCode === KeyCode.KEY_DOWN && this.state.isOpen && !this.state.openedViaKeyboard) {
```
nramamurth: Please leave code comments mentioning that the focus behaviour differs when accessed via click v/s keypress.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -57,29 +57,56 @@ class DropdownButton extends React.Component {
...
 
   /*
     In FireFox active styles don't get applied on space
    */
   handleKeyDown(event) {
-    if (event.keyCode === KeyCode.KEY_SPACE) {
+    if (event.keyCode === KeyCode.KEY_SPACE || event.keyCode === KeyCode.KEY_RETURN) {
       this.setState({ isActive: true });
+      this.setState({ openedViaKeyboard: true });
+    } else if (event.keyCode === KeyCode.KEY_DOWN && this.state.isOpen && !this.state.openedViaKeyboard) {
```
nramamurth: Is `isOpen` check merely a safety check here?

nramamurth:
The deployment seem to be outdated, recent changes need to be re-deployed. I pulled down the branch and noticed that the focus does not go to caret button upon dismissing the dropdown on Safari. 

focus

packages/terra-dropdown-button/tests/wdio/split-button-spec.js
supreethmr: not needed reverted to [previous code](https://github.com/cerner/terra-core/blob/3d4cbdca1a5c96ae211cdcf5bdb5847ecfdc965c/packages/terra-dropdown-button/tests/wdio/split-button-spec.js#L210)

packages/terra-dropdown-button/src/DropdownButton.jsx
supreethmr: I did not find any existing scenario where we are passing callback function to handleDropdownRequestClose() method and there is no callback for onClick is provided in DropdownButton for consumers to pass in callback. so am not sure when it is passed in as non-function.
 I feel this might be added has safety check to ensure callback that will triggered in handleDropdownRequestClose() is a function.

@StephenEsser would like to know your thoughts on callback usage in handleDropdownRequestClose() method

packages/terra-dropdown-button/src/SplitButton.jsx
supreethmr: I missed it, updated [here](https://github.com/cerner/terra-core/blob/3d4cbdca1a5c96ae211cdcf5bdb5847ecfdc965c/packages/terra-dropdown-button/src/SplitButton.jsx#L126-L127)

packages/terra-dropdown-button/src/SplitButton.jsx
supreethmr: nice catch :)  I forgot that length will `undefined` when children itself is item and not array.  
IMO when list has only one item then it will be considered as both first and last item of the list. so updated the same by using [ React.Chlidren.count()](https://github.com/cerner/terra-core/blob/3d4cbdca1a5c96ae211cdcf5bdb5847ecfdc965c/packages/terra-dropdown-button/src/_DropdownList.jsx#L153)

packages/terra-dropdown-button/src/SplitButton.jsx
supreethmr: To keep consistency with attributes had this value here previously.
Simplified to : [this.dropdownList.querySelector('[data-terra-dropdown-first-list-item]')](https://github.com/cerner/terra-core/blob/3d4cbdca1a5c96ae211cdcf5bdb5847ecfdc965c/packages/terra-dropdown-button/src/DropdownButton.jsx#L103)

packages/terra-dropdown-button/src/SplitButton.jsx
supreethmr: I have already [comments](https://github.com/cerner/terra-core/blob/3d4cbdca1a5c96ae211cdcf5bdb5847ecfdc965c/packages/terra-dropdown-button/src/SplitButton.jsx#L150) added for this do you want me change the text in it..?

packages/terra-dropdown-button/src/DropdownButton.jsx
supreethmr: yes, to ensure querySelector() is triggered only when dropdown is open.

supreethmr:
> The deployment seem to be outdated, recent changes need to be re-deployed. I pulled down the branch and noticed that the focus does not go to caret button upon dismissing the dropdown on Safari.
> 
> focus

Fixed it same as primary button : [here](https://github.com/cerner/terra-core/blob/3d4cbdca1a5c96ae211cdcf5bdb5847ecfdc965c/packages/terra-dropdown-button/src/SplitButton.jsx#L112)

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -57,29 +57,60 @@ class DropdownButton extends React.Component {
...
+    this.buttonNode = node;
   }
 
-  handleDropdownButtonClick() {
+  getButtonNode() {
+    return this.buttonNode;
+  }
+
+  handleDropdownButtonClick(event) {
+    if (this.state.isOpen) {
```
nramamurth: Um, can these be clubbed together? 

```
this.setState(prevState => ({ isOpen: !prevState.isOpen, openedViaKeyboard: !prevState.isOpen }));
```

packages/terra-dropdown-button/src/DropdownButton.jsx
nramamurth: Maybe, that could be removed and introduced whenever needed. I don't see its usage in the [original PR](https://github.com/cerner/terra-core/pull/2487) either.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -57,29 +57,60 @@ class DropdownButton extends React.Component {
...
   }
 
   handleDropdownRequestClose(callback) {
-    this.setState({ isOpen: false }, typeof callback === 'function' ? callback : undefined);
+    const callbackOnRequestClose = typeof callback === 'function' ? callback : undefined;
+    this.setState({ isOpen: false, openedViaKeyboard: false, isActive: false }, callbackOnRequestClose);
   }
 
   /*
     In FireFox active styles don't get applied on space
```
nramamurth: This needs to be updated now?

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -57,29 +57,60 @@ class DropdownButton extends React.Component {
...
    */
   handleKeyDown(event) {
-    if (event.keyCode === KeyCode.KEY_SPACE) {
+    if (event.keyCode === KeyCode.KEY_SPACE || event.keyCode === KeyCode.KEY_RETURN) {
       this.setState({ isActive: true });
+      this.setState({ openedViaKeyboard: true });
```
nramamurth: https://github.com/cerner/terra-core/pull/2766#discussion_r360082208

packages/terra-dropdown-button/src/SplitButton.jsx
supreethmr: I think here it was added with reference to [onClick() event of Item.jsx](https://github.com/cerner/terra-core/blob/3d4cbdca1a5c96ae211cdcf5bdb5847ecfdc965c/packages/terra-dropdown-button/src/Item.jsx#L51). Since onClick()  is on nested element `div` and not on `li` this could be added as safety check to prevent event from propagating further. 

I don't think we need it here and as well as in Item.jsx because I don't see any of the click() event defined on parent elements of Item and Primary Button. 

and I have also tested by removing this in all Browsers ( IE10 and 11 ) so I think for now we can remove this.

packages/terra-dropdown-button/src/SplitButton.jsx
supreethmr: Updated change-log and tests for metaData usage : ac055df370ea3dc545ea0f06f8c17ef7f24d3620

packages/terra-dropdown-button/src/DropdownButton.jsx
supreethmr: Edit : I found that callback which we will be receiving in handleDropdownRequestClose() function is onSelect() function of selected item. So on item selection `onSelect(event, metaData)` will be passed callback function and on outside click `mouse-event` will be passed as callback which is not of a type `function` hence we need this check here before invoking callback.

packages/terra-dropdown-button/src/DropdownButton.jsx
supreethmr: setting openedViaKeyboard as batch state update will break functionality of setting focus on list item. Since state update will not happen before DropDownList is rendered condition in focus-trap takes previous value of openedViaKeyboard and sets focus on button even when dropdown is opened by keyboard ( this happens when dropdown is initially opened with mouse click and reverse when opened with keyboard initially ).

packages/terra-dropdown-button/src/DropdownButton.jsx
supreethmr: addressed in this :  https://github.com/cerner/terra-core/pull/2766#discussion_r361656370

packages/terra-dropdown-button/src/DropdownButton.jsx
supreethmr: updated : 4676898c9beebbed5989f2709f5bf4e42543292b

packages/terra-dropdown-button/src/DropdownButton.jsx
supreethmr: updated : 4676898c9beebbed5989f2709f5bf4e42543292b

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -57,29 +57,66 @@ class DropdownButton extends React.Component {
...
-  /*
-    In FireFox active styles don't get applied on space
-   */
   handleKeyDown(event) {
-    if (event.keyCode === KeyCode.KEY_SPACE) {
-      this.setState({ isActive: true });
+    if (event.keyCode === KeyCode.KEY_SPACE || event.keyCode === KeyCode.KEY_RETURN) {
+      // In FireFox active styles don't get applied on space
+      this.setState({ isActive: true, openedViaKeyboard: true });
+    } else if (event.keyCode === KeyCode.KEY_DOWN && this.state.isOpen && !this.state.openedViaKeyboard) {
```
ShettyAkarsh: can you update this condition as below ?

else if ((event.keyCode === KeyCode.KEY_DOWN **|| event.keyCode === KeyCode.KEY_TAB** ) && this.state.isOpen && !this.state.openedViaKeyboard)

and remove the last elseif for TAB checking

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -96,13 +141,31 @@ class SplitButton extends React.Component {
...
+      // puts focus on first list element on down arrow key press when dropdown is opened by mouse click
+      const firstOption = this.dropdownList.querySelector('[data-terra-dropdown-first-list-item]');
+      firstOption.focus();
+      // required to prevent handleFocus() callback of DropdownList.
+      event.preventDefault();
+    } else if (event.keyCode === KeyCode.KEY_UP && this.state.isOpen && !this.state.openedViaKeyboard) {
+      // puts focus on last list element on up arrow key press when dropdown is opened by mouse click
+      const lastOption = this.dropdownList.querySelector('[data-terra-dropdown-last-list-item]');
+      lastOption.focus();
```
ShettyAkarsh: Needs confirmation on keyUp. should we implement this behavior of navigating from bottom to top ? when the dropdown is opened via mouseclick.

cc. @neilpfeiffer , @PaulSolDev

supreethmr: Added this functionality with reference to : https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-2/menubar-2.html

packages/terra-dropdown-button/src/terra-dev-site/test/dropdown-button/DefaultDropdownButton.test.jsx
```diff
@@ -8,9 +8,9 @@ const cx = classnames.bind(styles);
 export default () => (
   <div className={cx('container-spacing-wrapper')}>
     <DropdownButton label="Dropdown" id="dropdown">
-      <Item label="1st" onSelect={() => {}} />
-      <Item label="2nd" onSelect={() => {}} />
-      <Item label="3rd" onSelect={() => {}} />
+      <Item label="1st" metaData={{ key: '1st Option' }} onSelect={(event, metaData) => { console.log('Metdata of :', metaData.key); }} />
```
nramamurth: nit: metaData*

packages/terra-dropdown-button/src/terra-dev-site/test/dropdown-button/DefaultDropdownButton.test.jsx
```diff
@@ -8,9 +8,9 @@ const cx = classnames.bind(styles);
...
+      <Item label="2nd" metaData={{ key: '2nd Option' }} onSelect={(event, metaData) => { console.log('Metdata of :', metaData.key); }} />
```
nramamurth: Logs do not really help validate the functionality. I think you could enhance this to maybe enclose this in a `<p>` tag or something and show it on the page so that this can be caught in the test screenshot when either the primary button or an item is selected.

packages/terra-dropdown-button/src/_DropdownList.jsx
```diff
@@ -145,17 +149,28 @@ class DropdownList extends React.Component {
     return React.Children.map(this.props.children, (child, index) => React.cloneElement(child, {
       isActive: index === this.state.active,
       requestClose: this.props.requestClose,
+      ...(index === 0) && { 'data-terra-dropdown-first-list-item': true },
+      ...(index === React.Children.count(this.props.children) - 1) && { 'data-terra-dropdown-last-list-item': true },
```
StephenEsser: Something to consider. All dropdown list items could have the same data attribute instead of calculating a first and last item. In the scenario where a dropdown list contains only a single option both first and last data attributes would be applied.

```jsx
// Array index 0 is the first item and length - 1 is the last item. Assuming length is always >= 1. 
this.dropdownList.querySelectorAll('[data-terra-dropdown-list-item]'); 
```

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -57,29 +57,62 @@ class DropdownButton extends React.Component {
...
 
-  /*
-    In FireFox active styles don't get applied on space
-   */
   handleKeyDown(event) {
-    if (event.keyCode === KeyCode.KEY_SPACE) {
-      this.setState({ isActive: true });
+    if (event.keyCode === KeyCode.KEY_SPACE || event.keyCode === KeyCode.KEY_RETURN) {
+      // In FireFox active styles don't get applied on space
+      this.setState({ isActive: true, openedViaKeyboard: true });
```
StephenEsser: I'm seeing strange keyboard interactions when using the space and enter keys.

The dropdown open and closes when opened by long pressing the enter key.

1. Open the dropdown-button example page and use tab to navigate and focus a dropdown button.
2. Press and hold enter.
3. Dropdown opens
4. Release enter.
5. Observe the dropdown closes.

Repeat the steps above using spacebar instead of enter and the results will be different. Some of this is likely native browser behavior with keyboard interactions but the set state calls happening seem to put the component into a bad state.

packages/terra-dropdown-button/src/DropdownButton.jsx
StephenEsser: I'm wondering if openedViaKeyboard should be a state variable at all. Calling setState for openedViaKeyboard alone doesn't appear to change any of the rendering of the component. This may be a case where the variable makes sense as a class variable and not a state variable.

```jsx
this.openedViaKeyboard = false;
```

Does making a state change to openedViaKeyboard have an effect if the component is not in an isOpen state?

```jsx
this.setState({ openedViaKeyboard: false }); // Calling this but never setting isOpen to true/false
```

StephenEsser: There seems to be some sync issues with openedViaKeyboard and isOpen when the state is updated multiple times during different events. If those can be reconciled this can stay as a state variable.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -57,29 +57,62 @@ class DropdownButton extends React.Component {
...
+    } else if ((event.keyCode === KeyCode.KEY_DOWN || event.keyCode === KeyCode.KEY_TAB) && this.state.isOpen && !this.state.openedViaKeyboard) {
```
StephenEsser: We may need to get UX guidance on the use of the tab key to shift focus the dropdown. Using the example mentioned on the issue for reference the behavior is different.

(Mouse click to open dropdown and then press tab)

https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-2/menubar-2.html

packages/terra-dropdown-button/src/DropdownButton.jsx
supreethmr: It looked like default behaviour for RETURN Key press in html button that triggers click event of button. but when RETURN key was held for longer time keydown event and onClick event was being executed multiple times. this was causing dropdown to open and close by itself when  RETURN key was held for longer time.

Incase of Space key press keyup event was called following onClick callback which made dropdown to open after removing hold from space key. thus causing dropdown to open as expected.

hence prevented click call back in keyDown event and moved open logic to key-up event.

packages/terra-dropdown-button/src/DropdownButton.jsx
supreethmr: restored tab key to previous functionality of closing dropdown on Tab Key press. 

Thank you for pointing it out, I was misguided by initial WDIO test code changes  which had Tab key press logic implemented to access list items in both split button and dropdown button. 
Reverted WDIO changes to Previous code and replaced Tab with ArrowDown key press to access list items when opened with mouse click.

packages/terra-dropdown-button/src/_DropdownList.jsx
supreethmr: updated to queryselectorAll : [here](https://github.com/cerner/terra-core/blob/07833e70cea6c20f52e6cfb6b9bee03b4794740d/packages/terra-dropdown-button/src/DropdownButton.jsx#L112)

packages/terra-dropdown-button/src/terra-dev-site/test/dropdown-button/DefaultDropdownButton.test.jsx
supreethmr: Added <p> to display selected value : [code link](https://github.com/cerner/terra-core/blob/abe8f36b8488383183a4a06827b136a3c0b83a45/packages/terra-dropdown-button/src/terra-dev-site/test/dropdown-button/DefaultDropdownButton.test.jsx#L13-L16)

packages/terra-dropdown-button/src/terra-dev-site/test/dropdown-button/DefaultDropdownButton.test.jsx
supreethmr: [updated](https://github.com/cerner/terra-core/blob/dropdown-keyboard-mouse-a11y/packages/terra-dropdown-button/src/terra-dev-site/test/dropdown-button/DefaultDropdownButton.test.jsx)

PaulSolDev:
Functional verification complete.  No issues identified.

PR #2769, opened 2019-12-03
Fix intl Locale Loading for Sequential Locale Changing

### LINKED ISSUE #2743: [terra-i18n] Locale changes do not add localeData after initial IntlPolyfill load
# Bug Report
Locale data is not added when locale changes after initial load of the [intl polyfill.](https://github.com/andyearnshaw/Intl.js) This impacts IE10 usage of [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat) and [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DateTimeFormat) when the locale is changed such as terra-dev-site usage.  These are used by react-intl in as part of the number and date formatting APIs.

## Description
Very simple to reproduce, but it is in parts of the API that many consumers may not be needing to use regularly.

terra-i18n loads the polyfill upon initial load and correctly adds localeData to Intl.  Upon subsequent locale changes the localeData is not populated, due to the global being present.  This means that a locale change during operation in IE10 will cause portions of the page to be localized, but (in)direct usage of the Intl methods may not be correct if the locale does not honor the same formatting standards.

Formatter Component Example:
https://gist.github.com/PolymorphedSquirrel/c98eb499035fcd6125123ffe263f24fe

Test Example for use in TDS:
https://gist.github.com/PolymorphedSquirrel/b1162b169082c4b060f00f8b76df41c8

## Steps to Reproduce

1. Load page as en-US, change locale via TDS settings/locale to something that does not use , as thousands grouper and . as decimal delimiter.
2. this.props.intl.formatNumber(1234.1);
3. Put resulting string to the page.

Consider 'sv-SE' below:

intl.formatNumber(1234.1)
expected: 1 234,1
actual: 1,234.1

## Expected Behavior
Polyfilled Intl API formats data for locale if locale changes.

## Possible Solution
Check to see if the targeted locale is present in the locale data, and if not present, perform loadIntl(locale) despite the previous existence of global.Intl due to a prior polyfill load.  

[Current Implementation](https://github.com/cerner/terra-core/blob/master/packages/terra-i18n/src/i18nLoader.js#L14)

areIntlLocalesSupported is included in the example from [Intl.js](https://github.com/andyearnshaw/Intl.js#intljs-and-node) and should be able to help further refine the process.  Potentially looking something like https://gist.github.com/PolymorphedSquirrel/4dcccf4e8b272b672ca94099a2ebab73 and triggering the check when translations are loaded (indicating the locale change).

I validated something very similar with performing a dynamic require inside a utility component dependent upon locale and observed correct results in IE10, IE11, Edge, Firefox, Chrome, and Safari.
```
React.useEffect(() => {
    if (locale) {
      // TODO: validate locale not supported before adding
      if (global.IntlPolyfill) {
        require(`intl/locale-data/jsonp/${locale}`);
      }
    }
  }, [locale]);
```

## Environment

* Component Name and Version: terra-i18n 4.19.0
* Browser Name and Version: IE10 through compatibility mode
* Node/npm Version: n/a
* Webpack Version: n/a
* Operating System and version (desktop or mobile): Windows

## @ Mentions
Spoke with @emilyrohrbough about this and narrowed down the location in terra-i18n allowing for this bug to manifest.

emilyrohrbough:
### Summary
Fixes #2743

Load Locale data if it hasn't already been loaded.


packages/terra-i18n/src/intlLoaders.js
```diff
@@ -1,29 +1,53 @@
-/* eslint-disable */
+/* eslint-disable import/no-unresolved, compat/compat, no-console */
 import intlLoaders from 'intlLoaders';
 
+const hasIntlData = (locale) => {
```
emilyrohrbough: The `intl` documentation shows an[ example of checking if the local data has already been loaded](https://www.npmjs.com/package/intl#intljs-and-node) by using a module called `intl-locales-supported`. That module accepts a list of app locales to determine which locale data should be loaded. This is a flavor of the same tool. [Here is their code for reference](https://github.com/formatjs/formatjs/blob/master/packages/intl-locales-supported/src/index.ts#L7) If we want to eventually worry about internationalized pluralization rules, it might be beneficial to pull it in directly.

Thoughts?

packages/terra-i18n/src/terra-dev-site/doc/example/I18nDemo.jsx
```diff
@@ -30,7 +30,18 @@ class Demo extends React.Component {
...
+        <div>
+          <span> Example Message Format: </span>
+          <FormattedMessage id="Terra.ajax.error" />
+        </div>
+        <p>
+          <span> Example Number Format: </span>
+          <FormattedNumber value={parseFloat('1123432.123')} />
+        </p>
+        <p>
+          <span>Example Date Format: </span>
```
emilyrohrbough: Something to note about Intl Formatting, if the locale data is not loaded, it will fallback to the browser default. This is a Intl implementation detail: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#Locale_negotiation

packages/terra-i18n/src/I18nProvider.jsx
```diff
@@ -16,7 +14,7 @@ const propTypes = {
   /**
    * Translations messages object.
    */
-  messages: PropTypes.object.isRequired,
+  messages: PropTypes.shape({}).isRequired,
```
tbiethman: 👀

packages/terra-i18n/src/intlLoaders.js
tbiethman: https://www.npmjs.com/package/intl-locales-supported

0 dependencies, not much to it. I wouldn't be opposed to pulling it in directly if it suits our needs.

packages/terra-i18n/src/intlLoaders.js
```diff
@@ -1,29 +1,53 @@
...
+    Intl.NumberFormat,
+  ].filter((intlConstructor) => intlConstructor);
+
+  if (intlConstructors.length === 0) {
+    return false;
+  }
+
+  return intlConstructors.every((intlConstructor) => {
+    const supportedLocalesFromProvidedList = intlConstructor.supportedLocalesOf([locale], { localeMatcher: 'lookup' });
+    return !!supportedLocalesFromProvidedList.length && supportedLocalesFromProvidedList[0] === locale;
```
benbcai: I didn't know you can use `!!` to convert it.

packages/terra-i18n/src/intlLoaders.js
ryanthemanuel: +1 let's just pull it in ourselves.

packages/terra-i18n/src/intlLoaders.js
emilyrohrbough: I recently discovered it in another project!

PaulSolDev:

PR #2771, opened 2019-12-04
[terra-status-view] Updates message content to use semantic html element.

### LINKED ISSUE #2697: [terra-status-view] Accessibility issues with <div> usage
# Bug Report

The status view title and message should use appropriate semantic HTML elements instead of divs to meet accessibility requirements. At the very least, the divs should be replaced with paragraph elements. Ideally, the status view title would be able to map to various header elements.

## @ Mentions
@scottwilmarth @neilpfeiffer

lokesh-0813:
Fixes #2697


lokesh-0813:
There were some difference in few of the wdio screenshots and hence were re-generated.

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -118,9 +118,9 @@ const StatusView = ({
   let messageSection;
   if (message) {
     messageSection = (
-      <div className={cx('message')}>
+      <p className={cx('message')}>
```
StephenEsser: Regarding the screenshot changes, we should be able to style the paragraph tag to realign with the styles the div was providing previously.

lokesh-0813: Styles adjusted [here](544daee1958f769bd61b6c9d1d10882921de08ea).

packages/terra-status-view/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* `<div>` tag replaced with `<p>` in message content.
```
yuderekyu: can we specify that this change is for a11y?

jeremyfuksa:
In the issue discussion, @neilpfeiffer suggested both the error title and the message be rendered as `<p>` elements. I'm only seeing that the message was changed to `<p>`.

packages/terra-status-view/CHANGELOG.md
lokesh-0813: Updated here 8f59698af7fd91b2b9c32979400bc90749734ce6

lokesh-0813:
> In the issue discussion, @neilpfeiffer noted that both the error title and the message be rendered as `<p>` elements. I'm only seeing that the message was changed to `<p>`.
> 
> [Title](https://github.com/cerner/terra-core/blob/444e672dd2d5822b1d6c675b04e716ddb27e0431/packages/terra-status-view/src/StatusView.jsx#L112) needs to be changed as well, making sure that no global paragraph styling bleeds in.

@jeremyfuksa I misunderstood the comment.
Here is a change to title content 8f59698af7fd91b2b9c32979400bc90749734ce6

PR #2773, opened 2019-12-05
[terra-form-select] Fixed page jump within an Iframe/Embedded Browser

### LINKED ISSUE #2682: [terra-form-select] Clicking to view dropdown causes window to scroll to the top while embedded in mpage
# Bug Report

## Description
While in an mpage, when you click the terra-form-select(default variant) to display the drop down contents, the page will scroll all the way up making it impossible to see the results. The popup is open and you can choose an option via keyboard(without being able to see it) or use scroll wheel to scroll down, but if you click the scrollbar the dropdown closes.  This is behavior that will prevent consumers from performing this workflow.

HC Scrolling - status

## Steps to Reproduce

1. Open patient chart in solgm for a patient with health concerns. (Schreur,Grace)
2. Open OnePlan Plus Button mpage tab
3. Click a healthconcern in the list to open detail view.
4. Click the modify button(top right) to open the modify view.
5. Scroll down to "Health Concern Status" and click the drop down button.

The outermost mpage scroll bar will go from its current position all the way up.  Depending on resolution, this bar could be all the way up already and thus this issue may not bee seen.  But when using smaller resolutions, the user will see the drop down scroll out of view as in the gif. This most likely will happen normally in resolutions used on citrix by clients.
On a larger resolution monitor, you can reliably recreate the issue by shrinking down the size of the powerchart window.

Example on 4k monitor shrinking down the patient window slightly to recreate:
terra-form-select_scroll_issue_4k

When the default variant is used, the page will scroll up.  On the same health concerns modify view when the "expressed by" drop down is chosen, the page does not scroll.  That drop down is using the "combo box" variant.  

[Teams discussion](https://teams.microsoft.com/l/message/19:0adf7161dd244f87aa6415f7c5577836@thread.skype/1570042649980?tenantId=fbc493a8-0d24-4454-a815-f4ca58e8c09d&groupId=45f47bfa-38a9-444d-8933-9c20266ae38c&parentMessageId=1570042649980&teamName=Terra&channelName=Component%20Support&createdTime=1570042649980)

## Expected Behavior
Do not change the scroll position when clicking drop down or at least keep the drop down control in view if the scroll position reacts to opening the drop down.

## Environment
* Component Name and Version: Plan-js 1.2.1
* Browser Name and Version: IE11 11.737.17763.0
* Operating System and version (desktop or mobile): Windows 10 1809 (OS Build 17763.737)
* Powerchart: 2018.01 - solgm
* Mpages: 6.12 - http://ip85wodr12a/mpage-content/solgm.ip.devcerner.net/MPAGES_6_12

## @ Mentions
@StephenEsser

supreethmr:
### Summary
Resolves #2682 

### Additional Details
Page jump was happening at two scenarios where one is when single-select is opened before selecting dropdown items and another one is opening form-select when selection has been made.
- menu.focus()
- scrollIntoView()

Focus to menu(dropdown) was setting in opendropdown() method which triggers before PositionDropdown causing page to scroll to the top on form-select open.

scrollIntoView() is called to scroll into active option on dropdown open. ScrollIntoView causes page Jump as shown in Issue when used within Iframe. Hence replaced ScrollIntoView with ScrollTop of menu ( dropdown ).


packages/terra-form-select/src/single/Frame.jsx
```diff
@@ -195,38 +195,11 @@ class Frame extends React.Component {
...
     }
 
-    /**
-     * Avoids focusing the input if the toggle button is used to open the select menu.
-     * This is to avoid an issue with VoiceOver on iOS where shifting to toggle button while the
-     * input is focused / onScreen keyboard is open unexpected focus shift when the onScreen
-     * keyboard is closed
-     */
-    if (event && event.target
-      && (event.target.hasAttribute('data-terra-form-select-toggle-button')
```
supreethmr: Removing this condition check since data-terra-form-select-toggle-button and data-terra-form-select-toggle-button-icon are not included in single-select.

packages/terra-form-select/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ ChangeLog
 
 Unreleased
 ----------
+### Fixed 
+* Fixed Page Scroll Issue of form-select.
+* when form-select is used with-in Iframe / Embeded browser clicking on toggle button to open dropdown scrolls page to top of the window.
+* This error was discovered in powerchart application which uses Iframe to render Terra-components.
```
StephenEsser: We should remove internal specific implementation from the change log list

packages/terra-form-select/CHANGELOG.md
```diff
@@ -3,6 +3,10 @@ ChangeLog
 
 Unreleased
 ----------
+### Fixed 
+* Fixed Page Scroll Issue of form-select.
```
StephenEsser: Can these entries be consolidated? Each item in a list should be for a specific change, here there are three asterisks for one entry.

Example:

### Fixed 
* Fixed page scrolling when embedded within an iframe

supreethmr: sure, updated [change-log](https://github.com/cerner/terra-core/pull/2773/commits/a428f6c4bd960d219d453cc40d5239830ecb81cd)

packages/terra-form-select/CHANGELOG.md
supreethmr: removed : a428f6c4bd960d219d453cc40d5239830ecb81cd

ShettyAkarsh:
@supreethmr  can you add the wdio test case for testing this changes using iframe. with tiny or small viewports ?

supreethmr:
> @supreethmr can you add the wdio test case for testing this changes using iframe. with tiny or small viewports ?

Added here : 5d47bff7a2090a50ffcf735e9b887dead075cc0a

PR #2781, opened 2019-12-13
[action-footer] Update Docs to Use MDX

emilyrohrbough:
### Summary
Update action-footer docs to use MDX


packages/terra-action-footer/src/terra-dev-site/doc/action-footer/Standard.a.doc.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+```jsx
+import ActionFooter from 'terra-action-footer';
+```
+
+[View component source code](https://github.com/cerner/terra-core/tree/master/packages/terra-action-footer)
+
+## Examples
+<SingleStartActionExample title="Single Start Action" />
+<SingleEndActionExampleSrc title="Single End Action" />
```
StephenEsser: This example isn't rendering in the deployed site. There's a warning in the console:

> Component SingleEndActionExampleSrc was not imported, exported, or provided by MDXProvider as global scope

packages/terra-action-footer/src/terra-dev-site/doc/action-footer/Standard.a.doc.mdx
```diff
@@ -0,0 +1,59 @@
...
+
+Terra provides a Standard, Centered, and Block-style variations for Action Footer:
+* [Standard Action Footer](#usage)
+* [Centered Action Footer](/components/terra-action-footer/action-footer/centered)
+* [Block Action Footer](/components/terra-action-footer/action-footer/block/)
+
+## Getting Started
+
+- Install with [npmjs](https://www.npmjs.com):
+  - `npm install terra-action-footer`
```
StephenEsser: This may be an issue with the mdx loader, but this isn't being syntax highlighed. Not a huge deal, but something we might want to look into.  

image

packages/terra-action-footer/src/terra-dev-site/doc/action-footer/Block.c.doc.mdx
```diff
@@ -0,0 +1,27 @@
+import { Badge } from '../../../../package.json?dev-site-package';
```
StephenEsser: Personal preference, but as we uplift these documents, I think we should number the sort ordering. Instead of alphabetizing it. 

Block.c.doc.mdx -> Block.3.doc.mdx

emilyrohrbough: terra-clinical does that. Way easier to follow

packages/terra-action-footer/src/terra-dev-site/doc/action-footer/Standard.a.doc.mdx
emilyrohrbough: I played around with this, and this is the syntax style that is applied when in a list item. Moving to a paragraph / standalone renders how you'd expect

PR #2796, opened 2020-01-13
[terra-form-select] Announce OptGroup labels to improve screen reader accessibility 

### LINKED ISSUE #2663: [terra-form-select] Option group headings not read on desktop screen readers
# Bug Report

## Description
The option group headings in the form select examples aren't read by VoiceOver or JAWS. Works fine on mobile though.

## Steps to Reproduce
1. Using VO on Safari or JAWS on Chrome, navigate to https://engineering.cerner.com/terra-ui/components/terra-form-select/form-select/select
2. set focus to the Option Groups example dropdown
3. Navigate the dropdown menu items with arrow keys and note that the menu items are read but the option headings are not.

VoiceOver on iOS reads this text correctly.

## Expected Behavior
Screen readers should read the option group heading text.

lokesh-0813:
closes #2663 

Deployment Example: https://terra-core-deployed-pr-2796.herokuapp.com/components/terra-form-select/form-select/select

Found under **Ability to add Option Groups (applies to all)**


packages/terra-form-select/src/single/Menu.jsx
```diff
@@ -229,16 +229,25 @@ class Menu extends React.Component {
       }
     }
 
+    const optGroupElement = MenuUtil.getOptGroupElement(this.props.children, this.state.active);
     const element = MenuUtil.findByValue(this.props.children, this.state.active);
+
+    let displayText;
+    if (optGroupElement) {
+      displayText = `Group ${optGroupElement.props.label}, ${element.props.display}`;
```
emilyrohrbough: Does this need to be fully translated?

packages/terra-form-select/src/shared/_MenuUtil.js
```diff
@@ -126,6 +126,16 @@ class MenuUtil {
...
   }
 
+  /**
+   * Finds the parent optGroup element matching the requested value.
+   * @param {ReactNode} object - The node being searched.
+   * @param {string} value - The value of the target option.
+   * @return {ReactNode|undefined} - The optGroup. Returns undefined if not found.
+   */
+  static getOptGroupElement(object, value) {
+    return React.Children.toArray(object).find(element => element.type.isOptGroup && element.props.children.find(chidleElement => chidleElement.props.value === value));
```
supreethmr: React.Children.toArray() should also be added for `element.props.children` just like single opt-group in form-select there can be single item within opt-group.

lokesh-0813: Updated here edaea09de8e7b5b290819fa74435aa66ddf0e4c8

lokesh-0813:
> I did some testing in JAWS. Is there any way to get the screen reader to announce the group before the option when it moves into a new option group?
> 
> e.g. In the Single Select example, when switching from Aqua (Shade of blue) to Green (Shades of green), JAWS reads "Green. To move to an item, press the arrow keys. Group Shades of green, green."
> 
> Seems like it would be beneficial for JAWS to read, "Group Shades of green, green. To move to an item, press the arrow keys," instead. But only on the first item in an optgroup.

Hey @jeremyfuksa , this seems to be a common behaviour across all examples in terra-form-select for `single` variant when testing with JAWS. I have been looking into what is causing this. Will post an update if I figure out what's happening.

lokesh-0813:
> I did some testing in JAWS. Is there any way to get the screen reader to announce the group before the option when it moves into a new option group?
> 
> e.g. In the Single Select example, when switching from Aqua (Shade of blue) to Green (Shades of green), JAWS reads "Green. To move to an item, press the arrow keys. Group Shades of green, green."
> 
> Seems like it would be beneficial for JAWS to read, "Group Shades of green, green. To move to an item, press the arrow keys," instead. But only on the first item in an optgroup.

I tested native html select with optGroups in multiple websites with JAWS. This appears to be the common behaviour. The only key difference is in terra-form-select "To move to an item, press the arrow keys." is read out for all options and where as in native select its only read on first option. Also JAWS somehow fails to pick up optGroup titles on native select.

To put it in JAWS perspective


Blue. To move to an item, press the arrow keys. Group shades of blue, blue.
<--------------------------------------------> <------------------------->
```Default JAWS behaviour ------------------- Terra's visually hidden text```

jeremyfuksa:
Sounds like our hands are tied, then.

lokesh-0813:
Raised a JIRA for translations

[Jira Link](https://jira.cerner.com/browse/TRANS-7726)

ryanthemanuel:
@lokesh-0813 saw your comment on the translation JIRA.  We don't worry about finnish translations since it's not officially supported (see https://github.com/cerner/terra-core/blob/master/packages/terra-dropdown-button/translations/fi-FI.json) for an example of what we typically do.

packages/terra-form-select/src/single/Menu.jsx
lokesh-0813: Added here 7defe48c45eaa2e6265e7dfea86ffe9496c3afea
Type fixed here 9c75500046b25e53bb41fc7c6419936021756488

lokesh-0813:
> @lokesh-0813 saw your comment on the translation JIRA. We don't worry about finnish translations since it's not officially supported (see https://github.com/cerner/terra-core/blob/master/packages/terra-dropdown-button/translations/fi-FI.json) for an example of what we typically do.

@ryanthemanuel  Translations Added here 7defe48c45eaa2e6265e7dfea86ffe9496c3afea
Typo fixed here 9c75500046b25e53bb41fc7c6419936021756488.

I noticed that recent translations for Finnish have been ignored in terra-form-select. Except AR and FI everything else is updated.

lokesh-0813:
> Before merging I’d recommend confirming with the translations team that these translations can be combined in this format.
> 
> From the JIRA is appears the requested translations were specifically for the word “Group”
> 
> But the sentence translations structure is:
> 
> `Group ${groupName}, ${optionDisplay}`
> 
> or
> 
> `Group ${groupName}, ${optionDisplay} Selected.`
> 
> Not all languages share the same left to right sentence construction. The end result looks like this:
> 
> > Group Shade of blue, Blue
> > Group Shade of blue, Blue Selected.
> 
> Picture for reference:
> image

@StephenEsser 
Translations looks good for majority of the languages.

[JIRA LINK](https://jira.cerner.com/browse/TRANS-7726)

StephenEsser:
Thank you for verifying the translations @lokesh-0813

Issue #2727, opened 2019-10-28
terra-search-field onChange not triggered when the value is set programmatically


FloraChenCerner:
# Bug Report

## Description

When we were testing dictation (Nuance dictation) in one of our search field, after the character is added to the field by the dictation, the auto search is not triggered. Clicking on the search button doesn't work either. We believe what happened was terra-search-field onChange/onSearch not triggered when the value is set programmatically.

## Steps to Reproduce

Set the value of the search input programmatically and observe the auto search is not triggered.
Click on the search button and observe the search is not triggered

## Expected Behavior

We expect onChange/onSearch to be triggered when the search value is set programmatically, which means the auto search and clicking on the search button should trigger the search.
## Environment

* Component Name and Version: terra-search-field 3
* Browser Name and Version: iOS WKWebview (shouldn't matter what browser is used)
* Node/npm Version: [e.g. Node 8/npm 5] Node 8.7
* Webpack Version: 4
* Operating System and version (desktop or mobile): iOS 13 (shouldn't matter what system is used)

## @ Mentions

@dkasper-was-taken

lokesh-0813:
@FloraChenCerner The `input` used in `terra-search-field` is a wrapper around native `<input>` element. It is the default functionality of the native element to not fire `onChange` event on adding text programmatically. 
[Here](https://codesandbox.io/s/react-example-ibshc) is an example for the native behaviour.


> When we were testing dictation in one of our search field, after the character is added to the field by the dictation, the auto search is not triggered

For the dictation issue(if you are talking about the macOs dictation accessibility), it seems to work fine. Let me know if i am wrong here.
![ezgif com-video-to-gif (1)](https://user-images.githubusercontent.com/16148882/67761918-3853bc80-fa6a-11e9-83de-e93a7ab44032.gif)


@StephenEsser `handleSearch` in `terra-search-field` is only called on click of search icon or on `onChange` event for input. Any inputs on this?

FloraChenCerner:
When saying dictation, I refer to the Nuance dictation which is a third party dictation tool we use to dictate medical terms. If the macOS dictation is triggering the auto search, is it because it's not programmatically adding text? If the onChange event can't be triggered in this case, can we have something else triggered to make auto search work? Even clicking the search button doesn't work for us.

mjhenkes:
We need design input as to whether dictation should trigger auto search.

Clicking the search button should trigger a search and that is a minor bug. 
@lokesh-0813  You could take a look at this portion.

[...]

PR #2800, opened 2020-01-15
[terra-search-field] adding onInput prop

BenBoersma:
### Summary
Fixes https://github.com/cerner/terra-core/issues/2727
Adding an onInput prop so that users can programmatically set the value of search field, then trigger an onInput event to have the component update the internal variable to the new value.

@cerner/terra

[License]: ../LICENSE


packages/terra-search-field/src/SearchField.jsx
```diff
@@ -107,15 +111,15 @@ class SearchField extends React.Component {
     this.handleTextChange = this.handleTextChange.bind(this);
     this.handleSearch = this.handleSearch.bind(this);
     this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.handleInput = this.handleInput.bind(this);
     this.setInputRef = this.setInputRef.bind(this);
-    this.updateSearchText = this.updateSearchText.bind(this);
```
emilyrohrbough: I think we still want this to be bound

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -60,6 +60,11 @@ const propTypes = {
    */
   onChange: PropTypes.func,
 
+  /**
+   * Function to trigger when user inputs an value.
```
emilyrohrbough:
```suggestion
   * Function to trigger when user inputs a value.
```

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -107,15 +111,15 @@ class SearchField extends React.Component {
     this.handleTextChange = this.handleTextChange.bind(this);
     this.handleSearch = this.handleSearch.bind(this);
     this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.handleInput = this.handleInput.bind(this);
     this.setInputRef = this.setInputRef.bind(this);
-    this.updateSearchText = this.updateSearchText.bind(this);
 
     this.searchTimeout = null;
-    this.state = { searchText: this.props.defaultValue || this.props.value };
+    this.searchText = this.props.defaultValue || this.props.value;
```
emilyrohrbough: Why did we remove this from state?

packages/terra-search-field/src/SearchField.jsx
emilyrohrbough: Can we add info on the callback params consumers should expect?

packages/terra-search-field/src/terra-dev-site/doc/search-field/SearchField.1.doc.jsx
```diff
@@ -92,6 +94,11 @@ const DocPage = () => (
         example: <SearchFieldFocus />,
         source: SearchFieldFocusSrc,
       },
+      {
+        title: 'Search Field with setting a value programmatically',
+        example: <SearchFieldProgrammaticSet />,
+      source: SearchFieldProgrammaticSetSrc,
```
emilyrohrbough: do we need a first class example for consumers?

packages/terra-search-field/src/SearchField.jsx
BenBoersma: Since we now have both a handleInput and handleTextChange, when you type a value into the input they both fire. Since setState is asynchronous this meant most of the time both would run the updateSearchText forceUpdate calls.

packages/terra-search-field/src/terra-dev-site/doc/search-field/SearchField.1.doc.jsx
BenBoersma: Since there is a bit of complexity getting the expected functionality out of programmatically setting a value because of the need to create a custom 'input' event dispatch, @dkasper-was-taken and I wanted an example.

packages/terra-search-field/src/SearchField.jsx
BenBoersma: Fixed https://github.com/cerner/terra-core/pull/2800/commits/7f1b23a3ff00608c4dcf47ceae24a47d2b94db4d#diff-105fddc0e7a42eb008cb5c09faad5222R116

packages/terra-search-field/src/SearchField.jsx
BenBoersma: Fixed https://github.com/cerner/terra-core/pull/2800/commits/7f1b23a3ff00608c4dcf47ceae24a47d2b94db4d#diff-105fddc0e7a42eb008cb5c09faad5222R64

packages/terra-search-field/src/SearchField.jsx
emilyrohrbough: > Since setState is asynchronous this meant most of the time both would run the updateSearchText forceUpdate calls.

what do you mean?

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -258,6 +275,7 @@ class SearchField extends React.Component {
             ariaLabel={inputText}
             aria-disabled={isDisabled}
             onKeyDown={this.handleKeyDown}
+            onInput={this.handleInput}
```
StephenEsser: We may want to consider adding onInput as first class prop to the terra-form-input/Field so we don't ever lose track of this. Some teams may use dictation for regular inputs.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -162,8 +167,19 @@ class SearchField extends React.Component {
...
+      this.forceUpdate();
+    }
+  }
+
+  handleInput(event) {
+    const textValue = event.target.value;
+    this.updateSearchText(textValue);
+
+    if (this.props.onInput) {
+      this.props.onInput(event, textValue);
```
StephenEsser: Sending back two parameters seems a little strange. We'll want to be as consistent as we can with these callback parameters. If we do decide to add this to the input field we'll have to add additional function logic to return the two parameters as opposed to just passing the function straight through to the input like we do the onChange.

https://github.com/cerner/terra-core/blob/7f1b23a3ff00608c4dcf47ceae24a47d2b94db4d/packages/terra-form-input/src/Input.jsx#L161-L163

Personally, I would keep this native. It's passive to add it in the future, but non-passive to take it away.

```jsx
    if (this.props.onInput) {
      this.props.onInput(event); // Only returning the event.
    }
```

BenBoersma: Should I do the same thing to our onChange handling in search-field as well? Because I was keeping it consistent with how the 'handleTextChange' function handles it's onChange prop https://github.com/cerner/terra-core/pull/2800/files#diff-105fddc0e7a42eb008cb5c09faad5222R161

StephenEsser: No, we'll want to stay passive with the onChange callback. We can grab a few more opinions on this before making any changes. 

My vote is `onInput(event)`.  We can passively add the searchText into the API in the future if it's requested.

packages/terra-search-field/src/SearchField.jsx
BenBoersma: Typing the value into the components results in onInput and onChange being called in quick succession of each other. Fast enough that when I left it as a state and put in console messages to follow logic flow, I kept getting multiple setState calls on each keystroke.

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldProgrammaticSet.jsx
```diff
@@ -0,0 +1,81 @@
+import React from 'react';
+import Button from 'terra-button/lib/Button';
+import SearchField from '../../../SearchField';
+
+class CallbackSearchField extends React.Component {
```
supreethmr: nit: CallbackSearchField -> SearchFieldProgrammaticSet / SearchFieldProgrammaticSearch

ShettyAkarsh:
Found an issue like if we enter the text less than minimum required the message gets displayed saying 'Invalid Text'. this is not clearing out when we immediately add the text programmatically as shown below. 

SearchFieldOnInput

Probably we might have to call handleSearch here https://github.com/cerner/terra-core/blob/f1787a51b480734950a587084eaf853e09faabfd/packages/terra-search-field/src/SearchField.jsx#L182.

supreethmr:
nit : 
1. spell correction in [ReadMe](https://github.com/cerner/terra-core/blob/f1787a51b480734950a587084eaf853e09faabfd/packages/terra-search-field/docs/README.md#implementation-notes)
Screen Shot 2020-01-16 at 12 30 50 PM

2. button seems to be not Aligned in example and test page : 
Screen Shot 2020-01-16 at 11 51 31 AM

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -60,6 +60,11 @@ const propTypes = {
    */
   onChange: PropTypes.func,
 
+  /**
+   * Function to trigger when user inputs a value. Sends parameters {Event} event and {String} searchText
```
supreethmr: Do we need to be more specific with description here like :
>  Function to trigger when user programmatically changes a value OR Function to trigger when value changes programmatically OR something more that explains the purpose this function being added in Search-field.

BenBoersma: > Found an issue like if we enter the text less than minimum required the message gets displayed saying 'Invalid Text'. this is not clearing out when we immediately add the text programmatically as shown below.
> 
> SearchFieldOnInput
> 
> Probably we might have to call handleSearch here
> 
> https://github.com/cerner/terra-core/blob/f1787a51b480734950a587084eaf853e09faabfd/packages/terra-search-field/src/SearchField.jsx#L182
> 
> .

So we have two reasons we don't want handleInput to trigger the search function. 

The first reason is we don't want to double trigger the search when typing value, since onInput and onChange would fire. 

Second reason is that we don't want to give external control over when we search.

cc: @dkasper-was-taken

BenBoersma:
> Found an issue like if we enter the text less than minimum required the message gets displayed saying 'Invalid Text'. this is not clearing out when we immediately add the text programmatically as shown below.
> 
> SearchFieldOnInput
> 
> Probably we might have to call handleSearch here
> 
> https://github.com/cerner/terra-core/blob/f1787a51b480734950a587084eaf853e09faabfd/packages/terra-search-field/src/SearchField.jsx#L182
> 
> .

So we have two reasons we don't want handleInput to trigger the search function. 

The first reason is we don't want to double trigger the search when typing value, since onInput and onChange would fire. 

Second reason is that we don't want to give external control over when we search.

cc: @dkasper-was-taken

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldProgrammaticSet.jsx
BenBoersma: Fixed https://github.com/cerner/terra-core/pull/2800/commits/48780c9b67df1f5c0e433435270b039142582536#diff-5b54d3bd01a4e6076cc428ee0d369f6cR9 and https://github.com/cerner/terra-core/pull/2800/commits/48780c9b67df1f5c0e433435270b039142582536#diff-5b54d3bd01a4e6076cc428ee0d369f6cR9

BenBoersma:
> nit :
> 
> 1. spell correction in [ReadMe](https://github.com/cerner/terra-core/blob/f1787a51b480734950a587084eaf853e09faabfd/packages/terra-search-field/docs/README.md#implementation-notes)
>    Screen Shot 2020-01-16 at 12 30 50 PM

Fixed https://github.com/cerner/terra-core/pull/2800/commits/48780c9b67df1f5c0e433435270b039142582536#diff-0bbd442e9dea05d7972c8517ff7ce076R27

> 2. button seems to be not Aligned in example and test page :
>    Screen Shot 2020-01-16 at 11 51 31 AM

Fixed https://github.com/cerner/terra-core/pull/2800/commits/48780c9b67df1f5c0e433435270b039142582536

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -162,8 +167,19 @@ class SearchField extends React.Component {
   }
 
   updateSearchText(searchText) {
-    if (searchText !== undefined && searchText !== this.state.searchText) {
-      this.setState({ searchText });
+    if (searchText !== undefined && searchText !== this.searchText) {
```
yuderekyu: to cover all bases, is it worth updating this to a `typeof` check to protect from undeclared variables?
```suggestion
    if (typeof searchText !== 'undefined' && searchText !== this.searchText) {
```

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldProgrammaticSet.jsx
```diff
@@ -0,0 +1,85 @@
...
+  setTextClick() {
+    if (this.searchInput) {
+      this.searchInput.value = 'IpsumLorem';
+      let customEvent;
+      if (typeof (Event) === 'function') {
+        customEvent = new Event('input', {
+          bubbles: true,
+          cancelable: false,
+        });
+        customEvent.simulated = true;
```
yuderekyu: what's the simulated for? I don't see that in the event api

packages/terra-search-field/tests/wdio/search-field-spec.js
```diff
@@ -134,6 +134,47 @@ Terra.describeViewports('Search Field', ['medium'], () => {
...
+    Terra.it.validatesElement('cut');
+
+    it('should trigger search when pasting', () => {
+      // paste value from clipboard
+      browser.addValue('input', ['Control', 'v']);
+    });
+
+    Terra.it.validatesElement('paste');
+
+    it('should not trigger search when programatticaly setting value', () => {
```
yuderekyu: nit *programmatically

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldProgrammaticSet.jsx
BenBoersma: Actually I think I left that in on accident. When I was trying to get onChange to be simulated, it was supposed to let React know that it was a "simulated" onChange event and not do the checks that prevent onChange from being programmatically fired like this.

BenBoersma: Yep, totally left it in on accident, still got same behavior without it. Fixed here https://github.com/cerner/terra-core/pull/2800/commits/b86945e9e390e61f2836c59e2fe31688bc0cd737#diff-5b54d3bd01a4e6076cc428ee0d369f6cL35

packages/terra-search-field/tests/wdio/search-field-spec.js
BenBoersma: Fixed https://github.com/cerner/terra-core/pull/2800/commits/b86945e9e390e61f2836c59e2fe31688bc0cd737#diff-dbfac79b3a803165fdc0713a25464bcfR163

supreethmr:
> > Found an issue like if we enter the text less than minimum required the message gets displayed saying 'Invalid Text'. this is not clearing out when we immediately add the text programmatically as shown below.
> > SearchFieldOnInput
> > Probably we might have to call handleSearch here
> > https://github.com/cerner/terra-core/blob/f1787a51b480734950a587084eaf853e09faabfd/packages/terra-search-field/src/SearchField.jsx#L182
> > 
> > .
> 
> So we have two reasons we don't want handleInput to trigger the search function.
> 
> The first reason is we don't want to double trigger the search when typing value, since onInput and onChange would fire.
> 
> Second reason is that we don't want to give external control over when we search.
> 
> cc: @dkasper-was-taken

with reference to this example ( Added text using ref of search-field and not by using value prop of search-field hence we need to dispatch custom event for callback)  I would like to know what happens when text is added programmatically through value prop for search-field(controlled). 
Do consumer need to dispatch custom event even when value prop is changed programmatically..? 

Because i found that even with controlled search-field onChange will not be triggered when text is added programmatically. even search will not triggered until it invoked explicitly.

This could be due to component being unaware to the prop changes. so adding getDerivedStateFromProps() for search-field to update `searchText` whenever value prop is updated and calling handleSearch() would resolve this issue with controlled search-field.

```javascript
static getDerivedStateFromProps(nextProps, prevState) {
  if (nextProps.value !== prevState.prevPropsValue) {
      return {
         searchText: nextProps.value,
         prevPropsValue: nextProps.value,
      };
    }
    return null;
}

componentDidUpdate(prevProps) {
  if (prevProps.value !== this.props.value) {
      this.handleSearch();
  }
}

```

packages/terra-search-field/src/SearchField.jsx
BenBoersma: As per offline discussion, I have switched to only returning the even on the onInput prop call https://github.com/cerner/terra-core/pull/2800/commits/d02ec4b2a3e498a93dca1c78cf449791f00389d2

BenBoersma:
> with reference to this example ( Adding text using ref of search-field and not by using value prop of search-field hence we need to dispatch custom event for callback) I would like to know what happens when text is added programmatically through value prop for search-field(controlled).
> Do consumer need to dispatch custom event even when value prop is changed programmatically..?

In a controlled search-field where we change the state that is used for prop value, the re-render correctly displays the value, the clear button, and updates the internal variable. Invoking search by pressing the search button causes search-field to try to search with the new value. This has the correct functionality.

However, the current example uses the specific way in which the reported issue was programmatically changing the value. The dictation program they are using sets the value via a reference's value.

> Because i found that even with controlled search-field onChange will not be triggered when text is added programmatically. even search will not triggered until it invoked explicitly.
>
> This could be due to component being unaware to the prop changes. so adding getDerivedStateFromProps() for search-field to update `searchText` whenever value prop is updated and calling handleSearch() would resolve this issue with controlled search-field.

I do not believe we want to trigger a search when values are programmatically set.  In the use-case given to us by the issue reporter, I believe the dictation program they are using sets the value after each "word" is said, which means if we trigger search programmatically we would search on every word.

This was why we're just updating the internal variable so that an explicitly invoked search has the right value and a user could search once they are done dictating.

packages/terra-search-field/src/SearchField.jsx
BenBoersma: What if we wrote something like `Function to trigger when user inputs a value. Use when programmatically setting a value.` 
Though do we want to mention here that users need to manually dispatch the `input` event if they are programmatically setting?

packages/terra-search-field/src/SearchField.jsx
BenBoersma: I agree with this, but don't think it should be part of this PR. Will log an issue.

packages/terra-search-field/src/terra-dev-site/doc/example/SearchFieldProgrammaticSet.jsx
```diff
@@ -0,0 +1,84 @@
+import React from 'react';
+import classNames from 'classnames/bind';
+import Button from 'terra-button/lib/Button';
+import SearchField from '../../../SearchField';
+import styles from '../../test/search-field/examplestyles.scss';
+
+const cx = classNames.bind(styles);
+
+class SearchFieldProgrammaticSet extends React.Component {
```
yuderekyu: Not a big deal, but we should move away from class components and use functional components with hooks, whenever possible.

BenBoersma: Fixed here https://github.com/cerner/terra-core/pull/2800/commits/63069a3151a47224cf7d2c3c4ae05354b364304e#diff-5b54d3bd01a4e6076cc428ee0d369f6cR1

packages/terra-search-field/src/SearchField.jsx
BenBoersma: Set a more verbose message here https://github.com/cerner/terra-core/commit/63069a3151a47224cf7d2c3c4ae05354b364304e#diff-105fddc0e7a42eb008cb5c09faad5222R64

packages/terra-search-field/src/terra-dev-site/doc/search-field/SearchField.1.doc.jsx
```diff
@@ -29,6 +29,8 @@ import SearchFieldFilterNumeric from '../example/SearchFieldFilterNumeric';
 import SearchFieldFilterNumericSrc from '!raw-loader!../../../../src/terra-dev-site/doc/example/SearchFieldFilterNumeric';
 import SearchFieldFocus from '../example/SearchFieldFocus';
 import SearchFieldFocusSrc from '!raw-loader!../../../../src/terra-dev-site/doc/example/SearchFieldFocus';
+import SearchFieldProgrammaticSet from '../example/SearchFieldProgrammaticSet';
+import SearchFieldProgrammaticSetSrc from '!raw-loader!../../../../src/terra-dev-site/doc/example/SearchFieldProgrammaticSet';
```
mjhenkes: Since you're already updating this file you could switch it over to mdx

PaulSolDev:
Functional verification complete. No issues identified.

packages/terra-search-field/src/terra-dev-site/doc/search-field/SearchField.1.doc.jsx
BenBoersma: Switched here https://github.com/cerner/terra-core/pull/2800/commits/01caa1861a37d7efb6108958e4396e8fed923d80#diff-03a9a5c821fb01b08d491fecf3df12ccR1

PR #2804, opened 2020-01-21
[terra-profile-image] Added `fit` prop to specify object-fit style

### LINKED ISSUE #2798: ProfileImage sometimes clips images when resizing to fit container
# Feature Request

## Description
The ProfileImage is clipping images that are resized because the object-fit CSS style is being forced to "cover".  This property is being passed to the TerraImage that is being wrapped but cannot be modified by consumers.

Example: https://jira2.cerner.com/browse/PATSEARCH-407

## Steps to Reproduce
1. Load an image that must be scaled to fit the container

https://github.com/cerner/terra-core/blob/master/packages/terra-profile-image/src/ProfileImage.jsx#L55

## Expected Behavior
The expected behavior is that the image is scaled to fit into the container without being clipped.

## Possible Solution
Expose a prop type that allows the consumer to control the object-fit style or change the value passed to "fill" or "contain".

## @ Mentions
@ryanthemanuel

avinashg1994:
### Summary
Added `fit` prop to allows the consumer to control the object-fit style.

Resolves #2798


packages/terra-profile-image/src/ProfileImage.jsx
```diff
@@ -16,6 +16,10 @@ const propTypes = {
    * The text content that specifies an alternative text for an image.
    */
   alt: PropTypes.string,
+  /**
+  * Sets the `fit` prop of terra-image from the following values: `cover`, `contain`, `fill`, `scale-down`, `none`.
```
nramamurth: `terra-image` is internal implementation of this component. I'd not mention it in the documentation.

avinashg1994: I agree. Addressed it here : https://github.com/cerner/terra-core/pull/2804/commits/c0cbab974a3b0b0a47de3ffd04a401a78dfb3265

supreethmr:
Do we need to have example like [Toggle Image Fit](https://engineering.cerner.com/terra-ui/components/terra-image/image/image/) for Profile-Image Also...?
( this might help consumers to choose appropriate values among `{ cover, contain, fill, scale-down, none`} )

packages/terra-profile-image/tests/jest/ProfileImage.test.jsx
```diff
@@ -12,3 +12,9 @@ it('should render an avatar image without Source prop', () => {
   const wrapper = shallow(<ProfileImage alt="Default" height="75" width="75" />);
   expect(wrapper).toMatchSnapshot();
 });
+
+/* Profile Image with fit as contain */
+it('should render a Terra image with a placeholder prop', () => {
```
StephenEsser: The test description should be updated to describe the expectations of the test. It looks like this may have been copy pasted from above.

avinashg1994: Addressed here : https://github.com/cerner/terra-core/pull/2804/commits/87550e65421503ca5a0891c63b2619acaa3d2594

avinashg1994:
> Do we need to have example like [Toggle Image Fit](https://engineering.cerner.com/terra-ui/components/terra-image/image/image/) for Profile-Image Also...?

I agree, and i have added it here : https://github.com/cerner/terra-core/pull/2804/commits/87550e65421503ca5a0891c63b2619acaa3d2594

packages/terra-profile-image/src/ProfileImage.jsx
```diff
@@ -16,6 +16,10 @@ const propTypes = {
    * The text content that specifies an alternative text for an image.
    */
   alt: PropTypes.string,
+  /**
+  * Sets the `object-fit` style of the image from the following values; `cover`, `contain`, `fill`, `scale-down`, `none`.
```
neilpfeiffer: Let's add an additional warning about people photos and which props are recommended they use (and `fill` should never be used since it squishes a face out of proportion):

```suggestion
  * Sets the `object-fit` style of the image from the following values; `cover`, `contain`, `fill`, `scale-down`, `none`.
  * ![IMPORTANT](https://badgen.net/badge//IMPORTANT/CSS?icon=github) Anywhere the terra-profile-image is used to show images of PEOPLE, _only_ `cover` and `contain` are acceptable, `fill` should _never_ be used.
```

neilpfeiffer:
Ready for final review @dkasper-was-taken @tbiethman @mjhenkes @ryanthemanuel 

and will need requirements update @PaulSolDev

avinashg1994:
Changed `fit` prop to be as a string for the following reasons:

1. To prevent a "promise rejection" exception which is being thrown in props table, since the use of Object.values was for some reason causing the exception when using MDX.
2. Also for consistency, as per the [discussion](https://github.com/cerner/terra-framework/pull/663#discussion_r291202091), where we want to keep our prop types as strings

packages/terra-profile-image/src/ProfileImage.jsx
avinashg1994: Updated the warning https://github.com/cerner/terra-core/pull/2804/commits/c2ab5d24867dca1fd16b710775b4557f9ebcdeee, since i have removed the `fill` value from the list of possible values of fit.
Explanation here : https://github.com/cerner/terra-core/pull/2804#issuecomment-592416707

PaulSolDev:
In the example, fit options are not being applied correctly when selected from the dropdown in IE.  

Chrome, Firefox, Edge, and Safari display as expected.

avinashg1994:
> In the example, fit options are not being applied correctly when selected from the dropdown in IE.

This behaviour has now been fixed. The issue was from terra-image which is being consumed in terra-profile-image.
Fix: the [objectFitImages API](https://github.com/fregante/object-fit-images#objectfitimagesimages-options), should be called every-time an update has been made on the image w.r.t to fit property.

PaulSolDev:

packages/terra-profile-image/src/ProfileImage.jsx
```diff
@@ -16,6 +16,11 @@ const propTypes = {
    * The text content that specifies an alternative text for an image.
    */
   alt: PropTypes.string,
+  /**
+  * Sets the `object-fit` style of the image from the following values: `cover`, `contain`, `scale-down`, `none`.
+  * ![IMPORTANT](https://badgen.net/badge/UX/Design-Standards/blue) Anywhere the terra-profile-image is used to show images of People, _only_ `cover` and `contain` are acceptable.
+  */
+  fit: PropTypes.oneOf(['cover', 'scale-down', 'contain', 'none']),
```
cm9361: Why is "fill" not an option for ProfileImage?

packages/terra-profile-image/src/terra-dev-site/doc/example/common/FitTypesWrapper.jsx
```diff
@@ -0,0 +1,32 @@
...
+    render() {
+      return (
+        <div>
+          <WrappedComponent fit={this.state.fitType} {...this.props} />
+          <p><label htmlFor="fitType">Select a Fit Type:</label></p>
+          <select id="fitType" name="fitType" value={this.state.fitType} onChange={this.handleOnSelect}>
+            <option value="cover">cover</option>
+            <option value="contain">contain</option>
+            <option value="scale-down">scale-down</option>
+            <option value="none">none</option>
```
cm9361: Shouldn't there also be a test for the "fill" object fit type?

packages/terra-profile-image/src/ProfileImage.jsx
neilpfeiffer: It was decided by engineering to remove the `fill` prop option based on UX feedback that: 

>  for Profile Images that are intended to be photos of People - only `cover` and `contain` would be the acceptable values. `fill` would never be acceptable by UX as is changes the original image aspect ratio and distorts the face and likeness of the image.

Comment on the original issue here: [Issue #2798(comment)-584576876](https://github.com/cerner/terra-core/issues/2798#issuecomment-584576876) and PR here: [PR#2804(comment)-356539191](https://github.com/cerner/terra-core/pull/2804#pullrequestreview-356539191)

`Cover` is the recommended UX styling prop to use for profile-images and will match how it is currently built in new Avatar, the replacement for profile-image (consumers do not have the option to adjust it in avatar).

packages/terra-profile-image/src/terra-dev-site/doc/example/common/FitTypesWrapper.jsx
neilpfeiffer: Answered above: [PR 2804#discussion_r390656186](https://github.com/cerner/terra-core/pull/2804#discussion_r390656186)

Issue #2801, opened 2020-01-16
[terra-action-header] 'level' prop does not justify its purpose.


lokesh-0813:
# Bug Report

## Description
Providing value to `level` prop does not change the appearance of the text.

## Steps to Reproduce
set level={5}(or any value).
Notice that the heading appearance does not change.

## Expected Behavior
Appropriate heading should be applied if `level` prop is provided.

## Environment
* Component Version:2.34.0 - (January 7, 2020)

## Additional Context / Screenshots
`--terra-action-header-font-size` could be causing the issue.

emilyrohrbough:
@lokesh-0813 I think the level prop is used for accessibility with screen readers, not for altering the styles on the header

ShettyAkarsh:
so in that we might to Update the doc for level prop as following '**Optionally sets the heading level for Screen Readers. One of 1, 2, 3, 4, 5, 6. Default level=1.**. also even for Terra-Section-Headers https://github.com/cerner/terra-core/blob/master/packages/terra-section-header/src/SectionHeader.jsx#L23-L27

PR #2831, opened 2020-02-03
[terra-action-header] Updating Docs

lokesh-0813:
fixes : #2801


StephenEsser:
When creating a PR please use the PR templates that populate. The summary and description fields are helpful to quickly get an understanding of how, why, and what a PR is trying to accomplish without redirecting to the issue thread.

PR #2852, opened 2020-02-19
[terra-icon] Update docs to use mdx syntax

pranav300:
### Summary
updated docs to use MDX syntax.


packages/terra-icon/src/terra-dev-site/doc/icon/Icon.1.doc.mdx
```diff
@@ -18,15 +24,12 @@ This component requires the following peer dependencies be installed in your app
 | react-dom | ^16.8.5 |
 | react-intl | ^2.8.0 |
 
-
-
 <!-- AUTO-GENERATED-CONTENT:END -->
 
 ## Usage
 Each Icon can be imported individually.
```
emilyrohrbough: Should the the code be removed and the section mention looking at the static vs themeable file to see what icons can be imported?

pranav300: I don't think we should be doing that, as the documentation might look vague.
And Earlier we were not showing the toggle code button for the static and theamable, but now it is available. So the user can check the imports from there itself.

Issue #2790, opened 2019-12-27
[terra-button] Long press enables mouse click focus styles


StephenEsser:
# Bug Report

## Description

A [recent change](https://github.com/cerner/terra-core/pull/2740) introduced a timeout to disable focus styling when a button is clicked. The timeout is started on mouse down. However, if you long press before releasing the mouse the focus styles will be applied to the button.

## Steps to Reproduce

1. Navigate to https://engineering.cerner.com/terra-ui/components/terra-button/button/button
2. Click and hold any button example for longer than 300ms
3. Release the mouse while still hovering the button
4. Observe focus styles are applied

## Expected Behavior

The intention of #2740 was to disable button focus style when the button is clicked. 
 
## Possible Solution

Potentially track the event in a different way, without a timeout. 

## Additional Information

Closely related to https://github.com/cerner/terra-core/issues/2789

Issue #2789, opened 2019-12-27
[terra-button] Focus styles are not applied after closing modal manager


StephenEsser:
# Bug Report

## Description

Focus styles are not applied to the button that triggered opening the modal after the modal is closed.

A [recent change](https://github.com/cerner/terra-core/pull/2740) introduced a timeout to disable focus styling when a button is clicked. During WDIO testing it is possible to focus/blur/focus the same button within this timeout resulting in the focus styles not being applied.

## Steps to Reproduce

1. Navigate here https://stephenesser.github.io/button-focus-example/
2. Click the 10ms or 100ms timeout buttons
3. Observe the modal rapidly open and close without restoring the focus style to the button

Note: If you repeat steps 1-3 but long press the mouse click for the 10ms/100ms timeout buttons focus will work as expected. See #2790 for more details.

## Expected Behavior

When a terra-button is programmatically focused it should have focused styling.

## Possible Solution

Remove the timeout and restore the instance variable for shouldShowFocus when applicable.

## Additional details

Example demo: https://stephenesser.github.io/button-focus-example/

Example source code: https://github.com/StephenEsser/button-focus-example

## Example WDIO Test

Here is an example WDIO test example where this can sporadically fail. 

```jsx
  describe('should launch the modal and then dismiss the disclosure', () => {
    before(() => {
      browser.url('/#/raw/tests/modal-button');
      browser.click('#modal-button');
      browser.click('button[aria-label="Close"]');
    });
    Terra.should.matchScreenshot('example', { viewports });
  });
```

StephenEsser:
This issue is closely related to, but may not be identical to, https://github.com/cerner/terra-core/issues/2695

The programatic focus issues mentioned in this issue were introduced in terra-button v3.38.0. This issue is not reproducible in terra-button v3.37.0.

PR #2861, opened 2020-02-26
[terra-button] Fixes focusing issue on button long press, and focus not received on closing modals.

lokesh-0813:
### Summary
Removed timeout from a [previous change](https://github.com/cerner/terra-core/pull/2740) which failed wdio tests if the focus of the button was tested immediately after closing modals.

Removing timeout also fixes a issue where long pressing the button enabled click focus-styles.

closes : #2790 #2789


packages/terra-button/src/Button.jsx
```diff
@@ -127,6 +127,10 @@ class Button extends React.Component {
   }
 
   handleOnBlur(event) {
+    this.shouldShowFocus = false;
+    if (event.currentTarget.type === 'button') {
+      this.shouldShowFocus = true;
+    }
     this.setState({ focused: false });
```
emilyrohrbough: why is `shouldShowFocus` and the `focused` state different?

packages/terra-button/src/Button.jsx
```diff
@@ -199,8 +203,6 @@ class Button extends React.Component {
   handleMouseDown(event) {
     // Prevent button from showing focus styles when clicked
     this.shouldShowFocus = false;
-    // Wait until after onFocus has been triggered on browsers where it will get triggered for click
-    setTimeout(() => { this.shouldShowFocus = true; }, 300);
```
benbcai: A little background for this line of code...

Clicking on a button should not have the focus glow. However, we do want the focus glow to show if the focus is _programmatically_ set on the button. This line was added on #2531 to handle this.

This timeout was initially added by #2531 to allow showing the focus glow when programmatically setting focus on the button. The timeout on this line was increased to 300ms on #2740 because the onClick handler `handleClick` is added and needed to programmatically set focus on the button so that onBlur could get triggered (for FF, Safari, and IE) when the button loses focus. We set the focus in this case not to get the focus glow but. We set the focus on the button so that when it loses focus, onBlur would get called properly.

If we completely removed this time, this would break what was fixed on #2531 because after `handleMouseDown` is called, `this.shouldShowFocus` would still be false and subsequent calls to handleFocus would not be able to set the focus glow.


What I propose is to remove the 300ms and then add `this.shouldShowFocus = false;` and `setTimeout(() => { this.shouldShowFocus = true; });` in `handleClick` like this.

```
if (document.activeElement !== event.currentTarget) {
      this.shouldShowFocus = false;
      event.currentTarget.focus();
      setTimeout(() => { this.shouldShowFocus = true; });
    }
```

lokesh-0813: @benbcai Okay that makes sense. I was setting `this.shouldShowFocus = true` in `handleBlur` since `onBlur` is called before `onFocus` in the chain of events.

I will go ahead and set `this.shouldShowFocus = true` in `handleClick`, But I am curious to know the significance of setTimeOut here, why do we need to do `setTimeout(() => { this.shouldShowFocus = true; });` instead of just `this.shouldShowFocus = true`.

Changes made here 746bdbe9423f725930e2d631bc5b336be1cd55a0

packages/terra-button/src/Button.jsx
lokesh-0813: Few things have changed here 746bdbe9423f725930e2d631bc5b336be1cd55a0. Let me know if you are still confused.

packages/terra-button/package.json
```diff
@@ -42,5 +42,8 @@
     "keycode-js": "^2.0.1",
     "prop-types": "^15.5.8",
     "terra-icon": "^3.27.0"
+  },
+  "devDependencies": {
+    "terra-abstract-modal": "^3.21.0"
```
StephenEsser: The files in terra-dev-site directory are released with the component. This will have to be a hard dependency if we want to keep the example. We'll either need to remove the example or move this to be a hard dependency.

packages/terra-button/src/Button.jsx
```diff
@@ -139,7 +139,9 @@ class Button extends React.Component {
     // Button on Firefox, Safari and IE running on OS X does not receive focus when clicked.
     // This will put focus on the button when clicked if it is not currently the active element.
     if (document.activeElement !== event.currentTarget) {
+      this.shouldShowFocus = false;
       event.currentTarget.focus();
+      setTimeout(() => { this.shouldShowFocus = true; });
```
StephenEsser: Is the timeout necessary? Occasionally unset timeouts are used has a workaround for updates in certain lifecycles, but it may not be necessary here.

packages/terra-button/package.json
lokesh-0813: changed here b5ea39252dacda797d3d046bf3c4a1e6183b80e3

packages/terra-button/src/Button.jsx
lokesh-0813: changed here b5ea39252dacda797d3d046bf3c4a1e6183b80e3

packages/terra-button/src/Button.jsx
benbcai: @lokesh-0813 onFocus gets called immediately after handleMouseDown. If you just set this.shouldShowFocus = true in  handleMouseDown, onFocus will set the `focused` state to true when we don't want it to. Using setTimeout here will sets it back to true _after_ onFocus is called. You'll get a better understanding if you log out the value of this.shouldShowFocus in both onFocus and handleMouseDown and click on the button to see how the value changes.

lokesh-0813: I have updated here 7d43f4888a82a1b871514f0e41cc5390bcd61d28

@benbcai but I'm not sure if lines 200-203 are necessary here. I was assuming `handleMouseDown's` 'eventPreventDefault()' will prevent calling `handleFocus` immediately after mouse down and instead calls `handleClick` where we set `this.shouldShowFocus=false` and fire the focus event manually.

However consoling resulted in same output even in the absence of 200 - 203 lines. Let me know if I am wrong here.

packages/terra-button/package.json
StephenEsser: We should get a few more opinions on this. Adding a transitive dependency just for testing is usually something we should try to avoid when possible.

cc: @tbiethman @mjhenkes

packages/terra-button/src/Button.jsx
```diff
@@ -200,7 +202,7 @@ class Button extends React.Component {
     // Prevent button from showing focus styles when clicked
     this.shouldShowFocus = false;
     // Wait until after onFocus has been triggered on browsers where it will get triggered for click
-    setTimeout(() => { this.shouldShowFocus = true; }, 300);
+    setTimeout(() => { this.shouldShowFocus = true; });
```
StephenEsser: The zero second timeout has been added and removed a couple times. Can you explain why it is needed?

packages/terra-button/package.json
mjhenkes: Terra-button should not have a dependency on terra-abstract-modal.

packages/terra-button/src/Button.jsx
lokesh-0813: @StephenEsser I tried defending not to include. But @benbcai had other suggestions.

You can refer to this [comment](https://github.com/cerner/terra-core/pull/2861#discussion_r390464106) from above.

We need to wait until what @benbcai has to say on this.

packages/terra-button/src/Button.jsx
```diff
@@ -139,7 +139,9 @@ class Button extends React.Component {
     // Button on Firefox, Safari and IE running on OS X does not receive focus when clicked.
     // This will put focus on the button when clicked if it is not currently the active element.
     if (document.activeElement !== event.currentTarget) {
+      this.shouldShowFocus = false;
```
supreethmr: do we need this..? because `this.shouldShowFocus` is being set to true in below line.

lokesh-0813: Yes we need it. `this.shouldShowFocus` is set true after firing the focus event.

packages/terra-button/package.json
ShettyAkarsh: we could implement this example using html-dialog. to avoid the hard dependency on terra-abstract-modal. similar to what we have done for [Hookshot](https://engineering.cerner.com/terra-ui/components/terra-hookshot/hookshot/hookshot) examples before. by removing the hard dependency on terra-form-select.

lokesh-0813: @StephenEsser @mjhenkes @ShettyAkarsh  updated here 716688cc6b76a7f8f6ba276ed52158a5779e87f5

lokesh-0813:
> Terra-button cannot have a dependency on terra-abstract-modal. You'll have to find another way to test these changes.

updated here 716688cc6b76a7f8f6ba276ed52158a5779e87f5

benbcai:
After doing further investigations offline with @lokesh-0813 what we came to understand are the following:

- **handleMouseDown**
  - Previously, `setTimeout(() => { this.shouldShowFocus = true; });` [here](https://github.com/cerner/terra-core/blob/ddd0c422ba513b2d805402cd299a233320686e46/packages/terra-button/src/Button.jsx#L202-L203) was needed to reset `this.shouldShowFocus` back to `true` to prevent the button from getting the focus style when manually clicked because `handleFocus` would get called immediately after `handleMouseDown` (only in Chrome). However, with `event.preventDefault()` added, `handleFocus` would no longer get called automatically in Chrome. Instead, `handleFocus` would get called from `handleClick` for the need to programmatically set focus on the button without the focus style. Given that we have event.preventDefault()`, it is safe to remove `setTimeout(() => { this.shouldShowFocus = true; });` since it is no longer needed.

- **handleClick**
  - Button in Chrome automatically gets focus after clicked. However, FF, IE, and Safari do now. For this reason, we need to programmatically set focus on the button in handleClick. Additionally, since this is a manual click we do not want to apply the focus style on the button. Therefore, it is needed to set `this.shouldShowFocus` to false, before calling `.focus()` and setting `this.shouldShowFocus` back to true after it is done.

PR #2862, opened 2020-02-26
[terra-action-header] Add data attributes to the header buttons

emilyrohrbough:
### Summary
Add data attributes tot he header buttons to use as selectors in tests. Currently you need to know the button order of your render or need to use the class names to select them. The last release updated the buttons and changes the class names. This should help stabilize and clean up the tests that interact with the action-header.


emilyrohrbough:
@tbiethman - I was pretty confused by that as well, I wasn't able to get the snapshots to generated with the data-attribute. I have a feeling it's related to the fact these are custom props being spread on terra-button and we won't see them unless we use render instead of shallow

PR #2880, opened 2020-03-05
[terra-i18n] Update docs to use mdx syntax

pranav300:
### Summary
updated docs to use MDX syntax.


avinashg1994:
nit: For consistency we could number the upgrade guide and changelog as '2' and '3' respectively.

PR #2904, opened 2020-03-13
[terra-form-select] Fix onBlur event handling when selecting an option

### LINKED ISSUE #2845: [terra-form-select] Invalid onBlur events triggered when selecting an option
# Bug Report

## Description

The Select component is triggering inappropriate onBlur events when selecting options.

Switching the variant produces different onBlur results.

Default Variant:

A default variant will fire onBlur every time an option is selected. (Mobile and Desktop)

Multiple/Search/Tag/Combobox:

The onBlur fires when the first option is selected and the dropdown menu closes. (Mobile)

Example Code:

```jsx
import Select from 'terra-form-select';

const Example = () => {
  const [value, setValue] = useState(['blue']);
  const [count, setCount] = useState(0);

  return (
    <div>
      <p>{`Blur Count: ${count}`}</p>
      <Select variant="multiple" onBlur={() => { setCount(count + 1); }} value={value} onChange={setValue}>
        <Select.Option value="blue" display="Blue" />
        <Select.Option value="green" display="Green" />
        <Select.Option value="purple" display="Purple" />
        <Select.Option value="red" display="Red" />
        <Select.Option value="violet" display="Violet" />
      </Select>
    </div>
  );
};

export default Example;
```

## Steps to Reproduce

1. Create a demo page using the code snippet above.
2. Click on the arrow button of the Select to open the dropdown.
3. Select an option.
4. Observe the onBlur event count. 
5. Repeat steps 2-4 with each variant and observe they behave differently.

Note: The behavior is different when the dropdown is opened via clicking the arrow button vs clicking into the text/selection region.

Note: The behavior is different on desktop and mobile. Each variant should be testing on both desktop and mobile.

### Additional Notes

The mobile workflows were tested on mobile iOS Safari using an iPhone X.

## Expected Behavior

The onBlur event should only fire when focus navigates away from the Select. For the tag/multiple variants the dropdown menu should not close when selecting the first option.

pranav300:
### Summary
I have added condition, such that whenever `relatedTarget === this.input` onBlur event should not be triggered for mobile.

Resolves #2845

### DIFF
--- a/packages/terra-form-select/CHANGELOG.md
+++ b/packages/terra-form-select/CHANGELOG.md
@@ -4,7 +4,7 @@ ChangeLog
 Unreleased
 ----------
 ### Changed
-* Fixed `onBlur` event to not trigger when selecting an option
+* Fixed `onBlur` event to not be triggered when selecting an option
 
 5.44.0 - (March 10, 2020)
 ------------------
--- a/packages/terra-form-select/CHANGELOG.md
+++ b/packages/terra-form-select/CHANGELOG.md
@@ -4,7 +4,7 @@ ChangeLog
 Unreleased
 ----------
 ### Changed
-* Fixed `onBlur` event to not be triggered when selecting an option
+* Fixed `onBlur` event for mobile to not be triggered when selecting an option
 
 5.44.0 - (March 10, 2020)
 ------------------
--- a/packages/terra-form-select/CHANGELOG.md
+++ b/packages/terra-form-select/CHANGELOG.md
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Fixed `onBlur` event to not trigger when selecting an option
 
 5.44.0 - (March 10, 2020)
 ------------------
--- a/packages/terra-form-select/src/combobox/Frame.jsx
+++ b/packages/terra-form-select/src/combobox/Frame.jsx
@@ -329,7 +329,7 @@ class Frame extends React.Component {
     // and if the relatedTarget is falsey. The relatedTarget will be null when
     // dismissing the onscreen keyboard, else set to another element when
     // tapping elsewhere on the page
-    if (focusedByTouch && !relatedTarget) {
+    if (focusedByTouch && (relatedTarget === this.input)) {
       return;
     }
 
@@ -467,6 +467,8 @@ class Frame extends React.Component {
       isOpen: false,
       isAbove: false,
     });
+
+    this.input.focus();
 
     if (this.props.onSelect) {
       this.props.onSelect(value, option);
--- a/packages/terra-form-select/src/combobox/Frame.jsx
+++ b/packages/terra-form-select/src/combobox/Frame.jsx
@@ -326,9 +326,7 @@ class Frame extends React.Component {
 
     // Don't blur if we dismissed the onscreen keyboard
     // Determined by if we have have interacted with the frame via onTouchStart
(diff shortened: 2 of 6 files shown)


packages/terra-form-select/src/search/Frame.jsx
```diff
@@ -442,7 +442,7 @@ class Frame extends React.Component {
       searchValue: '',
       hasSearchChanged: false,
       isOpen: false,
-      isAbove: false,
+      // isAbove: false,
```
ShettyAkarsh: you could remove this line. if not required.

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -468,6 +468,8 @@ class Frame extends React.Component {
       isAbove: false,
     });
 
+    this.input.focus();
```
ShettyAkarsh: why do we need this to be set explicitly?

packages/terra-form-select/src/search/Frame.jsx
pranav300: updated here https://github.com/cerner/terra-core/pull/2904/commits/2d8de879a0187eb87b5dea6d6877d871738c1ef4

packages/terra-form-select/src/combobox/Frame.jsx
pranav300: Removed here https://github.com/cerner/terra-core/pull/2904/commits/2d8de879a0187eb87b5dea6d6877d871738c1ef4

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -329,7 +329,7 @@ class Frame extends React.Component {
     // and if the relatedTarget is falsey. The relatedTarget will be null when
     // dismissing the onscreen keyboard, else set to another element when
     // tapping elsewhere on the page
-    if (focusedByTouch && !relatedTarget) {
+    if (focusedByTouch && (relatedTarget === this.input)) {
```
nramamurth: This inline comment ⬇️ may need to be updated now.

> Determined by if we have have interacted with the frame via onTouchStart and if the relatedTarget is falsey.

packages/terra-form-select/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Fixed `onBlur` event to not be triggered when selecting an option
```
nramamurth: Does this need to be a bit more specific that this was done _just_ for mobile?

StephenEsser:
With these changes desktop is still experiencing invalid onBlur/onFocus. Are you wanting to split the changes into two separate PRs to fix desktop/mobile issues?

packages/terra-form-select/src/combobox/Frame.jsx
pranav300: Updated here https://github.com/cerner/terra-core/pull/2904/commits/09eb6be42fd6ff175b3d324f8d5ccbb9ca1583ec

packages/terra-form-select/CHANGELOG.md
pranav300: Updated here https://github.com/cerner/terra-core/pull/2904/commits/09eb6be42fd6ff175b3d324f8d5ccbb9ca1583ec

StephenEsser:
One of the WDIO screenshots for the select is failing. Do we know what caused the change and if the screenshot should be updated?

cc: @pranav300

PR #2908, opened 2020-03-18
[terra-paginator] Fixed cosmetic effects of text wrapping issue.

### LINKED ISSUE #2855: [terra-paginator] Cosmetic effects of text wrapping
## Description

At certain dimensions the Paginator text wrapping can be aggressive. We should get UX feedback on the best way to handle this scenario.

image

## Steps to Reproduce

1. Navigate to the terra-ui Paginator [example page](https://engineering.cerner.com/terra-ui/components/terra-paginator/paginator/paginator-controlled).
2. Slowly resize the window while observing the Paginator example. 
3. Once the text wraps  observe the visual state of the component.
4. Navigating to double digit page numbers will exaggerate the effect.

## Expected Behavior

We should reach out to UX and discuss what the expected behavior should be.

saket2403:
### Summary
Fixes #2855


packages/terra-paginator/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Fixed
+* Fixed cometic effects of text-wrapping
```
lokesh-0813: nit : typo here

saket2403: Fixed d30292e

packages/terra-paginator/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Fixed
+* Fixed cosmetic effects of text-wrapping
```
emilyrohrbough: Can we add in a test case to show these fixes?

saket2403: The effect was not visible for the widths which we use for wdio tests.So, I had not added the test

packages/terra-paginator/src/Paginator.module.scss
```diff
@@ -33,14 +33,12 @@
...
       min-width: var(--terra-paginator-nav-link-min-width, 2rem); // Helps ensure limited resizing as you page through large numbers.
       outline: none;
-      overflow-wrap: break-word; /* Modern browsers */
       padding: var(--terra-paginator-nav-link-padding, 0.35714rem 0.5rem);
       text-align: center;
       text-decoration: var(--terra-paginator-nav-link-text-decoration, none);
       touch-action: manipulation; // Enable fast tap interaction in webkit browsers; see https://webkit.org/blog/5610/more-responsive-tapping-on-ios/
       user-select: none; // Keeps user from highlighting and text selection
       white-space: normal;
-      word-wrap: break-word; /* For IE 10 and IE 11 */
```
StephenEsser: I'd like to confirm with UX they are okay with the next and previous buttons wrapping on Chrome:

image

On IE 10 the "Last" button can escape the bounds of the container:


image

jeremyfuksa: It's not ideal. I didn't notice that happening when I looked at it initially. Removing the seal of approval until after this gets addressed.

packages/terra-paginator/CHANGELOG.md
ryanthemanuel: Can you provide a container with a set width that *does* cause the issue?

packages/terra-paginator/src/Paginator.module.scss
saket2403: I think the wrapping in chrome is the correct function, but it is not happening in IE10 due to which the `Last` button goes out... Any css inputs  to fix it in IE10? cc/ @jeremyfuksa .

jeremyfuksa: `white-space: nowrap` gets the buttons to do what I think should be happening in Chrome. Removing `min-width` of `.nav-link` in IE 10 get the right spacing.
Screen Shot 2020-04-07 03 27 45 PM

jeremyfuksa: Changing `background-size: auto` for `.icon` in IE 10 also keeps the chevrons from being cut off.

saket2403: Updated 9ab38f7

packages/terra-paginator/CHANGELOG.md
saket2403: Added wdio test.

PaulSolDev:
**(Chrome, Safari, Edge, and Firefox)**
On resize there's a small width that seems to push the First/Last text outside of the footer container.

image

**(IE)**
On resize there is a small width where the 's' on previous is being cut off.  

image

saket2403:
> **(Chrome, Safari, Edge, and Firefox)**
> On resize there's a small width that seems to push the First/Last text outside of the footer container.
> **(IE)**
> On resize there is a small width where the 's' on previous is being cut off.

The above two issues can be fixed by changing the padding and margin values, I played around with the values and they seem to work, is there any specific values recommended for all 3 themes?  cc @jeremyfuksa 
> **(IE)**
> 
> At extremely small break points the text for "First" and "Last" gets cut-off.

For extremely small break-point should the texts be wrapped in IE or any specific values of padding can be applied? @jeremyfuksa .

StephenEsser:
Are there any updates on this issue? If the enhancement strategy is going to change we may want to close this issue in the interim to cleanup the PRs ready for review. 

https://github.com/cerner/terra-core/issues/2855#issuecomment-612938445

ryanthemanuel:
Let's close this for now pending a rethink of the strategy we want for this.

PR #2910, opened 2020-03-20
[terra-core] Added clinical-lowlight-theme wdio screenshots

avinashg1994:
### Summary
This PR includes the following changes after the release of https://github.com/cerner/terra-dev-site/pull/240:
1. Added `clinical-lowlight-theme` wdio screenshots
2. Un-commented the travis config code to run `clinical-lowlight-theme` tests.


packages/terra-table/src/terra-dev-site/test/table/Row.test.jsx
```diff
@@ -1,8 +1,12 @@
 import React from 'react';
 import Table from 'terra-table';
+import classNames from 'classnames/bind';
+import styles from './RowTest.module.scss';
+
+const cx = classNames.bind(styles);
 
 const RowExample = () => (
-  <React.Fragment>
+  <div className={cx('content-wrapper')}>
```
emilyrohrbough: Why is this change needed?

avinashg1994: Since the table border is being set to '0' in clinical-lowlight-theme as seen [here](https://github.com/cerner/terra-core/blob/master/packages/terra-table/themes/clinical-lowlight-theme/clinical-lowlight-theme.scss#L3). The **"endX is out of range"** error was being thrown when this wdio test was run.
This seems to be a special case, so i added the right border to prevent this error.

avinashg1994:
> I know color contrast is disabled for low-light, but this isn't very easy to read. Are we okay with this?
> Screen Shot 2020-03-24 at 1 10 52 PM

Not sure if this is the reason, but i think wdio screenshots are of lower resolution/quality due to which the text looks hazy. On the [dev-site](https://terra-core-deployed-pr-2910.herokuapp.com/tests/terra-table/table/row) it seems to be much better and readable

packages/terra-table/src/terra-dev-site/test/table/Row.test.jsx
emilyrohrbough: Why is this an issue with only this test example and not all of them if it's a reflection of the table border being zero? 

This does not seem like the appropriate fix IMO.

avinashg1994: I am not sure why this issues is only with this test. I could be wrong, but it seems to be an issue with webdriver.
The reason for this workaround is because this issue is only there for this particular test and that to only for clinical-lowlight-theme. Also all the theme values have been directly ported from the internal repo, so i did not intend on modifying them.
But if required, another solution is to update the border value.
@jeremyfuksa, your thoughts on this?

jeremyfuksa: Why is border set to 0 in Low Light when all the other themes have a 1px border?

avinashg1994: Not sure about the reason behind setting the border to 0px.
So should we change the value of border to 1px?

jeremyfuksa: Yes. Especially if it makes this one-off code adjustment unnecessary.

avinashg1994: Updated the border value for clinical-lowlight-theme to 1px : https://github.com/cerner/terra-core/pull/2910/commits/ccf6d5fd6fd5de326ea0ed2d06da9a47e75f88ad

PR #2920, opened 2020-04-07
[terra-form-select] Fixed detach issue for dropdowns with isTouchAccessible.

### LINKED ISSUE #2906: [terra-form-select] Detached dropdown using isTouchAccessible
# Bug Report

## Description

When using the isTouchAccessible prop the dropdown appearing above the field truncates and detaches from the component when filtering the list.

image

image

## Steps to Reproduce

1. Add a filterable Select (Tag, Multiple, Combobox, Search) to a view near the bottom of the window.
2. Add isTouchAccessible to the Select
3. Load the page and open the Select so the dropdown appears above the component. (It opens above when the component is near the bottom of the window)
4. Type into the input field to filter the list
5. Observe the filtered option detach from the component.

## Expected Behavior

The list should not detach from the component. 

## Possible Solution

Reposition the dropdown when the list height changes.

saket2403:
### Summary
The dropdown was not getting re-positioned when it got resized after a search operation was performed.

Closes #2906 

### Deployment Link

https://terra-core-deployed-pr-2920.herokuapp.com/components/terra-form-select/form-select/select

### Testing
Added wdio test for the concerned scenario.


supreethmr:
form-select-detach

Can we prevent this dropdown blinking that is happening once we clear the search value...?

packages/terra-form-select/src/shared/_FrameUtil.js
```diff
@@ -96,13 +100,13 @@ class FrameUtil {
...
    */
-  static shouldPositionDropdown(previousState, currentState, dropdown) {
+  static shouldPositionDropdown(previousState, currentState, dropdown, isTouchAccessible) {
     if (!currentState.isOpen) {
       return false;
     }
 
     const { bottom } = dropdown.getBoundingClientRect();
-    return previousState.isOpen === false || bottom > window.innerHeight;
+    return previousState.isOpen === false || bottom > window.innerHeight || (!previousState.resizeOnSearch && currentState.resizeOnSearch && isTouchAccessible);
```
ShettyAkarsh, line 109: Why do we need both the conditions here `!previousState.resizeOnSearch && currentState.resizeOnSearch` why not only this `(currentState.resizeOnSearch && isTouchAccessible)` ?

packages/terra-form-select/src/shared/_FrameUtil.js
```diff
@@ -25,19 +25,22 @@ class FrameUtil {
...
+  static dropdownPosition(props, target, dropdown, maxHeight, currentState, useSemanticDropdown = false) {
     const style = ({ ...props }).style || {};
     const { height } = dropdown.getBoundingClientRect();
     const { bottom, width, top } = target.getBoundingClientRect();
 
     const spaceBelow = window.innerHeight - bottom;
     const maximumHeight = parseInt(style.maxHeight || maxHeight, 10) || Infinity;
-    const canFitBelow = maximumHeight < spaceBelow || height < spaceBelow || spaceBelow > top;
+    let canFitBelow = (maximumHeight < spaceBelow || height < spaceBelow || spaceBelow > top);
+    if (useSemanticDropdown && currentState !== undefined && currentState.resizeOnSearch && currentState.isAbove && canFitBelow) {
```
supreethmr, line 40:
```jsx
if (useSemanticDropdown && currentState !== undefined && currentState.resizeOnSearch && currentState.isAbove && canFitBelow)
```
I think this if condition is added to prevent dropdownlist from displaying below form-select on search ( when dropdownlist has only one item ) . if this is the reason for having this condition then instead of using currentState and other variables, Can we modify the existing condition to like below to ensure dropdown will Fit below only when it has space to accommodate two or more options. 
```jsx
const canFitBelow = (maximumHeight < spaceBelow || ( height * 2 ) < spaceBelow || spaceBelow > top);
```

packages/terra-form-select/src/shared/_FrameUtil.js
```diff
@@ -25,19 +25,22 @@ class FrameUtil {
...
+      canFitBelow = false;
```
ShettyAkarsh, line 41: why are we not allowing the dropdown display below here ? will this logic not impact both the DOM and Portal rendering ? should we check the isTouchAccessible prop here ?

packages/terra-form-select/src/terra-dev-site/test/form-select/isTouchAccessibleBottom.test.jsx
```diff
@@ -0,0 +1,21 @@
...
+const IsTouchAccessibleBottom = () => (
+  <div className={cx('bottom-left')}>
+    <Select isTouchAccessible placeholder="Select a color" variant="tag" id="tag">
+      <Select.Option value="blue" display="Blue" />
+      <Select.Option value="green" display="Green" />
+      <Select.Option value="purple" display="Purple" />
+      <Select.Option value="red" display="Red" />
+      <Select.Option value="violet" display="Violet" />
+    </Select>
+    <button type="button" id="focusable">Focusable</button>
```
ShettyAkarsh, line 17: we could remove this if not used .

packages/terra-form-select/src/shared/_FrameUtil.js
saket2403: useSemanticDropdown is the isTouchAccessible prop

packages/terra-form-select/src/shared/_FrameUtil.js
saket2403: This condition is added so that if the dropdown has initially opened above it again opens above the select after search, that is why isAbove is also added

packages/terra-form-select/src/terra-dev-site/test/form-select/isTouchAccessibleBottom.test.jsx
saket2403: Removed 07f81a8

packages/terra-form-select/src/shared/_FrameUtil.js
saket2403: Removed 07f81a8

packages/terra-form-select/src/shared/_FrameUtil.js
supreethmr: without this condition dropdown displays below the form-select on search (when it has only one option in list) which seems to be expected behaviour of form-select when there is only one option to disaply but dropdown will not resize once it is displayed below the form-select as shown in below image. even after clearing the search value size of the dropdown remains unchanged. ( I am not sure why it is not resizing to it's default height on clearing the search value )


form-select-detach

packages/terra-form-select/src/shared/_FrameUtil.js
supreethmr: disregard my suggestion of having `height * 2` in condition as it doesn't serve the purpose of this condition. but with IsAbove being added as condition regardless of space available it will always display dropdown above the select like shown below.

form-select-detach

Is it expected  to display above form-select on search even though it has plenty of space below the form select...? 🤔

packages/terra-form-select/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Fixed
+* Fixed detach issue of dropdown of `search` variants with `isTouchAccessible` prop.
```
StephenEsser, line 7: Because `search` is one of the available variants I'd recommend we rephrase this changelog entry. 

Something similar to this perhaps:

```
* Fixed issue when using `isTouchAccessible` that could cause the dropdown to detach from the input
```

packages/terra-form-select/src/shared/_FrameUtil.js
```diff
@@ -51,6 +54,7 @@ class FrameUtil {
       isAbove,
       isPositioned: true,
       top: semanticTopWhenAbove,
+      resizeOnSearch: false,
```
StephenEsser, line 57: I think we can simplify the logic by applying a `bottom` instead of a `top` when the dropdown is appearing above and isTouchAccessible is enabled. 

```jsx
// Extract the height of the target.
const { height: targetHeight } = target.getBoundingClientRect();

// If isTouchAccessible is enabled and the dropdown should be above, apply a bottom with the same value as the target height.
const semanticBottomWhenAbove = (useSemanticDropdown && isAbove ? targetHeight : undefined);
```

The bottom value does not need to be recalculated when the height of the dropdown changes because it's bound to the bottom regardless of height it stays attached to the target. This also removes additional resize logic.

saket2403: I had actually tried it initially but without top being set the dropdown did not open above the select field.

StephenEsser: It works, there are a few other tweaks that need to be done. Likely the issue you were encountering was from this still being set:

https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/shared/_Dropdown.module.scss#L27

saket2403: Done a2c35f5

packages/terra-form-select/CHANGELOG.md
saket2403: Changed cb75c3a

packages/terra-form-select/src/shared/_FrameUtil.js
saket2403: Resolved a2c35f5

packages/terra-form-select/src/shared/_FrameUtil.js
saket2403: Fixed a2c35f5

saket2403:
> Can we prevent this dropdown blinking that is happening once we clear the search value...?

Fixed a2c35f5

packages/terra-form-select/src/shared/_FrameUtil.js
```diff
@@ -25,31 +28,31 @@ class FrameUtil {
...
     const availableSpace = canFitBelow ? spaceBelow : top;
-
     const availableMaxHeight = Math.min(maximumHeight, availableSpace - 10);
     const isAbove = !canFitBelow;
-    const semanticTopWhenAbove = (useSemanticDropdown && isAbove
-      ? -1 * Math.min(availableMaxHeight, height)
-      : undefined
-    );
+    const semanticBottomWhenAbove = (useSemanticDropdown && isAbove ? targetHeight : undefined);
+    const semanticTopWhenAbove = (useSemanticDropdown && !isAbove ? '100%' : undefined);
```
StephenEsser, line 48: This evaluation is backwards and a little confusing. The semantic top is set to undefined when above and 100% when not above. 

Because the top is not dynamic, it is always 100% when appearing below, I think this logic can stay in the dropdown css. It might need a new class.

```css
.is-touch-accessible.is-below {
  top: 100%;
}
```

saket2403: Initially I had tried to have a class for top, it worked fine on initial render but on re-opening the dropdown it went below the target, which should have opened above.

StephenEsser: Do you know what was causing it to open below the dropdown on subsequent opening?

saket2403: After the first. run the max height of the dropdown changes which is used from the previous run resulting in the dropdown to open below.
image

StephenEsser: I think this could be applied in CSS with more investigation, however, the result appears to be the same. Assuming this works as intended in all supported browsers I think this is fine.

packages/terra-form-select/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ ChangeLog
 Unreleased
 ----------
 ### Fixed
+* Fixed issue when using `isTouchAccessible` that could cause the dropdown to detach from the input
+
```
avinashg1994, line 8: nit: This extra line should be removed

packages/terra-form-select/src/shared/_FrameUtil.js
```diff
@@ -25,31 +28,31 @@ class FrameUtil {
    * @param {ReactNode} target - The select wrapper.
    * @param {ReactNode} dropdown - The dropdown.
    * @param {number} maxHeight - The maxHeight of the dropdown dropdown.
+   * @param {Object} currentState - The current component state.
```
avinashg1994, line 31: This should be removed, since currentState is no longer used

packages/terra-form-select/CHANGELOG.md
saket2403: Removed fb1ae9e

packages/terra-form-select/src/shared/_FrameUtil.js
saket2403: Removed fb1ae9e

PaulSolDev:
(mobile) Firefox and Chrome:

- unable to gain focus to the select for text entry without first selecting an option from the dropdown.   (This is not the case in the released version of the site)


- Dropdown gets hidden from view:

image

saket2403:
> (mobile) Firefox and Chrome:
> 
> * unable to gain focus to the select for text entry without first selecting an option from the dropdown.   (This is not the case in the released version of the site)
> * Dropdown gets hidden from view:

I am seeing similar behaviour in released version on engineering site also.
Engg Site



LocalHost

saket2403:
I am seeing similar behaviour on engg and localhost 
Engg Site

EnggSite-min

LocalHost
Local

PaulSolDev:
Thank you for following up.  On further review I'm seeing this behavior on the engg site with the combo and multi variants.  We can log a separate issue to address this.

Functional verification complete. No new issues identified.

PR #2922, opened 2020-04-10
[terra-form-select] Add maxHeight to all field variants

### LINKED ISSUE #2909: [terra-form-select] Add maxHeight prop to field components
# Feature Request

Add maxHeight Property to [SingleSelectField](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/SingleSelectField.jsx), [MultiSelectField](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/MultiSelectField.jsx), [SearchSelectField](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/SearchSelectField.jsx), [TagSelectField](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/TagSelectField.jsx)
## Description
[Max height ](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/MultiSelect.jsx#L50) is available as property in all forms of select components but not available in components that are in combination with field like [SingleSelectField](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/SingleSelectField.jsx), [MultiSelectField](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/MultiSelectField.jsx), [SearchSelectField](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/SearchSelectField.jsx), [TagSelectField](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/TagSelectField.jsx)

## Possible Solution
Add maxHeight to props of above mentioned components similar to how it is supported in [multiSelect component](https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/MultiSelect.jsx#L50).

saigorantla:
### Summary
Adding Max height as first class prop for all select fields.

Closes #2909 

### Testing
ezgif com-optimize


packages/terra-form-select/src/ComboboxField.jsx
```diff
@@ -191,6 +197,7 @@ const ComboboxField = ({
         isInvalid={isInvalid}
         isTouchAccessible={isTouchAccessible}
         defaultValue={defaultValue}
+        maxHeight={maxHeight}
```
StephenEsser, line 200: These additions will need to be passive with existing maxHeight implementations.

```jsx
<ComboboxField selectAttrs={{ maxHeight: 100 }} />
```

Please ensure each Field variant still supports the previous selectAttrs implementation. This may need to use a fallback.

```jsx
maxHeight={maxHeight || selectAttrs.maxHeight}
```

packages/terra-form-select/src/terra-dev-site/test/form-select/MaxHeight.test.jsx
```diff
@@ -2,11 +2,92 @@ import React from 'react';
...
+import ComboboxField from '../../../ComboboxField';
+import SearchSelectField from '../../../SearchSelectField';
+import MultiSelectField from '../../../MultiSelectField';
+import TagSelectField from '../../../TagSelectField';
 
 const cx = classNames.bind(styles);
 
 const MaxHeight = () => (
   <div className={cx('content-wrapper')}>
+    <MultiSelectField label="Multi Select Field" placeholder="Select a number" selectId="Multi-Select" defaultValue="1" maxHeight={300}>
```
StephenEsser, line 15: None of these added variants are being tested in the WDIO test. These can be removed.

packages/terra-form-select/tests/jest/SearchableFields.test.jsx
```diff
@@ -8,7 +8,7 @@ import TagSelectField from '../../src/TagSelectField';
 
 it('should render the selectId to the input field in ComboboxField component', () => {
   const select = (
-    <ComboboxField label="Label" defaultValue="blue" selectId="select-id">
+    <ComboboxField label="Label" defaultValue="blue" selectId="select-id" maxHeight={100}>
```
StephenEsser, line 11: Please create new snapshot tests for these that indicate the desired test. Adding maxHeight to existing tests makes future updates more confusing because the test description does not indicate this should test max height.

> should render the selectId to the input field in ComboboxField component

packages/terra-form-select/src/terra-dev-site/test/form-select/MaxHeight.test.jsx
saigorantla: Are you suggesting to move each component into separate tests? I am confused on your comment as I am only trying to test Height variant.

packages/terra-form-select/src/ComboboxField.jsx
saigorantla: will do

packages/terra-form-select/src/terra-dev-site/test/form-select/MaxHeight.test.jsx
StephenEsser: These variants were added to an existing test, but the test run wasn't updated. So it is still only testing the previous height. There was a change to the screenshot, but no new functionality is being tested.

image


image

The max height on the added variants isn't being tested in this test

saigorantla: Got it thank you you will add more tests

packages/terra-form-select/tests/jest/SelectFieldsMaxHeight.test.jsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
```
StephenEsser, line 1: It's a little strange to organize these changes into a combined spec file for only max height. Traditionally these would be  split out for each field variant and non-specific to any props/functionality.

/jest/ComboboxField.test.jsx
/jest/MultiSelectField.test.jsx
/jest/TagSelectField.test.jsx
/jest/SearchSelectField.test.jsx

It's not harmful to keep these tests in this file. It's a little less organized but the functionality is still being tested.

saigorantla: I was debating about that, I agree this is better approach. Updated test cases

packages/terra-form-select/src/ComboboxField.jsx
```diff
@@ -66,6 +66,10 @@ const propTypes = {
    */
   // eslint-disable-next-line react/forbid-prop-types
   labelAttrs: PropTypes.object,
+  /**
+   * The max height of the dropdown.
+   */
+  maxHeight: PropTypes.number,
```
emilyrohrbough, line 72: Should this be a string prop and accept the same options as the maxWidth prop?

StephenEsser: The maxWidth is a propType for the [field component](https://engineering.cerner.com/terra-ui/components/terra-form-field/form-field/form-field), the select maxHeight prop only accepts a number. I'd say for these changes we should keep it consistent with the existing <Select /> maxHeight API.

https://engineering.cerner.com/terra-ui/components/terra-form-select/form-select/select

emilyrohrbough: Is max height here a style value or maximum number of options it will display?

StephenEsser: It's the maximum height of the dropdown interpreted as pixels and is applied as an inline style on the dropdown. It's used in height calculations so the prop required being set as a number:

https://github.com/cerner/terra-core/blob/master/packages/terra-form-select/src/shared/_FrameUtil.js#L31-L55

This prevents a user setting a custom maxHeight that is larger than the actual available height on the screen.

emilyrohrbough: ohhh gottchya

emilyrohrbough: Thanks for the extra info!

StephenEsser:
Thank you, @saigorantla!

PR #2926, opened 2020-04-15
[terra-table] Adding style overflow: hidden to cell

### LINKED ISSUE #2854: [terra-table] Add overflow: hidden to cell
# Feature Request
Add css attribute overflow: hidden to terra-table's Cell.

## Description
In order to allow for a great number of content types, images, truncation etc, if a overflow:hidden; is added to the outer cell teams will need to set fewer custom styles when altering cell behavior from text wrapping.

## @ Mentions
@neilpfeiffer

lokesh-0813:
### Summary

To reduce the style count consumers provide when altering cell related to word-wrapping.

Closes #2854 

### Deployment Link

https://terra-core-deployed-pr-2926.herokuapp.com/components/terra-table/table/table

### Testing

Updated Jest and Added Wdio tests for long cell content with truncation.
### Additional Details

@cerner/terra


packages/terra-table/src/subcomponents/_Cell.jsx
```diff
@@ -55,7 +55,7 @@ const Cell = ({
   ...customProps
 }) => {
   const cellClassNames = cx('cell');
-  const contentClass = !removeInner ? cx('container') : undefined;
+  const contentClass = !removeInner ? cx('container') : cx('content-width');
```
dkasper-was-taken, line 58: We should probably move the 'content-width' class to:
https://github.com/cerner/terra-core/blob/master/packages/terra-table/src/subcomponents/_Cell.jsx#L75

That way this check is still valid:
```
if (ariaAttr || contentClass) {
```

It would probably look something like:
```
<div {...ariaAttr} className={contentClass || cx('content-width')}>
```

dkasper-was-taken: That was my mistake prior, I didn't consider that if check with my previous advice.

emilyrohrbough, line 58:
```suggestion
  const contentClass = cx([{
     'container': !removeInner,
     'content-width': removeInner,
  });
```

ryanthemanuel: If we do Dave's comment, this would not be needed.

PaulSolDev:
@dkasper-was-taken 
When checking the example in Edge on a Windows based device, the right column of the table appears to be cut off at small breakpoints.  The same does not occur while using Edge on Mac.

image

dkasper-was-taken:
@PaulSolDev Windows surface/mobile based edge has a minimum size to the browser, you'll notice that the other application buttons are beginning to go off the page as well. Looking at edge for windows desktop is displaying correcrtly. So I think you are safe to ignore that bit.

PaulSolDev:
@dkasper-was-taken Thank you for the clarification.  With this in mind, the functional review is complete and no issues were identified.

neilpfeiffer:
Reviewed: #2926 
Resolves: #2854

No significant UI changes, allows users to be able to add cell truncation.

| UX +1 |
|---|

PR #2933, opened 2020-04-20
[terra-button-group] Fix button losing focus when switching between windows

### LINKED ISSUE #2919: Button with Focus loses Focus style on revisiting from different window.
# Bug Report

## Description

When I'm using keyboard navigation to tab onto buttons in a ButtonGroup, the highlighted button loses the highlight on the right hand side of the button when in switch to and the back from another application, or another browser tab.
(This is happening in Safari and Chrome on Mac)

Screenshot 2020-04-06 at 14 22 50

I originally posted about this [on teams](https://teams.microsoft.com/l/message/19:0adf7161dd244f87aa6415f7c5577836@thread.skype/1586179701283?tenantId=fbc493a8-0d24-4454-a815-f4ca58e8c09d&groupId=45f47bfa-38a9-444d-8933-9c20266ae38c&parentMessageId=1586179701283&teamName=Terra&channelName=Component%20Support&createdTime=1586179701283)

## Steps to Reproduce

1. Open [this page](https://engineering.cerner.com/terra-core/components/terra-button-group/button-group/button-group)
2. Tab onto the Button Group example "ButtonGroup Is Block - width stretches to fill container"
3. With the first button highlighted, Cmd + tab to switch to another application and again to switch back
4. The highlight will be missing from one of the edges

## Expected Behavior
The button should retain the highlight around all 4 sides of the button

## Environment

* Component Name and Version: ButtonBroup.Button
* Browser Name and Version: Safari and Chrome on Mac
* Webpack Version:
* Operating System and version (desktop or mobile): Mac desktop

## @ Mentions

@supreethmr

pranav300:
### Summary

PR resolves for Button to not lose focus when switching between windows 

Closes #2919 

### Deployment Link

https://terra-core-deployed-pr-2933.herokuapp.com/

### Additional Details

@cerner/terra


packages/terra-button-group/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ ChangeLog
 
 Unreleased
 ----------
+### Fixed
+* Fixed button loosing focus when switching between windows
```
StephenEsser, line 7: Small typo on `loosing` -> losing

packages/terra-button-group/src/ButtonGroupButton.jsx
```diff
@@ -63,6 +67,17 @@ class ButtonGroupButton extends React.Component {
     }
   }
 
+  handleClick(event) {
+    if (document.activeElement === event.currentTarget) {
+      event.currentTarget.focus();
```
StephenEsser, line 72: Does this cause focus to be called twice each time the button is clicked?

packages/terra-button-group/src/ButtonGroupButton.jsx
```diff
@@ -63,6 +67,17 @@ class ButtonGroupButton extends React.Component {
     }
   }
 
+  handleClick(event) {
```
supreethmr, line 70: Do we need handleClick event..? if it is declared just to set `shouldShowFocus` to true then I see we are handling `shouldShowFocus` in keydown event as well which will set `shouldShowFocus` to true tab key press I think this should enough to hold the focus styles on selected button. 

having handleClick() here is making same value 'true' to set twice for `shouldShowFocus`.

packages/terra-button-group/src/ButtonGroupButton.jsx
```diff
@@ -80,13 +95,24 @@ class ButtonGroupButton extends React.Component {
...
     }
   }
 
+  handleFocus(event) {
+    if (this.shouldShowFocus) {
+      this.setState({ focused: true });
+    }
+
+    if (this.props.onFocus) {
+      this.props.onFocus(event);
```
supreethmr, line 112: the prop onFocus should be passed from ButtonGroup to here to avoid multiple onfocus events from getting called when Focus handler has added to both parent (ButtonGroup) and (ButtonGroupButton) as well.

packages/terra-button-group/src/ButtonGroupButton.jsx
ShettyAkarsh: @pranav300  correct me if i am wrong. @supreethmr I think in the case of first switching to another window or tabbing (which will make this shouldShowFocus to true). so once we come back to the button and click it. the variable shouldShowFocus will still be true. in that case we need handle click.

supreethmr: We can set `shouldShowFocus` to false in handleBlur to handle this scenario. 
For example :
```jsx
 handleOnBlur(event) {
    if (document.activeElement === event.currentTarget && this.state.focused) {
      this.shouldShowFocus = true;
    } else {
      this.shouldShowFocus = false;
      this.setState({ focused: false });
    }

    if (this.props.onBlur) {
      this.props.onBlur(event);
    }
  }
``` 
Since blur will be called on both scenarios like switching to new window or when we click outside / other button. handling `this.shouldShowFocus` in blur event would fix this issue. 

Above code will hold back the focus styles on button in blur event if it is previously focused ( on leaving the window ) else removes focus styles from button. So we might don't need handle click event to set `this.shouldShowFocus` to false it can be handled in handleBlur like shown above.

packages/terra-button-group/src/ButtonGroupButton.jsx
pranav300: Updated here https://github.com/cerner/terra-core/pull/2933/commits/530e5fb55778e29b3c26ef4896915987ed6a7351

packages/terra-button-group/src/ButtonGroupButton.jsx
pranav300: Removed onClick event here https://github.com/cerner/terra-core/pull/2933/commits/372398dc76e2c279355e476dee13fa8e5e587fc2

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -27,6 +27,10 @@ const propTypes = {
    * A list of keys of the buttons that should be selected.
    */
   selectedKeys: PropTypes.arrayOf(PropTypes.string),
+  /**
+   * Callback function triggered when button gains focus.
+   */
+  onFocus: PropTypes.func,
```
emilyrohrbough, line 33: Should this prop be added? I'm not seeing where it is being used

pranav300: It was added to avoid multiple onFocus events, for when the focus handler was added to both the ButtonGroup and ButtonGroupButton.

emilyrohrbough: I'm not sure I understand 100%. Did you add this to remove the onFocus from being used if a consumer passes it in? or will this be triggered instead?

pranav300: I added this thinking that if ButtonGroupButton and ButtonGroup both have onFocus prop, then ButtonGroup should not trigger an onFocus event because both will be calling it on the same target. But I didn't realize that it would also be inhibiting consumers from doing any onFocus events on the ButtonGroup.

Removed it here https://github.com/cerner/terra-core/pull/2933/commits/b5c32f607c81526c5b75e4fa4052a194427aa5c2

packages/terra-button-group/src/ButtonGroupButton.jsx
```diff
@@ -80,17 +88,29 @@ class ButtonGroupButton extends React.Component {
...
+    }
+  }
+
   render() {
     const {
       icon,
       isDisabled,
+      onFocus,
```
emilyrohrbough, line 113: We should add this to the propTypes & add a description

packages/terra-button-group/tests/wdio/button-group-spec.js
```diff
@@ -106,6 +106,19 @@ Terra.describeViewports('Button Group', ['huge'], () => {
 
     Terra.it.validatesElement('button3-unselected');
   });
+  describe('Button', () => {
```
emilyrohrbough, line 109:
```suggestion
  describe('Button Focus', () => {
```

packages/terra-button-group/tests/wdio/button-group-spec.js
```diff
@@ -106,6 +106,19 @@ Terra.describeViewports('Button Group', ['huge'], () => {
...
+    before(() => {
+      browser.url('/#/raw/tests/terra-button-group/button-group/button-group-multi-select');
+    });
+    it('should not lose focus when switching between windows', () => {
```
emilyrohrbough, line 113:
```suggestion
    Terra.it.validatesElement('before switching window');
    it('should not lose focus when switching between windows', () => {
```

packages/terra-button-group/tests/wdio/button-group-spec.js
```diff
@@ -106,6 +106,19 @@ Terra.describeViewports('Button Group', ['huge'], () => {
...
+      browser.keys('Tab');
+      browser.keys('Tab');
+      browser.keys('Space');
+      browser.newWindow('https://engineering.cerner.com/terra-core/raw/tests/terra-button-group/button-group/button-group-is-block', 'terra-core');
+      browser.switchTab();
+    });
+    Terra.it.validatesElement('should not lose focus when switching between windows');
```
emilyrohrbough, line 120:
```suggestion
    Terra.it.validatesElement('after switching windows');
```

packages/terra-button-group/src/ButtonGroupButton.jsx
pranav300: It is already present at [line 30](https://github.com/cerner/terra-core/blob/master/packages/terra-button-group/src/ButtonGroupButton.jsx#L30).

emilyrohrbough: oh 😄 easy update then 😋

packages/terra-button-group/tests/wdio/button-group-spec.js
pranav300: Update here https://github.com/cerner/terra-core/pull/2933/commits/11571ef1201eef61bab55418b2201cb70ecdff29.

packages/terra-button-group/tests/wdio/button-group-spec.js
pranav300: Update here https://github.com/cerner/terra-core/pull/2933/commits/11571ef1201eef61bab55418b2201cb70ecdff29.

packages/terra-button-group/tests/wdio/button-group-spec.js
pranav300: Update here https://github.com/cerner/terra-core/pull/2933/commits/11571ef1201eef61bab55418b2201cb70ecdff29.

PR #2934, opened 2020-04-20
[terra-form-select] Introduce native select control.

dkasper-was-taken:
### Summary
Added a native select control to the terra-form-select packages to transition teams away from the current default variant in order to address accessibility concerns.

### Deployment Link
https://terra-core-deployed-pr-2934.herokuapp.com/components/terra-form-select/form-select/native-select

### Testing
Jest & WDIO Included in change

### Additional Details
Finalization of theme color values is still needed as the control still relies on layered opacity. Orion-fusion-theme and clinical-low-light also fail contrast.

Trying to decide whether or not to flatten the placeholder, I think it depends on whether or not the primary use case is custom placeholder versus default.

@cerner/terra


packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,224 @@
...
+    { disabled },
+    { invalid: isInvalid },
+    { incomplete: required && isIncomplete },
+    { placeholder: isCurrentPlaceholder(selectAttrs.value, placeholder) },
+  );
+
+  return (
+    <div
+      {...customProps}
+      className={customProps.className ? `${nativeClassNames} ${customProps.className}` : nativeClassNames}
```
mjhenkes, line 190: @emilyrohrbough and I have been doing custom class names like this recently: https://github.com/cerner/terra-core/blob/master/packages/terra-card/src/Card.jsx#L45-L52
I think it's the cleanest pattern I've seen, but the bind thing is still kinda weird.

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,228 @@
...
+  const display = placeholder.display ? placeholder.display : intl.formatMessage({ id: 'Terra.form.select.defaultDisplay' });
+  const value = placeholder.value ? placeholder.value : defaultPlaceholderValue;
+  const attrs = placeholder.allowClear ? {} : { disabled: true, hidden: true };
+  return <option value={value} {...attrs}>{display}</option>;
+};
+
+const createOptions = options => options.map(current => {
+  if (current.childOptions) {
+    return (
+      <optgroup key={`${current.display} ${current.childOptions.length}`} label={current.display}>
```
tbiethman, line 120: Are the various displays expected to be unique? It'd be really confusing if they weren't unique, but I'm not sure if that's explicitly called out anywhere.

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,228 @@
...
+  isInvalid,
+  isIncomplete,
+  onChange,
+  options,
+  placeholder,
+  required,
+  value,
+  ...customProps
+}) => {
+  const [currentValue, setCurrentValue] = useState(defaultValue || getFirstValue(options, placeholder));
```
tbiethman, line 145: What would happen if, after the initial render, the NativeSelect was provided with a new options prop that doesn't include the current, er, `currentValue`? That might be an extreme edge case, but I'm not sure how the native select handles this scenario. 

If it will be a problem, you could validate the currentValue against the options during render and add an effect that resets the state to the new first value.

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: Updated to fix gap.

packages/terra-form-select/src/native-select/NativePropTypes.js
```diff
@@ -0,0 +1,45 @@
...
+   */
+  display: PropTypes.string.isRequired,
+  /**
+   * Whether the optgroup is disabled.
+   */
+  disabled: PropTypes.bool,
+  /**
+   * The array of select options.
+   */
+  childOptions: PropTypes.arrayOf(optionPropType).isRequired,
```
StephenEsser, line 34: How do you feel about naming this `options`? Did you chose this to avoid the clash with the root `options`?

dkasper-was-taken: I did change it so it was a little more explicit, but `options` makes sense too. Do you prefer `options`?

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: I believe the native select allows it, so I could adjust it do be a display + index. Thoughts?

ShettyAkarsh:
@dkasper-was-taken, so is there any plans for styling the display of html options and the dropdown as well ? something similar to existing form-select dropdown and options.

packages/terra-form-select/src/native-select/NativeSelect.jsx
ShettyAkarsh: Yeah only index should be good here. even the option also can have the same instead of value below.

dkasper-was-taken:
@ShettyAkarsh With a native select control we do not have control over the styling of the dropdowns visuals, that is largely dictated by the browser and OS. On certain browsers its possible to alter minor details, but it isn't consistent. So after discussions with the UX folks it was decided that this is a tradeoff we are willing to make to get the accessibility of the native select.

packages/terra-form-select/src/native-select/NativePropTypes.js
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: Updated with index of key occurrence method.

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,244 @@
...
+   */
+  onChange: PropTypes.func,
+  /**
+   * The array of select options and opt groups.
+   */
+  options: PropTypes.arrayOf(PropTypes.oneOfType([optionPropType, optGroupPropType])),
+  /**
+   * Placeholder data. The presence of this data object is used as the indicator for whether or not a placeholder is used.
+   */
+  placeholder: PropTypes.shape({
```
mjhenkes, line 77: @neilpfeiffer are we making a change to remove placeholder text in the future? Seems like i heard something to that effect but I may have misheard.

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,244 @@
...
+  ariaDescribedBy: PropTypes.string,
+  /**
+   * The aria label string value for the select.
+   */
+  ariaLabel: PropTypes.string.isRequired,
+  /**
+   * Additional attributes to spread onto the select.
+   */
+  // eslint-disable-next-line react/forbid-prop-types
+  attrs: PropTypes.object,
```
mjhenkes, line 40: Should we explain how `attrs` differs from custom props? spread on top level div vs spread on select component.

packages/terra-form-select/src/native-select/clinical-theme/NativeSelect.module.scss
```diff
@@ -0,0 +1,61 @@
+@import '~terra-mixins/lib/Mixins';
```
mjhenkes, line 1: clinical-theme? oh my ip!

packages/terra-form-select/src/terra-dev-site/doc/example/native-select-field/DefaultField.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import classNames from 'classnames/bind';
+import NativeSelectField from 'terra-form-select/lib/native-select/NativeSelectField';
+import styles from '../FormSelectDocCommon.module.scss';
```
mjhenkes, line 4: What styles are we adding?

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,244 @@
...
+import {
+  optionPropType,
+  optGroupPropType,
+} from './NativePropTypes';
+
+import styles from './NativeSelect.module.scss';
+
+const cx = classNamesBind.bind(styles);
+
+const propTypes = {
```
mjhenkes, line 27: Should we add a first class prop for max-width? We apply it to all the form select examples.

cc @StephenEsser

packages/terra-form-select/src/terra-dev-site/doc/example/native-select-field/DefaultField.jsx
```diff
@@ -0,0 +1,22 @@
...
+
+const cx = classNames.bind(styles);
+
+const DefaultFieldExample = () => (
+  <NativeSelectField
```
mjhenkes, line 9: Should this example have an aria label?

mjhenkes: Actually i assume this is handled by being a field and passing in the label.

packages/terra-form-select/src/terra-dev-site/doc/example/native-select-field/IncompleteField.jsx
```diff
@@ -0,0 +1,25 @@
...
+    label="Incomplete Field Text"
+    required
+    isIncomplete
+    options={[
+      { value: 'volvo', display: 'Volvo' },
+      { value: 'saab', display: 'Saab' },
+      { value: 'mercedes', display: 'Mercedes' },
+      { value: 'audi', display: 'Audi' },
+    ]}
+    placeholder={{}}
```
mjhenkes, line 20: What is this doing? Is placeholder required?

packages/terra-form-select/src/native-select/NativeSelect.jsx
mjhenkes: I suppose that would make it a prop not a class.... maybe nevermind here, I wish there was a better way to do this.

packages/terra-form-select/src/terra-dev-site/doc/example/native-select-field/IncompleteField.jsx
mjhenkes: I assume the empty object indicates the `-select-` text as the default.

packages/terra-form-select/src/terra-dev-site/doc/form-select/Select.01.doc.mdx
```diff
@@ -47,7 +47,10 @@ This component requires the following peer dependencies be installed in your app
 
 The Form-Select component must be composed inside the [Base][1] component with a locale in order for it to load the correct translation strings.
 
+As a result of accessibility issues, the default implementation of select should no be used. It will be fully removed in a future major version bump. To achieve the same functionality the [Native Select][2] should be consumed.
```
mjhenkes, line 50:
```suggestion
As a result of accessibility issues, the default implementation of select should not be used. It will be fully removed in a future major version bump. To achieve the same functionality the [Native Select][2] should be consumed.
```

packages/terra-form-select/src/terra-dev-site/doc/form-select/Select.01.doc.mdx
```diff
@@ -57,15 +60,15 @@ import Select from 'terra-form-select';
 
 ## Component Features
 
-- [Cross-Browser Support](https://engineering.cerner.com/terra-ui/contributing/terra-ui/component-standards#cross-browser-support)
+- [Cross-Browser Support](https://engineering.cerner.com/terra-ui/contributing/te87rra-ui/component-standards#cross-browser-support)
```
mjhenkes, line 63: wat?
```suggestion
- [Cross-Browser Support](https://engineering.cerner.com/terra-ui/contributing/terra-ui/component-standards#cross-browser-support)
```

packages/terra-form-select/src/terra-dev-site/doc/form-select/Select.01.doc.mdx
```diff
@@ -47,7 +47,10 @@ This component requires the following peer dependencies be installed in your app
 
 The Form-Select component must be composed inside the [Base][1] component with a locale in order for it to load the correct translation strings.
 
+As a result of accessibility issues, the default implementation of select should no be used. It will be fully removed in a future major version bump. To achieve the same functionality the [Native Select][2] should be consumed.
+
 [1]: https://engineering.cerner.com/terra-core/components/terra-base/base/base
+[2]: https://engineering.cerner.com/terra-core/components/terra-form-select/form-select/native-select
```
mjhenkes, line 53: Doing this should allow this link to work for both terra-ui and terra-core
```suggestion
[2]: /components/terra-form-select/form-select/native-select
```

packages/terra-form-select/tests/wdio/native-select-field-spec.js
```diff
@@ -0,0 +1,19 @@
+Terra.describeViewports('Native Select Field', ['tiny'], () => {
+  describe('Default Field', () => {
+    before(() => browser.url('/#/raw/tests/terra-form-select/form-select/native-select/default-field'));
```
mjhenkes, line 3: remove a redirect
```suggestion
    before(() => browser.url('/raw/tests/terra-form-select/form-select/native-select/default-field'));
```

packages/terra-form-select/src/terra-dev-site/doc/example/native-select-field/DefaultField.jsx
dkasper-was-taken: Yes the label serves double duty as the aria label as well. I believe it matches out existing field patterns.

packages/terra-form-select/src/terra-dev-site/doc/example/native-select-field/IncompleteField.jsx
dkasper-was-taken: The empty object indicates that a placeholder is desired, but no custom values are going to be passed.

packages/terra-form-select/src/terra-dev-site/doc/form-select/Select.01.doc.mdx
dkasper-was-taken: Looks like the Rini special. Updated

packages/terra-form-select/src/terra-dev-site/doc/form-select/Select.01.doc.mdx
dkasper-was-taken: Updated.

packages/terra-form-select/tests/wdio/native-select-field-spec.js
dkasper-was-taken: Updated.

packages/terra-form-select/src/terra-dev-site/doc/form-select/Select.01.doc.mdx
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,244 @@
...
+     */
+    allowClear: PropTypes.bool,
+    /**
+     * Optional override of the default display.
+     */
+    display: PropTypes.string,
+    /**
+     * Optional override of the default value.
+     */
+    value: PropTypes.string,
```
supreethmr, line 89: Curious to know the purpose of value prop here, does display and value serve different purpose..?

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,244 @@
...
+      onChange(event);
+    }
+    if (!refIsControlled.current) {
+      setCurrentValue(event.currentTarget.value);
+    }
+  };
+
+  let controlledValue;
+  if (refIsControlled.current) {
+    controlledValue = value || getFirstValue(options, placeholder);
```
supreethmr, line 182: I think this can be just ` controlledValue = value `. Since we are validating `const refIsControlled = useRef(!defaultValue && value);` I don't think value would be null when `if (refIsControlled.current)` validates to true and also we have one more check on line no. 192 where if `controlledValue` is null / undefined it will be replaced with `currentValue` ( which will be either `defaultValue` or `getFirstValue` )

packages/terra-form-select/tests/wdio/native-select-spec.js
```diff
@@ -0,0 +1,176 @@
...
+
+    Terra.it.validatesElement('initial');
+
+    it('should apply hover to the select', () => {
+      browser.moveToObject('#select-id', 5, 5);
+    });
+
+    Terra.it.validatesElement('hover');
+
+    it('should set focus to the select', () => {
```
supreethmr, line 38: I think this should be `'should not set focus to the select'`

packages/terra-form-select/tests/wdio/native-select-spec.js
```diff
@@ -0,0 +1,176 @@
...
+      browser.click('#root', 0, 0);
```
supreethmr, line 64: since select is still in its initial state. I don't think we need to click on root before tabbing. even if we just do Tab native-select will still receive  focus.

dkasper-was-taken: The tab key was not functioning without an initial interaction in the window.

packages/terra-form-select/src/native-select/NativeSelect.jsx
supreethmr: > if i user clears the selected value the select should reset to the first item.

Correct me if I am wrong. it resets to first item in line No. 192 also : ` value: controlledValue || currentValue`

dkasper-was-taken: That would only be correct if the list of options stayed static, as the default is only captured on initial mount, if the user adds/removes items in the select there is a potential an incorrect value would attempt to be restored, so the value needs to be captured anew is the value is cleared. There are other potential ways to account for this, but it is covered with this logic.

I can potentially update the state variable to `uncontrolledValue` rather that `currentValue` to make it clearer.

packages/terra-form-select/tests/wdio/native-select-spec.js
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: I updated the component to be a little more explicit about how it manages values.

packages/terra-form-select/CHANGELOG.md
```diff
@@ -29,6 +29,9 @@ Unreleased
 
 6.2.0 - (April 22, 2020)
 ------------------
+### Added
+* Added `NativeSelect` implementation to replace the default select variant.
+
```
StephenEsser, line 34: Nit: This was bumped into a released section

packages/terra-form-select/src/native-select/NativeUtils.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+  if (placeholder.value) {
+    return value === placeholder.value;
+  }
+  return value === defaultPlaceholderValue;
+};
+
+const getOptGroupKey = (display, index) => `${display}-${index}`;
+
+const geOptGroupKeyIndex = (display, currentKeys, index = 0) => {
```
ryanthemanuel, line 61:
```suggestion
const getOptGroupKeyIndex = (display, currentKeys, index = 0) => {
```

packages/terra-form-select/src/native-select/NativeUtils.js
```diff
@@ -0,0 +1,84 @@
...
+  if (currentKeys.indexOf(getOptGroupKey(display, index)) >= 0) {
+    return geOptGroupKeyIndex(display, currentKeys, index + 1);
```
ryanthemanuel, line 63:
```suggestion
    return getOptGroupKeyIndex(display, currentKeys, index + 1);
```

packages/terra-form-select/src/native-select/NativeUtils.js
```diff
@@ -0,0 +1,84 @@
...
+  return index;
+};
+
+export default {
+  defaultPlaceholderValue,
+  isCurrentPlaceholder,
+  getDisplay,
+  getFirstValue,
+  getOptGroupKey,
+  geOptGroupKeyIndex,
```
ryanthemanuel, line 74:
```suggestion
  getOptGroupKeyIndex,
```

packages/terra-form-select/src/native-select/NativeUtils.js
```diff
@@ -0,0 +1,84 @@
...
+};
+
+export {
+  defaultPlaceholderValue,
+  isCurrentPlaceholder,
+  getDisplay,
+  getFirstValue,
+  getOptGroupKey,
+  geOptGroupKeyIndex,
```
ryanthemanuel, line 83:
```suggestion
  getOptGroupKeyIndex,
```

packages/terra-form-select/tests/jest/native-select/NativeUtils.test.js
```diff
@@ -0,0 +1,128 @@
+import {
+  defaultPlaceholderValue,
+  isCurrentPlaceholder,
+  getDisplay,
+  getFirstValue,
+  getOptGroupKey,
+  geOptGroupKeyIndex,
```
ryanthemanuel, line 7:
```suggestion
  getOptGroupKeyIndex,
```

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,247 @@
...
+import classNamesBind from 'classnames/bind';
+import { injectIntl, intlShape } from 'react-intl';
+import ThemeContext from 'terra-theme-context';
+import {
+  defaultPlaceholderValue,
+  isCurrentPlaceholder,
+  getDisplay,
+  getFirstValue,
+  getOptGroupKey,
+  geOptGroupKeyIndex,
```
ryanthemanuel, line 16:
```suggestion
  getOptGroupKeyIndex,
```

packages/terra-form-select/src/native-select/NativeSelect.jsx
StephenEsser: The field wrappers have a `maxWidth` property consumers can utilize.  I don't think I'd vote for adding the additional prop here unless we plan to add it to all the other variants as well

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,247 @@
...
+      { placeholder: isCurrentPlaceholder(selectAttrs.value, placeholder) },
+    ),
+    customProps.className,
+  );
+
+  return (
+    <div
+      {...customProps}
+      className={nativeClassNames}
+      ref={refSelect}
```
tbiethman, line 214: Do we need to expose this ref for consumers? Or for the `select` element?

packages/terra-form-select/src/native-select/NativeSelect.module.scss
```diff
@@ -0,0 +1,173 @@
...
+    padding-left: var(--terra-form-select-native-frame-padding-left, 0.4286rem);
+    padding-right: 0;
+    padding-top: 0;
+    position: absolute;
+    top: 0;
+    width: 100%;
+    z-index: 0;
+  }
+
+  .select {
```
StephenEsser, line 125: The positioning seems a little misaligned on the deployed site. Is this the intended position of the dropdown when open?

image


Slightly offset to the left, a small reveal of the dropdown arrow

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,247 @@
...
+   */
+  defaultValue: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  /**
+   * Whether the input is disabled.
+   */
+  disabled: PropTypes.bool,
+  /**
+   * The Select identifier. Links the htmlFor of the field to the select identifier.
+   */
+  id: PropTypes.string.isRequired,
```
StephenEsser, line 53: This is a little tricky, should the ID be required when used outside of a field? There won't be an htmlFor to be bound to

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,247 @@
...
+  isInvalid,
+  isIncomplete,
+  onChange,
+  options,
+  placeholder,
+  required,
+  value,
+  ...customProps
+}) => {
+  const [uncontrolledValue, setUncontrolledValue] = useState(!value ? defaultValue || getFirstValue(options, placeholder) : undefined);
```
StephenEsser, line 156: The evaluation of the `value` prop may need to account for an empty string, in this case an empty string would always flip the select into an uncontrolled. Generally empty string is the placeholder/cleared value representation.

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,247 @@
...
+    refSelect.current.setAttribute('data-focus-interaction', 'none');
+  };
+  const handleOnFocus = () => {
+    if (refSelect.current.getAttribute('data-focus-interaction') !== 'mouse') {
+      refSelect.current.setAttribute('data-focus-interaction', 'keyboard');
+    }
+  };
+  const handleOnChange = event => {
+    if (onChange) {
+      onChange(event);
```
StephenEsser, line 174: Either the return value here needs to change or the onChange documentation in the propTypes needs to change. The documentation currently states the value will be return:

> Callback function triggered when the select value changes. function(value)

packages/terra-form-select/src/native-select/NativeSelectField.jsx
```diff
@@ -0,0 +1,204 @@
...
+  selectId,
+  showOptional,
+  value,
+  ...customProps
+}) => {
+  const helpText = (
+    <span>
+      {help}
+    </span>
+  );
```
StephenEsser, line 151: This is probably short enough to be inlined, but it doesn't make a difference either way. 

```jsx
const helpText = <span>{help}</span>;
```

Or

```jsx
<Field
  help={<span>{help}</span>}
/>
```

packages/terra-form-select/src/native-select/NativeSelectField.jsx
```diff
@@ -0,0 +1,204 @@
...
+    }
+  }
+
+  return (
+    <Field
+      {...customProps}
+      label={label}
+      labelAttrs={labelAttrs}
+      error={error}
+      help={helpText}
```
StephenEsser, line 172: Should this be conditionally added if help text is provided or do we want the empty span there for placeholder space?

mjhenkes:
@neilpfeiffer Should we remove the placeholder from this select now? or in a future release?

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.module.scss
dkasper-was-taken: This occurs depending on the browser. When altering the dropdown to span 100% of the width it breaks certain browsers alignment. After some discussion with UX, while it is annoying it acceptable.

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: !"" evaluates to true, so any value passed on the initial mount should satisfy the condition. Also, it's in progress, but the placeholder is being removed as well.

dkasper-was-taken:
@mjhenkes I believe Neil said in scrum to remove it. I just haven't made the change yet.

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: I will update the doc.

packages/terra-form-select/src/native-select/NativeSelect.jsx
StephenEsser: It looks like an empty string will set the uncontrolledValue as the first option in the list using `getFirstValue`. I think this will make `uncontrolledValue ` and `refIsControlled` ambiguous. It's a little goofy, I've seen a few teams in the past using an empty string as a value

```jsx
// (depending on if the empty string option is the first in the list)
// This will initialize `uncontrolledValue` as empty string and `refIsControlled` as true
<NativeSelect value='' />
```

```jsx
// (depending on if the empty string option is the first in the list)
// This will initialize `uncontrolledValue` as empty string and `refIsControlled` as true
<NativeSelect defaultValue='' />
```

packages/terra-form-select/src/native-select/NativeUtils.js
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeUtils.js
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeUtils.js
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeUtils.js
dkasper-was-taken: Updated.

packages/terra-form-select/tests/jest/native-select/NativeUtils.test.js
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: Updated to use refCallback.

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: Updated with:
```
  const refIsControlled = useRef(!!value);
```

<NativeSelect defaultValue='' />
// This results in:
// uncontrolledValue: ""
// refIsControlled: false

dkasper-was-taken: I see the issue now, my bad.

dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelectField.jsx
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelectField.jsx
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: Updated to be optional within the select, but required in the field.

packages/terra-form-select/CHANGELOG.md
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,278 @@
...
+  };
+  const handleOnFocus = (event) => {
+    if (refSelect.current.getAttribute('data-focus-interaction') !== 'mouse') {
+      refSelect.current.setAttribute('data-focus-interaction', 'keyboard');
+    }
+    if (onFocus) {
+      onFocus(event);
+    }
+  };
+  const handleOnChange = event => {
```
supreethmr, line 202: we should consider using useCallback for these funcitons :  https://reactjs.org/docs/hooks-reference.html#usecallback

dkasper-was-taken: Looking at some the implementation guides it appears that useCallback's optimization may be more expensive than its benefit as the inline function is still generated. The ideal usage would be more if the implementation relied more heavily on memoization.
https://dmitripavlutin.com/dont-overuse-react-usecallback/
https://kentcdodds.com/blog/usememo-and-usecallback

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,279 @@
...
+   */
+  defaultValue: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  /**
+   * Whether the input is disabled.
+   */
+  disabled: PropTypes.bool,
+  /**
+   * Whether a placeholder value should be included as an option.
+   */
+  hasPlaceholder: PropTypes.bool,
```
neilpfeiffer, line 59: @dkasper-was-taken I would rather this not be an exposed prop, since the placeholder text will always be defined internal to the component, only `allowClear` should have the ability to include the placeholder in the list of options and therefore this would be an internal only item to track with allowClear.

packages/terra-form-select/src/native-select/NativeSelect.module.scss
```diff
@@ -0,0 +1,177 @@
...
+    transition-property: background-size;
+    transition-timing-function: ease;
+    width: 100%;
+
+    &[data-focus-interaction='keyboard'] {
+      box-shadow: var(--terra-form-select-native-focus-keyboard-box-shadow, 0 0 0 2px rgba(76, 178, 233, 0.25));
+    }
+
+    &[data-focus-interaction='keyboard'],
+    &[data-focus-interaction='mouse'], {
```
neilpfeiffer, line 31: comma
```suggestion
    &[data-focus-interaction='mouse'] {
```

packages/terra-form-select/src/native-select/NativeSelect.module.scss
```diff
@@ -0,0 +1,177 @@
...
+      .arrow-icon {
+        background-image: var(--terra-form-select-native-invalid-arrow-icon-background-image, inline-svg('<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48" fill="#e50000" class="is-bidi"><path d="M48 12L24 36 0 12h48z"/></svg>'));
+      }
+
+      &[data-focus-interaction='keyboard'] {
+        box-shadow: var(--terra-form-select-native-invalid-focus-keyboard-box-shadow, 0 0 0 1px #e50000);
+      }
+  
+      &[data-focus-interaction='keyboard'],
+      &[data-focus-interaction='mouse'], {
```
neilpfeiffer, line 67: stray comma
```suggestion
      &[data-focus-interaction='mouse'] {
```

packages/terra-form-select/src/native-select/NativeSelect.module.scss
```diff
@@ -0,0 +1,177 @@
...
+      &:hover {
+        background-color: var(--terra-form-select-native-invalid-hover-background-color, #fff);
+        border-color: var(--terra-form-select-native-invalid-hover-border-color, #e50000);
+      }
+    }
+
+    &.placeholder {
+      color: var(--terra-form-select-native-placeholder-color, rgba(21, 33, 41, 0.62));
+      font-size: var(--terra-form-select-native-placeholder-font-size, 1.143rem);
+      font-style: var(--terra-form-select-native-placeholder-font-style, normal);
```
neilpfeiffer, line 81: Change to italics
```suggestion
      font-style: var(--terra-form-select-native-placeholder-font-style, italic);
```

packages/terra-form-select/src/clinical-lowlight-theme/NativeSelect.module.scss
```diff
@@ -0,0 +1,65 @@
...
+    --terra-form-select-native-background-size: 0 0;
+    --terra-form-select-native-border: 1px solid #181b1d;
+    --terra-form-select-native-border-radius: 3px;
+    --terra-form-select-native-color: #b2b5b6;
+    --terra-form-select-native-font-size: 1.143rem;
+    --terra-form-select-native-height: 2.142857rem;
+    --terra-form-select-native-line-height: 1.75;
+    --terra-form-select-native-placeholder-color: #6f7477;
+    --terra-form-select-native-placeholder-font-size: 1.143rem;
+    --terra-form-select-native-placeholder-font-style: normal;
```
neilpfeiffer, line 17:
```suggestion
    --terra-form-select-native-placeholder-font-style: italic;
```

packages/terra-form-select/src/clinical-lowlight-theme/NativeSelect.module.scss
```diff
@@ -0,0 +1,65 @@
...
+    --terra-form-select-native-arrow-icon-width: 0.714rem;
+
+    // focus
+    --terra-form-select-native-focus-background-color: #222a2e;
+    --terra-form-select-native-focus-background-size: 0 0;
+    --terra-form-select-native-focus-border-color: #004c76;
+    --terra-form-select-native-focus-transition-duration: 0;
+
+    // focus - keyboard
+    --terra-form-select-native-focus-keyboard-box-shadow: 0 0 1px 3px #004c76, 0 0 7px 4px #004c76;
```
neilpfeiffer, line 31: Update plus add:
```suggestion
    --terra-form-select-native-focus-keyboard-box-shadow: none;
    --terra-form-select-native-focus-keyboard-outline: 2px dashed #b2b5b6;
    --terra-form-select-native-focus-keyboard-outline-offset: 1px;
```

packages/terra-form-select/src/clinical-lowlight-theme/NativeSelect.module.scss
```diff
@@ -0,0 +1,65 @@
...
+    --terra-form-select-native-focus-hover-border-color: #004c76;
+
+    // incomplete
+    --terra-form-select-native-incomplete-background-color: #464232;
+
+    // invalid
+    --terra-form-select-native-invalid-background-color: #222a2e;
+    --terra-form-select-native-invalid-background-image: none;
+    --terra-form-select-native-invalid-border-color: #fb4c4c;
+    --terra-form-select-native-invalid-box-shadow: 0 0 0 1px #fb4c4c, 0 0 0 1px #fb4c4c;
```
neilpfeiffer, line 48: Update plus add:
```suggestion
    --terra-form-select-native-invalid-box-shadow: 0 0 0 1px #fb4c4c;
    --terra-form-select-native-invalid-focus-keyboard-box-shadow: 0 0 0 1px #fb4c4c;
    --terra-form-select-native-invalid-focus-keyboard-outline: 2px dashed #b2b5b6;
    --terra-form-select-native-invalid-focus-keyboard-outline-offset: 1px;
```

packages/terra-form-select/src/orion-fusion-theme/NativeSelect.module.scss
```diff
@@ -0,0 +1,65 @@
...
+    --terra-form-select-native-arrow-icon-width: 0.66667rem;
+
+    // focus
+    --terra-form-select-native-focus-background-color: #fff;
+    --terra-form-select-native-focus-background-size: 0 0;
+    --terra-form-select-native-focus-border-color: #4cb2e9;
+    --terra-form-select-native-focus-transition-duration: 0;
+
+    // focus - keyboard
+    --terra-form-select-native-focus-keyboard-box-shadow: 0 0 4px #a6d9f4;
```
neilpfeiffer, line 31: Add:
```
    --terra-form-select-native-focus-keyboard-outline: none;
    --terra-form-select-native-focus-keyboard-outline-offset: 0;
```

packages/terra-form-select/src/orion-fusion-theme/NativeSelect.module.scss
```diff
@@ -0,0 +1,65 @@
...
+    --terra-form-select-native-focus-hover-border-color: #4cb2e9;
+
+    // incomplete
+    --terra-form-select-native-incomplete-background-color: #fefd9a;
+
+    // invalid
+    --terra-form-select-native-invalid-background-color: #fff;
+    --terra-form-select-native-invalid-background-image: none;
+    --terra-form-select-native-invalid-border-color: #dedfe0;
+    --terra-form-select-native-invalid-box-shadow: none;
```
neilpfeiffer, line 48: Add:
```
    --terra-form-select-native-invalid-focus-keyboard-box-shadow: 0 0 4px #a6d9f4;
    --terra-form-select-native-invalid-focus-keyboard-outline: none;
    --terra-form-select-native-invalid-focus-keyboard-outline-offset: 0;
```

packages/terra-form-select/src/native-select/NativeSelect.module.scss
```diff
@@ -0,0 +1,177 @@
...
+    height: var(--terra-form-select-native-height, 2.142857rem);
+    line-height: var(--terra-form-select-native-line-height, 1.75);
+    position: relative;
+    transition-duration: 0s;
+    transition-property: background-size;
+    transition-timing-function: ease;
+    width: 100%;
+
+    &[data-focus-interaction='keyboard'] {
+      box-shadow: var(--terra-form-select-native-focus-keyboard-box-shadow, 0 0 0 2px rgba(76, 178, 233, 0.25));
```
neilpfeiffer, line 27: Update and add:
```suggestion      
      box-shadow: var(--terra-form-select-native-focus-keyboard-box-shadow, none);
      outline: var(--terra-form-select-native-focus-keyboard-outline, 2px dashed #000);
      outline-offset: var(--terra-form-select-native-focus-keyboard-outline-offset, 1px);
```

packages/terra-form-select/src/native-select/NativeSelect.module.scss
```diff
@@ -0,0 +1,177 @@
...
+      transition-duration: var(--terra-form-select-native-focus-transition-duration, 0);
+
+      &:hover {
+        border-color: var(--terra-form-select-native-focus-hover-border-color, #4cb2e9);
+      }
+    }
+
+    &:hover {
+      background-color: var(--terra-form-select-native-hover-background-color, #fff);
+      border-color: var(--terra-form-select-native-hover-border-color, #4cb2e9);
```
neilpfeiffer, line 44:
```suggestion
      border-color: var(--terra-form-select-native-hover-border-color, #dedfe0);
```

packages/terra-form-select/src/native-select/NativeSelect.module.scss
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    &.invalid {
+      background-color: var(--terra-form-select-native-invalid-background-color, #fff);
+      background-image: var(--terra-form-select-native-invalid-background-image, none);
+      border-color: var(--terra-form-select-native-invalid-border-color, #e50000);
+      box-shadow: var(--terra-form-select-native-invalid-box-shadow, 0 0 0 1px #e50000);
+
+      .arrow-icon {
+        background-image: var(--terra-form-select-native-invalid-arrow-icon-background-image, inline-svg('<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48" fill="#e50000" class="is-bidi"><path d="M48 12L24 36 0 12h48z"/></svg>'));
```
neilpfeiffer, line 59: color
```suggestion
        background-image: var(--terra-form-select-native-invalid-arrow-icon-background-image, inline-svg('<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48" fill="#1c1f21" class="is-bidi"><path d="M48 12L24 36 0 12h48z"/></svg>'));
```

packages/terra-form-select/src/native-select/NativeSelect.module.scss
```diff
@@ -0,0 +1,177 @@
...
+      }
+
+      &[data-focus-interaction='keyboard'] {
+        box-shadow: var(--terra-form-select-native-invalid-focus-keyboard-box-shadow, 0 0 0 1px #e50000);
```
neilpfeiffer, line 63: add:
```
        outline: var(--terra-form-select-native-invalid-focus-keyboard-outline, 2px dashed #000);
        outline-offset: var(--terra-form-select-native-invalid-focus-keyboard-outline-offset, 1px);
```

packages/terra-form-select/src/clinical-theme/NativeSelect.module.scss
```diff
@@ -0,0 +1,65 @@
...
+    --terra-form-select-native-arrow-icon-width: 0.714rem;
+
+    // focus
+    --terra-form-select-native-focus-background-color: #fff;
+    --terra-form-select-native-focus-background-size: 100% 0.142857rem;
+    --terra-form-select-native-focus-border-color: #4cb2e9;
+    --terra-form-select-native-focus-transition-duration: 0.33s;
+
+    // focus - keyboard
+    --terra-form-select-native-focus-keyboard-box-shadow: none;
```
neilpfeiffer, line 31: add:
```
    --terra-form-select-native-focus-keyboard-outline: none;
    --terra-form-select-native-focus-keyboard-outline-offset: 0;
```

packages/terra-form-select/src/clinical-theme/NativeSelect.module.scss
```diff
@@ -0,0 +1,65 @@
...
+    --terra-form-select-native-focus-hover-border-color: #4cb2e9;
+
+    // incomplete
+    --terra-form-select-native-incomplete-background-color: #ffc;
+
+    // invalid
+    --terra-form-select-native-invalid-background-color: #fcfdff;
+    --terra-form-select-native-invalid-background-image: none;
+    --terra-form-select-native-invalid-border-color: #e50000;
+    --terra-form-select-native-invalid-box-shadow: none;
```
neilpfeiffer, line 48: add:
```
    --terra-form-select-native-invalid-focus-keyboard-box-shadow: none;
    --terra-form-select-native-invalid-focus-keyboard-outline: none;
    --terra-form-select-native-invalid-focus-keyboard-outline-offset: 0;
```

packages/terra-form-select/src/clinical-theme/NativeSelect.module.scss
```diff
@@ -0,0 +1,65 @@
+@import '~terra-mixins/lib/Mixins';
+
+:local {
+  .clinical-theme {
```
neilpfeiffer, line 4: ⚠️ Remember to remove `clinical-theme` and keep internal before final.

packages/terra-form-select/src/native-select/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/clinical-lowlight-theme/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/clinical-lowlight-theme/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/clinical-lowlight-theme/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/orion-fusion-theme/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/orion-fusion-theme/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/orion-fusion-theme/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/clinical-theme/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/clinical-theme/NativeSelect.module.scss
dkasper-was-taken: Updated.

packages/terra-form-select/src/clinical-theme/NativeSelect.module.scss
dkasper-was-taken: Removed.

PaulSolDev:
Functional review complete.  The following concerns were identified:

- There seems to be an issue in Firefox where I'm making quick selections (in one or more select component) and occasionally the list will open (as expected) but then shut unexpectedly.

- Firefox and IE allow selections using tab and enter but not spacebar. All other browsers allow selections using spacebar and enter.

dkasper-was-taken:
@PaulSolDev The keyboard input inconsistencies are directly tied to the differences between the individual browsers, so we don't have control of their behavior given different inputs.

As far as the quick selections, can you describe the steps to reproduce or take a recording of the behavior?

dkasper-was-taken:
@PaulSolDev Looks like the occasionally closing is a bug logged to mozilla:
https://bugzilla.mozilla.org/show_bug.cgi?id=1430892

PaulSolDev:
Reviewed the above responses with @dkasper-was-taken and confirmed that these were known issues/differences that exist at the browser level.  

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -0,0 +1,272 @@
...
+   * The Select identifier to be applied to the html select node.
+   */
+  id: PropTypes.string,
+  /**
+   * @private
+   * The intl object to be injected for translations.
+   */
+  intl: intlShape.isRequired,
+  /**
+   * Whether the select input should use the filter style display, forcing a value to be selected.
```
neilpfeiffer, line 61: @dkasper-was-taken How does this sound to update the prop description? 
```suggestion
   * Whether the select input should use the filter style display, forcing a value to always be selected. It also removes the Terra placeholder and removes the ability for user to clear the value, returning the select to browser-native behavior.
```

packages/terra-form-select/src/terra-dev-site/doc/form-select/NativeSelect.13.doc.mdx
```diff
@@ -0,0 +1,56 @@
...
+- [Internationalization Support](https://engineering.cerner.com/terra-ui/contributing/terra-ui/component-standards#internationalization-i18n-support)
+- [Localization Support](https://engineering.cerner.com/terra-ui/contributing/terra-ui/component-standards#internationalization-i18n-support)
+- [LTR/RTL Support](https://engineering.cerner.com/terra-ui/contributing/terra-ui/component-standards#ltr--rtl-support)
+
+[View component source code](https://github.com/cerner/terra-core/tree/master/packages/terra-form-select)
+
+## Examples
+<UncontrolledExample title="Create an Uncontrolled Select" description="An example of implementing an uncontrolled Select." />
+<ControlledExample title="Create a Controlled Select" description="An example of implementing a controlled Select." />
+<FilterStyleExample title="Filter Style" description="An example of implementing a select to be user outside of form fields. (i.e. in a filter bar)" />
```
neilpfeiffer, line 42:
```suggestion
<FilterStyleExample title="Filter Style" description="For special use: an example of implementing a select outside of form fields (e.g. in a filter bar), which removes the placeholder, prevents the ability to be cleared, and forces a value to always be selected." />
```

dkasper-was-taken: Updated.

packages/terra-form-select/src/native-select/NativeSelect.jsx
dkasper-was-taken: Updated.

packages/terra-form-select/src/clinical-lowlight-theme/NativeSelect.module.scss
```diff
@@ -0,0 +1,70 @@
...
+    --terra-form-select-native-invalid-focus-keyboard-outline: 2px dashed #b2b5b6;
+    --terra-form-select-native-invalid-focus-keyboard-outline-offset: 1px;
+    --terra-form-select-native-invalid-hover-background-color: #222a2e;
+    --terra-form-select-native-invalid-hover-border-color: #fb4c4c;
+
+    // disabled
+    --terra-form-select-native-disabled-background-color: #222a2e;
+    --terra-form-select-native-disabled-border-color: #202427;
+    --terra-form-select-native-disabled-color: #b2b5b6;
+    --terra-form-select-native-disabled-opacity: 1;
```
neilpfeiffer, line 64:
```suggestion
    --terra-form-select-native-disabled-opacity: 0.3;
```

neilpfeiffer:
Tested: PR #2934

**Reviewed:**
- Follows pattern considerations documented in Cerner Design Standards: [Standard Drop-down: single-select](https://wiki.cerner.com/x/jGsVYQ)
- Follows UX defined design implementation for Terra Native Select
- Follows defined theme designs: terra default theme, clinical-lowlight-theme, orion-fusion-theme

**Followup:**
- Form field can be improved for screenreader accessibility, issue logged here: _{adding new bug issue}_

\
\

**Pass** |
-----------|
UX Review|

neilpfeiffer:
Accessibility reviewed.

PR #2938, opened 2020-04-22
Improves data-load performance of form-select-default variant

### LINKED ISSUE #2161: Form-Select speed handling large lists

supreethmr:
### Summary
To address the delay in load performance of form-select which was happening with large datasets implemented lazy loading for single-form-select. Options that will be visible on dropdown open will be rendered initially and rest of the options will be added to list as we scroll down the list.

With this approach the time taken to render the options has been considerably reduced compared to existing form-select. Since number of options rendered on open is reduced to certain count form-select takes less time to render the options.

**Time taken to load 1000 options by existing form-select** : 

Screen Shot 2020-04-23 at 4 55 42 PM

**Time taken to load 1000 options by new form-select** : 

Screen Shot 2020-04-23 at 4 59 51 PM

Number of Initial options to be rendered are calculated with comparison of dropdown `maxHeight` prop if provided else with parentWindow height which will be used to calculate the number of options to be rendered initially. Later as we scroll down the list same set of options gets added into dropdown to keep the scrolling behaviour aligned with existing behaviour.

Closes #2161 

### Deployment Link

https://terra-core-deployed-pr-2938.herokuapp.com/components/terra-form-select/form-select/single-select

### Testing
Tested with Chrome Profiler Tools by recording time-taken to load same number of options by both existing and new form-select. 

@cerner/terra

### DIFF
--- a/packages/terra-form-select/CHANGELOG.md
+++ b/packages/terra-form-select/CHANGELOG.md
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Added
+* Added `isLazyLoadEnabled` to enable lazy loading of options in default form-select
 
 6.2.0 - (April 22, 2020)
 ------------------
--- a/packages/terra-form-select/src/Select.jsx
+++ b/packages/terra-form-select/src/Select.jsx
@@ -107,6 +107,10 @@ const propTypes = {
    */
   value: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.array]),
   /**
+   *  Enables lazy loading options when set to true.
+   */
+  isLazyLoadEnabled: PropTypes.bool,
+  /**
    * The behavior of the select. One of `default`, `combobox`, `multiple`, `tag`, or `search`.
    */
   variant: PropTypes.oneOf([
--- a/packages/terra-form-select/src/SingleSelect.jsx
+++ b/packages/terra-form-select/src/SingleSelect.jsx
@@ -85,6 +85,10 @@ const propTypes = {
    * The selected value. Can be a string, number, or array of strings/numbers.
    */
   value: PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.array]),
+  /**
+   *  Enables lazy loading options when set to true.
+   */
+  isLazyLoadEnabled: PropTypes.bool,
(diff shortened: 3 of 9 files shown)


dkasper-was-taken:
We may want to re-evaluate this change.  The default select variant is being deprecated in favor of the native select. Also the terra-infinite-list is being deprecated because virtualization and lazy load scrolling are inaccessible concepts, so I don't know that introducing new similar concepts is a good idea.

supreethmr:
> We may want to re-evaluate this change. The default select variant is being deprecated in favor of the native select. Also the terra-infinite-list is being deprecated because virtualization and lazy load scrolling are inaccessible concepts, so I don't know that introducing new similar concepts is a good idea.

Thank you Dave for the information..! I was unaware of default select being deprecated. So went with this approach to fix [#2161](https://github.com/cerner/terra-core/issues/2161).
as part of investigation I also verified new native select with same set of options (1000 options). native select does improves performance to greater extent compare to existing form-select. 

This issue with performance load time does remains with other variants of form-select. I was trying to implement similar logic with other variants as well by keeping search and other functionalities to happen on virtual list instead of actual list that will displayed to user.

But having said that `virtualization and lazy load scrolling are inaccessible concepts` I would like to know if there are any alternative approaches that are being considered as replacement of  infinite-list going ahead.

mjhenkes:
Closing, the most performant single select list will be the native version, which will be our recommendation.

PR #2943, opened 2020-04-28
[terra-search-field] Fixed Invalid theme variables

### LINKED ISSUE #2936: terra-search-field invalid theme variables
# Bug Report

## Description
terra-search-field has invlaid theme variables without the double dash
https://github.com/cerner/terra-core/blob/master/packages/terra-search-field/src/SearchField.module.scss#L53-L54

## Possible Solution
add the double dash

saket2403:
### Summary
Fixed invalid theme variables without the double dash

Closes #2936 

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-search-field/src/SearchField.module.scss
```diff
@@ -50,8 +50,8 @@
       flex: 0 0 auto;
       margin: 0;
       position: absolute;
-      right: var(-terra-search-field-clear-right, 0);
-      top: var(-terra-search-field-clear-top, 0);
+      right: var(--terra-search-field-clear-right, 0);
+      top: var(--terra-search-field-clear-top, 0);
```
yuderekyu, line 54: do our themes actually use these variables? if they did, i'd expect us to update screenshots. i did a theme-wide search but got back 0 matches 😕perhaps we should remove these custom css variables instead?

emilyrohrbough: unless it's used by consumer or clinical... if we can't remove, we should add them to the orion-fusion-theme & clinical-lowlight-theme since we are trying to back-fill the partial themes

mjhenkes: I agree with @emilyrohrbough

yuderekyu: yeah, when i mentioned theme-wide search, that also included clinical and consumer themes :D unless i missed it, they weren't being used there

emilyrohrbough: Sounds like these can be removed based on Derek's search & usage.

saket2403: Removed d65e1ce

PR #2946, opened 2020-04-30
[terra-form-select] Fixes inconsistent hover interactions 

### LINKED ISSUE #2945: [terra-form-select] Non-enabled and invisible content is interactable and causing inconsistent hovers
# Bug Report

## Description

Hookshot's API provides an `isEnabled` property that can be toggled on and off. When isEnabled is false the content is positioned in the top left corner of the screen with an opacity of 0 to make it invisible to users. The invisible content is mouse interactable and overlays the screen.

## Steps to Reproduce

Here's a sample code sandbox: https://codesandbox.io/s/eager-maxwell-spi0e
1.  Open the code sandbox.
2. Click around on the example page, but do not click the button to open hookshot yet.
3. Observe all of the click events are being counted.
4. Click the button to open hookshot.
5. Attempt to click anywhere on the page. 
6. Observe the click count is not changing, but the background color is.

This example demonstrates interacting with invisible hookshot content. The invisible content is overlayed on top of the visible content. The visible content is no longer interactable. Any attempt to interact with it will interact with the invisible hookshot content.

This issue came up with the Select component. Opening the terra-form-select would occasionally cause inconsistent results with the hover interactions of the dropdown menu. Opening the Select dropdown would appear to cause a hover interaction with the list of options. This was happening because the the invisible dropdown content is still interactable and would invoke a mouse enter event on the non-enabled dropdown.
 
## Expected Behavior

Invisible content should not be mouse/keyboard interactable.

## Possible Solution

Set the default style of the hookshot content to `pointer-events: none` when invisible. Note: This is not fully supported in IE 10. An overlay or different combination of CSS may be necessary.

lokesh-0813:
### Summary
Terra-form-select hover styles are inconsistent when dropdown is continuously closed/opened by clicking above the select component when open.

The issue is consistent generally if the select component is at the top portion of the window. Issue arises because of the nature of hookshot content positioned at the top left of the window when its isEnabled prop is set to false.

Changes have been addressed in all the variants.

Closes #2945 

### Deployment Link

https://terra-core-deployed-pr-2946.herokuapp.com/home
### Testing
Wdio tests have been added for all variants and snapshots generated.

### Additional Details

@cerner/terra


packages/terra-form-select/src/single/Frame.jsx
```diff
@@ -265,7 +280,7 @@ class Frame extends React.Component {
           _onBlur();
         }
       }, 10);
-    // Modern browsers support event.relatedTarget
+      // Modern browsers support event.relatedTarget
```
supreethmr, line 283: nit : extra-space

packages/terra-form-select/src/tag/Frame.jsx
```diff
@@ -289,7 +299,7 @@ class Frame extends React.Component {
      */
     if (event && event.target
       && (event.target.hasAttribute('data-terra-form-select-toggle-button')
-      || event.target.hasAttribute('data-terra-form-select-toggle-button-icon'))) {
+        || event.target.hasAttribute('data-terra-form-select-toggle-button-icon'))) {
```
supreethmr, line 302: nit : extra-space

packages/terra-form-select/src/tag/Frame.jsx
```diff
@@ -722,21 +740,21 @@ class Frame extends React.Component {
...
-            isEnabled={this.state.isPositioned}
-            onResize={this.positionDropdown}
-            refCallback={(ref) => { this.dropdown = ref; }}
-            style={FrameUtil.dropdownStyle(dropdownAttrs, this.state)} // eslint-disable-line react/forbid-component-props
-          >
-            <Menu {...menuProps}>
-              {children}
-            </Menu>
-          </Dropdown>
+            <Dropdown
```
supreethmr, line 743: nit: extra-space

packages/terra-form-select/tests/wdio/select-spec.js
```diff
@@ -719,6 +719,22 @@ Terra.describeViewports('Select', ['tiny'], () => {
       });
 
       Terra.it.validatesElement('selected-option');
+
+      after(() => {
+        browser.refresh(); // remove selected option
+      });
+    });
+
+    describe('combobox should open close and open dropdown by clicking root and hover first option', () => {
```
supreethmr, line 728: nit: should hover on first option on reopen after closed by clicking on root. 
OR 
Something more appropriate

lokesh-0813: updated here b89c45958fa2c8099a9dd95b3e37a52e61c74105

packages/terra-form-select/src/single/Frame.jsx
lokesh-0813: I think they were generated after auto-fix, linter didn't complain though. So I think it's fine.

packages/terra-form-select/src/tag/Frame.jsx
lokesh-0813: I think they were generated after auto-fix, linter didn't complain though. So I think its fine.

packages/terra-form-select/src/tag/Frame.jsx
lokesh-0813: I think they were generated after auto-fix, linter didn't complain though. So I think its fine.

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -308,9 +318,17 @@ class Frame extends React.Component {
       return;
     }
 
+    const updateDropdownAttributes = () => {
+      if (this.state.isPositioned) {
+        this.dropdown.removeAttribute('inert');
+        this.dropdown.removeAttribute('aria-hidden');
```
ShettyAkarsh, line 324: why this is being removed here ?

ShettyAkarsh:
@lokesh-0813 , can we update/add the test case for default selected item also for one of the variant ?.  to make sure hover styles are not present on any of the items in dropdown expect the selected one.

packages/terra-form-select/src/tag/Frame.jsx
supreethmr: I don't think changing the existing formatting unless it is required is not recommended either 🤔🧐😄

StephenEsser: As long as these changes are from the auto-formatting feature of VS Code using the terra configurations I think these changes are okay.

Any time I open these specific terra-form-select files my VSCode formatter reformats these code blocks. I have to turn off the plugin or manually unstage the changes block by block to prevent auto format updates.

supreethmr: Good to know that. 🙂

packages/terra-form-select/src/combobox/Frame.jsx
lokesh-0813: We are adding the `inert` properties while mounting, all this while the dropdown will be in-accessible. Removing the inert properties will grant access to the dropdown.

If you are asking related to the `aria-hidden` property, aria-hidden will be added as a by-product when we include `inert` so it has to be removed while removing `inert` as well.

ShettyAkarsh: Yeah i was asking related to aria-hidden.  thanks for the details

lokesh-0813:
> @lokesh-0813 , can we update/add the test case for default selected item also for one of the variant ?. to make sure hover styles are not present on any of the items in dropdown.

Added here ffd3bf24987392f0dd84cab37eec581d0c4840b0

PR #2955, opened 2020-05-06
[terra-action-footer] Add theme-context

pranav300:
### Summary

This pr updates terra-action-footer to use the new theme-context to apply themes.

@cerner/terra


packages/terra-action-footer/src/clinical-lowlight-theme/BlockActionFooter.module.scss
```diff
@@ -0,0 +1,6 @@
+:local {
+  .clinical-lowlight-theme {
```
emilyrohrbough, line 2: Did we want to back-fill all missing css vars with these changes?

packages/terra-action-footer/package.json
```diff
@@ -23,7 +23,8 @@
   "homepage": "https://github.com/cerner/terra-core#readme",
   "peerDependencies": {
     "react": "^16.8.5",
-    "react-dom": "^16.8.5"
+    "react-dom": "^16.8.5",
+    "terra-theme-context": "^1.0.0"
```
emilyrohrbough, line 27: Shouldn't this be a dependency?

pranav300: Updated here https://github.com/cerner/terra-core/pull/2955/commits/6146fb9e5371674f76085d02d1f01acd52f3d3b4

packages/terra-action-footer/src/clinical-lowlight-theme/BlockActionFooter.module.scss
pranav300: Updated here https://github.com/cerner/terra-core/pull/2955/commits/6146fb9e5371674f76085d02d1f01acd52f3d3b4

pranav300:
> What about Standard Action Footer and Centered Action Footer?

Standard and Centered Action Footer are just a wrapper around Block Action Footer Wrapper with no themeable vars. So, even if we provide theme-context to them, it will always be the same and get overridden by blockActionFooter.

ShettyAkarsh:
@pranav300 , add a jest test to check whether the theme classnames are applied properly.

PR #2956, opened 2020-05-07
[terra-avatar] Add theme-context

pranav300:
### Summary

This pr updates terra-avatar to use the new theme-context to apply themes.

@cerner/terra


packages/terra-avatar/src/clinical-lowlight-theme/Avatar.module.scss
```diff
@@ -0,0 +1,42 @@
+:local {
+  .clinical-lowlight-theme {
```
emilyrohrbough, line 2: I did control + f  for `--terra` on `packages/terra-avatar/src/common/Avatar.module.scss` and on this file. I think we are missing two vars in both lowlight & fusion theme

packages/terra-avatar/src/variants/Avatar.jsx
```diff
@@ -100,14 +104,17 @@ class Avatar extends React.Component {
...
-      'avatar',
-      setColor(alt, color, hashValue),
-      { 'fallback-icon': this.state.fallback },
-      { image: Boolean(image) },
-      { 'is-deceased': isDeceased },
+    const avatarClassNames = classNames(
+      cx([
+        'avatar',
+        setColor(alt, color, hashValue),
+        { 'fallback-icon': this.state.fallback },
```
supreethmr, line 111: Do we still need `'fallback-icon'` in Avatar..? Since with latest changes default fallback will be initials.
cc : @yuderekyu

packages/terra-avatar/src/variants/Avatar.jsx
```diff
@@ -121,5 +128,6 @@ class Avatar extends React.Component {
 
 Avatar.propTypes = propTypes;
 Avatar.defaultProps = defaultProps;
+Avatar.contextType = ThemeContext;
```
emilyrohrbough, line 131: Do we want to add the context like this?

tbiethman: It's the only way to access the new context API with class components aside from using an HOC. It should be fine, up until we need to consume more that one context value here. Then we might want to look into moving this to a hooks-based implementation.

(This will require the jest tests to fully mount the component to populate the context value. Not ideal, but given Avatar's limited depth, shouldn't be a huge deal.)

packages/terra-avatar/src/variants/Avatar.jsx
yuderekyu: this class stops the color variant's box-shadow and background-color from appearing with the image. we could re-name this class to `error-fallback` to sync with the latest fallback changes?

Screen Shot 2020-05-14 at 9 37 54 AM

PR #2961, opened 2020-05-11
Set placeholders opacity to 1 to make firefox color-contrast consistent with other browsers

benbcai:
### Summary
Firefox adds a lower opacity to the placeholders in terra-form-input, terra-form-select and terra-form-textarea which causes tests to fail in Firefox due to insufficient color-contrast.

See https://www.w3schools.com/howto/howto_css_placeholder.asp

### Deployment Link
https://terra-core-deployed-pr-2961.herokuapp.com/tests/terra-form-input/form-input/placeholder-form-input

### Additional Details

- Left: Firefox without change (insufficient color-contrast).
- Middle: Firefox with change.
- Right: Chrome with or without change (no effect).

Screen Shot 2020-05-12 at 9 57 26 AM

@cerner/terra


neilpfeiffer:
Since opacity is already handled with the RGBA color, should be no issue adding `opacity:1`  

@jeremyfuksa if you want to investigate more and/or approve.

jeremyfuksa:
@benbcai you can also target Firefox through the `::-moz-placeholder` vendor prefix: https://css-tricks.com/almanac/selectors/p/placeholder/

benbcai:
@jeremyfuksa Is it preferred to use `::-moz-placeholder` over the current change? If so, I can make the change. Otherwise, I'll keep what I currently have.

jeremyfuksa:
> @jeremyfuksa Is it preferred to use `::-moz-placeholder` over the current change? If so, I can make the change. Otherwise, I'll keep what I currently have.

I re-read your original post and noticed that the change didn't affect Chrome, so this is fine. Don't worry about the change.

PR #2969, opened 2020-05-14
[terra-dropdown-button] Add theme-context

pranav300:
### Summary

This pr updates terra-dropdown-button to use the new theme-context to apply themes.

@cerner/terra


emilyrohrbough:
Did we need to update SplitButton.jsx as well? 

https://github.com/cerner/terra-core/blob/master/packages/terra-dropdown-button/src/SplitButton.jsx#L206

I noticed there are 13 theme vars in https://github.com/cerner/terra-core/blob/master/packages/terra-dropdown-button/src/SplitButton.module.scss that did not get added to orion-fusion-theme/SplitButtion.module.scss or to cerner-lowlight-theme/SplitButton.module.scss

packages/terra-dropdown-button/src/clinical-lowlight-theme/DropdownButtonBase.module.scss
```diff
@@ -0,0 +1,116 @@
+@import '~terra-mixins/lib/Mixins';
+
+:local {
```
emilyrohrbough, line 3: DropdownButtonBase only has 5 theme vars: https://github.com/cerner/terra-core/blob/master/packages/terra-dropdown-button/src/_DropdownButtonBase.module.scss

Did you add these in because of the mixin/vars in this package? Would it make sense to include theme in their respective SplitButton & Dropdown button files?

pranav300: updated here https://github.com/cerner/terra-core/pull/2969/commits/31134b68f70230a96a41562108620fcdc7807c02.

PR #2971, opened 2020-05-15
[terra-form-checkbox] Add theme-context

pranav300:
### Summary

This pr updates terra-form-checkbox to use the new theme-context to apply themes.

@cerner/terra


packages/terra-form-checkbox/src/Checkbox.module.scss
```diff
@@ -1,3 +1,9 @@
+@import '~terra-mixins/lib/Mixins';
```
mjhenkes, line 1: do we need to add mixins here?

mjhenkes: I think i see what you're doing, I think it makes more sense to add the mixins include to each  of the theme scss files because thats where the mixin is used.

emilyrohrbough:
Looks good to me after the mixin import change Matt suggested. 👍

packages/terra-form-checkbox/src/Checkbox.module.scss
pranav300: Update here https://github.com/cerner/terra-core/pull/2971/commits/1502a47693f807fab75f3bb20f9dc7048900d901.

PR #2978, opened 2020-05-19
[terra-form-textarea] Add theme context.

pranav300:
### Summary

This pr updates terra-form-textarea to use the new theme-context to apply themes.

@cerner/terra


packages/terra-form-textarea/src/orion-fusion-theme/Textarea.module.scss
```diff
@@ -0,0 +1,98 @@
...
+    --terra-form-textarea-disabled-color: #9b9fa1;
+    --terra-form-textarea-disabled-opacity: 1;
+
+    // Invalid
+    --terra-form-textarea-error-background-image: none;
+    --terra-form-textarea-error-background-color: #fff;
+    --terra-form-textarea-error-border-color: #dedfe0;
+    --terra-form-textarea-error-border-style: solid;
+    --terra-form-textarea-error-border-width: 1px;
+    --terra-form-textarea-error-box-shadow: none;
```
emilyrohrbough, line 62: It looks like this theme var is used in two places: https://github.com/cerner/terra-core/blob/master/packages/terra-form-textarea/src/Textarea.module.scss#L126

Do you know which other theme var is being used is 2 places? I'm noticing 75 theme vars in theTextarea.module.scss file, but only 73 theme vars in this theme file

pranav300: The other one is `--terra-form-textarea-focus-opacity`.

emilyrohrbough: tricky tricky 😄

packages/terra-form-textarea/src/Textarea.jsx
```diff
@@ -210,17 +212,22 @@ class Textarea extends React.Component {
       ...customProps
     } = this.props;
 
+    const theme = this.context;
```
supreethmr, line 215: shouldn't these be React.useContext(ThemeContext); ...?

pranav300: For the class Component we can use `class.contextType` to set ThemeContext to variable `this.context`.

You can check it out over here https://reactjs.org/docs/context.html#classcontexttype

PR #2982, opened 2020-05-21
[terra-breakpoints] Handle mobile device orientation change

### LINKED ISSUE #2911: [terra-breakpoint] breakpoint not set when device orientation changes
# Bug Report

## Description

We found this [issue](https://jira2.cerner.com/browse/PCT-6936) where the application navigation bar doesn't change between tab and menu button when rotating iPad from landscape to portrait or vice versa. I noticed the resize event [here](https://github.com/cerner/terra-core/blob/master/packages/terra-breakpoints/src/ActiveBreakpointProvider.jsx#L32) is not triggered when changing the orientation, which explains why the layout is not updated.

## Steps to Reproduce

1. Load an iOS app, such as PCT, to display a view that should flex when viewport change
2. Change the orientation of the device and observe the layout not changing.

Please refer to the screenshots in https://jira2.cerner.com/browse/PCT-6936

## Expected Behavior

Breakpoint should be set when device orientation is changed.

## Possible Solution

Based on https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/Managing_screen_orientation, it may be resolved by listening to orientationchange event, or handled by css.

## Environment

* Component Name and Version: terra-breakpoint
* Browser Name and Version: AppleWebKit/537.36
* Node/npm Version: [e.g. Node 8/npm 5] Node 10 npm 6
* Webpack Version: 4
* Operating System and version (desktop or mobile): iOS 13

benbcai:
### Summary
Handle the `orientationchange` event on mobile device rotation since the `resize` event is not triggered.

Closes #2911

### Deployment Link

### Testing
Tested locally on mobile simulator. See [screenshots](https://github.com/cerner/terra-core/issues/2911#issuecomment-631843754). There is no test updates because this can only be tested on a mobile device with rotation.

@cerner/terra


packages/terra-breakpoints/src/ActiveBreakpointProvider.jsx
```diff
@@ -23,10 +23,14 @@ class ActiveBreakpointProvider extends React.Component {
 
   componentDidMount() {
     window.addEventListener('resize', this.setActiveBreakpoint);
+
+    // This event is needed to handle mobile device rotation since the resize event is not triggered.
+    window.addEventListener('orientationchange', this.setActiveBreakpoint);
```
tbiethman, line 28: In case anyone was wondering, and I'm not sure if this is even possible, but if both handlers were to execute, setActiveBreakpoint will consolidate the calls into a single update.

benbcai: While debugging I did see that `resize` is called if and after `orientationchange` is called so it's great that they are consolidated.

PR #2997, opened 2020-05-29
[terra-profile-image] Add theme context

pranav300:
### Summary

This pr updates terra-profile-image to use the new theme-context to apply themes.

@cerner/terra


packages/terra-profile-image/src/ProfileImage.module.scss
```diff
@@ -1,4 +1,6 @@
-@import '~terra-mixins/lib/Mixins';
```
emilyrohrbough, line 1: Do we need this import for inline-svg in this file?

pranav300: Added here https://github.com/cerner/terra-core/pull/2997/commits/4067560a79a06a40a6588dcc2dd27bf85f1a2214.

PR #3001, opened 2020-06-01
[terra-search-field] Add theme context

pranav300:
### Summary

This pr updates terra-search-field to use the new theme-context to apply themes.

@cerner/terra


packages/terra-search-field/src/SearchField.jsx
```diff
@@ -231,11 +233,16 @@ class SearchField extends React.Component {
       ...customProps
     } = this.props;
 
-    const searchFieldClassNames = cx([
-      'search-field',
-      { block: isBlock },
+    const theme = this.context;
```
emilyrohrbough, line 236: Could we be using the themeContext hook?

emilyrohrbough: Synced with @pranav300.  didn't realize this wasn't a functional component when looking at the few lines that changed so we can't use the useContext hook here.

PR #3011, opened 2020-06-03
[terra-action-footer] Fix text in footer not wrapping in ie 10

### LINKED ISSUE #3009: [terra-action-footer] Text in footer doesn't wrap in ie 10
# Bug Report

## Description

When using the terra `Text` component inside the terra `ActionFooter` component, the text doesn't wrap properly in internet explorer 10. Issue still occurs when `isWordWrapped` prop on `Text` component is set to true.

## Steps to Reproduce

1. Create `ActionFooter` component with `Text` component for the `end` prop. 
```jsx
<ActionFooter
    end={(
      <Text
        isWordWrapped
      >
        {'This is some very long text. This is some very long text. This is some very long text. This is some very long text.'}
      </Text>
    )}
/>
```
2. Load footer in ie 10.
3. Shrink screen enough that text should wrap.
4. Observe text does not wrap and continues off screen, creating a horizontal scroll bar.

From [this stack overflow post](https://stackoverflow.com/questions/16808815/ie10-flexbox-p-element-non-wrapping), issue appears to be how IE handles word wrapping for flex boxes. And the terra action footer does use flexbox. The issue isn't observed when only using `Text` component outside `ActionFooter`.

Screenshot of issue reproduced with terra dev site:
ie-issue

## Expected Behavior

Text should wrap properly, since `Text` component is being used with `isWordWrapped` prop set to true.

## Possible Solution

From above stack overflow post, solution is to add css rule `flex: 0 1 auto` to the element wrapping the text element, which would be the action footer "start-actions" and "end-actions" css classes [here](https://github.com/cerner/terra-core/blob/7058b54cf297f2534a3163c643477388fe66a1e6/packages/terra-action-footer/src/ActionFooter.module.scss#L19). When testing this I confirmed it fixes the issue. Screenshot:
ie-issue-fixed

## Environment

* Component Name and Version: Reproduced on latest version of terra master branch
  * terra-text: v4.30.0
  * terra-action-footer: v2.44.0
* Browser Name and Version: IE 10
* Node/npm Version: Node v12.16.3. NPM v6.14.4
* Webpack Version: v4.43.0
* Operating System and version (desktop or mobile): Windows 10 VM. Desktop.

JacksonJN:
### Summary

Fixes issue of text in action footer not wrapping in ie 10. Occurs when using the terra `Text` component inside terra `ActionFooter` component, even when setting `isWordWrapping` prop to true of the `Text` component.

Issue is due to bug in IE with word wrapping inside a flexbox, which action-footer uses. 

Closes #3009 

### Deployment Link

### Testing

Ran linter and jest tests locally. Manually tested in local build of terra dev site in various browsers and resolutions.

### Additional Details

More details in issue logged #3009.

@cerner/terra

### DIFF
--- a/packages/terra-action-footer/src/ActionFooter.module.scss
+++ b/packages/terra-action-footer/src/ActionFooter.module.scss
@@ -17,5 +17,6 @@
   .start-actions,
   .end-actions {
     align-self: center;
+    flex: 0 1 auto; //Needed for IE 10 to wrap things properly
   }
 }
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -85,6 +85,7 @@ Cerner Corporation
 - Saket Bajaj [@saket2403]
 - Prachi Gotkhindikar [@prachigotkhindikar1]
 - Jeff Merten [@jeffmerten]
+- Jackson Nowotny [@JacksonJN]
 
 Community
 
@@ -196,3 +197,4 @@ Community
 [@saket2403]:https://github.com/saket2403
 [@vobango]:https://github.com/vobango
 [@jeffmerten]: https://github.com/jeffmerten
+[@JacksonJN]: https://github.com/JacksonJN
--- a/packages/terra-action-footer/CHANGELOG.md
+++ b/packages/terra-action-footer/CHANGELOG.md
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Fixed
+* Text in footer not wrapping in IE 10
 
 2.44.0 - (June 2, 2020)
 ------------------


packages/terra-action-footer/src/ActionFooter.module.scss
```diff
@@ -17,5 +17,6 @@
   .start-actions,
   .end-actions {
     align-self: center;
+    flex: 0 1 auto; //Needed for IE 10 to wrap things properly
```
supreethmr, line 20: This can also be just `flex: 1` to make all the flexible items to be the same length, regardless of its content. 
reference : https://www.w3schools.com/cssref/css3_pr_flex.asp

JacksonJN: I tried testing this, and it changed the layout of the end section of the action footer. It aligns the end content to the middle of the footer in both chrome and IE, as opposed to right aligning it.

I found some sources [1][2] which suggest that `flex: 1` is the same as `flex: 1 1 0`. Which is what I am seeing in chrome:
Screen Shot 2020-06-04 at 1 11 24 PM

[1]https://stackoverflow.com/questions/37386244/what-does-flex-1-mean
[2]https://www.w3.org/TR/css-flexbox-1/#flex-common

jvarness:
@mjhenkes @supreethmr are we good with this PR? I'm assuming the comments here can be addressed with [this comment](https://github.com/cerner/terra-core/issues/3009#issuecomment-640904254) and we can merge this for the next release.

packages/terra-action-footer/src/ActionFooter.module.scss
emilyrohrbough, line 20:
```suggestion
    flex: 0 1 auto; // Needed for IE 10 to wrap things properly
```

JacksonJN:
Can this be merged now?

mjhenkes:
@JacksonJN, We need to have our UX designers take a look before we merge.

packages/terra-action-footer/src/ActionFooter.module.scss
neilpfeiffer: @supreethmr -- setting this to `flex: 1` would actually be counter-productive since it changes the component _away_ from modern browser default/initial behavior, which is what we want to set explicitly for IE10. As @JacksonJN mentions, `flex: 1` is the same as `flex: <positive-number>` which equates to `flex: <positive-number> 1 0`.

Specifying the full initial value: `flex: 0 1 auto` is the best fix to "reset" IE10 to shim it to match the default `flex` behavior in all the other supported browsers. For more reasons why - read up on the history of IE10 implementing a very early draft spec for Flexbox in 2012 https://github.com/cerner/terra-core/issues/3009#issuecomment-640904254

neilpfeiffer:
UX +1 |
-----------|

PR #3014, opened 2020-06-04
[terra-toolbar] Add theme-context

pranav300:
### Summary

This pr updates terra-toolbar to use the new theme-context to apply themes.
Apart of #3017 

@cerner/terra


packages/terra-toolbar/src/Toolbar.jsx
```diff
@@ -23,7 +25,16 @@ const defaultProps = {
...
-  let toolbarClassNames = cx('toolbar', `${align}-align`);
+  const theme = React.useContext(ThemeContext);
+
+  let toolbarClassNames = classNames(
+    cx(
+      'toolbar',
+      `${align}-align`,
+      theme.className,
+    ),
+    customProps.className,
```
emilyrohrbough, line 36: CustomProps wasn't pass previously. Not sure we should add it now.

pranav300: CustomProps were being passed before but something like [this](https://github.com/cerner/terra-core/pull/3014/files#diff-ff210e5dbf27c1d1bc3b7bcee1f88c3dL27) 

I did miss to make toolbarClassName to be const and forgotten to remove its reassignment. Updated in this commit https://github.com/cerner/terra-core/pull/3014/commits/5095b7040c6c0b0fad6c5126cf956d2de734402c

StephenEsser:
Adding `resolves` into the PR summary prior to merging will auto close the [root issue](https://github.com/cerner/terra-core/issues/3017). To keep from closing/reopening each time a theme context PR is merged choose a word not on this list to use in the summary:

https://help.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

PR #3020, opened 2020-06-05
Update eslint config terra dependency

### LINKED ISSUE #2999: Update eslint-config-terra v3.0.0 to @cerner/eslint-config-terra v^4.0.0
# Feature Request

## Description
We recently updated our eslint config & released it as a scoped package. We should move to the latest version.

DMcginn:
### Summary
Updated the eslint-config-terra dependency from v^3.0.0 to @cerner/eslint-config-terra v^4.0.0, along with the associated config. Additionally updated multiple files to fit eslint rules and to ensure that the lint tests pass.

Closes #2999 

@cerner/terra


package.json
```diff
@@ -25,7 +25,7 @@
     "extends browserslist-config-terra"
   ],
   "eslintConfig": {
-    "extends": "terra",
+    "extends": "@cerner/terra",
     "overrides": [
```
emilyrohrbough, line 29: Just released this when looking at something else, but can't these overrides be removed?
https://github.com/cerner/terra-toolkit/blob/master/packages/eslint-config-terra/eslint.config.js#L81-L87

emilyrohrbough:
@DMcginn sorry 😅

packages/terra-base/src/terra-dev-site/doc/base/GlobalBaseStylesExample.jsx
```diff
@@ -14,7 +14,7 @@ const BaseExample = () => (
     <ol>
       <li>Ordered list item</li>
       <li>
-Ordered list item
+        Ordered list item
```
mjhenkes, line 17: Does JSX now ignore these tabs or something? I'm pretty sure all the lack of tabs was to prevent tabs from showing up on the dom. Is that not the case anymore?

DMcginn: Screen Shot 2020-06-08 at 11 18 14 AM

The list looks like this with and without the tabs in the JSX file, so I'm assuming it just ignores the tabs now since there's no change?

package.json
DMcginn: Do you mean the eslint overrides in both terra-toolkit and terra-core or just terra-toolkit?

PR #3021, opened 2020-06-05
Condense wdio config

emilyrohrbough:
### Summary

Mimiks changes made in:
- https://github.com/cerner/terra-clinical/pull/642
- https://github.com/cerner/terra-framework/pull/1094

There are a lot of wdio configuration in this project. It was a little confusing to me to keep track of which config was being used or not. I took the existing package-level wdio logic and pushed to a root wdio level.

I also add npm run test to the package-level test scripts.

### Testing
running tests locally + travis builds

### Additional Details

@cerner/terra


.npmrc
```diff
@@ -1 +1,2 @@
 package-lock=false
+node-options=--max_old_space_size=2048
```
mjhenkes, line 2: add a line break

packages/terra-action-footer/package.json
```diff
@@ -39,11 +39,11 @@
...
     "precompile": "rm -rf lib",
-    "test": "npm run jest && npm run wdio",
+    "test": "npm run lint && npm run jest && npm run wdio",
     "jest": "jest --config ../../jestconfig.js",
-    "wdio-default": "cd ../.. && npx wdio ./packages/$npm_package_name/wdio.conf.js",
-    "wdio-lowlight": "cd ../.. && THEME=clinical-lowlight-theme npx wdio ./packages/$npm_package_name/clinical-lowlight-wdio.conf.js",
-    "wdio-fusion": "cd ../.. && THEME=orion-fusion-theme npx wdio ./packages/$npm_package_name/wdio.conf.js",
+    "wdio-default": "cd ../.. && npx wdio",
+    "wdio-lowlight": "cd ../.. && THEME=clinical-lowlight-theme npx wdio",
+    "wdio-fusion": "cd ../.. && THEME=orion-fusion-theme npx wdio",
```
StephenEsser, line 46: Any interest in exposing bin commands from root or a similar concept to simplify the maintenance/uniformity of these packages?

tc = terra-core
```json
    "test": "tc-test",
    "jest": "tc-jest",
    "wdio-default": "tc-wdio-default",
    "wdio-lowlight": "tc-wdio-lowlight",
    "wdio-fusion": "tc-wdio-fusion",
```

emilyrohrbough: I think this would go into the bucket of scripts we could add if we created a terra-scripts package. running wdio, jest, the changelog update script, etc. are scripts we've been duplicating across terra-core, framework and clinical.

If you have a strong opinion to do it now, I'd be more than happy to make the change.

mjhenkes: If the scripts package contains scripts like this we should be clear that it's for mono-repo/internal use only and not intended for end users.

.npmrc
mjhenkes, line 2: do we still need this?

packages/terra-action-footer/package.json
emilyrohrbough: 100% agreed. It'd be for ourselves.

.npmrc
emilyrohrbough: I assume so? I remove for these scripts: https://github.com/cerner/terra-core/pull/3021/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2L81-L83

packages/terra-action-footer/package.json
StephenEsser: I don't have a strong opinion about making this change. There are advantages and disadvantages in obscuring the testing scripts. If we're interested in the idea we can talk more about the terra-script package in more detail down the road. 

We don't make changes to the scripts/commands too often, but when we do we make updates to all packages, it seems like a hassle to get right/write a script to make the updates to each package individually.

A little wordy, but the mono repo specific scripts could be namespaced. (If we wanted commands for non-mono and mono)

```
tc-mono-wdio-default

tcm-wdio-default
```

emilyrohrbough: I don't think it's a bad idea. Let's keep it in mind and we can do that work on a new PR. I'll merge this for now & we can solidify what we thinking might be best. I agree, it's a little annoying to update 15+ package.json for a simple script

PR #3022, opened 2020-06-07
[terra-html-table] Add theme context

### LINKED ISSUE #3017: [terra-core] Add theme-context and directly include packaged themes
# Feature Request

## Description

Add `terra-theme-context` and directly include packaged themes for each component

pranav300:
### Summary

This pr updates terra-cell-grid to use the new theme-context to apply themes.
Added clinical-lowlight theme file.
partially resolves #3017

@cerner/terra


packages/terra-html-table/src/Table.module.scss
```diff
@@ -94,14 +99,6 @@
...
       }
     }
-
-    tfoot {
-      tr {
-        border-bottom: var(--terra-html-table-footer-row-border-bottom, 1px solid #dedfe0);
-        border-top: var(--terra-html-table-footer-row-border-top, 0);
-        text-align: left;
-      }
-    }
```
StephenEsser, line 104: Why was this is removed?

pranav300: Currently, html-table does not use any className as tfoot. So I thought of removing the extra styles, which are not being consumed.

PR #3030, opened 2020-06-10
[terra-dropdown-button] Fixed dropdown not opening on pressing space in firefox.

### LINKED ISSUE #3008: [terra-dropdown-button] Dropdown Menu does not Open with Space Key in Firefox
# Bug Report

## Description

The dropdown menu does not display when the space key is used during keyboard navigation to open the dropdown.

Discovered when testing against firefox: 
- https://github.com/cerner/terra-core/blob/master/packages/terra-dropdown-button/tests/wdio/dropdown-button-spec.js#L76
- https://github.com/cerner/terra-core/blob/master/packages/terra-dropdown-button/tests/wdio/split-button-spec.js#L99

### Dropdown Button Behavior
1. Go to https://engineering.cerner.com/terra-core/raw/tests/terra-dropdown-button/dropdown-button/callback-dropdown-button
2. Tab to dropdown button
3. Hit space key. Menu does not open. 
4. Repeat in chrome and this works. 

### Split Button Behavior
1. Go to https://engineering.cerner.com/terra-core/raw/tests/terra-dropdown-button/dropdown-button/callback-split-button
2. Tab to split button arrow button
3. Hit space key. Menu does not open. 
4. Repeat in chrome and this works. 

- Dropdown Button: https://share.getcloudapp.com/geuwvnXN
- SplitButton Button: https://share.getcloudapp.com/qGuKjk8q

## Expected Behavior
Dropdown menu should open when the space key is used. This should open the menu for both the dropdown button and split button.

Running the following should pass:
```
BROWSERS=firefox npm run wdio-default -- --spec packages/terra-dropdown-button/tests/wdio/dropdown-button-spec.js
BROWSERS=firefox npm run wdio-default -- --spec packages/terra-dropdown-button/tests/wdio/split-button-spec.js
```

## Environment

* Component Name and Version: terra-dropdown-button v1.14.0
* Browser Name and Version: firefox

saket2403:
### Summary
Fixed issue regarding dropdown not opening on pressing space in firefox browser.
Closes #3008 

### Deployment Link
https://terra-core-deployed-pr-3030.herokuapp.com/components/terra-dropdown-button/dropdown-button/dropdown-button

### Additional Details

@cerner/terra


packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -100,7 +100,7 @@ class DropdownButton extends React.Component {
   handleKeyDown(event) {
     if (event.keyCode === KeyCode.KEY_SPACE || event.keyCode === KeyCode.KEY_RETURN) {
       // In FireFox active styles don't get applied on space
-      this.setState({ isActive: true, openedViaKeyboard: true });
+      this.setState(prevState => ({ isActive: true, openedViaKeyboard: true, isOpen: !prevState.isOpen }));
```
supreethmr, line 103: I think this issue was due to the difference in event handling between browsers. As stated in [comments ](https://github.com/cerner/terra-core/blob/34658ed61ae8f5d0c19c01173e505da855716d6c/packages/terra-dropdown-button/src/DropdownButton.jsx#L104-L109) above `handleKeyDown` method `event.preventDefault()` was added to prevent `onClick` from getting called but it looks like `onClick` is still called in Firefox after keyUp event. May be firefox expects explicit `preventDefault` even in `onKeyUp` event.

Since we have call for `openDropdown()` method in both `onClick` and `onKeyUp`  dropdown is getting  open and close immediately.

I think we can fix this by just moving `openDropDown()` method to `onKeyDown` and placing it right after the [event.preventDefault()](https://github.com/cerner/terra-core/blob/34658ed61ae8f5d0c19c01173e505da855716d6c/packages/terra-dropdown-button/src/DropdownButton.jsx#L110) so that it would make more clear on the purpose of having `openDropdown` method in `onKeyDown` event.

packages/terra-dropdown-button/src/SplitButton.jsx
```diff
@@ -149,7 +149,7 @@ class SplitButton extends React.Component {
   handleCaretKeyDown(event) {
     if (event.keyCode === KeyCode.KEY_SPACE || event.keyCode === KeyCode.KEY_RETURN) {
       // In FireFox active styles don't get applied onKeyDown
-      this.setState({ caretIsActive: true, openedViaKeyboard: true });
+      this.setState(prevState => ({ caretIsActive: true, openedViaKeyboard: true, isOpen: !prevState.isOpen }));
```
supreethmr, line 152: Setting isOpen in keydown will cause dropdown to close immediately as we leave space OR return key after holding it for some time( long press ) . To adrress this issue `openDropdown()` is being handled only in `onKeyUp` event. 

So with this change we can see that dropdown gets open on long press of space OR return key and it closes the dropdown as soon as we leave the key.
dropdownbutton

packages/terra-dropdown-button/src/DropdownButton.jsx
saket2403: Fixed 4170c40. Renamed `openDropDown` to `toggleDropDown`, hope it sounds better.

packages/terra-dropdown-button/src/SplitButton.jsx
saket2403: Removed setState and fixed with preventDefault 4170c40

packages/terra-dropdown-button/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Fixed dropdown not opening on pressing space in firefox.
```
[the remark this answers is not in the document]
saket2403: Yes it works for IE as well.

PaulSolDev:

PR #3037, opened 2020-06-15
Clinical Theme Build Issue Corrections

ShettyAkarsh:
Fixes the clinical theme build issues for Form-select Multiple and Tag variant

@cerner/terra

### DIFF
--- a/packages/terra-form-select/tests/wdio/select-spec.js
+++ b/packages/terra-form-select/tests/wdio/select-spec.js
@@ -1140,6 +1140,10 @@ Terra.describeViewports('Select', ['tiny'], () => {
 
     describe('multiple placeholder should be shown initially', () => {
       Terra.it.validatesElement();
+
+      after(() => {
+        browser.refresh();
+      });
     });
 
     describe('multiple placeholder should disappear after selection and blur', () => {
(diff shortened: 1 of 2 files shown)


packages/terra-form-select/tests/wdio/select-spec.js
```diff
@@ -1140,6 +1140,10 @@ Terra.describeViewports('Select', ['tiny'], () => {
 
     describe('multiple placeholder should be shown initially', () => {
       Terra.it.validatesElement();
+
+      after(() => {
+        browser.refresh();
```
emilyrohrbough, line 1145: These were removed to fix test errors when running in firefox. Does reverting this change break the firefox test runs? https://github.com/cerner/terra-core/commit/837514e05df9f889ed4dcc15b0ef68a06f0e175f#diff-80bc530bcf881e722aea305d312913ddL1143

ShettyAkarsh: yeah code is reverted with browser.refresh(). Tested this in firefox and the test is passing.

ryanthemanuel: i ran our internal nighly jenkins job against this branch on firefox and it passed.  Looks good.

PR #3066, opened 2020-07-06
[terra-form-select] Fix input focus on mobile for tag and multiselect

### LINKED ISSUE #2964: [terra-form-select] On mobile, the Select with combo/multi variant has inconsistent behavior
# Bug Report

## Description

On mobile: A select with a variant of combo or multi has varied behavior around keyboard focus for text entry depending on browser.

**Safari/Edge: expected behavior**

- Select dropdown arrow for the select (dropdown opens)

- Tap on the input (keyboard appears and input gains focus/text entry supported)

**Firefox/Chrome: bug** 
Scenario 1-

- Select dropdown arrow for the select (dropdown opens)

- Tap on the input (no response)

- Select on option in the dropdown (option selected, keyboard appears, and input gains focus/text entry supported)

Scenario 2-

- Select the input for the select (dropdown opens, keyboard appears, input gains focus/text entry supported)

## Environment

* Component Name and Version: Select
* Browser Name and Version: mobile Firefox and Chrome
* Webpack Version:
* Operating System and version (desktop or mobile): iOS 13.4.1

## @ Mentions

@StephenEsser

StephenEsser:
### Summary

This PR fixes an issue with onClick and input focus on mobile devices for the tag and multi-select variants.

The original issue was input focus on mobile. This was caused by the `point-events: none` being set on the div around the input. After removing `point-events` it re-introduced an onClick issue where onClick would not fire on Firefox and Safari. 

onClick would not fire because onMouse Down would reveal an input and onMouseUp would happen on the input instead of the original div. This suppressed the onClick events. This can be reproduced here in Firefox and Safari

https://codesandbox.io/s/happy-snyder-39dso?file=/src/styles.css

The fix is to use onClick instead of onMouseDown to open/close/toggle the dropdown menu. This allows the events to fire as expected prior to the input being rendered overlayed on the original div.

Fixes:

- On mobile the input can be focused when the dropdown is opened via the dropdown arrow
- Click events fire appropriately for all browsers
- On Safari the mobile interactions focus in the control appropriately

Closes #2964

### Deployment Link

https://terra-core-deployed-pr-3066.herokuapp.com/

@cerner/terra


packages/terra-form-select/src/multiple/Frame.jsx
```diff
@@ -585,7 +599,8 @@ class Frame extends React.Component {
         ]);
 
         return (
-          <div data-terra-form-select-toggle className={toggleClasses} onMouseDown={this.handleToggleMouseDown}>
+          // eslint-disable-next-line jsx-a11y/click-events-have-key-events
```
mjhenkes, line 602: Could you give some reasoning here for disabling this?

StephenEsser: The linter wants the div to have all of the key events (space/enter) that a button would have so it can be activated via the keyboard. This element cannot receive focus so the keyboard events wouldn't work either way. This was sneaking around the rule previously by using mouse down instead of click to avoid the rule, but the functionality/result is the same conceptually.

mjhenkes: What would you think about adding that comment to the file?

StephenEsser: Added some comments to the disables:

https://github.com/cerner/terra-core/pull/3066/commits/2c58713b37a0c2b29f07671ff04a1de239e6f471

PaulSolDev:

PR #3070, opened 2020-07-08
Initial terra docs repo

mjhenkes:
### Summary
Initializes the terra-docs repo for housing general use react components for documentation.

This was all generated by an updated generator that I will be putting out a PR for. This is marked to be ignored by the publish command.

A subsequent pull request will update this component and add functionality to it, but i figured it would be easier to see the changes when separated from this generated content.

### Deployment Link

### Testing
Automated

### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-docs/.npmignore
+++ b/packages/terra-docs/.npmignore
@@ -1,0 +1,8 @@
+*.log
+target
+reports
+tests/**/__snapshots__/
+aggregate_translations
+dev-site-config
+screenshots
+errorScreenshots
--- a/packages/terra-docs/LICENSE
+++ b/packages/terra-docs/LICENSE
@@ -1,0 +1,203 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+   2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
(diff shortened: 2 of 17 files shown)


packages/terra-docs/.npmignore
```diff
@@ -0,0 +1,8 @@
+*.log
```
emilyrohrbough, line 1: Should we start adding a "files" key to the package.json to only release docs, lib, and tests?

packages/terra-docs/LICENSE
```diff
@@ -0,0 +1,203 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
emilyrohrbough, line 190: year?

packages/terra-docs/README.md
```diff
@@ -0,0 +1,24 @@
+# Terra Docs
+
+[![NPM version](https://badgen.net/npm/v/terra-docs)](https://www.npmjs.com/package/terra-docs)
+[![Build Status](https://badgen.net/travis/cerner/terra-core)](https://travis-ci.com/cerner/terra-core)
+
+{insert description}
```
emilyrohrbough, line 6: Need description

packages/terra-docs/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "terra-docs",
+  "main": "lib/Docs.js",
+  "version": "0.0.0",
+  "private": true,
+  "description": "{insert description}",
```
emilyrohrbough, line 6: add description

packages/terra-docs/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "terra-docs",
+  "main": "lib/Docs.js",
+  "version": "0.0.0",
+  "private": true,
```
emilyrohrbough, line 5: false?

packages/terra-docs/src/Docs.jsx
```diff
@@ -0,0 +1,31 @@
...
+
+const propTypes = {
+  /**
+   * Content to be displayed as the name
+   */
+  name: PropTypes.string,
+};
+
+const defaultProps = {
+  name: 'default',
```
emilyrohrbough, line 16: Based on PR description, Is this just a template at this point?

packages/terra-docs/.npmignore
mjhenkes: we need to release src too for code examples

packages/terra-docs/README.md
mjhenkes: coming soon, when content is actually added.

packages/terra-docs/src/Docs.jsx
mjhenkes: yes

packages/terra-docs/package.json
mjhenkes: coming soon, when content is actually added.

packages/terra-docs/LICENSE
mjhenkes: I'll update the generator

mjhenkes:
Also are we 100% sure on the name terra-docs?

emilyrohrbough:
> Also are we 100% sure on the name terra-docs?

🤷‍♀️  but changes look good otherwise

packages/terra-docs/.npmignore
emilyrohrbough: right. Still might be less to configure than adding all generated files & config files to npmignore

packages/terra-docs/tests/wdio/docs-spec.js
```diff
@@ -0,0 +1,9 @@
+// Terra.describeViewports('Docs', ['tiny', 'medium', 'large'], () => {
```
mjhenkes, line 1: The tests are commented out because it seems that by setting the package to private, the package is ignored when compiling to lib so it's not in the testing site.

packages/terra-docs/package.json
ryanthemanuel, line 5: We will want to do something like this to prevent this from attempting a release in lerna: https://github.com/cerner/terra-toolkit/blob/main/lerna.json#L5-L7

mjhenkes: afik setting this to private will already prevent the lerna release

mjhenkes: I'm right:
>Lerna will never publish packages which are marked as private ("private": true in the package.json).

https://github.com/lerna/lerna/tree/master/commands/publish#usage

packages/terra-docs/.npmignore
StephenEsser: The files key can be a little dangerous too, there's been a few situations where a new file is created and is ignored during the publish that crashes the package. These situations are rare, but I don't know of a good way to detect them automatically.

packages/terra-docs/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright <%= currentYear %> Cerner Innovation, Inc.
```
StephenEsser, line 190: Year wasn't subbed in here

packages/terra-docs/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+- [Getting Started](#getting-started)
+- [Documentation](https://github.com/cerner/terra-core/tree/main/packages/terra-docs/docs)
```
StephenEsser, line 9: Starting a petition to redo all the branch renames `main` -> `base`... just kidding.

packages/terra-docs/src/Docs.jsx
```diff
@@ -0,0 +1,31 @@
...
+};
+
+const Docs = ({ name, ...customProps }) => {
+  const DocsClassNames = cx([
+    'docs',
+    customProps.className,
```
StephenEsser, line 22: I can log this against the generator, but I wanted to point out the className here is bound. The generator might need a few updates to handle this if we want to be more cognizant of it when generating components

packages/terra-docs/LICENSE
mjhenkes: whoops

packages/terra-docs/src/Docs.jsx
mjhenkes: I have a generator PR i could make some minor updates to

mjhenkes: https://github.com/cerner/generator-terra-module/pull/147

StephenEsser: I logged an issue and commented on the PR incase we decide to handle them separately:

https://github.com/cerner/generator-terra-module/issues/148

packages/terra-docs/package.json
mjhenkes: I added the change to the command... it seems like lerna still tries to version the package even if it's not releasing?

PR #3072, opened 2020-07-09
Travis ci without docker

mjhenkes:
### Summary
Testing to see if we could remove docker from travis-ci and if that would give us any performance gains.

Closes #

### Deployment Link

### Additional Details
The initial terra docs repo PR:
Screen Shot 2020-07-12 at 10 49 14 AM

This PR:
Screen Shot 2020-07-12 at 10 48 22 AM

This PR: grouping like times:
Screen Shot 2020-07-13 at 10 32 47 AM

Compare the `ran for` times, total time indicates waiting in executors (i think)

@cerner/terra


.travis.yml
```diff
@@ -7,106 +7,157 @@ branches:
...
-    - stage: compile - default theme
-      script: travis_wait docker-compose run test-ci npm run compile:prod -- --env.disableAggregateThemes
+    - stage: danger lint jest and compile
+      script:
+        # clear the cache of any old build artifacts
+        - rm -rf ./travis-build
+        - npm run danger
+        - npm run lint
+        - npm run jest
+        # These can't be built in parallel because the cache is a single object and you can't add to it in parallel
```
ryanthemanuel, line 23: Can we run all 3 of these in parallel with danger lint and jest?

wdio.conf.js
```diff
@@ -8,4 +8,12 @@ if (process.env.npm_package_name !== 'terra-core') {
   wdioConfig.specs = [path.join(__dirname, 'packages', process.env.npm_package_name, 'tests', 'wdio', '**', '*-spec.js')];
 }
 
+const ci = process.env.CI;
+
+if (ci) {
+  wdioConfig.host = 'localhost';
+}
+
+// console.log(wdioConfig);
```
ryanthemanuel, line 17: Remove commented console log.

.travis.yml
ryanthemanuel: Never mind. I think I understand your comment now.

mjhenkes: Yeah at one point i had all three compile jobs running in parallel but only the last to complete would be cached.

We could get a couple of minutes back by deprecating terra-markdown and terra-doc-template (the only remaining   components that use theme aggregation)

wdio.conf.js
```diff
@@ -8,4 +8,10 @@ if (process.env.npm_package_name !== 'terra-core') {
   wdioConfig.specs = [path.join(__dirname, 'packages', process.env.npm_package_name, 'tests', 'wdio', '**', '*-spec.js')];
 }
 
+const ci = process.env.CI;
+
+if (ci) {
+  wdioConfig.host = 'localhost';
```
mjhenkes, line 14: This line won't work in the docker dev env.... if you use it and i think it's only me... comment this out to run tests

package.json
```diff
@@ -47,7 +47,7 @@
   },
   "scripts": {
     "bootstrap:hoist": "lerna bootstrap --hoist --nohoist=one-cerner-style-icons",
-    "clean": "lerna clean --yes && rm -rf node_modules",
+    "clean": "lerna clean --yes && rm -rf node_modules || true",
```
mjhenkes, line 50: this was added to allow this check to succeed even if lerna isn't installed.... no need to clean then right?

wdio.conf.js
mjhenkes: i changed this to look for a different env `travis` this will keep docker dev env working

Issue #2992, opened 2020-05-27
[terra-form-input] Insufficient color-contrast for placeholder


benbcai:
# Bug Report

## Description
The placeholder color in terra-form-input and terra-form-textarea is [rgba(25, 32, 35, 0.53)](https://github.com/cerner/terra-core/blob/fac7744f171f4336c980791547a61caa54878962/packages/terra-form-input/src/Input.module.scss#L26). The equivalent hex color of this is `#85898a`. This color on a white background does not meet the minimum color contrast ratio of `4.5:1`. 

The placeholder color in terra-form-select is [rgba(21, 33, 41, 0.61)](https://github.com/cerner/terra-core/blob/fac7744f171f4336c980791547a61caa54878962/packages/terra-form-select/src/shared/_Frame.module.scss#L243). The equivalent hex color of this is `#70777d` and this has a contrast ratio of `4.54 : 1` and is sufficient.

terra-form-input and terra-form-textarea should match terra-form-select to be consistent and meet the contrast ratio. What is unknown is Chrome/FF does not complain about the insufficient color contrast. IE seems to be the only browser able to truly detect the insufficient color contrast.

There are plans in the near future to get rid of placeholders. Until then, we should use a darker placeholder color to provide sufficient color contrast.

## Steps to Reproduce
1. Many online rgba to hex color converters do not take the opacity into account. 
2. Go to https://cssgenerator.org/rgba-and-hex-color-generator.html and enter the rgba values.
3. Take a screenshot of the generated color.
4. Update the screenshot to https://imagecolorpicker.com/en/.
5. The hex color of `#85898a` should be returned.
6. Go to https://dequeuniversity.com/rules/axe/3.5/color-contrast?application=axeAPI and enter `#85898a` as the foreground color and `#ffffff` as the background color.
7. Note that a contrast ratio of `3.53 : 1` does not meet the minimum ratio of `4.5:1`.

## Additional Context / Screenshots

```
1) [enormous] Onset Picker has [date granularity] [default] is accessible and is within the mismatch tolerance:
expected no accessibility violations but got:
	{
  "description": "Ensures the contrast between foreground and background colors meets WCAG 2 AA contrast ratio thresholds",
  "help": "Elements must have sufficient color contrast",
  "helpUrl": "https://dequeuniversity.com/rules/axe/3.5/color-contrast?application=axeAPI",
  "id": "color-contrast",
  "impact": "serious",
  "nodes": [
    {
      "all": [],
      "any": [
[...]

PR #3075, opened 2020-07-12
[terra-form-input][terra-form-textarea] Remove prop placeholder

pranav300:
### Summary

This PR:

- Deprecates prop placeholder from form-input and form-textarea
- Removes/updates tests and examples using prop placeholder
- Removes themeable variables:
** `--terra-form-input-placeholder-color`
** `--terra-form-input-placeholder-font-style`
** `--terra-form-input-placeholder-disabled-color`
** `--terra-form-input-placeholder-disabled-font-style`
** `--terra-form-textarea-placeholder-color`
**  `--terra-form-textarea-placeholder-font-style`
** `--terra-form-textarea-placeholder-disabled-color`
** `--terra-form-textarea-placeholder-disabled-font-style`
- Updates wdio snapshots in search-field for changes made in form-input

Closes #2992 along with #3078

### Deployment Link

https://terra-core-deployed-pr-3075.herokuapp.com/

### Additional Details

Thank you for contributing to Terra.
@cerner/terra

PR #3078, opened 2020-07-15
[terra-search-field] Removed the terra-form-input and replaced it with a native input

pranav300:
### Summary

This PR removes the use of `terra-form-input` from the search-field and replaces it with `native input`.
All the styles necessary for the input are taken directly from `terra-form-input`.

This is done to support placeholder in search-field after PR #3075 has been merged, which removes placeholder from the form-input and form-textarea.

Part of #2992 

### Deployment Link

https://terra-core-search-field-asffbr.herokuapp.com/components/terra-search-field/search-field/search-field

### Additional Details

@cerner/terra


packages/terra-search-field/src/SearchField.jsx
```diff
@@ -6,7 +6,7 @@ import ThemeContext from 'terra-theme-context';
 import Button from 'terra-button';
 import * as KeyCode from 'keycode-js';
 import IconSearch from 'terra-icon/lib/icon/IconSearch';
-import Input from 'terra-form-input';
+// import Input from 'terra-form-input';
```
lokesh-0813, line 9: You forgot to remove this comment.

pranav300: Removed here https://github.com/cerner/terra-core/pull/3078/commits/e87bc5c5b9760c13bda762afd10b3f163229b3f5.

emilyrohrbough:
@pranav300 I'm confused, why do we need to replace terra-form-input with native input to fix the placeholder issue? Don't we need to filter out and remove the placeholder prop from being passed down to the terra-form-input?

pranav300:
> @pranav300 I'm confused, why do we need to replace terra-form-input with native input to fix the placeholder issue? Don't we need to filter out and remove the placeholder prop from being passed down to the terra-form-input?

In search-field, we do want to display placeholder and with the changes done in this PR https://github.com/cerner/terra-core/pull/3075 it won't be possible with the form-input.

This PR is not ready for review, I had raised it so that we can discuss whether this approach is appropriate or not.

emilyrohrbough:
> This PR is not ready for review, I had raised it so that we can validate the approach.

This PR is not ready for review, can we move this to draft mode? and/or add a no not merge label on this PR. There was nothing in the summary to indicate this wan't ready for review.

Since this isn't ready, who is suppose to be validating this? Could that person have looked at the branch diff?

packages/terra-search-field/CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ ChangeLog
 
 Unreleased
 ----------
+### Removed
+* Removed `terra-form-input` from dependencies.
+
+### Changed
+* Updated `terra-form-input` to `native input`.
```
StephenEsser, line 10: I think this entry should be something similar to "Removed the terra-form-input and replaced it with a native input"

packages/terra-search-field/src/SearchField.module.scss
```diff
@@ -103,6 +87,83 @@
...
+      outline: none;
+
+      [dir=ltr] & {
+        box-shadow: var(--terra-search-field-input-focus-ltr-box-shadow);
+      }
+
+      [dir=rtl] & {
+        box-shadow: var(--terra-search-field-input-focus-rtl-box-shadow);
+      }
+    }
```
StephenEsser, line 167: We should reduce these styles/variables as much as possible and apply only the ones necessary for theming the search field appropriately. Are all terra-form-input variables required to theme the search field?

packages/terra-search-field/src/SearchField.module.scss
```diff
@@ -103,6 +87,83 @@
...
+    padding-left: var(--terra-search-field-input-padding-left, 0.28571rem);
+    padding-right: var(--terra-search-field-input-padding-right, 2.28571rem); // Match the padding to width of the clear button
+    padding-top: var(--terra-search-field-input-padding-top, 0.28571rem);
+    transition-duration: var(--terra-search-field-input-transition-duration, 0s);
+    transition-property: background-size;
+    transition-timing-function: var(--terra-search-field-input-transition-timing-function, ease);
+    width: 100%;
+
+    &::placeholder {
+      color: var(--terra-search-field-input-placeholder-color, rgba(25, 32, 35, 0.53));
```
benbcai, line 127: Looks like this is the same placeholder color that was used in form-input. Wouldn't this continue to have the insufficient color contrast with the background that we originally had in form-input?

packages/terra-search-field/src/SearchField.module.scss
pranav300: > Are all terra-form-input variables required to theme the search field?

I haven't imported all the input variables, only the important ones.

packages/terra-search-field/src/SearchField.module.scss
pranav300: UPDATED HERE https://github.com/cerner/terra-core/pull/3078/commits/7c7364dc819baa811c938fa9d5d90b6a204f0d02.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -273,17 +272,17 @@ class SearchField extends React.Component {
...
             onChange={this.handleTextChange}
             disabled={isDisabled}
-            ariaLabel={inputText}
+            aria-label={inputText}
             aria-disabled={isDisabled}
             onKeyDown={this.handleKeyDown}
             onInput={this.handleInput}
-            refCallback={this.setInputRef}
+            ref={this.setInputRef}
             {...additionalInputAttributes}
```
benbcai, line 286: Previously, form-input will automatically add `aria-required` if `required` was passed to `inputAttributes`. Since this is no longer leveraging form-input, it would lose this logic. Should [this](https://github.com/cerner/terra-core/blob/cb664693fa409abe01ad4118a95877015afbdabf/packages/terra-form-input/src/Input.jsx#L147-L149) be added here?

pranav300: I and @neilpfeiffer discussed this and decided that it is unneeded to make `search-field` a required field.

dkasper-was-taken, line 286: It maybe not be part of this scope, but in general we spread pass-through attributes before our implementation specific ones so the behaviors cannot be overridden. So I'd like to see this moved.

neilpfeiffer:
Tested: PR #3078
Resolves: Issue #2992 along with #3075

**Reviewed:**
- Follows pattern considerations documented in Cerner Design Standards: [Search: Global Search](https://wiki.cerner.com/x/NG8VYQ)
- Follows UX defined design implementation for Global Search
- Follows defined theme designs: terra-default-theme, orion-fusion-theme, clinical-lowlight-theme

\
**Fix with Followup Issue:**
- [[search-field] investigate alignment of search-field, input, select, buttons #3085](https://github.com/cerner/terra-core/issues/3085)

\
**Related Updates (also will be updating usage of placeholders):**
[terra-core: issue #3083](https://github.com/cerner/terra-core/issues/3083)
[terra-clinical: issue #680](https://github.com/cerner/terra-clinical/issues/680)
[terra-framework: issue #1193](https://github.com/cerner/terra-framework/issues/1193)

\
\
Verified by @neilpfeiffer, ready to merge, merge before https://github.com/cerner/terra-core/pull/3075, 

**Pass** |
-----------|
UX Review|

packages/terra-search-field/src/SearchField.jsx
neilpfeiffer: Thanks @dkasper-was-taken  - logged this new issue here to move the spread to first in the list: https://github.com/cerner/terra-core/issues/3086

PR #3082, opened 2020-07-22
[terra-form select] Prevent opening of multiple dropdowns and moved focus to select menu when opened via toggle.

### LINKED ISSUE #3035: [terra-form-select] Search variant doesn't take focus on clicking toggle button on mobile
# Bug Report

## Description
I've only tested this in the search variant, but I imagine it exists in other variants as well.

(In 6.8.0) When rendered in mobile, the toggleButton does not open the dropDown, but relies on the handleMouseDown event. This [event](https://github.com/cerner/terra-core/blob/ff8271d46fd2519bc59bd1ad7bd11a718dec326c/packages/terra-form-select/src/search/Frame.jsx#L381) is implemented with a preventDefault call, in order to stop the search input from losing focus.

However, that also stops the select control from ever taking focus in the first place, resulting in opening the dropDown while an entirely different element has focus:

image

This cascades into more significant issues for our application, which relies on onBlur events as indications the user has finished editing some field which we subsequently need to update (the services we call can have logic that needs to run on a per detail basis).

## Steps to Reproduce
Easiest way to see this would be to visit https://engineering.cerner.com/terra-ui/components/terra-form-select/form-select/select on a mobile device, and select a component using the drop down. For the first selection, you'll notice that the input box does not get highlighted.

image
image

You can see the difference in behavior by clicking back and forth between inputs, which will always take focus and close the other input, and then clicking back and forth between drop down arrows, which will often leave both open and not have focus.

Probably related to #2898 and #2964 

## Expected Behavior
Clicking the drop down arrow should take focus and apply it to the component.

## Possible Solution
Adding 
`if (this.input) {
    this.input.focus();
}`
to onMouseDown before the preventDefault call (after may work as well) seems to fix this issue, as does moving that same snippet outside of the "if (this.state.isOpen)" block in handleToggleButtonMouseDown.

I imagine this would affect accessibility stuff in some way I don't understand though.

## Environment

* Component Name and Version: terra-form-select 6.8.0
* Browser Name and Version: Chrome 83.0.4103.101 (on Android)
* Webpack Version:
* Operating System and version (desktop or mobile): Mobile

saket2403:
### Summary
Fixed an issue to prevent multiple dropdowns from opening in mobile and move focus to select menu when opened via toggle button.

Closes #3035 

### Deployment Link

### Testing
Updated jest and wdio tests.

### Additional Details

@cerner/terra


packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -348,7 +353,7 @@ class Frame extends React.Component {
     // Don't blur if we dismissed the onscreen keyboard
     // Determined by if we have have interacted with the frame via onTouchStart
     // and if the focus is on input.
-    if (focusedByTouch && (relatedTarget === this.input)) {
+    if (relatedTarget === this.input) {
```
supreethmr, line 356: don't we need `focusedByTouch` here to differentiate between touch and non-touch device...??

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -324,7 +330,7 @@ class Frame extends React.Component {
       if (this.state.isPositioned) {
         this.dropdown.removeAttribute('inert');
         this.dropdown.removeAttribute('aria-hidden');
-        document.querySelector(this.selectMenu).setAttribute('tabIndex', '0');
+        this.selectMenu.setAttribute('tabIndex', '0');
```
supreethmr, line 333: I know this is not part of your code changes.. but menu has explicit [tabIndex=0](https://github.com/cerner/terra-core/blob/f937194e0b1d2e8b64bd4cabbbc0454d1d76cabb/packages/terra-form-select/src/combobox/Menu.jsx#L455)  so why do we need to set same tabIndex=0 again for menu.

packages/terra-form-select/src/multiple/Frame.jsx
```diff
@@ -428,6 +433,20 @@ class Frame extends React.Component {
...
+   * @param {event} event - The mouse down event.
+   */
+  handleMouseDown(event) {
+    // Preventing default events stops the search input from losing focus.
+    // The default variant has no search input therefore the mouse down gives the component focus.
+    event.preventDefault();
+    if (event.target.hasAttribute('data-terra-form-select-input')) {
+      this.input.focus();
+    }
+    this.openDropdown(event);
```
supreethmr, line 447: Since we are calling `this.openDropdown` on mouseDown  focus to input will be set at [openDropdown()](https://github.com/cerner/terra-core/blob/f937194e0b1d2e8b64bd4cabbbc0454d1d76cabb/packages/terra-form-select/src/multiple/Frame.jsx#L314) method. Do we still need to set the focus here as well...?

packages/terra-form-select/src/multiple/Frame.jsx
```diff
@@ -354,7 +359,7 @@ class Frame extends React.Component {
     // Don't blur if we dismissed the onscreen keyboard
     // Determined by if we have have interacted with the frame via onTouchStart
     // and if the focus is on input.
-    if (focusedByTouch && (relatedTarget === this.input)) {
+    if (relatedTarget === this.input) {
```
supreethmr, line 362: here as well I assume `focusedByTouch` was added specifically added to prevent Blur for touch-devices. bit confused how are we handling now this scenario by removing `focusedByTouch`?

supreethmr:
With new code changes there difference in existing focus behavior on dropdown close by clicking toggle-icon.
Prior to these code change on dropdown close focus was setting back to form-select-input as shown below (active element is still input ): 

form-select-ext-bvr

with these code changes focus moves out of form-select on close by clicking toggle-icon / toggle-button ( body becomes active element on clicking toggle icon )

form-select-new-bvr

Same behavior we can see in generated wdio screenshots as well focus styles are missing from form-select-input.

packages/terra-form-select/src/combobox/Frame.jsx
saket2403: This change is done so that the dropdown does not close when user opens it via toggle and clicks on input to focus it in desktop. No side effects in mobile.

packages/terra-form-select/src/multiple/Frame.jsx
saket2403: This change is done so that  when user opens it via toggle and clicks on input to focus it in desktop.Since it is already open this [line](https://github.com/cerner/terra-core/blob/f937194e0b1d2e8b64bd4cabbbc0454d1d76cabb/packages/terra-form-select/src/multiple/Frame.jsx#L287)  is executed in openDropdown

packages/terra-form-select/src/multiple/Frame.jsx
saket2403: Same as [above](https://github.com/cerner/terra-core/pull/3082#discussion_r458882200)

saket2403:
> With new code changes there difference in existing focus behavior on dropdown close by clicking toggle-icon.
> Prior to these code change on dropdown close focus was setting back to form-select-input as shown below (active element is still input ):

Fixed 13f1be6

packages/terra-form-select/CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ ChangeLog
 
 Unreleased
 ----------
+### Fixed 
+* Fixed am issue to prevent multiple dropdowns from opening in mobile and move focus to select menu when opened via toggle button.
```
StephenEsser, line 7: Small typo, 

`am` -> `an`

This entry should probably be split into two entries for each of the different fixes

packages/terra-form-select/tests/wdio/select-spec.js
```diff
@@ -1734,17 +1732,14 @@ Terra.describeViewports('Select', ['tiny'], () => {
       });
 
       Terra.it.validatesElement('select-closed', { selector: '#root' });
+      after(() => browser.click('#root'));
```
StephenEsser, line 1735: Can you please provide a description/explanation of these test changes? Why are they necessary, what is causing them to be necessary, and what happens/breaks if they are removed?

packages/terra-form-select/CHANGELOG.md
saket2403: Oopps Fixed here 3c82c03

packages/terra-form-select/tests/wdio/select-spec.js
saket2403: Looks like this hook is not required hence removed 3c82c03

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -303,10 +309,10 @@ class Frame extends React.Component {
     } else {
       // Allows time for state update to render select menu DOM before shifting focus to it
       setTimeout(() => {
-        if (document.querySelector(this.selectMenu)) {
-          document.querySelector(this.selectMenu).focus();
+        if (this.selectMenu) {
+          this.selectMenu.focus();
         }
-      }, 10);
+      }, 100);
```
yuderekyu, line 315: curious, how did you land on 100?

yuderekyu: hmm for reuse, can we extract this setTimeout and focus logic into a function? maybe we call it `shiftFocusToMenu` or something

saket2403: Moved setTimeout and focus to a function in FrameUtil, 875db90.
Stephen had done an initial investigation and suggested to go for a 100ms value [here](https://github.com/cerner/terra-core/issues/3035#issuecomment-644888940).

packages/terra-form-select/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+Fixed
```
StephenEsser, line 5: There's been a few minor changes in the formatting of the change log entries:

https://github.com/cerner/terra-core/pull/3106

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -497,6 +492,7 @@ class Frame extends React.Component {
    */
   toggleDropdown(event) {
     if (this.state.isOpen) {
+      this.input.focus();
```
StephenEsser, line 495: Why was this added?

packages/terra-form-select/src/multiple/Frame.jsx
```diff
@@ -428,6 +423,20 @@ class Frame extends React.Component {
...
+   */
+  handleMouseDown(event) {
+    // Preventing default events stops the search input from losing focus.
+    // The default variant has no search input therefore the mouse down gives the component focus.
+    event.preventDefault();
+    if (event.target.hasAttribute('data-terra-form-select-input')) {
+      this.input.focus();
+    }
+    this.openDropdown(event);
+  }
```
StephenEsser, line 438: This function is never called. The function being called in this file is the `static handleMouseDown` defined above the constructor. The Multiple and Tag variants were updated to open and close the dropdown on click events, rather than mouse down events.

https://github.com/cerner/terra-core/pull/3066/

packages/terra-form-select/src/multiple/Frame.jsx
```diff
@@ -698,8 +708,10 @@ class Frame extends React.Component {
     const descriptionId = `terra-select-screen-reader-description-${uniqueid()}`;
     const customAriaDescribedbyIds = customProps['aria-describedby'];
     const ariaDescribedBy = customAriaDescribedbyIds ? `${descriptionId} ${customAriaDescribedbyIds}` : descriptionId;
+    const selectMenuId = `terra-select-menu-${uniqueid()}`;
```
StephenEsser, line 711: I'm curious if these should be defined in the constructor rather than on every render.

packages/terra-form-select/src/shared/_FrameUtil.js
```diff
@@ -139,6 +139,19 @@ class FrameUtil {
     }
     return '-1';
   }
+
+  /**
+   * Allows time for state update to render select menu DOM before shifting focus to it
+   * @param {Object} frame - The frame object.
+   */
+
```
StephenEsser, line 147: Nit: extra new line

packages/terra-form-select/src/shared/_FrameUtil.js
```diff
@@ -139,6 +139,19 @@ class FrameUtil {
...
+  static shiftFocusToMenu(frame) {
+    setTimeout(() => {
+      if (frame.selectMenu) {
+        frame.selectMenu.focus();
+      }
+    }, 100);
+  }
```
StephenEsser, line 154: Nit: As a static util method I think this should work as an agnostic utility and be passed the element to be focused or at the very least pass the item to focus as a parameter and not the frame. 

```js
  static focusMenu(menu) { // setMenuFocus, delayMenuFocus, delayFocus, setFocus
    setTimeout(() => {
      if (menu) {
        menu.focus();
      }
    }, 100);
  }
```

packages/terra-form-select/src/tag/Frame.jsx
```diff
@@ -450,6 +444,20 @@ class Frame extends React.Component {
...
+   */
+  handleMouseDown(event) {
+    // Preventing default events stops the search input from losing focus.
+    // The default variant has no search input therefore the mouse down gives the component focus.
+    event.preventDefault();
+    if (event.target.hasAttribute('data-terra-form-select-input')) {
+      this.input.focus();
+    }
+    this.openDropdown(event);
+  }
```
StephenEsser, line 459: Same comment mentioned above for the multiple variant, this function is never called

saket2403: Removed

packages/terra-form-select/src/shared/_FrameUtil.js
saket2403: I had tried passing this.selectMenu here but it was passing undefined (weird), hence did it this way.

packages/terra-form-select/src/multiple/Frame.jsx
saket2403: Moved to constructor

packages/terra-form-select/src/multiple/Frame.jsx
saket2403: Removed

packages/terra-form-select/src/combobox/Frame.jsx
saket2403: removed

packages/terra-form-select/CHANGELOG.md
saket2403: Updated to new format

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -286,27 +291,19 @@ class Frame extends React.Component {
...
     if (this.input) {
       this.input.focus();
     } else {
-      // Allows time for state update to render select menu DOM before shifting focus to it
-      setTimeout(() => {
-        if (document.querySelector(this.selectMenu)) {
-          document.querySelector(this.selectMenu).focus();
-        }
-      }, 10);
+      FrameUtil.shiftFocusToMenu(this.selectMenu);
```
StephenEsser, line 306: This probably isn't working correctly using the menu reference

packages/terra-form-select/wdio.conf.js
```diff
@@ -1,6 +1,6 @@
 const path = require('path');
 const wdioConf = require('../../node_modules/terra-toolkit/config/wdio/wdio.conf');
 
-wdioConf.config.specs = [path.join(__dirname, 'tests', 'wdio', '**', '*-spec.js')];
+wdioConf.config.specs = [path.join(__dirname, 'tests', 'wdio', '**', 'select-spec.js')];
```
StephenEsser, line 4: I'd revert this, this is a _frame-spec.js that wouldn't match this regex. Was the regex/matcher causing issues?

packages/terra-form-select/src/shared/_FrameUtil.js
StephenEsser: This is probably an issue with maintaining references. I think leaving it the way it is will be fine. My concerns were it creates implicit naming convention of instance variables between unrelated classes and passes a reference to the entire class to call focus on a dom node. This being an internal util this isn't too big of deal.

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -180,9 +180,10 @@ class Frame extends React.Component {
     this.handleToggleMouseDown = this.handleToggleMouseDown.bind(this);
     this.handleToggleButtonMouseDown = this.handleToggleButtonMouseDown.bind(this);
     this.handleTouchStart = this.handleTouchStart.bind(this);
+    this.id = `terra-select-menu-${uniqueid()}`;
```
StephenEsser, line 183: This isn't too big of deal but this should probably be `menuId` or something that indicates it's the id specifically for the menu

saket2403: Done b5171d2

packages/terra-form-select/wdio.conf.js
saket2403: Ohh I forgot to revert that done b5171d2

packages/terra-form-select/src/combobox/Frame.jsx
saket2403: Forgot to revert to `this`, was testing passing selectMenu. Done b5171d2

StephenEsser:
Seeing a few issues doing some brief functional testing. 

On desktop clicking the arrow should still move focus into the input. It does not with these changes.

image

<hr />

Opening the isTouchAccessible for the combobox site example now behaves differently:

Existing on [terra-ui](https://engineering.cerner.com/terra-ui/components/terra-form-select/form-select/combobox):

image

[Deployment](https://terra-core-form-select--olb1zo.herokuapp.com/components/terra-form-select/form-select/combobox) behavior:

image

When opening the isTouchAccessible example on this PR deployment the dropdown jumps and forces a scroll.

saket2403:
> Seeing a few issues doing some brief functional testing.
> 
> On desktop clicking the arrow should still move focus into the input. It does not with these changes. Comparing the terra-ui deployment to the PR deployment to see differences in behavior.

> Opening the isTouchAccessible for the combobox site example now behaves differently:
> Existing on [terra-ui](https://engineering.cerner.com/terra-ui/components/terra-form-select/form-select/combobox):

> When opening the isTouchAccessible example on this PR deployment the dropdown jumps and forces a scroll.

Fixed 0fa748e

PaulSolDev:

PR #3091, opened 2020-07-28
[terra-search-field] Move additional attributes to the top of the attribute list.

### LINKED ISSUE #3086: [search-field] move attribute spread
## Description

Per comment here https://github.com/cerner/terra-core/issues/3086 from @dkasper-was-taken, move the spread attributes `{...additionalInputAttributes}` to first in the list of attributes, 
https://github.com/cerner/terra-core/blob/4aebf91d808076c9756a1ef3a6e6af52a77c497a/packages/terra-search-field/src/SearchField.jsx#L276-L288

saket2403:
### Summary
Moved `additionalInputAttributes` to the top of the list so it does. not overrides the implementation specific behaviours.
Closes #3086 

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-search-field/CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ ChangeLog
 
 Unreleased
 ----------
+### Changed
+* Moved `additionalInputAttributes` to the top of the attributes list.
```
StephenEsser, line 7: `additionalInputAttributes` is an internal instance variable. This entry may not make sense to consumers reading the changelog. Something along the lines of `Moved custom input attributes to be spread first on the input element` may make sense to users.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -283,7 +284,6 @@ class SearchField extends React.Component {
             onKeyDown={this.handleKeyDown}
             onInput={this.handleInput}
             ref={this.setInputRef}
-            {...additionalInputAttributes}
```
StephenEsser, line 286: This has the potential to non-passively change the API and break consumers. I do think this is an appropriate change to fix the attribute priority, but it may be worth a discussion on whether or not to major version bump, treat the change as a bug fix, or move forward if we have reason to believe few people (or none) would be impacted by the change.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -273,6 +273,7 @@ class SearchField extends React.Component {
       <div {...customProps} className={searchFieldClassNames}>
         <div className={cx('input-group')}>
           <input
+            {...additionalInputAttributes}
```
yuderekyu, line 276: can we add a test case for this?

packages/terra-search-field/src/SearchField.jsx
ryanthemanuel: I think the big thing we care about here is the className so let's keep applying any custom class names that teams might be passing in.  Anything else, I would argue we should never have been doing in the first place so (with my recommendation) I would be comfortable calling this passive.

packages/terra-search-field/CHANGELOG.md
saket2403: Changed c92ed69

packages/terra-search-field/src/SearchField.jsx
saket2403: Added jest to show that custom class does not override terra styles. c92ed69

packages/terra-search-field/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+Changed
```
StephenEsser, line 5: This heading missing some syntax formatting

packages/terra-search-field/src/SearchField.jsx
emilyrohrbough: @saket2403 I see you resolved this conversation, but I'm not sure moving the attribute spread pulls the custom class name out to ensure custom styles arn't accidentally broken.

packages/terra-search-field/CHANGELOG.md
StephenEsser: Taking a closer look, this seems to have changed recently with the change log lint resolution PR. I think this may be correct as is, but I'd like to confirm with @mjhenkes on the changes. Whether this should be unformatted headings or triple #'s,

```
Changed
```

```
### Changed
````

mjhenkes: This is intended, repeating headings are not recommend because they will break hash linking to sections in the document.

saket2403:
> > I see you resolved this conversation, but I'm not sure moving the attribute spread pulls the custom class name out to ensure custom styles arn't accidentally broken.
> 
> Following up on Emily's comment above. We'll need to make sure the custom classNames passed through inputAttributes are merged into the `<input />` classNames

Ahh I misinterpreted the conversation there .... Updated f933599

packages/terra-search-field/CHANGELOG.md
mjhenkes: Sorry to go around on this but we ended up updating this to a list style, see #3106 for more details

PR #3094, opened 2020-07-28
[terra-form-select] Update allowClear, isFilterStyle from SingleSelect and placeholder from all the variants.

### LINKED ISSUE #3083: [terra-select] Update placeholders and allowClear
# Feature Request

## Description
As part of removing placeholders in form inputs for better accessibility, the Form Select inputs will also have a higher level of Cerner UX Design Standard requirements to provide a common and consistent placeholder pattern that is no longer editable by consumers. 

Similar to Native Select, `allowClear` now becomes a default feature included with SingleSelect, and `isFilterStyle` is added for cases where a value must always be selected and not cleared, commonly found in filters and toolbars.

Related to 
[[terra-form-input] Insufficient color-contrast for placeholder #2992](https://github.com/cerner/terra-core/issues/2992)

pranav300:
### Summary

This PR updates:
1. Removes `placeholder` prop from all the variants. And passes a default placeholder text `- Select -`.
2. Removes `allowClear` prop from SingleSelect making the functionality the default behavior.
3.  Adds prop `isFilterStyle` to SingleSelect, to match native-select that removes the ability to have no value selected, showing a placeholder, and the ability to clear an already selected value.
4. Updated color for placeholder.

Closes #3083 

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-form-select/src/ComboboxField.jsx
```diff
@@ -170,6 +164,11 @@ const ComboboxField = ({
     }
   }
 
+  if (customProps.placeholder) {
+    // eslint-disable-next-line no-param-reassign
+    customProps.placeholder = null;
```
supreethmr, line 169: I think instead of assigning to null we can delete the prop 
```js
delete customProps.placeholder;
```

packages/terra-form-select/src/MultiSelect.jsx
```diff
@@ -214,7 +206,7 @@ class MultiSelect extends React.Component {
         display={this.display()}
         onDeselect={this.handleDeselect}
         onSelect={this.handleSelect}
-        placeholder={selectPlaceholder}
+        placeholder={intl.formatMessage({ id: 'Terra.form.select.defaultDisplay' })}
```
supreethmr, line 209: Do we need `placeholder={intl.formatMessage({ id: 'Terra.form.select.defaultDisplay' })}` prop assignment here ...?? because I see input of multi-select always refers to ['Terra.form.select.defaultDisplay'](https://github.com/cerner/terra-core/blob/cf1f413a32cc4702892ba571ceae14c7e0a45404/packages/terra-form-select/src/multiple/Frame.jsx#L253) and `placeholder` prop has no reference in `frame.jsx`. and we can see [placeholder prop is being set to undefined as well](https://github.com/cerner/terra-core/blob/cf1f413a32cc4702892ba571ceae14c7e0a45404/packages/terra-form-select/src/multiple/Frame.jsx#L710).

packages/terra-form-select/src/ComboboxField.jsx
pranav300: Updated here https://github.com/cerner/terra-core/pull/3094/commits/d3d0df8877f5742edf80995be6aae5ba19b5ff6b

packages/terra-form-select/src/MultiSelect.jsx
pranav300: Oops!! forgot to remove it. Done here https://github.com/cerner/terra-core/pull/3094/commits/da8765dc02ec816efca769497d1f26ba9503f98d

packages/terra-form-select/src/ComboboxField.jsx
```diff
@@ -170,6 +164,11 @@ const ComboboxField = ({
     }
   }
 
+  if (customProps.placeholder) {
+    // eslint-disable-next-line no-param-reassign
```
supreethmr, line 168: did you forgot to remove this linter disable 🧐

pranav300: with `delete` also it is basically reassigning. Therefore linter disable is required.

packages/terra-form-select/src/SingleSelect.jsx
```diff
@@ -170,41 +166,40 @@ class SingleSelect extends React.Component {
...
-    const selectPlaceholder = placeholder === undefined ? defaultPlaceholder : placeholder;
-    let clearOptionDisplay;
+    let frameValue;
 
-    if (allowClear) {
-      if (selectPlaceholder.length === 0) {
-        clearOptionDisplay = defaultPlaceholder;
+    if (!isFilterStyle || this.state.value || value) {
+      frameValue = SelectUtil.value(this.props, this.state);
+    } else if (children) {
```
supreethmr, line 183: can this be simplify to :
```js
else {
frameValue = ( children && children[0].props.children ) ? children[0].props.children[0].props.value : children[0].props.value;
```

packages/terra-form-select/src/Select.jsx
```diff
@@ -145,24 +148,27 @@ function Select(props) {
 
   switch (variant) {
     case Variants.COMBOBOX: {
-      const { maxSelectionCount, ...comboboxProps } = otherProps;
+      const { maxSelectionCount, isFilterStyle, ...comboboxProps } = otherProps;
```
supreethmr, line 151: do we need `isFilterStyle` to be passed for all the variants...??

packages/terra-form-select/src/SingleSelect.jsx
pranav300: This condition fails if  there is no option/opt-group provided for the select.

packages/terra-form-select/src/Select.jsx
pranav300: I think this is to extract `isFilterStyle` from all the props i.e `otherProps`. And later to the components only component specific props are being passed i.e `...comboboxProps` for combobox.

supreethmr: yes, but I am bit confused about usage of this prop. Correct me if i am wrong, why do we need extract if it is not used by these variants..? I see this prop is declared only for single-select variant

packages/terra-form-select/src/SingleSelect.jsx
```diff
@@ -170,41 +166,40 @@ class SingleSelect extends React.Component {
...
       value,
       intl,
+      isFilterStyle,
       ...otherProps
     } = this.props;
 
-    const defaultPlaceholder = intl.formatMessage({ id: 'Terra.form.select.defaultDisplay' });
-    const selectPlaceholder = placeholder === undefined ? defaultPlaceholder : placeholder;
-    let clearOptionDisplay;
+    let frameValue;
```
lokesh-0813, line 179: nit: This can be renamed something different.

packages/terra-form-select/src/Select.jsx
pranav300: > why do we need extract if it is not used by these variants..?

If not extracted this will be passed to the variants where it's not supported and become part of the spread operator in there local files. This would cause an unsupported and unrecognized prop to be present in the DOM.

packages/terra-form-select/src/SingleSelect.jsx
pranav300: Updated here https://github.com/cerner/terra-core/pull/3094/commits/7f7548bb813d8756aaa233b31d5ae4965d81d33d.

packages/terra-form-select/src/ComboboxField.jsx
jmsv6d: Why are you deleting `placeholder` from the custom props? The console would log a warning if it got applied to a dom element. Removing the prop would make this a major version bump, so it would be on consumers to not be passing in a prop that doesn't exist.

packages/terra-form-select/src/Combobox.jsx
```diff
@@ -8,7 +8,7 @@ import SelectUtil from './shared/_SelectUtil';
 
 const propTypes = {
   /**
-   * Whether a clear option is available to clear the selection, will use placeholder text if provided.
+   * Whether a clear option is available to clear the selection, will use placeholder text.
```
jmsv6d, line 11: Could you add what the provided placeholder text is here?

pranav300: Updated `allowClear` description to be -
`Whether a clear option is available to clear the selection, will use  - Select - as the clear option.`

packages/terra-form-select/src/ComboboxField.jsx
StephenEsser: I agree with Jaime, I don't think there is value in deleting the placeholder prop here. Consumers upgrading should be removing the prop from their code, deleting the prop will suppress any warnings that invalid props are being passed.

packages/terra-form-select/src/Select.jsx
```diff
@@ -32,6 +33,12 @@ const propTypes = {
    */
   // eslint-disable-next-line react/forbid-prop-types
   dropdownAttrs: PropTypes.object,
+  /**
+   * Whether the select input should use the filter style display, forcing a value to always be selected.
+   * This also removes the placeholder and removes the ability for user to clear the value, returning the select to browser-native behavior.
+   * This is only applicable to the `default` variant.
+   */
+  isFilterStyle: PropTypes.bool,
```
StephenEsser, line 41: To be consistent with native select I think we should keep the prop name the same, but I'm having a hard time understanding how the prop name correlates to the functionality it toggles on and off. I know what the prop does, but I don't know what the prop name actually signifies.

packages/terra-form-select/src/SingleSelect.jsx
```diff
@@ -170,41 +166,40 @@ class SingleSelect extends React.Component {
...
 
-    const defaultPlaceholder = intl.formatMessage({ id: 'Terra.form.select.defaultDisplay' });
-    const selectPlaceholder = placeholder === undefined ? defaultPlaceholder : placeholder;
-    let clearOptionDisplay;
+    let selectValue;
 
-    if (allowClear) {
-      if (selectPlaceholder.length === 0) {
-        clearOptionDisplay = defaultPlaceholder;
+    if (!isFilterStyle || this.state.value || value) {
```
StephenEsser, line 181: This will need to account for falsey, but real, values. empty string and the number zero for example. Generally these are done by checking for undefined.

packages/terra-form-select/src/SingleSelect.jsx
```diff
@@ -170,41 +166,40 @@ class SingleSelect extends React.Component {
...
+      selectValue = SelectUtil.value(this.props, this.state);
+    } else if (children) {
+      if (children[0].props.children) {
+        selectValue = children[0].props.children[0].props.value;
```
StephenEsser, line 185: I'm wary of these chained prop/array accessors being safe. I'd recommend borrowing/simplifying some of the accessors being used in the menu util. Children(OptGround/Options) have types that can be evaluated.

https://github.com/cerner/terra-core/blob/main/packages/terra-form-select/src/shared/_MenuUtil.js

packages/terra-form-select/src/terra-dev-site/test/form-select/FilterStyle.test.jsx
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import classNames from 'classnames/bind';
+import Select from '../../../Select';
+import styles from './common/Select.test.module.scss';
+
+const cx = classNames.bind(styles);
+
+class FilterStyle extends React.Component {
```
StephenEsser, line 8: New examples should use hooks when possible for best practice. Hooks will help simplify this test example

packages/terra-form-select/src/SingleSelect.jsx
pranav300: updated condition here https://github.com/cerner/terra-core/pull/3094/commits/5c0d19afda050f822e2d6e7fcd04abefa15886b6

packages/terra-form-select/src/SingleSelect.jsx
pranav300: Removed the use of chained prop/array here https://github.com/cerner/terra-core/pull/3094/commits/5c0d19afda050f822e2d6e7fcd04abefa15886b6

packages/terra-form-select/src/terra-dev-site/test/form-select/FilterStyle.test.jsx
pranav300: Updated the example here https://github.com/cerner/terra-core/pull/3094/commits/5c0d19afda050f822e2d6e7fcd04abefa15886b6

packages/terra-form-select/src/Select.jsx
ShettyAkarsh: +1, yup we should provide a prop name that correlates with the functionality.

packages/terra-form-select/src/single/Frame.jsx
```diff
@@ -184,12 +186,21 @@ class Frame extends React.Component {
...
-      : <div id={placeholderId} className={cx('placeholder')}>{placeholder || '\xa0'}</div>
-    );
+    if (!isFilterStyle && !display) {
+      return (<div id={placeholderId} className={cx('placeholder')}>{intl.formatMessage({ id: 'Terra.form.select.defaultDisplay' }) || '\xa0'}</div>);
+    }
+
+    if (display && MenuUtil.findByDisplay(children, display)) {
+      return (<span id={displayId}>{display}</span>);
+    }
+    if (children) {
```
ShettyAkarsh, line 200: what happens in the case if isFilterStyle is set to false and the display values is not found  ? which fails the above condition.  i guess in that case we add the first item in the display with this logic. which shouldn't be the case as isFilterStyle is set to false.

packages/terra-form-select/src/Select.jsx
ShettyAkarsh: +1, we should keep the prop name that correlates with the functionality

packages/terra-form-select/src/single/Frame.jsx
pranav300: Resolved it here https://github.com/cerner/terra-core/pull/3094/commits/29f1e7283f86fac21d66bd5e739ab2339f154507

PR #3096, opened 2020-07-28
[terra-icon] Fixed how class names are identified with css modules to support themes properly

### LINKED ISSUE #3095: [terra-icon] Themes are broken
# Bug Report

## Description

When we made the change to terra-icon to isolate themeable variables to only the icons they were used in, we inadvertently broke css modules for this component.

ryanthemanuel:
### Summary

Fixed how class names are identified with css modules to support themes properly
Fixes #3095


mjhenkes:
Lets get the tests updated and it would also be nice if the documentation example page supported the lowlight theme

Issue #3012, opened 2020-06-03
terra-status-view UX Audit


ryanthemanuel:
# terra-status-view

##  Known Aliases
Document any known aliases of the component, decide if another name makes more sense.
- [X] Has known aliases
  - Error view
  - No data view
- [ ] Requires name change

## Design Standard Doc
Any documentation needed to be added to terra-ui describing the make up and usage of the component
Any doc examples that need to be updated?
- [X] Missing design standard documentation linkage.

## Guides
Any guides needed to be added to terra-ui describing the usage of this component when used with other components.
- [ ] Missing Guides
- [X] Missing UX recommended Usage.
  - Call out how to handle multiple views on the page
  - Call out not to use the full page views for a smaller space
- [ ] Deprecation guide

## Accessibility Guides
- [ ] Missing Accessibility Guides

## Behaviours
- [X] Missing behaviours
  - Have the ability to have show hide for more detailed information
  - Evaluate responsiveness at small breakpoints (e.g. font sizes)
  - Evaluate a max width on the message and title
- [X] Contains bad practice behaviours that should be removed
  - Possibly split out a custom version and have this one completely locked down
  - Possibly change this one to custom only and use the application status view to be the locked down version

## Accessibility
- [ ] Meets wcag 2.0/section 508 standards
  - Need to evaluate how we're presenting the title and the message.  Both p's now and probably need to be hierarchical
  - Need to evaluate the expand collapse functionality when it gets added
- [ ] Meets wcag 2.1/en 301 549 standards

## Deprecation
- [ ] Component is a bad pattern and should be deprecated.

## Related Issues
- https://github.com/cerner/terra-core/issues/3000
- https://github.com/cerner/terra-application/issues/30

PR #3151, opened 2020-09-07
[terra-status-view] Remove focus from glyphs in IE10

### LINKED ISSUE #3149: [terra-status-view] Status view glyphs are focusable in IE10
# Bug Report

## Description

Status view glyphs are focusable in IE10.

## Expected Behavior

Glyphs should not be focusable

## Possible Solution

Add focusable='false' to the svg element. We should evaluate other components to see if they have a similar problem

saket2403:
### Summary
Removed focus from glyphs in IE10.
Closes #3149 

### Deployment Link

### Additional Details

@cerner/terra


ShettyAkarsh (approved):

avinashg1994 (approved):

packages/terra-status-view/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+* Changed
+  * Removed focus from glyphs in IE10 .
+
```
avinashg1994, line 5: Should be moved under `Unreleased` section

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -99,7 +99,7 @@ const StatusView = ({
   } else if (variant && !isGlyphHidden) {
     glyphSection = (
       <div className={cx('glyph')}>
-        <svg className={cx(variant)} />
+        <svg focusable={false} className={cx(variant)} />
```
yuderekyu, line 102: lets also recommend this pattern in our documentation for consumers using the custom glyph prop

jeremyfuksa (approved):

packages/terra-status-view/src/StatusView.jsx
saket2403: Done 322f801

packages/terra-status-view/CHANGELOG.md
saket2403: Done 322f801

lokesh-0813 (approved):

emilyrohrbough (approved):

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -99,7 +100,7 @@ const StatusView = ({
...
```
StephenEsser, line 103: Does this need to be an SVG tag? Does switching to a div break/fix expected functionality? On initial glance this is just an empty container that is used to apply a CSS background image.

saket2403: On replacing svg with div, no image shows up.

supreethmr (approved):

packages/terra-status-view/src/StatusView.jsx
StephenEsser: Do you know why no image shows up? We use divs/spans for background SVG images throughout the component library. The root issue for this bug is using the empty SVG container, if we need the SVG container then this fix works, but if we don't need the SVG container a more appropriate fix is likely swapping out the SVG tag for a div or span.

yuderekyu: @saket2403 when you replaced the svg with div, did you also update the style selector? those styles would affect how the markup is rendered: https://github.com/cerner/terra-core/blob/main/packages/terra-status-view/src/StatusView.module.scss#L49

saket2403: Updating the css selectors did work, replaced svg with div fe0e024

packages/terra-status-view/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
```
StephenEsser, line 5: This should be under a Fixed section if the issue was logged as a bug

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -28,6 +28,7 @@ const propTypes = {
 
   /**
    * Display a custom glyph. Overrides a variant's default glyph.
+   * Set `focusable=false` for svg element used as `customGlyph`.
```
StephenEsser, line 31: This comment might not be necessary anymore, but it's not harmful to leave it if users are using a a custom SVG

packages/terra-status-view/src/StatusView.jsx
```diff
@@ -99,7 +100,7 @@ const StatusView = ({
   } else if (variant && !isGlyphHidden) {
     glyphSection = (
       <div className={cx('glyph')}>
-        <svg className={cx(variant)} />
+        <div className={cx(variant, 'bg-common')} />
```
StephenEsser, line 103: We should use a more descriptive class name here, glyph-icon, glyph-container, icon or something more descriptive. bg-common feels a little too generic, almost like a boostrap/global css value

saket2403: Changed b469734

packages/terra-status-view/CHANGELOG.md
saket2403: Changed b469734

pranav300 (approved):

StephenEsser (approved):

packages/terra-status-view/src/StatusView.module.scss
```diff
@@ -46,7 +46,8 @@
   .glyph {
     padding-bottom: 0.714rem;
 
-    > svg {
+    > svg,
```
yuderekyu, line 49: this is out of scope and probably a breaking change, but i wonder if we can eventually remove the `svg` selector. currently, it only applies if a consumer passes an svg via the `customGlyph` prop. its probably better if we let consumers handle their own svg styles. 

tho, i think our plan is to eventually remove custom functionality from the status-view in some matter. https://github.com/cerner/terra-core/issues/3012

emilyrohrbough: I think you are right that we can probably remove this when we remove the support for custom svgs. It's prob fine for now to leave as is

yuderekyu (approved):

PR #3158, opened 2020-09-14
[terra-form-field] Fixed voiceover issues in label-group.

### LINKED ISSUE #3152: [terra-form-input] Required inputs do not announce correctly on VoiceOver Safari
# Bug Report

## Description

Input fields do not appropriately announce labels when set to `required`. The issues appears to be related to inserting the required asterisk in front of the label.

## Steps to Reproduce

1. (In Safari) Navigate to https://engineering.cerner.com/terra-ui/components/terra-form-input/form-input/input-field
2. Activate VoiceOver
3. Navigate to the required input field example
4. Observe the label text is not announced. It announces "star and one more item", where the label is the "one more item"

Attaching the text output of VoiceOver Safari for one of the terra-ui examples. Notice that label is not announced. VoiceOver announces "star and one more item" instead of the label.

image

## Expected Behavior

Required input fields should appropriately announce labels on mobile Safari when using VoiceOver

## Possible Solution

The issue appears to happen because of the dom markup:

```html
<label for="requiredField" class="Field-module__label___2QGDW">
  <div class="Field-module__required___3Csnx">*</div>
  Required Input Field
</label>
```

Potentially hiding the div containing the asterisk from screen readers would resolve the issue.

pranav300:
### Summary

So the issue was found in the form-field. And a similar issue was seen for `optionalText` where the voiceover was reading `Label TextAnd one more Item`. 
The `optionalText` issue was fixed by changing `div` to `span`. Whereas for the required `*` div I have added `aria-hidden` and currently `*` is not being read. Now, if we want `*` to be read then we can change [this](https://github.com/cerner/terra-core/pull/3158/files#diff-329b2171cca50d6f3e9df40c1bb35d87R181) also to a `span`. 

Closes #3152 

### Deployment Link

https://terra-core-required-lab-ztue0y.herokuapp.com/

### Additional Details

@cerner/terra


StephenEsser (approved):
I tested these changes on desktop Safari for the input field and it appears to be working. We should do a more thorough test prior to merging for the other field elements and mobile.

Edit:

We'll also need to test the "optional" text announces correctly along with the label

emilyrohrbough (approved):

tbiethman (approved):

jeremyfuksa (approved):

saket2403 (approved):

PR #3172, opened 2020-09-24
[terra-alert] Fixing text alignment at ultrawide browser widths

### LINKED ISSUE #3171: [terra-alert] Long text is misaligned at widths > 1785px
# Bug Report

## Description

Alert long text falls out of alignment with the icon and title at ultrawide browser widths.

## Steps to Reproduce

Resize the browser display to any width greater than 1785px.

Screen Shot 2020-09-24 12 20 58 AM

jeremyfuksa:
### Summary

Cleaning up the appearance of alerts in browsers set to display on ultrawide monitors.

Closes #3171

### Deployment Link

https://terra-core-terra-alert--6yh41r.herokuapp.com/components/terra-alert/alert/alert


jeremyfuksa:
Screen Shot 2020-09-24 02 08 24 PM

This still isn't exactly right. I have an idea, but it needs media queries and I'm seeing that CSS Modules doesn't play well with them.

packages/terra-alert/src/Alert.module.scss
```diff
@@ -68,7 +69,6 @@
 
   .title {
     display: var(--terra-alert-title-display);
-    float: var(--terra-alert-title-float, left);
```
neilpfeiffer, line 71: Not a fan of removing this or adjusting the component, only need to update the example contents.

packages/terra-alert/src/terra-dev-site/doc/example/LongTextExample.jsx
```diff
@@ -12,6 +12,7 @@ const LongTextExample = () => {
         Four score and seven years ago our fathers brought forth on this continent, a new nation, conceived in Liberty, and dedicated to the proposition that all men are created equal.
         <Button
           variant="de-emphasis"
+          isCompact
```
neilpfeiffer, line 15: Rather than trying to fix Button, which will never have proper alignment inline with text, might be easiest to  resolve by switching out the Toggle in the example for the `terra-show-hide` component, even though it adds a new line.

packages/terra-alert/src/Alert.module.scss
jeremyfuksa: What was the intention behind this? If you are using flexbox, floats don't act on the box model.

packages/terra-alert/src/terra-dev-site/doc/example/LongTextExample.jsx
neilpfeiffer: Backup option could be we swap out the toggle to use a hyperlink (which would stay inline) which launches a basic javascript alert with the hidden/full text.

jeremyfuksa: I replaced toggle with show/hide on the Heroku deployment.

packages/terra-alert/src/Alert.module.scss
```diff
@@ -55,6 +55,7 @@
   }
 
   .section {
+    display: flex;
```
neilpfeiffer, line 58: This causes the layout of the section contents to break and not display as inline any longer (the title line breaks and wraps, inner blocks layout horizontally as flexes which in not intended). Removing will correct the layout.

packages/terra-alert/src/Alert.module.scss
neilpfeiffer: It wasn't so much due to browsers not being able to support flex, but more due to uncontrollable nature of the possible children content that can be provided. 

Traditional float was used instead of flex so the title can be inserted before the message text (or any children content) and essentially function as if is was set as `inline` and delivered with the children/message. It allows for normal block content to be inserted as the alert content, which since it is pure children, is largely unable to be controlled or understood.

packages/terra-alert/src/terra-dev-site/doc/example/LongTextExample.jsx
```diff
@@ -1,31 +1,34 @@
 import React, { useState } from 'react';
-import Button from 'terra-button';
-import Toggle from 'terra-toggle';
+import ShowHide from 'terra-show-hide';
 import Alert from 'terra-alert';
+import classNames from 'classnames';
+import classNamesBind from 'classnames/bind';
```
neilpfeiffer, line 5: Duplicative import with previous line, can remove line 5 and update line 9 to just:
```
```

packages/terra-alert/src/terra-dev-site/doc/example/LongTextExample.jsx
```diff
@@ -1,31 +1,34 @@
...
-          </p>
-          <p>
-            But, in a larger sense, we can not dedicate -- we can not consecrate -- we can not hallow -- this ground. The brave men, living and dead, who struggled here, have consecrated it, far above our poor power to add or detract. The world will little note, nor long remember what we say here, but it can never forget what they did here. It is for us the living, rather, to be dedicated here to the unfinished work which they who fought here have thus far so nobly advanced. It is rather for us to be here dedicated to the great task remaining before us -- that from these honored dead we take increased devotion to that cause for which they gave the last full measure of devotion -- that we here highly resolve that these dead shall not have died in vain -- that this nation, under God, shall have a new birth of freedom -- and that government of the people, by the people, for the people, shall not perish from the earth.
-          </p>
-        </Toggle>
-      </span>
+      <ShowHide
+        preview={previewText}
```
neilpfeiffer, line 18: Need to either re-include the `{previewText}` as the first item with the children/contents or can remove the preview prop and provide the non-hidden text as just a span ahead of the `<ShowHide... />` component so is visible both before and after the show/hide.

packages/terra-alert/src/terra-dev-site/doc/example/LongTextExample.module.scss
```diff
@@ -0,0 +1,7 @@
+:local {
+  .long-text-example {
+    p:first-child {
+      margin-top: 0;
+    }
```
neilpfeiffer, line 5: can also add no margin-bottom on last `<p>` so it is brings the show/hide button closer to the text and saves some extra vertical height.
```
    p:last-child {
      margin-bottom: 0;
    }

```

jeremyfuksa:
Feedback addressed in 6a90935.

packages/terra-alert/src/terra-dev-site/doc/example/LongTextExample.module.scss
```diff
@@ -0,0 +1,11 @@
+:local {
+  .long-text-example {
```
neilpfeiffer, line 2: Since teams might look to copy this as their exact example, let's tighten up default style from 1em(14px) to 0.71428rem(10px) on the  basic `<p>` tag, which will make it look like default styling/standard spacer sizes.
```
p {
      margin: 0.71428rem 0;
}
```

neilpfeiffer (approved):
Updates to the LongTextExample look good, nice work @jeremyfuksa 
I added one last style update suggestion: https://github.com/cerner/terra-core/pull/3172#pullrequestreview-496765480

UX approved after last css is applied, ready for additional reviews.

| UX +1 |
| ---- |

emilyrohrbough (commented):
Is this a bug with terra-alert or our example? I was expecting changes to the component's styles

jeremyfuksa:
> Is this a bug with terra-alert or our example? I was expecting changes to the component's styles

It ended up being an example problem and not a bug.

emilyrohrbough (approved):

ryanthemanuel (approved):

PR #3187, opened 2020-10-11
[terra-icon] Remove CSP violation

### LINKED ISSUE #3182: IconBase causes CSP inline style violation
# Bug Report

## Description
it appears that `terra-overlay` requires `IconBase` where IconBase is inserting inline-style [here](https://github.com/cerner/terra-core/blob/master/packages/terra-icon/src/IconBase.jsx#L75)

Screen Shot 2020-10-06 at 1 12 11 PM

## Steps to Reproduce
1. Add `terra-overlay` in an application
2. Turn on CSP without `unsafe-inline` for `style-src` directive

## Expected Behavior
* No CSP violation coming from terra-application without unsafe-inline for style-src directive

## Environment

* Component Name and Version: "terra-icon-3.28.0", "terra-overlay-3.56.0",
* Browser Name and Version: Chrome Version 85.0.4183.102 (Official Build) (64-bit)
* Node/npm Version: [e.g. Node 8/npm 5] node 10
* Webpack Version: webpack ^4.41
*Operating System and version (desktop or mobile): desktop

pranav300:
### Summary

This PR removes the code to append a style tag to the head. And passes the animation styles as global styles.

Closes #3182 

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-icon/src/Icon.module.scss
```diff
@@ -21,3 +21,66 @@
     transform: scaleX(-1);
   }
 }
+
+:global {
```
mjhenkes, line 25: Could you make this :local?

pranav300: If we change it to :local, then CSS Modules changes the keyframe name [here](https://github.com/cerner/terra-core/pull/3187/files/138e706ceb6f1bb20470bacc1fb2f13f8937cd3f#diff-3ca2b30d1d2a7907939e40bd50431ce8R78) by appending the fileName and hash key but doesn't do anything to the keyframe declaration. Hence, the keyframe is not recognized.

That's why I have passed it in global scope.

mjhenkes: Could you make line 75 and below local?

pranav300: Similar behavior is being seen if I put the CSS under the line 75 into local.

packages/terra-icon/src/IconBase.jsx
```diff
@@ -62,22 +62,6 @@ const IconBase = ({
...
-      const head = document.head || document.getElementsByTagName('head')[0];
-      const style = document.createElement('style');
-      style.type = 'text/css';
-      style.id = 'terra-icon-animation';
-      // If we haven't written the styles to the DOM yet, add them, otherwise don't do this for subsequent spinner icons.
```
emilyrohrbough, line 72: do we need to keep this logic of adding a node to attach these styles to?

packages/terra-icon/src/Icon.module.scss
```diff
@@ -21,3 +21,66 @@
...
+  /* stylelint-disable at-rule-no-vendor-prefix */
+  /* stylelint-disable property-no-vendor-prefix */
+  @-webkit-keyframes tui-spin-ltr {
```
yuderekyu, line 28: do we get this for free with autoprefixer? also can someone tell me if this is necessary at all? caniuse says it's supported by the browsers that we support https://caniuse.com/?search=%40keyframes

packages/terra-icon/src/IconBase.jsx
pranav300: I don't think we need to add a node to attach these styles as the global scope takes care of it.

And the reason behind the CSP violation was the style node in head without any nonce or hash attribute.

packages/terra-icon/src/Icon.module.scss
pranav300: removed all the prefixers here https://github.com/cerner/terra-core/pull/3187/commits/bf44dee3d4a6952d87319321968f0386aff790e0.

packages/terra-icon/src/Icon.module.scss
```diff
@@ -21,3 +21,34 @@
...
+      transform: rotate(0deg);
+    }
+    
+    100% {
+      transform: rotate(-359deg);
+    }
+  }
+
+  /* stylelint-disable selector-class-pattern */
+  [dir=ltr] .tui-Icon.is-spin {
```
mjhenkes, line 47: You can do this, which will scope the selector but not apply the scope to the animation name
```suggestion
  :local([dir=ltr] .tui-Icon.is-spin) {
```

packages/terra-icon/src/Icon.module.scss
```diff
@@ -21,3 +21,34 @@
...
+    animation: tui-spin-ltr 3s infinite linear;
+  }
+
+  [dir=rtl] .tui-Icon.is-spin {
```
mjhenkes, line 51:
```suggestion
  :local([dir=rtl] .tui-Icon.is-spin) {
```

packages/terra-icon/src/Icon.module.scss
pranav300: Updated here https://github.com/cerner/terra-core/pull/3187/commits/c51499d2d550c87e7551e34717068ae4a895a861

packages/terra-icon/src/Icon.module.scss
pranav300: Updated here https://github.com/cerner/terra-core/pull/3187/commits/c51499d2d550c87e7551e34717068ae4a895a861

mjhenkes (approved):

emilyrohrbough (approved):

supreethmr (approved):

PR #3188, opened 2020-10-12
[terra-form-select] Added mutation observer shim to better support inert in IE10.

### LINKED ISSUE #3184: Add the mutation observer shim wherever we are including the inert polyfill
# Bug Report

## Description

Because we support IE10, we need to add the mutation observer shim wherever we are including the inert polyfill.  This includes inside all variants of the select. For example:

https://github.com/cerner/terra-core/blob/main/packages/terra-form-select/src/combobox/Frame.jsx#L194

We should mimic what we're doing here:

https://github.com/cerner/terra-framework/blob/4880772b8b92bb05f99ef3f3084a8d1e4c389bdc/packages/terra-abstract-modal/src/AbstractModal.jsx#L5

saket2403:
### Summary
Added mutation observer shim and node contains polyfill to better support inert polyfill in IE10.

Closes #3184 

### Deployment Link

### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-form-select/package.json
+++ b/packages/terra-form-select/package.json
@@ -31,6 +31,7 @@
     "classnames": "^2.2.5",
     "keycode-js": "^2.0.1",
     "lodash.uniqueid": "^4.0.1",
+    "mutationobserver-shim": "0.3.3",
     "prop-types": "^15.5.8",
     "react-lifecycles-compat": "^3.0.2",
     "terra-form-field": "^4.9.0",
--- a/packages/terra-form-select/CHANGELOG.md
+++ b/packages/terra-form-select/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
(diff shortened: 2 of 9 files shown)


ShettyAkarsh (approved):

tbiethman (approved):

packages/terra-form-select/package.json
```diff
@@ -31,6 +31,7 @@
     "classnames": "^2.2.5",
     "keycode-js": "^2.0.1",
     "lodash.uniqueid": "^4.0.1",
+    "mutationobserver-shim": "0.3.3",
```
benbcai, line 34: Is there a reason why using 0.3.3 instead of the latest 0.3.7?

ryanthemanuel: Let's keep them all consistent with what we're already using for now.  We can log separate issues to update this later.

benbcai (approved):

PR #3203, opened 2020-10-26
[terra-section-header] Updated dev site examples to use hooks

avinashg1994:
### Summary

Updated terra-dev-site examples to use hooks.

### Deployment Link

https://terra-core-section-head-prwguq.herokuapp.com/

### Testing

Jest and WDIO

### DIFF
--- a/packages/terra-section-header/src/terra-dev-site/doc/example/AccordionSectionHeader.jsx
+++ b/packages/terra-section-header/src/terra-dev-site/doc/example/AccordionSectionHeader.jsx
@@ -1,29 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import SectionHeaderExampleTemplate from 'terra-section-header/lib/terra-dev-site/doc/example/SectionHeaderExampleTemplate';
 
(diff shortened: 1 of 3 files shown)


packages/terra-section-header/src/terra-dev-site/doc/example/AccordionSectionHeader.jsx
```diff
@@ -1,29 +1,22 @@
-import React from 'react';
+import React, { useState } from 'react';
 import SectionHeaderExampleTemplate from 'terra-section-header/lib/terra-dev-site/doc/example/SectionHeaderExampleTemplate';
```
StephenEsser, line 2: Outside the scope of this PR, but I think in the future we should rework these examples to not abstract the section header out of the example. I recently looked at these examples to implement the section header and found them to not be very helpful since you cannot see how to implement the section header itself.

StephenEsser (approved):

emilyrohrbough (approved):

tbiethman (approved):

PR #3285, opened 2020-12-03
react-intl upgrade

### LINKED ISSUE #3208: Update components to work with react-intl v2-v5
# Feature Request
In an endeavor to minimize the number of major version bumps we want to make the update for our components to work with react-intl v5 wherever possible.

## Description
For this update the peer dependencies need to be updated. `intlShape` Is no longer exported since v3, so the imports for that must be removed and PropType for intl object updated. An undocumented non-passive change was made in react-intl for the `FormattedMessage` component, when a function is passed as a child of the component the function is now given the param of an array of strings instead of the joined string. This causes many of the components that we are rendering in these functions to have prop type warnings as they are expecting a string. There are a few approaches that we could take to make these components continue to work as expected for all the react-intl version. I believe for most of these cases we can switch from using the `FormattedMessage` component to `injectIntl` + `intl.formatMessage()` or the function provided to `FormattedMessage` can do a check if the param is an array and join the strings, if so.

I have begun this work at https://github.com/cerner/terra-core/tree/react-intl-upgrade
This work should be able to be completed and released prior to the framework portions of the upgrade being completed.
Note: terra-base and terra-i18n should not be updated with this change.

jmsv6d:
### Summary

Updates components to be compatible with react-intl v2-v5

Closes #3208 

### Deployment Link

### Testing

Verified locally that components work latest version of react-intl without errors.

### Additional Details

@cerner/terra


mjhenkes (approved):

ryanthemanuel (approved):

packages/terra-alert/src/Alert.jsx
```diff
@@ -55,6 +55,10 @@ const propTypes = {
    * Callback function triggered when Dismiss button is clicked. The presence of this prop will cause the Dismiss button to be included on the alert.
    */
   onDismiss: PropTypes.func,
+  /**
```
supreethmr, line 58: I think `@private` is missing here....!

packages/terra-form-radio/src/RadioField.jsx
```diff
@@ -27,6 +27,7 @@ const propTypes = {
    * Whether or not to hide the required indicator on the legend.
    */
   hideRequired: PropTypes.bool,
+  intl: PropTypes.shape({ formatMessage: PropTypes.func }),
```
supreethmr, line 30: missing prop description for `intl`

packages/terra-show-hide/src/ShowHide.jsx
```diff
@@ -16,7 +16,7 @@ const propTypes = {
   /**
    * The intl object to be injected for translations. Provided by the injectIntl function.
    */
-  intl: intlShape.isRequired,
+  intl: PropTypes.shape({ formatMessage: PropTypes.func }).isRequired,
```
supreethmr, line 19: `@private` seems to be missing here too..!

benbcai (approved):

PR #3356, opened 2021-02-17
[CONNECT-30881] Allow terra-list-item to render an id in the dom

Andrew-Shoemake:
### Summary

Updating terra-list-item to allow passing an id to the li of terra-list-item element for easier feature/element tracking using Pendo

### Deployment Link

N/A

### Testing

Added an additional unit test to validate the id is being set, inspected the dom to validate the id being set, and ran wdio tests to ensure no UI broke.

### Additional Details

@cerner/terra


Andrew-Shoemake:
Had a quick chat with @StephenEsser, we may not actually need this and achieve the same thing via customProps. Hold off on review for the time being.

Andrew-Shoemake:
It has been determined this is not necessary due to terra-list-item having custom props

PR #3384, opened 2021-03-03
[terra-form-select] Fixed focus not getting remove in IE for Search and Combobox variants.

### LINKED ISSUE #3366: [terra-form-select] Select focus border-color is not getting removed on moving out the focus
# Issue Description

In the IE11 browser, the select focus border-color is not getting removed when the focus is moved out of the element. If we continue to select other select elements in the view then all previously selected elements will remain highlighted in light blue.  

Able to reproduce the issue in the documentation site and have attached the screenshot for reference.
Screenshot 2021-02-24 at 10 14 56 AM

pranav300:
### Summary

This PR fixes focus `not getting removed` in IE for search and combobox variants by making `relatedTarget` to fallback to `document.activeElement` when `event.relatedTarget` is null.

In IE the relatedTarget for Search and Combobox is `null` due to which [this](https://github.com/cerner/terra-core/blob/main/packages/terra-form-select/src/search/Frame.jsx#L217) condition returns `true` due to `this.selectMenu` for these 2 variants is also null.

Closes #3366 

### Deployment Link

### Testing

Tested on:
IE and Chrome in Windows10.
Safari and Chrome in Mac.

### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-form-select/src/search/Frame.jsx
+++ b/packages/terra-form-select/src/search/Frame.jsx
@@ -204,7 +204,7 @@ class Frame extends React.Component {
    * Handles the blur event.
    */
   handleBlur(event) {
-    const { relatedTarget } = event;
+    const relatedTarget = event.relatedTarget || document.activeElement;
 
     // The check for dropdown.contains(activeElement) is necessary to prevent IE11 from closing dropdown on click of scrollbar in certain contexts.
     if (this.dropdown && (this.dropdown === document.activeElement && this.dropdown.contains(document.activeElement))) {
--- a/packages/terra-form-select/CHANGELOG.md
+++ b/packages/terra-form-select/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fixed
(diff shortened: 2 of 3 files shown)


ShettyAkarsh (approved):

lokesh-0813 (approved):

StephenEsser (approved):

benbcai (approved):

packages/terra-form-select/src/search/Frame.jsx
```diff
@@ -204,7 +204,7 @@ class Frame extends React.Component {
    * Handles the blur event.
    */
   handleBlur(event) {
-    const { relatedTarget } = event;
+    const relatedTarget = event.relatedTarget || document.activeElement;
```
supreethmr, line 207: I think we need similar changes for tag and multiple variant as well : 

Screen Shot 2021-03-17 at 12 35 45 PM

mjhenkes: If we need more changes for the other variants, could you log a new issue to address them?

PR #3385, opened 2021-03-05
[terra-form-select] Allow VO to read index value for select options

### LINKED ISSUE #3325: [terra-form-select] Screen Reader is not reading the index value for options
# Bug Report

## Description
Voice Over starts reading by announcing select has 5 options and then reads out the options as we tab through them without announcing "1 off 5", "2 off 5", etc along with the name.

## Steps to Reproduce

1. Open any example on [dev-site](https://engineering.cerner.com/terra-core/components/terra-form-select/form-select/combobox).
2. Navigate to the select 
3. Turn on VO and tab through the options.

## Expected Behavior
It should read index values along with the option name.

## Environment

* Component Name and Version:
* Browser Name and Version:
* Webpack Version:

## @ Mentions
@ryanthemanuel , @neilpfeiffer

saket2403:
### Summary

Closes #3325 

### Deployment Link

### Additional Details

https://user-images.githubusercontent.com/42207428/110085199-d40be000-7db6-11eb-835b-972c31cade01.mov

The VO  reads all variants like the screen recording above except the native select

@cerner/terra


packages/terra-form-select/src/combobox/Menu.jsx
```diff
@@ -356,14 +356,16 @@ class Menu extends React.Component {
     }
 
     if (element) {
+      const index = active.getAttribute('index');
+      const totalOptions = optGroupElement ? optGroupElement.props.children.length : document.querySelectorAll('[data-terra-select-option]').length;
```
StephenEsser, line 360: I don't think we should query the entire document. This might work due to the conditional render of the dropdown only if open, but it feels really dangerous. We should find a safer way to get the child count of this specific menu

packages/terra-form-select/src/combobox/Menu.jsx
```diff
@@ -356,14 +356,16 @@ class Menu extends React.Component {
...
       if (element.props.display === '' && element.props.value === '') {
         // Used for case where users selects clear option and opens
         // dropdown again and navigates to clear option
         visuallyHiddenComponent.current.innerText = clearSelectTxt;
       } else if (this.isActiveSelected()) {
-        visuallyHiddenComponent.current.innerText = intl.formatMessage({ id: 'Terra.form.select.selectedText' }, { text: displayText });
+        visuallyHiddenComponent.current.innerText = `${intl.formatMessage({ id: 'Terra.form.select.selectedText' }, { text: displayText })}${index}of${totalOptions}`;
```
StephenEsser, line 366: This new addition to the sentence will probably need translated

packages/terra-form-select/src/combobox/Menu.jsx
pranav300: When we don't have any `optGroup` or if we have `multiple optGroups`, then these changes work fine.

But when we have combinations of `non-optGroups and optGroups`, I think there is an issue. Attaching a gif to display the problem.

ezgif com-gif-maker
Here the `totalOptions` jumps from 10 to 5, the moment an optGroup is encountered.

Proposed solution:

ezgif com-gif-maker (1)
We can change the sentencing, But I believe this will be a better approach.

Here's the snippet for this:
```jsx
if (element) {
  const index = active.getAttribute('index');
  const totalOptions = this.props.children[1].length;
  const optGroupElementsLength = optGroupElement ? optGroupElement.props.children.length : undefined;
  if (element.props.display === '' && element.props.value === '') {
    // Used for case where users selects clear option and opens
    // dropdown again and navigates to clear option
    visuallyHiddenComponent.current.innerText = clearSelectTxt;
  } else if (this.isActiveSelected()) {
    visuallyHiddenComponent.current.innerText = optGroupElementsLength ? `${intl.formatMessage({ id: 'Terra.form.select.selectedText' }, { text: displayText })}, option ${index} of ${optGroupElementsLength} group options, of total ${totalOptions} options` : `${intl.formatMessage({ id: 'Terra.form.select.selectedText' }, { text: displayText })} ${index} of total ${totalOptions} options`;
  } else {
    visuallyHiddenComponent.current.innerText = optGroupElementsLength ? `${`${displayText}, option ${index}`} of ${optGroupElementsLength} group options, of total ${totalOptions} options` : `${`${displayText}, option ${index}`} of total ${totalOptions} options`;
  }
}
```

This will also resolve the issue of using the `document` mentioned above.

packages/terra-form-select/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fixed
+  * Allow voice-over to read index values for options.
```
StephenEsser, line 6: I left a [comment](https://github.com/cerner/terra-core/issues/3325#issuecomment-792947555) on the issue. Is there a web standards document that defines how this interaction should be working?

ShettyAkarsh:
Adding 'do not merge' label as we are waiting on the UX and accessibility inputs on how the format of the text which VO reads should be and also how the options under the Group should be read.

scottwilmarth:
@ShettyAkarsh @StephenEsser @neilpfeiffer @mjhenkes [I noted on a previous issue](https://github.com/cerner/terra-core/issues/3325#issuecomment-791724069) that Terra should NOT try to "fix" any failures of any of the assistive technology like VoiceOver when it comes to a native HTML element. This should be logged to Apple VoiceOver as an issue. User agents are responsible for the accessibility of a native component. If Terra customizes a native HTML element, Cerner would become responsible and liable for the accessibility of it. Terra would also be setting itself up to maintain the fixes should Apple fix the issue.

neilpfeiffer:
@scottwilmarth point of clarification that I think you might have missed as context - this issue/pr is not trying to increase the accessibility of the native terra-form-select, the intent is to keep the native-select as native. This PR will be addressing increasing the accessibility of _all the other types_ of dropdown selects in this package which are complete custom creations (combobox, searchable select, multiselect, and tag multiselect).

scottwilmarth:
If this is regarding other non-native versions, please provide more context of what exactly I should be looking at to review.

saket2403:
As Neil mentioned, in this PR we are implementing index for the custom variants of form select. As of now I have implemented the index to read as `x of y`. Please test the deployed examples [here](https://terra-core-select-setsi-quyhnj.herokuapp.com/components/terra-form-select/form-select/combobox), and suggest changes for the phrase `x of y` read out by the VO, if required.  As of now this PR reads `index of total options` for normal options and `index of  total option in the group` for opt-groups. Which can be ambiguous in the scenario where the select has a combination of both options and opt-groups as pointed out in this [comment](https://github.com/cerner/terra-core/pull/3385#discussion_r589602228). So we need some input on what the VO should read for both options and opt-groups. Anything you would like to add @neilpfeiffer.
cc @scottwilmarth

packages/terra-form-select/src/combobox/Menu.jsx
neilpfeiffer: We can use the existing translations we already have in [terra-paginator](https://github.com/cerner/terra-core/tree/main/packages/terra-paginator/translations) since the format and usage is the same, leaving out the word for "Page":
https://github.com/cerner/terra-core/blob/2511a56ab492730cc93fafd2c029e81d7115ec17/packages/terra-paginator/translations/en-US.json#L7

And since we are concatenating it with the existing string/s, we'll also want to add space and punctuation (parentheses work for all the Terra supported languages) so that the final text is read back more appropriately : e.g. "Group Shades of Green, Forest Green (1 of 5),"   

\
\
Sample i18n with parens:
```
  "Terra.form.select.displayPositionInSetSize": "({indexNumber} of {totalNumber})",
```

\
\
The concatenation can keep the space which might get lost in the translation:
```
visuallyHiddenComponent.current.innerText = `${intl.formatMessage({ id: 'Terra.form.select.selectedText' }, { text: displayText })} ${intl.formatMessage({ id: 'Terra.form.select.displayPositionInSetSize' }, { indexNumber: index, totalNumber: totalOptions })},`;
```

and

```
visuallyHiddenComponent.current.innerText = `${displayText} ${intl.formatMessage({ id: 'Terra.form.select.displayPositionInSetSize' }, { indexNumber: index, totalNumber: totalOptions })},`;
```

packages/terra-form-select/src/combobox/Menu.jsx
neilpfeiffer: +1 to the intent of the proposals above and the problems found.

After much testing and researching, we'll want to change the options within option groups from being announcing with their own index count and total, but flatten the list to sound like it is only one list overall, with a single total and unique index for each option item, similar as to how we are already flattening the OptGroup label as part of the `displayText`.

For your example listed above with the combination of options and option-groups, the displayText plus setsize information would read back as:
```
Blue (1 of 10),
Cyan (2 of 10),
Azul (4 of 10),
Aqua (5 of 10),
Group Shades of Green, Green (6 of 10),
Group Shades of Green, Forest Green (7 of 10),
Group Shades of Green, Neon Green (9 of 10),
Group Shades of Green, Emerald 10 of 10),
```

_Side note: We may need to revisit disabled option items, since they are currently never read back in either the menu or frame and the count ._

If we needed a static total count (e.g. for the legacy single-select) the value could be passed in to the menu as a prop from the frame, which already has it defined.

But since the combobox, multi, and tag can all be filtered based on what is typed in the input, the totalOptions, the setsize of the list changes dynamically, so it could make the most sense to just set both the index/posinset and the total/setsize as attributes on each option, such that when the list is filtered, the attribute values can be updated, something more like:

```
    if (element) {
      const active = this.menu.querySelector('[data-select-active]');
      const posinset = active.getAttribute('aria-posinset');
      const setsize = active.getAttribute('aria-setsize');
      if (element.props.display === '' && element.props.value === '') {
        // Used for case where users selects clear option and opens
        // dropdown again and navigates to clear option
        visuallyHiddenComponent.current.innerText = clearSelectTxt;
      } else if (this.isActiveSelected()) {
        visuallyHiddenComponent.current.innerText = `${intl.formatMessage({ id: 'Terra.form.select.selectedText' }, { text: displayText })} ${intl.formatMessage({ id: 'Terra.form.select.displayPositionInSetSize' }, { indexNumber: posinset, totalNumber: setsize })},`;
      } else {
        visuallyHiddenComponent.current.innerText = `${displayText} ${intl.formatMessage({ id: 'Terra.form.select.displayPositionInSetSize' }, { indexNumber: posinset, totalNumber: setsize })},`;
      }
    }
```

So using the same list above, typing "Gr" in the input, then navigating down into the list would read:
```
Group Shades of Green, Green (1 of 4),
Group Shades of Green, Forest Green (2 of 4),
Group Shades of Green, Neon Green (4 of 4),
```

_Again, how to handle disabled option items may need to be revisited later._

saket2403:
PR has been updated with all the comments and suggestions as discussed with UX.

mjhenkes (approved):

neilpfeiffer:
Notes on final updates applied by @saket2403 :
- The form-selects for search, combobox, multiple, tag, and legacy-single select have all been updated to for screen readers to announce the option `{item name}` followed by `( {position} of {total} )`, _e.g. Blue (1 of 8)_, as the user navigates through the list of options.
- If the option item is selected, it is announced as `{item name} Selected. ( {position} of {total} )`,  _e.g. Blue Selected. (1 of 8)_
- For OptionGroups, the option group label is not read separately, instead it is combined with the each option item name and the position information, _e.g. Group Shades of Blue, Aqua (1 of 8)_, the `{total}` is the overall total of all options in each option group. As the user navigates from one option group to the next, the `{position}` is the position of the option item relative to the total flattened list, not the position only within its' option group.
- The `totalOptions` (number of options initially presented the menu) in each of the `_frame` private subcomponents were removed from the announced strings for `List of {total} options.`, since it was found to not always accurately represent the number of options presented dynamically in the list (filtering the list, disabling items in the list, adding new options in combo and tag, etc). This can be revisited in the future if a better way to update the option total in the `_frame` as the available options change in the `_menu`.
- The native-select was left unchanged with no updates.

\
\
No further input is needed at this time and all current outstanding UX questions have been resolved. 
cc: @scottwilmarth

packages/terra-form-select/CHANGELOG.md
neilpfeiffer: There isn't a great single standard to base off of. The guidance we are going off of so far are composited between the [samples you have referenced](https://github.com/cerner/terra-core/issues/3325#issuecomment-792947555), patterns for related ui patterns like accessible list filtering (even though they aren't a dropdown, navigating a list is still the same), and industry standard products like Microsoft Teams, etc.

neilpfeiffer (approved):
Tested: PR #3385
Resolves: Issue #3325

**Reviewed:**
- Tested and verified changes on VoiceOver for Mac/Safari and JAWs for Windows/IE



\
\
Verified by @neilpfeiffer, complete pending any engineering comments

**UX Reviewed** |
-----------|

packages/terra-form-select/src/shared/_MenuUtil.js
```diff
@@ -198,6 +198,18 @@ class MenuUtil {
     return index === -1 ? null : options[Math.min(index + 1, options.length - 1)].props.value;
   }
 
+  /**
+   * Finds the index of option following the active option.
+   * @param {ReactNode} object - The node being flattened.
+   * @param {string} value - The value of the active option.
+   * @return {integer} - The value of the index of option.
+   */
+  static checkIndex(object, value) {
```
StephenEsser, line 207: This isn't an exposed API so the naming isn't a big deal, but ordinarily I'd expect this to be something like `findIndex` or `getIndex`.

StephenEsser (approved):

ShettyAkarsh (approved):

supreethmr (approved):
Do we need changes in single-select also...?? I am thinking it would be better if we don't consider single-select for this code changes as it is marked for deprecation and also since this code changes doesn’t brings up any changes in VO Behavior.

this inconsistency in VO could be due  to single-select structure that causes VO to skip the option's index in single-select and reads out aria-label value and placeholder value repeatedly. placeholder text and the aria-label of `display` div takes priority over text added to visually-hidden-component. 

cc: @neilpfeiffer

saket2403:
> Do we need changes in single-select also...?? I am thinking it would be better if we don't consider single-select for this code changes as it is marked for deprecation and also since this code changes doesn’t brings up any changes in VO Behavior.

Just confirmed with Neil, we will be deprecating the single select soon, so changes are not required as users are directed to use native select instead. @supreethmr.

packages/terra-form-select/src/shared/_MenuUtil.js
saket2403: Done 1369bb9

PR #3386, opened 2021-03-05
[terra-form-input] Update input to fallback to native input for unsupported types.

### LINKED ISSUE #3378: terra-form-input for file input not working
# Bug Report

## Description
terra-form-input is used for file upload and the input file button does not show up as expected in IE.

## Steps to Reproduce
Load the document site in IE-11
Here is the link: https://engineering.cerner.com/terra-ui/components/terra-form-input/form-input/input
The file input shows up as in screenshot:
image

## Expected Behavior
image

## Environment

* Component Name and Version:
* Browser Name and Version:
* Webpack Version:

pranav300:
### Summary

This PR updates input to remove className `form-input` to make it look like native input when `type="file"`.

Closes #3378

### Deployment Link

**Review these examples:**
- Supported Types -> Input: https://terra-core-change-input-xisti2.herokuapp.com/tests/terra-form-input/form-input/supported-types-input
- Supported Types -> InputField: https://terra-core-change-input-xisti2.herokuapp.com/tests/terra-form-input/form-input/supported-types-input-field
- Unsupported Types -> Input: https://terra-core-change-input-xisti2.herokuapp.com/tests/terra-form-input/form-input/unsupported-types-input
- Unsupported Types -> InputField: https://terra-core-change-input-xisti2.herokuapp.com/tests/terra-form-input/form-input/unsupported-types-input-field

### Additional Details

- Investigation and Proposal: https://github.com/cerner/terra-core/pull/3386#discussion_r590259305

- Decided Final Changes: https://github.com/cerner/terra-core/pull/3386#issuecomment-820194408


packages/terra-form-input/src/Input.jsx
```diff
@@ -122,7 +122,7 @@ class Input extends React.Component {
     const theme = this.context;
     const formInputClassNames = classNames(
       cx(
-        'form-input',
+        type === 'file' ? '' : 'form-input',
```
supreethmr, line 125: Nit : To keep it consistent ( Unless changing like this throws lint error ).
```suggestion
       'form-input': (type !== 'file'),
```

packages/terra-form-input/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Removed
```
StephenEsser, line 5:
```suggestion
* Changed
```

This is a change to conditionally apply the class name. It doesn't remove functionality. Removed is slightly misleading.

packages/terra-form-input/src/Input.jsx
StephenEsser: ^ adding on to Supreeth's comment. The syntax below shows an example for conditional assignment using classNames.

```js
// These classes are only applied if the logic statement is truthy
{ 'form-error': isInvalid },
{ 'form-incomplete': (isIncomplete && required && !isInvalid) },
```

However, removing `form-input` will remove the majority of the [themable properties](https://github.com/cerner/terra-core/blob/50b6591cb451ee123e7e81d0ec082fe8151a5af1/packages/terra-form-input/src/Input.module.scss#L7-L64) of the component. This is probably not what we want to do, but we can track down the specific properties causing the styling issues to see if we can create a class to conditionally apply them.

Is there a specific set of styles causing the issue?

supreethmr: +1 
keeping  class `form-input` as it was before and moving only specific styles causing truncation in IE to new conditional class looks more appropriate here.

StephenEsser: @neilpfeiffer to clarify our decision here so we can link back to it in future discussions. Is your recommendation to remove all terra styles and custom properties when the type is file? Specifically, does this also include removing the incomplete and error classes?

packages/terra-form-input/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+ * Removed className `form-input` when `type="file"`.
```
benbcai, line 6: It would be generally easier to understand what is changed if description is something like what you have on the PR title:

`Update input to be native input when the type is file`

pranav300: Updated here https://github.com/cerner/terra-core/pull/3386/commits/fda676b3d846449ad4561fa467c7cbbaabb8bd97

packages/terra-form-input/src/Input.jsx
neilpfeiffer: Thanks @StephenEsser, adding details here for the confirmed decisions and future reference.

---

**Changes:**

Since specific input types have native UI provided by browsers by default, like `type="file"`, and only `type="text"`, `type="number"`, and `type="password"` were originally designed for and tested as the terra-form-input component with intentional styling and behavior handling, we've decided to create a separation for types that are 'supported' (has Terra specific styling and behaviors) and those that are 'unsupported'. 

Rather than completely blocking those types that are unsupported, the decision is to conditionally skip binding all terra classes: the base `'form-input'`, plus the conditional `'form-error'`, `'form-incomplete'`, and the `theme`, yet still pass through any consumer provided classNames. 

After testing, the final list of 'supported' and 'unsupported' `types` are as follows:

- **Supported Types** = 'text', 'number', 'password', 'email', 'search', 'tel', 'url', 'hidden'.

- **Unsupported Types** =  'date', 'datetime-local', 'month', 'time', 'week', 'color', 'file', 'range', 'checkbox', 'radio', 'button', 'reset', 'submit'.

---

Note: If future enhancements provide Terra styling and behavior handling for any of the unsupported types, they can be added back to the list of supported types.

neilpfeiffer (changes requested):
@pranav300 - I've added the series of comments and suggestions for updates based on our change decisions provided in this comment: https://github.com/cerner/terra-core/pull/3386#discussion_r613809173

I've also created a branch based on this one with suggested changes and tests needed to validate which input types should be considered 'supported' vs. 'unsupported', branch comparisons are here: 
- https://github.com/cerner/terra-core/compare/change-input-with-type-file-to-native...non-supported-input-type-tests

neilpfeiffer:
Adding details here for the confirmed decisions and future reference:

---

**Changes:**

Since specific input types have native UI provided by browsers by default, like `type="file"`, and only `type="text"`, `type="number"`, and `type="password"` were originally designed for and tested as the terra-form-input component with intentional styling and behavior handling, we've decided to create a separation for types that are 'supported' (has Terra specific styling and behaviors) and those that are 'unsupported'. 

Rather than completely blocking those types that are unsupported, the decision is to conditionally skip binding all terra classes: the base `'form-input'`, plus the conditional `'form-error'`, `'form-incomplete'`, and the `theme`, yet still pass through any consumer provided classNames. 

After testing, the final list of 'supported' and 'unsupported' `types` are as follows:

- **Supported Types** = 'text', 'number', 'password', 'email', 'search', 'tel', 'url', 'hidden'.

- **Unsupported Types** =  'date', 'datetime-local', 'month', 'time', 'week', 'color', 'file', 'range', 'checkbox', 'radio', 'button', 'reset', 'submit'.

---

Note: If future enhancements provide Terra styling and behavior handling for any of the unsupported types, they can be added back to the list of supported types.

pranav300:
I have merged all the changes suggested by @neilpfeiffer in this [branch](https://github.com/cerner/terra-core/tree/non-supported-input-type-tests). And added wdio tests for the new test.

packages/terra-form-input/src/terra-dev-site/test/form-input/UnsupportedTypesInput.test.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+const unsupportedTypes = ['date', 'datetime-local', 'month', 'time', 'week', 'color', 'file', 'range', 'checkbox', 'radio', 'button', 'reset', 'submit'];
+
+const getInputTypeExamples = (arr) => {
+  const inputTypeExamples = [];
+  for (let i = 0; i < arr.length; i += 1) {
+    inputTypeExamples.push((
+      <div className={cx('content-wrapper')}>
+        <label htmlFor={`html-original-${arr[i]}-input`}>{`Input type=${arr[i]}`}</label>
+        <Input name={`html-original-${arr[i]}-input`} type={arr[i]} ariaLabel={`Input type=${arr[i]}`} />
```
neilpfeiffer, line 16: Same as comment above for `SupportedTypesInput` Test.

packages/terra-form-input/src/terra-dev-site/test/form-input/UnsupportedTypesInputField.test.jsx
```diff
@@ -0,0 +1,84 @@
...
+            inputId={`unsupported-${arr[i]}-input`}
+            error="Error message."
+            help="Help message."
+            isInvalid={this.state.isInvalid}
+            isIncomplete={this.state.isIncomplete}
+            required={this.state.required}
+            inputAttrs={{
+              name: `html-original-${arr[i]}-input`,
+              type: arr[i],
+              ariaLabel: `Input type=${arr[i]}`,
```
neilpfeiffer, line 39: Same as comment above for `SupportedTypesInputField` Test.

packages/terra-form-input/src/terra-dev-site/test/form-input/SupportedTypesInput.test.jsx
```diff
@@ -0,0 +1,29 @@
...
+
+const supportedTypes = [undefined, 'text', 'number', 'password', 'email', 'search', 'tel', 'url', 'hidden'];
+
+const getInputTypeExamples = (arr) => {
+  const inputTypeExamples = [];
+  for (let i = 0; i < arr.length; i += 1) {
+    inputTypeExamples.push((
+      <div className={cx('content-wrapper')}>
+        <label htmlFor={`supported-${arr[i]}-input`}>{`Input type=${arr[i]}`}</label>
+        <Input name={`supported-${arr[i]}-input`} type={arr[i]} ariaLabel={`Input type=${arr[i]}`} />
```
neilpfeiffer, line 16: The `ariaLabel` is redundant and not needed since these examples already include the preferred `label` element with the `htmlFor` implementation (something our other input tests & examples should have used instead, which we can fix on future enhancement of the full component). Other raw/standalone input examples for terra-form-input don't use a label element, instead they (1)either have no text above or (2)use a div with no connecting aria, so a quick workaround at the time was to use `ariaLabel`  on those freestanding (non-form-field) examples to pass axe checks, but using `aria-label` is not our Terra recommended A11Y implementation for inputs, instead using the explicit visible label as the preferred method.
- https://dequeuniversity.com/rules/axe/4.1/label-title-only
- https://www.deque.com/blog/anatomy-of-accessible-forms-best-practices/
- https://www.w3.org/WAI/tutorials/forms/labels/
- https://www.a11ywithlindsey.com/blog/introduction-accessible-labeling

Either can update to remove the ariaLabel:
``` diff
-        <Input name={`html-original-${arr[i]}-input`} type={arr[i]} ariaLabel={`Input type=${arr[i]}`} />
+        <Input name={`html-original-${arr[i]}-input`} type={arr[i]}  />
```

or less ideal, but would match existing examples, change the label to a div:
```
-        <label htmlFor={`html-original-${arr[i]}-input`}>{`Input type=${arr[i]}`}</label>
+        <div>{`Input type=${arr[i]}`}</div>
```

packages/terra-form-input/src/terra-dev-site/test/form-input/SupportedTypesInputField.test.jsx
```diff
@@ -0,0 +1,84 @@
...
+            inputId={`unsupported-${arr[i]}-input`}
+            error="Error message."
+            help="Help message."
+            isInvalid={this.state.isInvalid}
+            isIncomplete={this.state.isIncomplete}
+            required={this.state.required}
+            inputAttrs={{
+              name: `supported-${arr[i]}-input`,
+              type: arr[i],
+              ariaLabel: `Input type=${arr[i]}`,
```
neilpfeiffer, line 39: Our field implementations already provide for proper accessibility connecting the inputs and labels, so this become redundant, and we don't want to suggest consumers should also provide an `aria-label` on the inputs.

```diff
-              ariaLabel: `Input type=${arr[i]}`,
```

packages/terra-form-input/src/terra-dev-site/test/form-input/UnsupportedTypesInputField.test.jsx
neilpfeiffer, line 39: Same as comment for `SupportedTypesInputField` Test.

packages/terra-form-input/tests/wdio/form-input-spec.js
```diff
@@ -27,6 +27,20 @@ Terra.describeViewports('Form-Input', ['tiny', 'large'], () => {
...
+    $('#root').moveTo({ xOffset: 0, yOffset: 700 }); // add to ensure consistent hover styles
+
+    Terra.validates.element('supported types');
+  });
+
+  it('should display Form Input unsupported types', () => {
+    browser.url('/tests/terra-form-input/form-input/unsupported-types-input');
+    $('#root').moveTo({ xOffset: 0, yOffset: 700 }); // add to ensure consistent hover styles
+
+    Terra.validates.element('unsupported types');
```
neilpfeiffer, line 41: I am wondering about the usefulness of having WDIO snapshots for the unsupported types - since they are already considered `unsupported` and will be dependent on browser changes, not terra changes. Plus they don't all fit in the capture area.  My suggestion would be to remove this test.

packages/terra-form-input/tests/wdio/form-input-spec.js
```diff
@@ -128,6 +142,20 @@ Terra.describeViewports('Form-Input', ['tiny', 'large'], () => {
...
+
+      Terra.validates.screenshot('input field supported types');
+    });
+
+    it('should display InputField unsupported types', () => {
+      browser.url('raw/tests/terra-form-input/form-input/unsupported-types-input-field');
+      $('#root').moveTo({ xOffset: 0, yOffset: 700 }); // add to ensure consistent hover styles
+
+      Terra.validates.element('input field unsupported types');
+    });
```
neilpfeiffer, line 158: Similar as comment above: 

I am wondering about the usefulness of having WDIO snapshots for both of these tests for InputFields, since the test is really meant primarily for unscripted testing and validation. 

I would suggest removing these two tests, since all the types don't all fit within the image capture area (screenshots don't capture all of the types included), and we already are validating the styling of the supported types on the standalone input test.

packages/terra-form-input/src/Input.jsx
```diff
@@ -120,13 +125,18 @@ class Input extends React.Component {
 
     const attributes = { ...customProps };
     const theme = this.context;
+
+    // unsupportedTypes: 'date', 'datetime-local', 'month', 'time', 'week', 'color', 'file', 'range', 'checkbox', 'radio', 'button', 'reset', 'submit'
+    const supportedTypes = [undefined, 'text', 'number', 'password', 'email', 'search', 'tel', 'url', 'hidden'];
```
benbcai, line 130: I think what you have here works and we can keep it as is. I just wonder if it would be less confusing to not include `undefined` as a supported type since it is not really an input type. Instead, we can explicitly include the check for `type === undefined` on line 132.

packages/terra-form-input/src/terra-dev-site/test/form-input/SupportedTypesInput.test.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import classNames from 'classnames/bind';
+import Input from '../../../Input';
+import styles from './common/Input.test.module.scss';
+
+const cx = classNames.bind(styles);
+
+const supportedTypes = [undefined, 'text', 'number', 'password', 'email', 'search', 'tel', 'url', 'hidden'];
```
benbcai, line 8: Should we exclude `undefined` as a supported type in this test since `undefined` is technically not an input type? This would actually show `type=undefined` in the DOM. If `type` was not set as a prop, then type would just be `undefined` and there would _not_ be a type attribute in the DOM since it would just be the default type of text as illustrated in the existing [default-form-input test example](https://engineering.cerner.com/terra-core/tests/terra-form-input/form-input/default-form-input).

Screen Shot 2021-04-19 at 10 58 15 AM

packages/terra-form-input/src/Input.jsx
```diff
@@ -120,13 +125,18 @@ class Input extends React.Component {
...
+
+    const cxSupportedTypeClassNames = supportedTypes.includes(type) && cx(
```
neilpfeiffer, line 132: @benbcai something like this?
```suggestion
    const cxSupportedTypeClassNames = (supportedTypes.includes(type) || type === undefined ) && cx(
```

benbcai: @neilpfeiffer Yes!

packages/terra-form-input/src/terra-dev-site/test/form-input/SupportedTypesInput.test.jsx
neilpfeiffer: it's not a type, but since Terra doesn't set the type to "text" as default (even though HTML does when it isn't provided), it was added here as an extra test case because the component can be generated without a type since it is undefined by default and still needs to work. I see now the bad effect - generating `type=undefined` is undesired.

Since other tests already check for when the type isn't set, let's remove it.

packages/terra-form-input/src/Input.jsx
neilpfeiffer: +1 - remove `undefined` and instead put it with the check on 132 with something like https://github.com/cerner/terra-core/pull/3386#discussion_r615980437
```suggestion
    const supportedTypes = ['text', 'number', 'password', 'email', 'search', 'tel', 'url', 'hidden'];
```

packages/terra-form-input/src/Input.jsx
neilpfeiffer: Curious - would this short circuit faster and skip needing to do the array.includes() if type is not set? 
Since `undefined` is defaulted and will probably be the most common case:

```suggestion
    const cxSupportedTypeClassNames = (type === undefined || supportedTypes.includes(type)) && cx(
```

tbiethman: Would a defaultProp value to negate this case help? Or do we want to rely on the native type fallback?

packages/terra-form-input/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Update input to fallback to native defaults when the type is date, datetime-local, month, time, week, color, file, range, checkbox, radio, button, reset, or submit as they were never tested or originally intended to be supported.
```
ShettyAkarsh, line 6:
```suggestion
  * Update input to fallback to native defaults when the type is date, datetime-local, month, time, week, color, file, range, checkbox, radio, button, reset or submit as they were never tested or originally intended to be supported.
```

packages/terra-form-input/src/terra-dev-site/test/form-input/SupportedTypesInput.test.jsx
pranav300: I have updated wdio tests to be only for `supported` tests. And fixed the accessibility issue by replacing the `name` prop with `id`.

packages/terra-form-input/src/terra-dev-site/test/form-input/SupportedTypesInputField.test.jsx
pranav300: updated here https://github.com/cerner/terra-core/pull/3386/commits/451995dda70a0c87b88b1a69e7e802692ae84c2d

packages/terra-form-input/src/terra-dev-site/test/form-input/UnsupportedTypesInputField.test.jsx
pranav300: updated here https://github.com/cerner/terra-core/pull/3386/commits/451995dda70a0c87b88b1a69e7e802692ae84c2d

packages/terra-form-input/tests/wdio/form-input-spec.js
pranav300: Updated here https://github.com/cerner/terra-core/pull/3386/commits/451995dda70a0c87b88b1a69e7e802692ae84c2d.

packages/terra-form-input/tests/wdio/form-input-spec.js
pranav300: Removed the WDIO tests here https://github.com/cerner/terra-core/pull/3386/commits/2d2817ef44b58f6de8d9b1c0912b5958da012b5e

packages/terra-form-input/src/Input.jsx
pranav300: Updated here https://github.com/cerner/terra-core/pull/3386/commits/6ccb549abcd88c1ad6d0af5ea0318a0d38da4652

packages/terra-form-input/src/terra-dev-site/test/form-input/SupportedTypesInput.test.jsx
pranav300: Updated here https://github.com/cerner/terra-core/pull/3386/commits/6ccb549abcd88c1ad6d0af5ea0318a0d38da4652.

packages/terra-form-input/src/Input.jsx
pranav300: Updated here https://github.com/cerner/terra-core/pull/3386/commits/6ccb549abcd88c1ad6d0af5ea0318a0d38da4652.

benbcai (approved):

ShettyAkarsh (approved):

neilpfeiffer (approved):
Tested: PR #3386
Resolves: Issue #3378

**Reviewed:**
- Updated terra-form-input to provide Terra ui styling and behaviors to the types listed as supported, and allows for browser native fallback for the types listed as unsupported. https://github.com/cerner/terra-core/pull/3386#issuecomment-820194408

 **UX Reviewed** |
-----------|

tbiethman (approved):

supreethmr (approved):

PR #3400, opened 2021-03-18
Consume jest-config-terra

mjhenkes:
### Summary

This is an example upgrade to the proposed jest-config-terra configuration package.

Notes on the added config:
The i18n package vicariously tests the output of terra-aggregate translations. With the jest-config-terra package no other package needs to run it before jest testing.

Enzyme setup is not included in jest-config-terra as I'm trying to be agnostic to the enzyme/testing-library:react debate and the jest config should be able to be used in no react modules without extraneous dependencies.

The jest config has clear mocks set to true, this uncovered at least one incorrect test that relied on a mocked value in a previous test.

I removed the file mock in favor of just using a string in the affected test files.

### Deployment Link

### Additional Details

@cerner/terra


StephenEsser (approved):

packages/terra-form-select/CHANGELOG.md
```diff
@@ -11,6 +11,9 @@
   * Fixed `focus not getting removed` in IE for search and combobox variants.
   * Fixed `focus not getting removed` in IE for multiple and tag variants.
 
+* Changed
+  * Updated Jest command
```
benbcai, line 15: Do we want to combine this with the above `Changed` section or leave it like this to save a build?

mjhenkes: yeah, good call.

packages/terra-base/CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Changed
   * Removed unnecessary `wdio.conf.js` file
   * Upgrade tests to terra-functional-testing and WebDriverIO v6
+  * Updated jest tests verify the custom prop validation in another way.
```
benbcai, line 8: We can ignore this if we want to save a build.

```suggestion
  * Updated jest tests to verify the custom prop validation in another way.
```

benbcai (approved):

ryanthemanuel (approved):

PR #3404, opened 2021-03-19
[terra-icon] Bump to OCS Icons v1.38.0

jeremyfuksa:
### Summary

Bumped to v1.38.0 and got low light tests to pass.

Changes:
I've updated the generator to not create static lowlight icons with themeing, the lowlight icons cannot be themed and are locked into the lowlight version. I also removed the empty lowlight scss files that threw warnings in webpack.

I originally intended to alias the lowlight icons to their themable non-lowlight variants, but the lowlight theme and the lowlight static icons are not equivalent. There is more work to do to theme icons for lowlight, but for now I thought we should move forward with adding the new icons.


package.json
```diff
@@ -110,5 +110,9 @@
     "webpack-cli": "^3.3.2",
     "webpack-dev-server": "^3.3.1",
     "webpack-merge": "^4.1.2"
+  },
+  "dependencies": {
+    "@cerner/jest-config-terra": "^1.0.0",
+    "@cerner/terra-aggregate-translations": "^2.0.1"
```
mjhenkes, line 116: i think these can be removed

packages/terra-icon/src/icon/IconAlertLowLight.jsx
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 import classNamesBind from 'classnames/bind';
 import ThemeContext from 'terra-theme-context';
 import IconBase from '../IconBase';
-import styles from '../IconAlertLowLight.module.scss';
+import styles from '../clinical-lowlight-theme/IconAlert.module.scss';
```
mjhenkes, line 7: this change doesn't seem right, we shouldn't be pulling the lowlight theme directly

packages/terra-icon/scripts/src/generate-icon/renderJsx.js
```diff
@@ -24,7 +25,11 @@ const staticIcons = iconData
 const renderJsx = icon => new Promise((resolve, reject) => {
   let templatePath;
   if (staticIcons.includes(icon.name)) {
-    templatePath = staticIconTemplatePath;
+    if (icon.name.endsWith('LowLight')) {
```
mjhenkes, line 28: We now generate lowlight icons differently than normal static icons because they don't support theming.

packages/terra-icon/CHANGELOG.md
```diff
@@ -5,10 +5,8 @@
 ## 3.42.0 - (April 6, 2021)
 
 * Changed
-  * Removed unnecessary `wdio.conf.js` file
-  * Upgrade tests to terra-functional-testing and WebDriverIO v6
-  * Updated Jest command
-  * Updated screenshots with elements that are out of bound
+  * Bumped `one-cerner-style-icons` to v1.38.0
+  * Minor version bumps to dependencies
```
StephenEsser, line 9: I think there are a few conflicts here. Some items were removed and the new changes are in a previously release section

StephenEsser (approved):

packages/terra-icon/scripts/src/generate-icon/staticIconLowLightTemplate.txt
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable */
+import React from 'react';
+import classNames from 'classnames';
+import IconBase from '../IconBase';
+
+const SvgIcon = (customProps) => {
+  const attributes = Object.assign({}, customProps);
+  const iconClassNames = classNames(
+    customProps.className,
+  );
```
StephenEsser, line 10: Might not be worth the effort but at this point we could ditch class names entirely and let them pass through with custom props

mjhenkes: i did think about this, but didn't want to mess with it

tbiethman (approved):

packages/terra-icon/CHANGELOG.md
mjhenkes:
```suggestion
  * Removed unnecessary `wdio.conf.js` file
  * Upgrade tests to terra-functional-testing and WebDriverIO v6
  * Updated Jest command
  * Updated screenshots with elements that are out of bound
```

jeremyfuksa (commented):
+1

PR #3410, opened 2021-03-24
[terra-toolbar] Removed extra padding due to null children

### LINKED ISSUE #3401: [terra-toolbar] Padding applied to null children in Toolbar
# Bug Report

## Description

In the typical use-case of conditionally rendering a child component to a terra-toolbar Toolbar component (i.e. setting the child to null so it will not be rendered), the default padding is still being applied to that child component.

## Steps to Reproduce

Code sample:
```jsx
render() {
  const condition = /* some condition */ false;
  return (
    <Toolbar>
      {condition ? <Button text="First" /> : null}
      <Button text="Second" />
    </Toolbar>
  );
}
```

Two children rendered normally:
image

Only rendering second child (first is null):
image

image

## Expected Behavior

No extra padding is added when a child component is null/falsy.

## Possible Solution

Checking if the child is falsy before applying padding.

## Environment

* Component Name and Version: terra-toolbar 1.20
* Browser Name and Version: Chrome 89
* Node/npm Version: [Node 10.16.0/npm 6.9.0]
* Webpack Version: 4.46.0
* Operating System and version (desktop or mobile): macOS Catalina 10.15.7

## @ Mentions

@tbiethman

saket2403:
### Summary
Fix to remove extra padding that was being applied due to conditional rendering of a null child .
Closes #3401 

### Deployment Link

### Additional Details

@cerner/terra


mjhenkes:
@saket2403 could you add a test for this?

saket2403:
> @saket2403 could you add a test for this?

Added c4724c2

packages/terra-toolbar/tests/wdio/toolbar-spec.js
```diff
@@ -18,6 +18,17 @@ Terra.describeViewports('Toolbar', ['tiny'], () => {
     browser.url('/raw/tests/terra-toolbar/toolbar/center-align-toolbar');
     Terra.validates.element('center align');
   });
+
+  describe('Conditional Toolbar', () => {
+    it('displays two buttons', () => {
+      browser.url('/raw/tests/terra-toolbar/toolbar/conditional-toolbar');
+      Terra.validates.element('both buttons');
+    });
+    it('displays one buttons', () => {
```
benbcai, line 27:
```suggestion
    it('displays one button', () => {
```

mjhenkes (approved):

benbcai (approved):

packages/terra-toolbar/src/Toolbar.jsx
```diff
@@ -36,7 +36,7 @@ const Toolbar = ({
     customProps.className,
   );
 
-  const items = React.Children.map(children, item => (<div className={cx('item')}>{item}</div>));
+  const items = React.Children.map(children, item => (item ? <div className={cx('item')}>{item}</div> : null));
```
StephenEsser, line 39: Does this only happen for `null` children? Does it also happen for `undefined` children?

```jsx
<Toolbar>
      {showButton && <Button text="Button 1" />}
 </Toolkit>

Comment (emilyrohrbough) [reply packages/terra-toolbar/src/Toolbar.jsx:39]: Could you check if (React.Children.count(children) > 0)?
```js
let items;
if (React.Children.count(children) > 0) {
  items = React.Children.map(children, item => (<div className={cx('item')}>{item}</div>));
}
```

yuderekyu: or variation of above that just checks for a truthy value with a ternary
```js
const items = React.Children.count(children) ? React.Children.map(children, item => (<div className={cx('item')}>{item}</div>)) : null;
```

saket2403: > Does this only happen for `null` children? Does it also happen for `undefined` children? What happens in this scenario before/after the fix above?
 
The scenario in the snippet works fine with the changes I have made.

> Could you check if (React.Children.count(children) > 0)?

It is not working as expected with React.children.count(), reason being that the child count is displaying same even if it is a null / undefined child. So I checked directly if the child itself is a truthy value.

ShettyAkarsh (approved):

packages/terra-toolbar/src/Toolbar.jsx
ryanthemanuel: Let's return 'item' instead of null here.
```suggestion
  const items = React.Children.map(children, item => (item ? <div className={cx('item')}>{item}</div> : item));
```

saket2403: Done

supreethmr (approved):

lokesh-0813 (approved):

PR #3391, opened 2021-03-09
[terra-signature] Fix touch screen interaction on Windows

bbanothu:
### Summary
 
The reason for these changes is to fix an issue with signing on windows touch screen devices. 

**A description of the original issue:**
esignature on the web would not allow for signing via touch on any windows based device with touch screen capabbilities (i.e surface pro)

**Steps to reproduce the original issue:**
1. Boot up the terra dev site.
2. go into signature. 
3 then use touch inputs to sign on any windows device that allows for touch inputs.

**A description of the changes that have been made, why they were made, and how they are resolving the issue:**
1. removed the part that listens to "onTouchStart". This makes sure that touch inputs and mouse inputs work at all times.
2. Added checks for touchInputs vs mouseInputs because the x & y coordinates are in different objects here. this makes sure that there are not null values. 

**Steps to validate the changes are working as expected:**
1. Boot up the terra dev site.
2. Go into signature. 
3. Attempt to sign using mouse and then by touch. - on a windows touch device
4. Use a chrome emulator or repeat steps 1-3 on an ipad/iphone/phone as well. 

### Testing

This was done manually on:
Ipad, Surface pro, and Mac using safari, chrome, IE, and Edge

### Additional Details

Checking for onStart on page load does not allow for windows touch base devices (i.e Surface) to be able to use touch inputs. 

Thank you for contributing to Terra.
@cerner/terra

PR #3436, opened 2021-05-07
mouseDown touch-input NaN/Null Fix

bbanothu:
### Summary
 
The reason for these changes is to fix an issue with signing on windows touch screen devices. On mouseDown was not done correctly which leads to Nan Values. 

**A description of the original issue:**
esignature on the web would not allow for signing via touch on any windows based device with touch screen capabbilities (i.e surface pro). Fixed a bug an a preexisting pr. 

**Steps to reproduce the original issue:**
1. Boot up the terra dev site.
2. go into signature. 
3. add console log for the lineSegements on mouse down.
4. attempt to make two seperate lines with touch inputs and then check the values of the line segments. it contains Nan's. 

**A description of the changes that have been made, why they were made, and how they are resolving the issue:**
2. Added checks for touchInputs vs mouseInputs on mouseDown. 

**Steps to validate the changes are working as expected:**
1. Boot up the terra dev site.
2. go into signature. 
3. add console log for the lineSegements on mouse down.
4. attempt to make two seperate lines with touch inputs and then check the values of the line segments. it should no longer contain nan's. 

### Testing

This was done manually on:
Ipad, Surface pro, and Mac using safari, chrome, IE, and Edge

### Additional Details

A fix on a previous Pr -> https://github.com/cerner/terra-core/pull/3391 

@cerner/terra


vikonava (approved):

CONTRIBUTORS.md
```diff
@@ -1,5 +1,6 @@
 Cerner Corporation
 - Bharath Banothu [@bbanothu]
+- Bharath Banothu [@bbanothu]
```
mjhenkes, line 3: This looks like a duplicate entry.

packages/terra-signature/src/Signature.jsx
```diff
@@ -115,7 +115,11 @@ class Signature extends React.Component {
 
     this.canvasRect = this.canvas.getBoundingClientRect();
 
-    this.addLine(event.pageX - this.canvasRect.left, event.pageY - this.canvasRect.top, false);
+    if (event.type === 'touchstart') {
```
mjhenkes, line 118: are there any tests you could write for this?

CONTRIBUTORS.md
bbanothu: Fixed, I thought I needed to reAdd my name again for a new merge.

packages/terra-signature/src/Signature.jsx
bbanothu: https://github.com/cerner/terra-core/pull/3436/files/61110e29625b4f0be9cdb948403e8abf0c183ce6#diff-101895006107b419f432f22c93db224b2ab5f26689e324d4a7e89f1914ed3604R104

This PR just contains one check that wasn't made during the last PR for this bug. Similar to all of these changes in mouseInBounds and mouseXY. We were not able to identify or write any of tests for this last time so I do not think there are any.

dkasper-was-taken (approved):

PR #3463, opened 2021-07-09
[DO NOT MERGE] terra-pills test PR.

ShettyAkarsh:
PR is to test in IE browser using the deployment link generated


packages/terra-core-docs/src/terra-dev-site/doc/pills/example/DisclosurePill.jsx
```diff
@@ -0,0 +1,56 @@
+import React, { useRef, useState } from 'react';
+import uniqueid from 'lodash.uniqueid';
+import Popup from 'terra-popup';
```
tbiethman, line 3: It doesn't look like terra-core-docs has an explicit dependency on terra-popup, we should add one.

PR #3464, opened 2021-07-12
[terra-form-select] Fix focus issue in menu with large number of options in mobile devices

### LINKED ISSUE #3451: terra-form-select 'search' variant does not focus when there is a large number of options
# Bug Report

## Description

When the terra-form-select 'search' variant has a large number of options (400+), the menu is not focused correctly which results in the dropdown remaining open when clicking on other fields.

## Steps to Reproduce

1. Create a search select field with 400+ options
2. Click the dropdown arrow to open the dropdown
3. Click elsewhere on the screen - the dropdown does not close because it was not focused
4. *Happens most frequently on the first opening of the dropdown

Video:
https://user-images.githubusercontent.com/85512188/121061339-bb361080-c791-11eb-9124-dd06a25dbb71.mov

## Expected Behavior

The dropdown should still be focused even if there is a large number of options.

## Possible Solution

The menu is not yet painted to the DOM when trying to focus here https://github.com/cerner/terra-core/blob/main/packages/terra-form-select/src/shared/_FrameUtil.js#L150. In other words, that 100 ms timeout is not long enough. One solution I have noticed to fix the issue is adding a requestAnimationFrame inside the timeout, around the frame.selectMenu.focus();. 

## Environment

* Component Name and Version: terra-form-select 6.29.0
* Browser Name and Version: Chrome 91.0.4472.77 
* Node/npm Version: [e.g. Node 8/npm 5] Node 10/npm 6
* Webpack Version: Webpack v4
* Operating System and version (desktop or mobile): Mac, issue only occurs on mobile

## @ Mentions

@supreethmr

pranav300:
### Summary

This PR fixes the delay in rendering the menu options which was causing focus issues in mobile devices.

Closes #3451

### Deployment Link

Link to the test: [many-options](https://terra-core-fix-select-bujg16vl.herokuapp.com/tests/terra-form-select/form-select/many-options)

### Testing

Added a new test example to verify the changes.

### Additional Details

@cerner/terra


packages/terra-form-select/src/terra-dev-site/test/form-select/SelectVariantsWithChildrenGreaterThan500.test.jsx
```diff
@@ -0,0 +1,46 @@
+import React, { useState } from 'react';
```
pranav300, line 1: This file will be removed once the changes are verified.

packages/terra-form-select/src/terra-dev-site/test/form-select/SelectVariantsWith500Options.test.jsx
```diff
@@ -0,0 +1,46 @@
+import React, { useState } from 'react';
```
pranav300, line 1: This file is just to verify the changes and will be removed before merging them.

packages/terra-form-select/src/search/Frame.jsx
```diff
@@ -433,7 +433,7 @@ class Frame extends React.Component {
       || event.target.hasAttribute('data-terra-form-select-toggle-button-icon'))) {
       this.setState({ isOpen: true, isPositioned: false });
 
-      FrameUtil.shiftFocusToMenu(this);
+      setTimeout(() => FrameUtil.shiftFocusToMenu(this), 100);
```
supreethmr, line 436: FrameUtil.shiftFocusToMenu() has already has wait time of 100ms. this will add up 100ms more wait time to that function. if increasing time is the only way to fix this then we can do it in FrameUtil itself instead adding it in each frame files.

Have you tried using useLayoutEffect() hook to identify whether the menu component is loaded up with items and then shifting focus to menu..?

packages/terra-form-select/src/terra-dev-site/test/form-select/SelectVariantsWith500Options.test.jsx
supreethmr: we might need to keep this file to test the PR code changes with WDIO.

pranav300: I have removed this test. Changes can be verified in this test https://github.com/cerner/terra-core/pull/3464/files#diff-5285632fbb1d1102a5455105fc54dbfe0e7adf24e64a59f1eeb38ea2cdfd4736R8

packages/terra-form-select/src/terra-dev-site/test/form-select/ManyOptions.test.jsx
```diff
@@ -29,7 +38,7 @@ const ManyOptionsExample = () => {
       <p>
         Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
       </p>
-      <Select id="frmSelectSingle" placeholder="Select a number" variant="default" maxHeight={200}>
+      <Select id="frmSelectSingle" placeholder="Select a number for Tag" variant="tag" maxHeight={200}>
```
supreethmr, line 41: Id needs to be changed it still refers to single form-select. I think it can be removed as it is not used in any of the WDIO specs.

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -486,6 +488,11 @@ class Frame extends React.Component {
         this.dropdown.removeAttribute('inert');
         this.dropdown.removeAttribute('aria-hidden');
         this.selectMenu.setAttribute('tabIndex', '0');
+
+        // sets focus to select menu after select menu is positioned when opened by toggle-button click
+        if (this.selectMenu && this.shouldFocusDropdown) {
+          this.selectMenu.focus();
```
benbcai, line 494: Do we still need to call `FrameUtil.shiftFocusToMenu(this)` which "[allows time for state update to render select menu DOM before shifting focus to it](https://github.com/cerner/terra-core/blob/4e5dbb730f93dfafb61f23a539cf55186e591f8c/packages/terra-form-select/src/shared/_FrameUtil.js#L144)" instead of calling `this.selectMenu.focus()` directly?

supreethmr (approved):

packages/terra-form-select/src/terra-dev-site/test/form-select/ManyOptions.test.jsx
pranav300: Reverted the changes made to this file.

packages/terra-form-select/src/combobox/Frame.jsx
pranav300: I don't think it is needed now. As earlier we were calling `FrameUtil.shiftFocusToMenu(this)` inside the `openDropdown`. But now we are passing focus inside the `updateDropdown` which is called by `positionDropdown` triggered by `onResize` event. So in that case the dropdown is already rendered.

benbcai (approved):

mjhenkes (approved):

ShettyAkarsh (approved):

PR #3475, opened 2021-08-07
[terra-core] Upgrade to terra functional testing v2

### LINKED ISSUE #3460: [terra-core] Upgrade to terra-functional-testing@2.0.0
# Story

## Description

As a developer for terra, I'd like to be able to utilize the latest version of terra-functional-testing (and thus the latest version of wdio)

## Motivation

So that we can eventually be on the latest stable release of Node to ensure we continue to receive supportability for Node

## Definition of Done

We'll know we're done when we've upgraded to 2.0.0 of terra-functional-testing and the code is merged to main

pranav300:
### Summary

This PR upgrades terra-core to use terra-functional-testing v2.

Closes #3460 

@cerner/terra


package.json
```diff
@@ -95,6 +95,7 @@
     "gh-pages": "^2.0.1",
     "glob": "^7.1.2",
     "jest": "^26.6.3",
+    "ky": "^0.12.0",
```
pranav300, line 98: This was added because a different version of it was getting installed in the root node_modules due to `terra-functional-testing -> webdriverio -> webdriver`. This was causing the danger command to fail in Travis.

mjhenkes: interesting, thanks for noting why it was added.

packages/terra-overlay/tests/wdio/overlay-spec.js
```diff
@@ -39,7 +39,8 @@ Terra.describeViewports('Overlay', ['huge'], () => {
 
       it('Clicks on the Full screen overlay button', () => {
         $('#trigger_fullscreen').click();
-        expect($('#root')).toHaveAttribute('inert', ['', 'true']); // chrome returns true, firefox returns ''
```
benbcai, line 42: What error did you get from expect-webdriverio V3? Does it no longer allow to use an array in the second argument for `toHaveAttribute`? I wonder if we should add an example in the [terra-functional-testing V2 upgrade guide](https://engineering.cerner.com/terra-ui/dev_tools/cerner-terra-toolkit-docs/terra-functional-testing/upgrade-guides/version-2-upgrade-guide) so that consumers can follow.

pranav300: > What error did you get from expect-webdriverio V3?

`Expected ['', true] but received 'true'` for chrome.  

> Does it no longer allow to use of an array in the second argument for toHaveAttribute?

It does allow to enter an array as a second argument. But now it doesn't allow to pass an array with possible expectations, rather the whole array should be matching.

Earlier, we used [toHaveAttributeFn](https://github.com/webdriverio/expect-webdriverio/blob/v1.4.1/src/matchers/element/toHaveAttribute.ts#L17). Which used to check if the expectation is present in the array or not.

But now, we use [toHaveAttributeAndValueFn](https://github.com/webdriverio/expect-webdriverio/blob/v1.4.1/src/matchers/element/toHaveAttribute.ts#L17). And this expects the values to be matching. So, when we pass an array as an expectation, then it must be matched with an array.

benbcai: Thanks for the explanation. I can't find any team using `toHaveAttribute` in their tests so they shouldn't run into this so probably don't need to add this in the upgrade guide.

benbcai (approved):

ShettyAkarsh (approved):

PR #3492, opened 2021-09-28
[terra-grid] Updated css to avoid dart sass deprecation warning.

mjhenkes:
### Summary

Dart sass warns on the use of the `/` operator and suggests that we switch it to `math.div`, however `math.div` is only supported in dart sass and not in node sass. To avoid a major version bump and to support older and newer versions of webpack-config-terra. i've removed the division case written out what the results would be.

Closes #

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-grid/src/Grid.module.scss
```diff
@@ -28,22 +27,57 @@
...
+        max-width: 8.3333333333%;
+      }
+      .column-#{$size-name}-2 {
+        flex-basis: 16.6666666667%;
+        max-width: 16.6666666667%;
+      }
+      .column-#{$size-name}-3 {
+        flex-basis: 25%;
+        max-width: 25%; }
+      .column-#{$size-name}-4 {
```
benbcai, line 46:
```suggestion
        max-width: 25%;
      }
      .column-#{$size-name}-4 {
```

packages/terra-grid/src/Grid.module.scss
```diff
@@ -28,22 +27,57 @@
     padding-left: var(--terra-grid-gutter-padding-left, 1rem);
   }
 
-  // Create column classes
-  @for $i from 1 through 12 {
-    .column-#{$i} {
```
benbcai, line 33: Are these column classes no longer needed and written out like below?

mjhenkes: no, they were never used

tbiethman (approved):

benbcai (approved):

ryanthemanuel (approved):

PR #3515, opened 2021-10-29
Allow tabIndex prop to be set.

elliott-hoffman-cerner:
Before this change, Buttons would ignore the tabIndex prop and set tabIndex soley by the isDisabled prop.

### Summary
With this change, users can set the tabIndex prop if they wish. isDisabled will still force TabIndex to -1 for passivity. I'm making this change as part of a larger change to Time Input, where I need a button to be enabled yet skipped over in tab order because I'm emulating the keyboard commands of a radio button group.

### Deployment Link
https://terra-core-tabindex-ibh6efjmea.herokuapp.com/tests/cerner-terra-core-docs/button/tab-index

### Testing
I tested by setting the isDisabled and tabIndex props on a button.


packages/terra-button/src/Button.jsx
```diff
@@ -303,11 +303,11 @@ class Button extends React.Component {
 
     return (
       <ComponentType
+        tabIndex={isDisabled ? '-1' : undefined}
```
benbcai, line 306: I think the change makes sense. The pattern that we typically follow is to add `tabIndex` as a first class prop and then set the tabIndex if one is provided.

`tabIndex={isDisabled ? '-1' : tabIndex}`

elliott-hoffman-cerner: Since tabIndex is a normal DOM attr, why would we expose it as a 1st class prop in our component? As a pattern - there are so many DOM attrs we implicitly support, how do we decide which make it into 1st class prop?

benbcai: I don't think we have any hard rules to determine whether or not a prop must be first-class or can just be a passthrough. Generally speaking, we'd want to expose a prop as first-class to consumers when the component is more opinionated about this prop or want to have more control on how this prop is applied. 

One scenario this change may not work as intended is when the consumer passed in a tabIndex of 1 with `isDisabled` set to true. Previously it would set tabIndex to -1 but now it would honor the tabIndex to 1 that is passed in when it is disabled. We also usually spread the customProps (`{...customProps}`) before setting all the first-class props so that these first-class props have higher precedence over all the passthrough props.

packages/terra-core-docs/src/terra-dev-site/test/button/TabIndex.test.jsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import Button from 'terra-button';
```
benbcai, line 2: Given that we now have this test file, we should add a test step in https://github.com/cerner/terra-core/blob/main/packages/terra-button/tests/wdio/button-spec.js to actually use this test example. Otherwise, this isn't being tested except just having the test example.

packages/terra-button/src/Button.jsx
```diff
@@ -82,6 +82,10 @@ const propTypes = {
    * Callback ref to pass into the dom element.
    */
   refCallback: PropTypes.func,
+  /**
+   * Just like normal tabIndex, but if isDisabled then it will be forced to -1.
+   */
+  tabIndex: PropTypes.string,
```
yuderekyu, line 88: should this be a number or string? i've seen design systems support both.

elliott-hoffman-cerner: No, it is not valid React to pass in a number like ```-1``` as a prop like that. At least not according to my linter settings.

packages/terra-button/src/Button.jsx
neilpfeiffer: Generally, Terra recommends that consumers and projects do not set custom tabindex for interactive elements (even though they may be tempted to), as it destroys proper accessibility.

- `-1` on non-disabled + interactive elements is bad, unless for very rare and specific places that only receive programmatic focus, which usually are higher-order components and usages that already have a very keen understanding of proper accessibility details needing to be included (i.e. not general `terra-button` use).
- `0` on <button> and <a> elements is superfluous and redundant, as such, is bad.
- `1` (or any number greater than 1) is bad, as it supersedes the natural flow for DOM order keyboard navigation, and should not be used ever by anyone https://webaim.org/techniques/keyboard/tabindex and violates [jsx-eslint/.../ tabindex-no-positive.md](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/tabindex-no-positive.md) and would need overriding.
  
 \
 \
Which is why this prop was never exposed before as first-class. However, it also was probably why it was never tested via spreading customProps as @benbcai mentioned, to see if we were swallowing it or honoring it if provided.
  
Without very specific reusability reasons, I would prefer rather than supporting this first-class, it would be safer to write separate guide docs for how spreading customProps works, include an example for advanced component use where this specific case is needed for `terra-button`, and explain risks of what can be screwed up if done incorrectly. 

And correct the swallowing of the tabindex value if provided by consumers via customProps and honor it (unless is disabled) by changing to this:

```
        tabIndex={isDisabled ? '-1' : customProps.tabIndex}
```

packages/terra-button/src/Button.jsx
```diff
@@ -307,7 +313,7 @@ class Button extends React.Component {
         className={buttonClasses}
         type={type}
         disabled={isDisabled}
-        tabIndex={isDisabled ? '-1' : undefined}
+        tabIndex={isDisabled ? '-1' : tabIndex}
```
neilpfeiffer, line 316:
```suggestion
        tabIndex={isDisabled ? '-1' : customProps.tabIndex}
```

neilpfeiffer (changes requested):

packages/terra-button/src/Button.jsx
yuderekyu: 🤷 
* [semantic ui](https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Checkbox/Checkbox.js#L294)
* [material ui](https://github.com/mui-org/material-ui/blob/master/packages/mui-material-next/src/Button/Button.js#L639)
* [carbon design](https://github.com/carbon-design-system/carbon/blob/main/packages/react/src/components/Tooltip/Tooltip.js#L191)
* [dash](https://github.com/plotly/dash/blob/dev/components/dash-core-components/src/components/Textarea.react.js#L195)

packages/terra-button/src/Button.jsx
benbcai: That's a good point @neilpfeiffer. If we don't want consumers to set tabIndex, then we shouldn't make it a first-class prop.

packages/terra-button/src/Button.jsx
elliott-hoffman-cerner: TIL, thank you. I'm going to remove the top-level prop all together based on @neilpfeiffer's comment.

elliott-hoffman-cerner: TIL, thank you.

elliott-hoffman-cerner:
Well, I've screwed up this PR somehow. Going to start fresh and open another based on the feedback from this PR.

packages/terra-button/src/Button.jsx
neilpfeiffer: Interesting that Semantic and Dash still allow for one of type number or string, Material and Carbon are fine with just number.

PR #3535, opened 2021-12-01
Enabled child buttons are toggle buttons

elliott-hoffman-cerner:
### Summary
Now a button group's buttons will render aria-pressed unless they are disabled. Previously, only selected buttons would render aria-pressed - which meant screen readers would read the selected button as "Toggle button" but the other enabled buttons as "Button".  Now both selected and unselected buttons will read as "toggle button". Disabled buttons will still be read as "Button"

### Deployment Link
TBD

### Testing
See new jest test and snap.

### Additional Details
I came across this while testing my A11Y Time Input enhancement.


packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -87,7 +87,7 @@ class ButtonGroup extends React.Component {
       const cloneChild = React.cloneElement(child, {
         onClick: this.wrapOnClick(child),
         className: cx([{ 'is-selected': isSelected && !child.props.isDisabled }, child.props.className]),
-        'aria-pressed': isSelected || null,
+        'aria-pressed': child.props.isDisabled ? null : isSelected,
```
benbcai, line 90: It makes sense that aria-pressed should be true for an active and selected button. If the button is disabled, should aria-pressed be `null` like how it is with the change or should it literally be set to false?

elliott-hoffman-cerner: Initially I thought it should be false if the button is disabled, but then I found the PR of that line from three years ago: https://github.com/cerner/terra-core/pull/2231#issue-410470597, which indicated we don't want it that way.

benbcai (approved):

jmsv6d (approved):

PR #3550, opened 2021-12-22
[terra-search-field] Migrate Docs to terra-core-docs

saket2403:
### Summary

Migrate Docs to terra-core-docs

Closes #

### Deployment Link

### Additional Details

@cerner/terra


ShettyAkarsh (approved):

ShettyAkarsh:
`https://github.com/cerner/terra-core/blob/search-docs/packages/terra-search-field/package.json#L33-L34` can you check if these dependencies are used only for doc purposes and remove it in case yes ?

saket2403:
> https://github.com/cerner/terra-core/blob/search-docs/packages/terra-search-field/package.json#L33-L34 can you check if these dependencies are used only for doc purposes and remove it in case yes ? #

Both dependencies are used for the search button.

benbcai (approved):

PR #3575, opened 2022-02-02
Anchor Button A11y

jmsv6d:
### Summary

Added `button` role for anchor buttons. 

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-button/src/Button.jsx
```diff
@@ -319,6 +319,7 @@ class Button extends React.Component {
         onFocus={this.handleFocus}
         href={href}
         ref={refCallback}
+        role={href ? 'button' : undefined}
```
supreethmr, line 322: Can <button> be used instead of <a> and `handleClick` be updated similar to an `handleKeyDown()` to update [window.location.href](https://github.com/cerner/terra-core/blob/e9221d84673c1a5c58f23b9121cd8e9830dffe3d/packages/terra-button/src/Button.jsx#L170) on click. 
So if it is possible using button rather than the button role will remove the any additional customization needed, as native HTML buttons are supported by all user agents and assistive technology and provide keyboard and focus requirements by default, without need for additional customization.

jmsv6d: I'm not sure I'm understanding what you're asking. Is the suggestion to remove the logic of flexing to render an `a` element when the `href` prop is present and to do manual link navigation onClick for the `button` element?

xenoworf (approved):

packages/terra-button/src/Button.jsx
supreethmr: yes that's correct !! instead of relying on role `button` we can use <button> element. native button element is better supported by all user agents and assistive technology compare to role button.

jmsv6d: So, that change of approach would change the scope of this story. This was intended to be a small change while we evaluate the need for an anchor buttons. In future, we were going to potentially break the anchor button out to being it's own component, we could investigate this approach as part of that work.

sdadn (approved):

supreethmr (approved):

jeremyfuksa (approved):
This change is approved.

PR #3576, opened 2022-02-08
[terra-icon] OCS icons v1.42.1

jeremyfuksa:
### Summary

Bumps OCS icon library dependency to v1.42.1.


eawww (changes requested):
Can't comment on binary files but the new icons are missing from the wdio snapshots.

packages/terra-icon/package.json
```diff
@@ -23,7 +23,7 @@
   "devDependencies": {
     "csvtojson": "^2.0.8",
     "jsdom": "^15.1.1",
-    "one-cerner-style-icons": "https://github.com/cerner/one-cerner-style-icons/archive/v1.41.2.tar.gz",
+    "one-cerner-style-icons": "https://github.com/cerner/one-cerner-style-icons/archive/refs/tags/v1.42.1.tar.gz",
```
xenoworf, line 26: Seems odd to pull this from github instead of npm. @jmsv6d or @benbcai why don't we put this on NPM?

packages/terra-icon/src/icon/IconVolumeSetDecrease.jsx
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable */
```
xenoworf, line 1: Seems like a bad practice to disable eslint entirely. I'd prefer leaving it on and opting out of _some_ rules only if they're really a problem. Tell me: why turn off linting?

packages/terra-icon/src/svg/volumeSet_decrease.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 48 48"><path d="M29.121 48.059L15.087 34.024H.05V13.976h15.037L29.121-.059zm6.323-35.406l-2.131 2.131a13.009 13.009 0 010 18.43l2.131 2.131a16.039 16.039 0 000-22.692z" data-name="256"/></svg>
```
xenoworf, line 1: Please add a newline to the end of all files.

packages/terra-icon/package.json
jeremyfuksa: Chalk this one up to, "This is how it was always done." I'd be interested to hear why this isn't on NPM as well.

packages/terra-icon/src/svg/volumeSet_decrease.svg
xenoworf: Same comment about generated files. N/m: not worth it. Maybe we can fix the script to emit newline. But low value if these are generated.

packages/terra-icon/src/icon/IconVolumeSetDecrease.jsx
xenoworf: @jeremyfuksa told me these files are generated, so n/m: low value in linting them.

xenoworf (approved):

packages/terra-icon/src/icon/IconVolumeSetDecrease.jsx
eawww: Aw I had a detailed comment typed up and everything. I should refresh more often. Oh well, no sense letting documentation and context go to waste:
These files are generated automatically by [this script](https://github.com/cerner/terra-core/blob/main/packages/terra-icon/scripts/src/generate-icon/renderJsx.js) using these templates: [template.txt](https://github.com/cerner/terra-core/blob/main/packages/terra-icon/scripts/src/generate-icon/template.txt), [staticIconTemplate.txt](https://github.com/cerner/terra-core/blob/main/packages/terra-icon/scripts/src/generate-icon/staticIconTemplate.txt). [All of the icons](https://github.com/cerner/terra-core/tree/main/packages/terra-icon/src/icon) are this way.
Without this, it flags tons of silly deviations from our linting rules that can vary depending on the raw SVG transpiled into the templates from [one-cerner-style-icons](https://github.com/cerner/one-cerner-style-icons/tree/main/src/icons). I 💯 agree that it's bad practice, especially since it could let just any ol' code through from the raw SVGs. However, this would require changing the templates to meet terra's linting standards, ensuring all the SVG in the OCS icons transpiles correctly, and re-generating the jsx files for all of the icons, which seems like it's a separate issue and beyond the scope of this PR. Backlog?

eawww (approved):

packages/terra-icon/src/svg/volumeSet_decrease.svg
jeremyfuksa: Fixed the migration script to inject new lines.

packages/terra-icon/package.json
benbcai: I think the reason is because one-cerner-style-icons is for Cerner use only and not intended to be consumed directly. It should only be consumed by terra-icon, which is published to NPM.

benbcai (approved):

jeremyfuksa:
Closing in favor of #3601.

PR #3581, opened 2022-02-18
[terra-core-docs] Feature: Icon Collection Doc Makeover (🚧WIP)

eawww:
### Summary

Uplifting the icon doc site because we've identified the following issues with it:
- Separating icons out into "static" and "themable" doesn't really make sense because many of the "static" icons have "themable" elements
- It currently offers no guidance as to the intended uses for each icon
- Icons are listed in the order they were added to the library, which is isn't super helpful for finding anything.
- Having a column for the entire import statement is a waste of space
- This is a frequently used reference across a variety of roles and it deserves some love

Final product will include several interactive elements for quality-of-life. Here is a before/after with a rough mockup of our concept so far:
_Before:_
image

_After:_
image

Feedback on the concept or requests for additional features can be left as comments on this PR.

For code review digestibility, work will be completed in discrete chunks and merged into this branch.

- [ ] Basic Table Structure #3582
- [ ] Search Functionality
- [ ] Import statement copy button
- [ ] Function column filtering
- [ ] Categories/sets view
- [ ] Data pull from [One Cerner Style Icons](https://github.com/cerner/one-cerner-style-icons/)
  - We are still working on updating this repo with the new required data
  - [ ] Update that repo with the new required data


neilpfeiffer:
Questions/Comments:
1. What will be the layout when the page content area is less than 800-900px or so? Do you have a responsive small/mobile layout, since this table already is pretty full for the desktop layout and will need an alternate presentation other than fluid reflow. 
2. The search bar (would follow the "Component / Scoped Search" pattern in Design Standards) feels like it could use more design and integration into the table/page rather than simply stretching the full width across the page, include an informational header with number of matching results count, clear search action, etc. Will need to confirm if you will need to use the explicit search button, not simply reacting after {x} letter presses.
3. The import statement icon/link copy feature seems a little out of place with the current site. Perhaps  a better option is to add a "<code>" toggle in the last column, similar to the current doc-example template, with an expand/collapse section below the row like an accordion to show the single icon use code sample, or import statement faceup.
4. We'll need to have documentation explaining the Function Types ("Action", "Status", "Toggle").
4. The "Themeable Elements" column isn't quite clear, as it's not a direct relationship to theming. Consumers can provide a color (which that color they _should_ include with scoped theme variables and colors), so it may require a better link to explaining theme responsibilities for consumers. It also makes it sound like the items not-listed are "not theme-able" when actually they are already themed, but controlled by the component.

eawww:
@neilpfeiffer 
> What will be the layout when the page content area is less than 800-900px or so?
Right now, the intended behavior is for it to scroll horizontally because the content we intend to squeeze in here won't possibly fit in a smaller viewport. My assumption is that this will rarely be used on mobile devices since it is primarily a used as reference for work that is done in a desktop context.

> The search bar (would follow the "Component / Scoped Search" pattern in Design Standards) feels like it could use more design and integration into the table/page rather than simply stretching the full width across the page...
You're 100% right. I'll need to do more research on this.

> The import statement icon/link copy feature seems a little out of place with the current site. Perhaps a better option is to add a "`<code>`" toggle in the last column, similar to the current doc-example template, with an expand/collapse section below the row like an accordion to show the single icon use code sample
This seems much too complicated for something that has a very simple desired outcome. As a dev using this table, I'm not interested in seeing the entire import statement for every icon. Devs using this reference know what an import statement is and how to construct it (in the event that they don't, it'll be explained right above the table). A designer using the table won't be interested in the import statement at all. Plus, it takes up space and adds noise around the juicy information that we're here for: the terra-icon aliases.

They won't be copying different sections of the import statement. When it's copied, it's pretty much always going to be copied in its entirety so there's no need to afford copying sections of it. Making a someone highlight a whole string to copy it seems cruel when it's so easy to do it automatically.

I do understand that it's out of place for the current site, but perhaps we could discuss going the other way on it and using this pattern across the rest of the site?

> We'll need to have documentation explaining the Function Types ("Action", "Status", "Toggle").
That'll be in the column descriptions right above the table. I had this sort of worked out on my other PR before I lost my mind and ruined the whole thing 😅 ([see the diff on the deleted file](https://github.com/cerner/terra-core/pull/3582/commits/649b88795d2e8ac02605b69b8ccdcc9e5d3db9af))

> The "Themeable Elements" column isn't quite clear
I totally agree. I'm open to suggestions on what that column should be called. The meaning of it will be explained in the documentation above the table.

eawww:
@benbcai thanks for your comment and sorry it's taken me so long to respond.
> When I first see the svg icon on the left side in the new layout I personally want to know immediately what the name of this icon is so I know what to call it. But I would have to look farther to the right to then get to know the name in the Import Name column. Would it be better to left shift the Import Name column to be between the SVG and Description columns?

The thesis of this uplift is to present design standards/recommended usage front and center to encourage consumers to be more thoughtful with how they use the icons. So we wanted to present that information as up-front as we could to passively intervene in someone scrolling down, grabbing whatever icon they think looks alright, and running with it.
Also this table condenses icons that are duplicated in Terra and provides information on the icon as a whole before presenting the option of which alias to use.

> What exactly is the Meaning column used for? It seems similar to Description but I'm unclear how to distinguish the difference between these two columns.
The Meaning column gives guidance on the intended use for the icon from Cerner's design standards. The Description column is a physical description of the icon's image for accessibility and searchability. Though I do see your point and I think the Description column isn't as important to show right up front.

xenoworf (approved):

benbcai:
> @benbcai thanks for your comment and sorry it's taken me so long to respond.
> 
> > When I first see the svg icon on the left side in the new layout I personally want to know immediately what the name of this icon is so I know what to call it. But I would have to look farther to the right to then get to know the name in the Import Name column. Would it be better to left shift the Import Name column to be between the SVG and Description columns?
> 
> The thesis of this uplift is to present design standards/recommended usage front and center to encourage consumers to be more thoughtful with how they use the icons. So we wanted to present that information as up-front as we could to passively intervene in someone scrolling down, grabbing whatever icon they think looks alright, and running with it. Also this table condenses icons that are duplicated in Terra and provides information on the icon as a whole before presenting the option of which alias to use.
> 

@eawww I think this table showing the relevant information about the icons is great and a big improvement. After thinking a bit more about my earlier question, the purpose of the "Import Name" column is to provide information about how to import the icon instead of displaying the official name associated to this icon. So I think it makes sense to keep the "Import Name" column where it is now.

PR #3602, opened 2022-03-21
[A11y-MVB] Build error fixes

sdadn:
### Summary

This PR addresses Travis build errors.

Closes #

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-icon/src/IconBaseDec.jsx
```diff
@@ -50,7 +50,6 @@ const IconBase = ({
   children,
```
benbcai, line 50: We should alphabetize the props in this file to be consistent. It's out of the scope of this PR and should have been updated on the [original creation PR](https://github.com/cerner/terra-core/pull/3602/files#diff-4c928a6653a1808a855a6620e9d9c72f4215734657cf63e4fad2b75520e65f26).

benbcai (approved):

packages/terra-icon/src/IconBaseDec.jsx
sdadn: I plan to update one of the icon examples so that it's more clear; i can do this as part of that PR.

neilpfeiffer (approved):

sdadn:
Note: the updated screenshots are just to resolve wdio test snapshot differences caused by css bleed. This can be fixed by updating snapshot selectors to be more robust and specific to the element.

Issue #3572, opened 2022-01-27
Unable to dismiss keyboard for the Terra Search Field


ccehlers:
# Bug Report

## Description
On a Mobile view when using the Terra Search Field, a user is unable to dismiss the keyboard when the enter key is pressed, leaving the user stuck on the screen.  We've had consistent client feedback of this being defective.  

## Steps to Reproduce

1. Enter a mobile view with the Terra Search Field or Terra Input Field and the keyboard present
2. Select the 'Enter' button on the keyboard
3. Keyboard does not dismiss and leaves the user confused on how to access the rest of the page content.

## Additional Context / Screenshots
![image](https://user-images.githubusercontent.com/12070453/151453349-c0250bea-1ee6-4a84-87d6-c901648653b6.png)

## Expected Behavior
Upon hitting the Enter key, the user again sees the rest of the page content and the keyboard dismisses.  The expectation would be to have a UX that manages both to be intuitive to the user as well as meet Accessibility requirements.  

## Possible Solution
The challenge with this will be Accessibility concerns, as the reason it behaves this way is due to the need for the field to always be in context and not blur out the field to dismiss the keyboard.  However this is behavior is also consistently being reported as defective by our clients within Nursing workflows on mobile devices.

## Environment
terra-search-field (3.74.0) + terra-form-input (4.17.0)
Chrome for Android, Safari for iOS - any version
node 10/npm 6
webpack 4
Mobile Android + iOS

## Notes
This can be easily reproduced by visiting one of the terra component documentation pages in a mobile browser (ex. https://engineering.cerner.com/terra-ui/components/terra-form-input/form-input/input)

## @ Mentions
@scottwilmarth

benbcai:
Logged UXPLATFORM-6428 to track this issue.

PR #3620, opened 2022-05-18
Fixed mobile keyboard not closing on search-input

sdadn:
### Summary

Closes:
#3572 
UXPLATFORM-6428

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-search-field/src/SearchField.jsx
```diff
@@ -172,6 +174,7 @@ class SearchField extends React.Component {
 
   handleKeyDown(event) {
     if (event.nativeEvent.keyCode === KeyCode.KEY_RETURN) {
+      this.searchBtnRef.focus();
```
xenoworf, line 178: Thanks, this teaches me the value of using refs more clearly than the doc.

packages/terra-search-field/CHANGELOG.md
```diff
@@ -2,20 +2,23 @@
...
 ## Unreleased
 
+* Changed
+  * Fixed mobile keyboard not closing after pressing enter key on mobile devices.
+
 ## 3.74.5 - (May 3, 2022)
 
 * Changed
-  * Minor dependency version bump
+  * Minor dependency version bump.
```
benbcai, line 11: The `Minor dependency version bump` log is generated by this [script](https://github.com/cerner/terra-toolkit/blob/%40cerner/terra-open-source-scripts%401.12.0/packages/terra-open-source-scripts/src/terra-cli/prepare-for-release/updateChangelogForPackage.js#L26). If we prefer to add a period, we should make the change in this script instead of manually here to avoid inconsistencies with all other changelogs.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -172,6 +174,9 @@ class SearchField extends React.Component {
 
   handleKeyDown(event) {
     if (event.nativeEvent.keyCode === KeyCode.KEY_RETURN) {
+      // set focus to search button to hide keyboard on mobile devices
+      this.searchBtnRef.focus();
```
benbcai, line 178: Can we add some tests to cover this change?

supreethmr: A WDIO test to ensure button is focused on return key press !!

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -304,6 +309,7 @@ class SearchField extends React.Component {
           icon={Icon}
           isIconOnly
           isCompact
+          ref={this.searchBtnRef}
```
supreethmr, line 312: Button supports[ `refcallback` prop](https://engineering.cerner.com/terra-ui/components/cerner-terra-core-docs/button/about#button-props) to provide an callback ref to the dom. we can use `refcallback` to access dom.
```jsx
  <Button
    ------
    ------
    refCallback={(ref) => { this.searchBtnRef.current = ref; }}
  />
```

packages/terra-search-field/src/SearchField.jsx
supreethmr: `this.searchBtnRef.focus();` this will cause an error since the dom node of button will be available on `current` attribute of ref ( as mentioned [here](https://reactjs.org/docs/refs-and-the-dom.html#accessing-refs)). Calling focus() method directly on ref object will throw `Err: focus() is not a function`.

packages/terra-search-field/CHANGELOG.md
sdadn: Undid the changes since it's probably better as a separate conversation & PR and just adds noise here:
f8e54a4ab93abd094046383931e0a3c9c0a4dff7

packages/terra-search-field/src/SearchField.jsx
sdadn: Thanks for the explanation!
Updated here: f4db4098cfb69d72c9f50e870151647eae859479

packages/terra-search-field/src/SearchField.jsx
sdadn: Added test here:
f4db4098cfb69d72c9f50e870151647eae859479

packages/terra-search-field/src/SearchField.jsx
sdadn: Changed here: f4db4098cfb69d72c9f50e870151647eae859479

benbcai (approved):

supreethmr (approved):

eawww:
✅ Accessibility review completed.

To answer @supreethmr 's concern: while focus is given to the search button after an 'enter' or 'search' key press, focus/visual indicator is easily moved back to the search input field in the same way that it would if the user had advanced focus to the search button themselves.

eawww (approved):

chrismichalewicz:
Functionally Reviewed

PR #3639, opened 2022-07-05
Publish on 07/05/22

BenBoersma:
### Summary
The following packages will be released:
- terra-action-header@2.74.3
- terra-alert@4.60.7
- @cerner/terra-core-docs@1.14.0
- terra-demographics-banner@3.58.2
- terra-dialog@2.63.6
- @cerner/terra-docs@1.7.1
- terra-form-field@4.20.5
- terra-form-fieldset@2.63.5
- terra-form-input@4.17.5
- terra-form-select@6.35.5
- terra-form-textarea@5.19.5
- terra-hyperlink@2.52.0
- terra-icon@3.48.0
- terra-list@4.51.5
- terra-overlay@3.69.5
- terra-paginator@2.80.2
- terra-responsive-element@5.35.2
- terra-search-field@3.75.0
- terra-show-hide@2.55.3
- terra-status-view@4.48.6
- terra-status@4.50.3
- terra-table@4.31.2
- terra-tag@2.52.3
- terra-toggle-button@3.61.4
- terra-toggle-section-header@2.58.3
- terra-toggle@3.50.3

Closes #

### Deployment Link

### Additional Details

@cerner/terra


BenBoersma:
> I think the version bump of the package should match the version bump of the dependency e.g. minor version change for dependency -> minor version bump for package. Or just make everything a minor version change to make it easier.

I mean I'm fine with this, I believe standard practice up to this point has been that minor version updates to dependencies are patch updates.

BenBoersma:
> > I think the version bump of the package should match the version bump of the dependency e.g. minor version change for dependency -> minor version bump for package. Or just make everything a minor version change to make it easier.
> 
> I mean I'm fine with this, I believe standard practice up to this point has been that minor version updates to dependencies are patch updates.

Actually thinking on it more, if we matched dependency version bumps wouldn't that mean consuming a major version bump would result in a major version bump on something that may not be a breaking change? I think we usually do
Major = breaking change
Minor = bugfix or new feature
patch = dependency update (that doesn't break the above)
I assume on some level we were following https://semver.org/spec/v2.0.0.html

sdadn:
> Actually thinking on it more, if we matched dependency version bumps wouldn't that mean consuming a major version bump would result in a major version bump on something that may not be a breaking change? I think we usually do Major = breaking change Minor = bugfix or new feature patch = dependency update (that doesn't break the above) I assume on some level we were following https://semver.org/spec/v2.0.0.html

A major version bump of a dependency would usually mean that a backwards incompatible change is introduced. Even if there are no changes to the API of the dependency, there may be other changes such as dropping support for older versions of react/node etc. Therefore it is safer to match MVB of dependencies with MVB of releases. 

DevDependencies are the exception. Consuming a new major version of e.g. eslint doesn't impact the API contract we have with consumers (nor should it be a releasable change), so upgrading shouldn't warrant a MVB.

sdadn (approved):

benbcai:
> Minor = bugfix or new feature

I would generally consider a bug fix as PATCH version if the fix itself doesn't introduce any new breaking changes.

PR #3633, opened 2022-06-22
[terra-hyperlink] Fixed focus box bug

saket2403:
### Summary

Allow users to set programmatic focus to the hyperlink. This fix shows the focus box on the hyperlink when a popup is closed and the active element is the hyperlink component.

### Deployment Link

https://terra-core-deployed-pr-#.herokuapp.com/

### Additional Details

Thank you for contributing to Terra.
@cerner/terra

PR #3642, opened 2022-07-12
[terra-hyperlink] Add support for onMouseDown function passed by users

saket2403:
### Summary

Added support for onMouseDown function passed by users.

Closes #

### Deployment Link

### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-hyperlink/src/Hyperlink.jsx
+++ b/packages/terra-hyperlink/src/Hyperlink.jsx
@@ -77,6 +77,10 @@ const propTypes = {
    */
   onKeyUp: PropTypes.func,
   /**
+   * Callback function triggered when mouse is pressed.
+   */
+  onMouseDown: PropTypes.func,
+  /**
    * Sets the hyperlink variant. One of `default`, `external`, `image`, `video`, `audio`, `document`.
    */
   variant: PropTypes.oneOf(['default', 'external', 'image', 'video', 'audio', 'document']),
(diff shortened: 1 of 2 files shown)


benbcai (approved):

benbcai:
PR https://github.com/cerner/terra-core/pull/3633 already had some tests that cover onMouseClick. Are those existing jest/wdio tests sufficient for this additional change or do we need to add a new one to cover this?

sdadn (approved):

packages/terra-hyperlink/src/Hyperlink.jsx
```diff
@@ -76,6 +76,10 @@ const propTypes = {
    * Callback function triggered when key is released.
    */
   onKeyUp: PropTypes.func,
+  /**
+   * Callback function triggered when mouse is pressed.
+   */
+  onMouseDown: PropTypes.func,
```
benbcai, line 82: From the Teams conversation, it sounds like using onMouseDown in this component is discouraged from an accessibility standpoint and may be removed in the future for enhance accessibility. In that case, we should avoid display this new prop in the props table either by:

1. Making this as a private prop.
2. Completely remove it as a first-class prop but would need to pass it to `handleMouseDown` for the callback.

saket2403: marked prop as private [0019324](https://github.com/cerner/terra-core/pull/3642/commits/00193249dfa8ae7cf535f4a41ae5239d52702fef)

PR #3645, opened 2022-07-26
[terra-action-header] Add support to allow users to set A11y label to action buttons.

saket2403:
### Summary

Closes #

### Deployment Link

### Additional Details

@cerner/terra


packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -10,10 +10,18 @@ import styles from './ActionHeader.module.scss';
 const cx = classNames.bind(styles);
 
 const propTypes = {
+  /**
+   * Accessibility label for Back button. To be used with onBack prop.
+   */
+  backButtonLabel: PropTypes.string,
```
supreethmr, line 16:
```suggestion
  backButtonA11yLabel: PropTypes.string,
```

saket2403: Done [c3fd549](https://github.com/cerner/terra-core/pull/3645/commits/c3fd5495440c65ccc397d1e1c90290c255ceab0a)

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -24,6 +30,10 @@ const propTypes = {
    * Changing 'level' will not visually change the style of the content.
    */
   level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
+  /**
+   * Accessibility label for Next button. To be used with with onNext prop
```
benbcai, line 34:
```suggestion
   * Accessibility label for Next button. To be used with onNext prop
```

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -59,6 +69,10 @@ const propTypes = {
    * Callback function for when the previous button is clicked. The previous-next button group will display if either this or onNext is set but the button for the one not set will be disabled.
    */
   onPrevious: PropTypes.func,
+  /**
+   * Accessibility label for Previous button. To be used with with onPrevious prop.
```
benbcai, line 73:
```suggestion
   * Accessibility label for Previous button. To be used with onPrevious prop.
```

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -88,10 +106,29 @@ const ActionHeader = ({
...
+  backButtonA11yLabel,
+  nextButtonA11yLabel,
+  prevButtonA11yLabel,
   ...customProps
 }) => {
   const theme = React.useContext(ThemeContext);
 
+  const descriptiveText1 = `On Clicking the close button, ${text} will be closed`;
+  const descriptiveText2 = `On Clicking the maximize button, ${text} will be maximized`;
+  const descriptiveText3 = `On Clicking the minimize button, ${text} will be minimized`;
```
benbcai, line 118: Should these text be translated?

saket2403: Yes, I have asked for Neil's approval for the text and will log the translation JIRA and replace these.

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -88,10 +106,25 @@ const ActionHeader = ({
...
 
+  const closeButtonId = `terra-action-header-close-button-${uniqueid()}`;
+  const maximizeButtonId = `terra-action-header-maximize-button-${uniqueid()}`;
+  const minimizeButtonId = `terra-action-header-minimize-button-${uniqueid()}`;
+
+  const visuallyHiddenComponent = (
+    <>
+      <VisuallyHiddenText aria-hidden id={closeButtonId} text={intl.formatMessage({ id: 'Terra.actionHeader.close.description' }, { text })} />
+      <VisuallyHiddenText aria-hidden id={maximizeButtonId} text={intl.formatMessage({ id: 'Terra.actionHeader.maximize.description' }, { text })} />
+      <VisuallyHiddenText aria-hidden id={minimizeButtonId} text={intl.formatMessage({ id: 'Terra.actionHeader.minimize.description' }, { text })} />
```
supreethmr, line 124: don't we need similar descriptive texts for Back button, Next button and Previous button.

supreethmr:
It would be nice if we have some examples updated with the usage of this new props.

packages/terra-action-header/src/ActionHeader.jsx
saket2403: The descriptive texts will be passed as props for these Back, Next and Prev buttons to icon Button's text prop. Since we are not sure of the context in which these buttons will be used.

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -88,10 +106,25 @@ const ActionHeader = ({
...
+  backButtonA11yLabel,
+  nextButtonA11yLabel,
+  prevButtonA11yLabel,
   ...customProps
 }) => {
   const theme = React.useContext(ThemeContext);
 
+  const closeButtonId = `terra-action-header-close-button-${uniqueid()}`;
+  const maximizeButtonId = `terra-action-header-maximize-button-${uniqueid()}`;
+  const minimizeButtonId = `terra-action-header-minimize-button-${uniqueid()}`;
```
supreethmr, line 118: Lets use same Id for all.
```suggestion
  const buttonId = uniqueid();
  const closeButtonId = `terra-action-header-close-button-${buttonId}`;
  const maximizeButtonId = `terra-action-header-maximize-button-${buttonId}`;
  const minimizeButtonId = `terra-action-header-minimize-button-${buttonId}`;
```

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -88,10 +106,25 @@ const ActionHeader = ({
...
+
+  const visuallyHiddenComponent = (
+    <>
+      <VisuallyHiddenText aria-hidden id={closeButtonId} text={intl.formatMessage({ id: 'Terra.actionHeader.close.description' }, { text })} />
```
supreethmr, line 122: Is `aria-hidden` added for any specific screen reader and browser combination..??

saket2403: Without aria-hidden screen reader will read these text twice

supreethmr (approved):

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -196,7 +234,7 @@ const ActionHeader = ({
   return (
     <ActionHeaderContainer
       {...customProps}
-      startContent={leftButtons}
+      startContent={leftButtons || visuallyHiddenComponent}
```
benbcai, line 237: Is this what is causing the title in the action header to shift to the left in the screenshots. It looks like there isn't much left padding in the new screenshots. Is this expected?

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -59,6 +69,10 @@ const propTypes = {
    * Callback function for when the previous button is clicked. The previous-next button group will display if either this or onNext is set but the button for the one not set will be disabled.
    */
   onPrevious: PropTypes.func,
+  /**
+   * Accessibility label for Previous button. To be used with onPrevious prop.
+   */
+  prevButtonA11yLabel: PropTypes.string,
```
xenoworf, line 75: Happy to see this convention spreading!

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -75,6 +89,10 @@ const defaultProps = {
   onNext: undefined,
   onPrevious: undefined,
   children: undefined,
+  backButtonA11yLabel: undefined,
```
xenoworf, line 92: Why must a required prop specify a default value?

saket2403: It is not a required prop.

packages/terra-action-header/src/ActionHeader.jsx
saket2403: Yes the visually-hidden-text has a width of 1px which is causing a slight shift.

packages/terra-action-header/src/ActionHeader.jsx
```diff
@@ -88,10 +106,26 @@ const ActionHeader = ({
...
   ...customProps
 }) => {
   const theme = React.useContext(ThemeContext);
 
+  const buttonId = uniqueid();
+  const closeButtonId = `terra-action-header-close-button-${buttonId}`;
+  const maximizeButtonId = `terra-action-header-maximize-button-${buttonId}`;
+  const minimizeButtonId = `terra-action-header-minimize-button-${buttonId}`;
+
+  const visuallyHiddenComponent = (
```
xenoworf, line 121: Please use the AccessibleValue component, which I think will solve this double-reading problem and was purpose built for this kind of scenario: https://github.com/cerner/terra-framework/blob/c92fd06dc900c13f4c3ce8457066c7db15a5a2f6/packages/terra-time-input/src/_AccessibleValue.jsx#L25

That component was created as part of adding a11yLabels for TimeInput.

xenoworf (changes requested):
Requesting to use AccessibleValue component. Other than that this code looks fine.

packages/terra-action-header/src/ActionHeader.jsx
benbcai: Is this approved by UX? Otherwise, I think we need to adjust the styles to make it look how it used to be.

One possible option is to add the appropriate padding/margin to this instance of `visuallyHiddenComponent` to match how it used to be.

packages/terra-action-header/src/ActionHeader.jsx
xenoworf: If you'd prefer, perhaps ship-as-is and then go back and use this component? I think that would be passive and might help you ship faster. Your call. If you'd like to do the iterative approach, tell me so and I'll flip my Request Change to Approved :).

saket2403: Yes I think that would be better, also if the AccessibleValue is a reusable component for these scenarios, I think it would be better to move it to a common folder instead of importing it from time input or adding the same code for using it in another component.

xenoworf: Agreed. A third option is to use AccessibleValue as-is, and then move it later - which would still be passive. 

So I take it you will be shipping as-is and then turning around to use AccessibleValue in some form?

saket2403: Yes. For the time being I will be shipping as is and then change to AccessibleValue at a later stage.

xenoworf (approved):

packages/terra-action-header/src/ActionHeader.jsx
neilpfeiffer: Putting the visually hidden content in the startContent is the wrong location to insert. By including it here is it what is causing the current screenshot changes and the look is undesired UX changes. 

We intentionally created different left margin alignments depending on if there are action-header buttons or not in the start content, shifting between 10px if no contents, to 5px if buttons are present (has to do with how the no border buttons visually look to best align). The css that applies these differences is here: https://github.com/cerner/terra-core/blob/a7c88211833e8f180dec97341ef28bbdc590355e/packages/terra-action-header/src/ActionHeaderContainer.module.scss#L69

I would recommend inserting the visually hidden components within their own 0 height parent row, that will render on its own row below the entire current action-header contents flex row.

neilpfeiffer (dismissed):
Requesting changes to where the visually hidden content is inserted in order to restore correct visual screenshots: [see comment for ActionHeader.jsx: line 237](https://github.com/cerner/terra-core/pull/3645/files?file-filters%5B%5D=.jsx&file-filters%5B%5D=.snap&show-viewed-files=true#r935064769)

PR #3678, opened 2022-11-02
[terra-toolbar] Add programmatic context for what toolbar controls

sycombs:
### Summary
This change adds ariaControls, ariaLabel, and ariaLabelledBy props so that consumers of terra-toolbar can associate a toolbar to the content it controls and screen readers can communicate what the toolbar is in context of.

UXPLATFORM-7624

### Deployment Link

- Doc examples: https://terra-core-toolbar-cont-g5yzlb.herokuapp.com/components/cerner-terra-core-docs/toolbar/about
- Tests: https://terra-core-toolbar-cont-g5yzlb.herokuapp.com/tests/cerner-terra-core-docs/toolbar/default-toolbar

### Additional Details

@cerner/terra


sdadn (approved):

saket2403 (approved):

packages/terra-toolbar/src/Toolbar.jsx
```diff
@@ -36,9 +50,28 @@ const Toolbar = ({
...
+    additionalToolbarProps['aria-labelledby'] = ariaLabelledBy;
+  }
+
   const items = React.Children.map(children, item => (item ? <div className={cx('item')}>{item}</div> : item));
 
-  return (<div {...customProps} className={toolbarClassNames} role="toolbar">{items}</div>);
+  return (
+    <div
+      {...customProps}
+      {...additionalToolbarProps}
```
neilpfeiffer, line 68: > "Will be used over ariaLabel if both are present."

This approach does not use ariaLabelledBy over ariaLabel if both are provided as the prop description suggests, but instead spreads them both. 

Unless you were meaning that a screenreader may choose to read the 'aria-labelledby' over the 'aria-label', which may or may not be consistently true: 
https://www.davidmacd.com/blog/does-aria-label-override-static-text.html

```suggestion      
      aria-label={!ariaLabelledBy ? ariaLabel : undefined}
      aria-labelledby={ariaLabelledBy}
```

\
\
Also updating or adding new WDIO tests, and using the props on the doc examples showing best practice usage recommendations, would help to be able to test them with screenreaders.

packages/terra-toolbar/src/Toolbar.jsx
```diff
@@ -12,6 +12,16 @@ const propTypes = {
...
+  /**
+   * String that labels the content the toolbar controls.
+   */
+  ariaLabel: PropTypes.string,
+  /**
+   * String that labels the content the toolbar controls. References the
+   * id of the element the toolbar controls. Will be used over ariaLabel if
+   * both are present.
+   */
+  ariaLabelledBy: PropTypes.string,
```
neilpfeiffer, line 24: > References the id of the element the toolbar controls.

This is not accurate. It's not the ID of the element that the toolbar controls. The ID provided to `ariaLabelledBy` should be the visible text label -- usually is a label div, span, heading, etc. -- that describes what the toolbar does (e.g. "Text Formatting"). If no visible label is present, then `ariaLabel ` should be used instead.


Another attribute + prop that we should add to appropriately add programmatic context for what element the toolbar does control would be `'aria-controls'`:
https://www.w3.org/WAI/ARIA/apg/example-index/toolbar/toolbar.html

packages/terra-toolbar/src/Toolbar.jsx
```diff
@@ -12,6 +12,16 @@ const propTypes = {
    * Content alignment. Align to start, end, or center.
    */
   align: PropTypes.oneOf(['start', 'end', 'center']),
+  /**
+   * String that labels the content the toolbar controls.
+   */
+  ariaLabel: PropTypes.string,
```
neilpfeiffer, line 18: This prop description is vague and unclear, not guiding the consumer as best as it could.

It also should be indicated that `ariaLabel` should be used only if a visible label for the toolbar is not already present, and that `ariaLabelledBy` should be the first choice when available to use instead.

packages/terra-toolbar/src/Toolbar.jsx
neilpfeiffer: Resolved: https://github.com/cerner/terra-core/pull/3678/commits/fe5ebc7bfe746948753b4b606b5b9aa4e2088250

packages/terra-toolbar/src/Toolbar.jsx
neilpfeiffer: Resolved: https://github.com/cerner/terra-core/pull/3678/commits/5052341bb4dfab157f5f4163a9b3b3de3cbd4172

packages/terra-toolbar/tests/jest/Toolbar.test.jsx
```diff
@@ -35,6 +35,23 @@ describe('Toolbar', () => {
...
+  });
+
+  it('should render component with aria-labelledby', () => {
+    const wrapper = shallow(
+      <div>
+        <Toolbar ariaLabelledBy="test-id" />
+        <div id="test-id">
+          Test content
+        </div>
+      </div>,
```
neilpfeiffer, line 50: Closer, but the construction with "test content" is similar to what an `aria-controls` test would look like. A visible label describing the toolbar tools/buttons would typically come before the toolbar, and it's purpose is to describe the type of tools within the toolbar:
```suggestion
      <div>
        <h4 id="test-id">Image Tools</h4>
        <Toolbar ariaLabelledBy="test-id" />
      </div>,
```

neilpfeiffer:
Per my comment https://github.com/cerner/terra-core/pull/3678#discussion_r1015429786:

- [ ] Let's add a prop for `ariaControls` so that consumers can directly link the id of the content area that toolbar is directly controlling, as shown by the WAI example - https://www.w3.org/WAI/ARIA/apg/example-index/toolbar/toolbar.html

- [ ] We should have the docsite example include a sample "ariaControls" area/div below it, and make a new WDIO test we can use for unscripted testing. The sample "content" div can simply use the placeholder inside it, with text that says "Sample Content Area for Toolbar": https://engineering.cerner.com/terra-ui/dev_tools/cerner-terra-docs/documentation-components/placeholder

packages/terra-toolbar/tests/jest/Toolbar.test.jsx
neilpfeiffer: Not the best visual reference, but an example of a set of toolbars side-by-side, each with their own visible label above each toolbar region (e.g. "Color Mode, "Highlight", "Screen Settings", "Zoom", and "Volume Booster" would all be a visible labels for the toolbar region below each): https://ps.w.org/accessibility-enabler/assets/screenshot-2.png?rev=2023391

neilpfeiffer: Updated: https://github.com/cerner/terra-core/pull/3678/commits/aa1daf85b135ceb3179d872c812f095d0b5eb5d6#diff-abc5c772f120a26032f43c2cd33cc35d110d3e2559953ea75eaf4b18bb1abe82

@sycombs While less critical to be descriptive in Jest tests, everywhere else we typically we want to avoid using "Test" or "Example" or "Lorem Ipsum" in our sample content, because even tests are informal documentation and (1) makes it difficult to communicate best practices to consumers, and (2) makes it difficult to use for unscripted testing with screenreaders which should mimic real situations.

packages/terra-toolbar/tests/jest/Toolbar.test.jsx
```diff
@@ -35,6 +35,16 @@ describe('Toolbar', () => {
     expect(wrapper.hasClass('align-center'));
   });
 
+  it('should render component with aria-label', () => {
+    const wrapper = shallow(<Toolbar ariaLabel="test" />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('should render component with aria-labelledby', () => {
+    const wrapper = shallow(<Toolbar ariaLabelledBy="test-id" />);
```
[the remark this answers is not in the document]
neilpfeiffer: Resolved: https://github.com/cerner/terra-core/pull/3678/commits/aa1daf85b135ceb3179d872c812f095d0b5eb5d6

packages/terra-toolbar/tests/jest/Toolbar.test.jsx
sycombs: Did some renaming:
[cd05571](https://github.com/cerner/terra-core/pull/3678/commits/cd055714965430fb0530d05315a699ba7613b73c)

packages/terra-toolbar/src/Toolbar.jsx
```diff
@@ -12,6 +12,24 @@ const propTypes = {
...
+   */
+  ariaControls: PropTypes.string,
+  /**
+   * String that labels the content the toolbar controls. `ariaLabel` only be used
+   * when no visible label for the toolbar is present. `ariaLabelledBy` should be
+   * used over `ariaLabel` when possible.
+   */
+  ariaLabel: PropTypes.string,
+  /**
+   * String that labels the content the toolbar controls. Should reference
```
supreethmr, line 27:
```suggestion
   * String that labels the content of the toolbar controls. Should reference
```

packages/terra-core-docs/src/terra-dev-site/doc/toolbar/About.1.doc.mdx
```diff
@@ -1,5 +1,8 @@
 import { Badge } from 'terra-toolbar/package.json?dev-site-package';
 import DefaultToolbar from './example/DefaultToolbar?dev-site-example';
+import AriaLabelToolbar from './example/AriaLabelToolbar?dev-site-example';
+import AriaLabelledByToolbar from './example/AriaLabelledByToolbar?dev-site-example';
+import AriaControlsToolbar from './example/AriaControlsToolbar?dev-site-example';
```
neilpfeiffer, line 5: Great effort in adding the 3 new ones here, but having now 4 examples (default, option a, option b, option c...) that only show one part/prop at a time is probably too many for the docsite page and may be confusing to consumers, since we are trying to shift away from showing fragments of each al-a-carte option without context, now towards only showing either the single best composition or multiple full variation examples, that each equally follow best practices so that any would be acceptable if seen in live projects.

Since you have these three ad-hoc and individual tests exactly the same already also as [test files](https://github.com/cerner/terra-core/tree/cd055714965430fb0530d05315a699ba7613b73c/packages/terra-core-docs/src/terra-dev-site/test/toolbar) (these fit better as tests how they are done currently anyway), let go ahead and remove them, and update the "DefaultToolbar" example with fragments of these approaches so we just have one single docsite example that shows the new props in use:
- since we don't have a great stylized visible label UX design that we recommend to go with toolbars, for this docsite example for now, lets focus on using the `ariaLabel` prop, rather than ariaLabelledBy, with label value of "Documentation Tools", 
- plus add the new `ariaControls` and the div+placeholder sample content area below, the div might also help to add an additional aria-label: 
```
      <div id="test-content-id" aria-label="Add Progress Note Documentation">
        <Placeholder title="Sample Content Area for Toolbar" />
      </div>
``` 

\
\
Also - I have added in a separate comment below, we can provide additional content to the docsite example wrapper component (title and description) to give consumers more context.

packages/terra-core-docs/src/terra-dev-site/doc/toolbar/About.1.doc.mdx
```diff
@@ -40,6 +43,9 @@ import Toolbar from 'terra-toolbar';
 
  ## Examples
  <DefaultToolbar />
+ <AriaLabelToolbar />
+ <AriaLabelledByToolbar />
+ <AriaControlsToolbar />
```
neilpfeiffer, line 48: Adds a more descriptive title and description sentence:

```suggestion
 <DefaultToolbar title="A Basic Toolbar using the aria-label and aria-controls attributes for accessibility" 
 description="For assistive technology users, this example adds the aria-label attribute to name the type tools this 
 toolbar contains, and adds the aria-controls attribute to indicate which page region the toolbar applies to." />
```

packages/terra-core-docs/src/terra-dev-site/test/toolbar/AriaLabelToolbar.test.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import Button from 'terra-button';
+import IconEdit from 'terra-icon/lib/icon/IconEdit';
+import IconAdd from 'terra-icon/lib/icon/IconAdd';
+import IconAttachment from 'terra-icon/lib/icon/IconAttachment';
+import Toolbar from 'terra-toolbar';
+
+export default () => (
+  <Toolbar ariaLabel="image tools">
```
neilpfeiffer, line 9: Even though the aria-label is not visible, grammatically they follow standard Title Casing similar to as if is was visible, so that it is announced in the same way to assistive technology, and will show visibly as a caption to user, so it appears the same as if it were ordinarily visible.
```suggestion
  <Toolbar ariaLabel="Image Tools">
```

packages/terra-toolbar/src/Toolbar.jsx
sycombs: Updated the wording for clarity [d92fdde](https://github.com/cerner/terra-core/pull/3678/commits/d92fdde1469050de5ac17f97563b115a8d4162a4)

packages/terra-core-docs/src/terra-dev-site/doc/toolbar/About.1.doc.mdx
sycombs: Updated [d92fdde](https://github.com/cerner/terra-core/pull/3678/commits/d92fdde1469050de5ac17f97563b115a8d4162a4)

neilpfeiffer (approved):
UX Reviewed

packages/terra-core-docs/src/terra-dev-site/doc/toolbar/About.1.doc.mdx
```diff
@@ -39,7 +39,11 @@ import Toolbar from 'terra-toolbar';
  * [LTR/RTL Support](https://engineering.cerner.com/terra-ui/about/terra-ui/component-standards#ltr--rtl)
 
  ## Examples
- <DefaultToolbar />
+ <DefaultToolbar
+  title="A Basic Toolbar using the aria-label and aria-controls attributes for accessibility" 
+  description="For assistive technology users, this example adds the aria-label attribute to name the type tools this
```
neilpfeiffer, line 44:
```suggestion
  description="For assistive technology users, this example adds the aria-label attribute to name the type of tools this 
```

supreethmr (approved):

packages/terra-core-docs/src/terra-dev-site/doc/toolbar/About.1.doc.mdx
sycombs: [3acd70b](https://github.com/cerner/terra-core/pull/3678/commits/3acd70ba90dbf11f01be77e86173f66328a0e562)

PR #3685, opened 2022-11-15
Clicking parent row in a table will collapse/expand its child rows

tingting-han:
### Summary
When we click a parent row in a table, it will expand or contract its child rows. This is for https://jira2.cerner.com/browse/RCBACM-20374. 

Closes #

### Deployment Link

### Additional Details

@cerner/terra


neilpfeiffer:
Closing. Expandable table rows in terra-table is not something Terra will support currently.

PR #3696, opened 2022-12-02
[form-select] form-single-select accessiblity updates

sugan2416:
### Summary

Removed `isLabelHidden ` and `hideRequired ` props in single-select-field.
Added `aria-label` to list options.
Added/Updated aria attributes in single select frame.
Improved screen reader response for VO(Safari+Chrome) and JAWS.

Closes #

### Deployment Link

### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-form-select/src/single/Frame.jsx
+++ b/packages/terra-form-select/src/single/Frame.jsx
@@ -113,20 +113,6 @@ const defaultProps = {
   value: undefined,
 };
 
-/**
-   * Determines compatible aria-labelledby IDs based on active variant
-   * Used with default, multiple, and tag variants to provide information about the label, displayed
-   * value and or the placeholder value.
-   */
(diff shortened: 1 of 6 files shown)


packages/terra-form-select/src/single/Frame.jsx
```diff
@@ -400,8 +386,16 @@ class Frame extends React.Component {
    * Determines compatible role attribute to apply to select based on active variant and disabled prop
    */
   role() {
+    let role;
     const { disabled } = this.props;
-    return disabled ? undefined : 'combobox';
+
+    if (SharedUtil.isMac()) {
+      role = SharedUtil.isSafari() ? 'group' : 'button';
```
supreethmr, line 393: Curios !! how is `isMac` different from `isSafari` ??

supreethmr (approved):

PR #3707, opened 2023-01-02
[form-select] form-single-select accessiblity updates

sugan2416:
### Summary

Added/Updated `hideRequired`, `isLabelHidden`, `onChange`  as private prop
Updated `aria-label` with meaningful text.
Fixed issue related to expanded state for VO(Chrome+Safari)

Closes #

### Deployment Link

### Testing

Updated jest snapshots

### Additional Details

@cerner/terra


packages/terra-form-select/src/shared/_Option.jsx
```diff
@@ -94,6 +94,12 @@ const Option = ({
     }
   }
 
+  let label = display;
+  if (customProps.index && customProps.totalOptions) {
+    label = `${display} (${customProps.index} of ${customProps.totalOptions})`;
```
supreethmr, line 99:
```suggestion
    label = `${display} (${customProps.index} of ${customProps.totalOptions})`;
```
did we missed to get translations for string "of" here

sugan2416: we have the translation for the string "of". Will update this to use it.

sugan2416: Updated in https://github.com/cerner/terra-core/pull/3707/commits/33798ba649bf547f477821356fb9f65d9f824385

packages/terra-form-select/src/shared/_Option.jsx
```diff
@@ -94,6 +94,12 @@ const Option = ({
     }
   }
 
+  let label = display;
+  if (customProps.index && customProps.totalOptions) {
+    label = `${display} (${customProps.index} ${customProps.ofText} ${customProps.totalOptions})`;
+  }
+  const itemLabel = isSelected || customProps.index === 1 ? `${customProps.expandedStateText} ${label}` : label;
+
```
supreethmr, line 102: Let's create a private prop definitions for each of these props since they are known props and being passed down menu.jsx it's better to have prop declared for each of the above values(index, ofText, totalOptions and expandedStateText).
we need to ensure customProps should contain only props passed by consumer for <SingleSelect.Option> .

Including this props in customProps will add them to dom as well as shown in below screenshot
( all the above attributes are not required to be included on `li` dom) 
Screenshot 2023-01-19 at 11 27 56 PM

sugan2416: Updated to private props in https://github.com/cerner/terra-core/pull/3707/commits/81fc0fed4e8d45e376fbb56046a06dc142177cf8

packages/terra-form-select/src/single/Frame.jsx
```diff
@@ -396,12 +382,29 @@ class Frame extends React.Component {
...
+    return label || this.ariaLabel();
+  }
+
   /**
    * Determines compatible role attribute to apply to select based on active variant and disabled prop
    */
   role() {
     const { disabled } = this.props;
-    return disabled ? undefined : 'combobox';
+    const role = SharedUtil.isSafari() ? 'group' : 'button';
```
supreethmr, line 405:
```suggestion
    const role = SharedUtil.isSafari() ? 'group' : 'combobox';
```
Should this be combobox instead of button ..?? 
Because this would make VO to read form-select as button when it focused with keyboard.

sugan2416: This was set to "button" for chrome as VO had issues wherein it was announcing the selected item twice. This is because `aria-hidden` wasn't taking effect when role was set to "combobox". With this change, though the role is button, the custom response would mention "Expanded/Collapsed combobox" first and at the end of response it would finally read as button. However this issue doesn't occur in safari, as it's using `aria-live=polite`

packages/terra-form-select/CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
 # Changelog
 
 ## Unreleased
+* Added
+  * Added `aria-label` to list options.
+
+* Changed
+  * Changed `hideRequired`, `isLabelHidden`, `onChange` props to private.
+  * Updated aria attributes.
```
sdadn, line 9: Can this be a little more descriptive?

packages/terra-form-select/src/single/Frame.jsx
sugan2416: Yes, tested on edge.

packages/terra-form-select/CHANGELOG.md
sugan2416: https://github.com/cerner/terra-core/pull/3707/commits/4700debfb340ff2e8b654e94a9fb801accb3bdd0

supreethmr (approved):

saket2403 (approved):

ShettyAkarsh (approved):

PR #3726, opened 2023-02-16
[terra-button] Fix Aria label bug

saket2403:
### Summary

Fixed aria label bug introduced in 15/2 release [here](https://github.com/cerner/terra-core/commit/177f6bf21400124e784cb81c2cfa36c677077573#diff-39f6c87ccf92298263c905de24e815f96ac1c87e46309e7d88ac4513f8bb7193R310-R312).

Closes #

### Deployment Link

### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-button/src/Button.jsx
+++ b/packages/terra-button/src/Button.jsx
@@ -308,9 +308,9 @@ class Button extends React.Component {
     );
 
     // ignore aria-label for button with decorative icons.
-    let ariaLabel = customProps['aria-label'] ? customProps['aria-label'] : text;
-    if ((isIconOnly || variant === 'utility') && iconType === IconTypes.DECORATIVE) {
-      ariaLabel = undefined;
+    let ariaLabel = customProps['aria-label'];
+    if (isIconOnly || variant === 'utility') {
+      ariaLabel = (iconType === IconTypes.INFORMATIVE) ? ariaLabel || text : undefined;
     }
 
     let ComponentType = 'button';
--- a/packages/terra-button/src/Button.jsx
+++ b/packages/terra-button/src/Button.jsx
@@ -308,9 +308,9 @@ class Button extends React.Component {
     );
 
     // ignore aria-label for button with decorative icons.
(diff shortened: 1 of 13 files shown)


packages/terra-button/src/Button.jsx
```diff
@@ -308,9 +308,9 @@ class Button extends React.Component {
     );
 
     // ignore aria-label for button with decorative icons.
-    let ariaLabel = (iconType !== IconTypes.DECORATIVE) ? customProps['aria-label'] : undefined;
-    if ((isIconOnly || variant === 'utility') && iconType === IconTypes.INFORMATIVE) {
-      ariaLabel = ariaLabel || text;
+    let ariaLabel = customProps['aria-label'] ? customProps['aria-label'] : text;
+    if ((isIconOnly || variant === 'utility') && iconType === IconTypes.DECORATIVE) {
+      ariaLabel = undefined;
     }
```
supreethmr, line 314: Let's update this condition to like below so that it avoids adding unexpected aria-label to components like it has added for terra-alert and other components
```js
    let ariaLabel = customProps['aria-label'];
    if (isIconOnly || variant === 'utility') {
      ariaLabel = (iconType === IconTypes.INFORMATIVE) ? ariaLabel || text : undefined;
    }
```

packages/terra-button/src/Button.jsx
```diff
@@ -308,9 +308,9 @@ class Button extends React.Component {
     );
 
     // ignore aria-label for button with decorative icons.
-    let ariaLabel = (iconType !== IconTypes.DECORATIVE) ? customProps['aria-label'] : undefined;
-    if ((isIconOnly || variant === 'utility') && iconType === IconTypes.INFORMATIVE) {
-      ariaLabel = ariaLabel || text;
+    let ariaLabel = customProps['aria-label'] ? customProps['aria-label'] : text;
```
supreethmr, line 311: this will add aria-label to all variants. where as existing implementation was to add aria-label only when it is provided for other variants except iconOnly and utility. for iconOnly and utility it will add the area label even when consumer does not pass it.

supreethmr (approved):

PR #3768, opened 2023-03-31
Update terra-text usage guide

swetasingh9714:
### Summary
Update terra-text usage guide to include information on decorative, supportive and meaningful text attributes.

@cerner/terra


swetasingh9714:
@scottwilmarth @supreethmr 

Screenshot 2023-03-31 at 12 30 06 PM

As per point 2.2 in the screenshot above(of Scott's comment in my recently closed spike), some information HAS to be conveyed to the users. 

Regardless of it being the text style itself or the meaning behind the style, the ask that something has to be conveyed by the screenreader means that there should be a usage of an aria-* attribute or a dummy span that won't be displayed so that this info is conveyed. 

So doesn't this warrant a code change to accommodate this new requirement/enhancement/accessibility update.

sugan2416 (approved):

sdadn:
Can we add an entry to `terra-core-docs/CHANGELOG.md` to keep track of this change?

swetasingh9714 (commented):
@sdadn I have addressed your comment. Please approve if there are no further queries.

saket2403 (approved):

supreethmr:
@sdadn  it looks like we need to update actions to prevent deploy action from running. This check is not prevent for forked repo's 
 : https://github.com/cerner/terra-core/blob/ad8fc28ab57875c16bd20b8e24188a3e1470687c/.github/workflows/ci-cd.yml#L67

PR #3773, opened 2023-04-06
[terra-button] Enable visible focus when button is active

sugan2416:
### Summary

- Added visible focus to button when the active dom element is button
- This change was made to allow focus on button when `terra-menu` is closed

Closes #

### Additional Details

@cerner/terra

### DIFF
--- a/packages/terra-button/src/Button.jsx
+++ b/packages/terra-button/src/Button.jsx
@@ -173,7 +173,7 @@ class Button extends React.Component {
     }
 
     // Add focus styles for keyboard navigation
-    if (event.nativeEvent.keyCode === KeyCode.KEY_SPACE || event.nativeEvent.keyCode === KeyCode.KEY_RETURN) {
+    if (event.nativeEvent.keyCode === KeyCode.KEY_SPACE || event.nativeEvent.keyCode === KeyCode.KEY_RETURN || document.activeElement === event.currentTarget) {
       this.setState({ focused: true });
     }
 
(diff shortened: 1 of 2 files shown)


sdadn:
Are there existing tests that cover the change in if statement?

packages/terra-button/src/Button.jsx
```diff
@@ -173,7 +173,7 @@ class Button extends React.Component {
     }
 
     // Add focus styles for keyboard navigation
-    if (event.nativeEvent.keyCode === KeyCode.KEY_SPACE || event.nativeEvent.keyCode === KeyCode.KEY_RETURN) {
+    if (event.nativeEvent.keyCode === KeyCode.KEY_SPACE || event.nativeEvent.keyCode === KeyCode.KEY_RETURN || document.activeElement === event.currentTarget) {
```
supreethmr, line 176: I think event.currentTarget will always match document.active element !! Curious are there any scenario where event.currentTarget would not be equal to document.activeElement?? 

If event.currentTarget always matches document.active element then probably you should update this condition to 
```js
 ( document.activeElement === event.currentTarget && ( event.nativeEvent.keyCode === KeyCode.KEY_SPACE || event.nativeEvent.keyCode === KeyCode.KEY_RETURN ) )
```

sdadn: Can some unit tests also be added to test the if statement and any potential edge cases?

sugan2416: If we consider the case of terra-menu, upon Esc press, an opened menu closes and the programmatic focus is returned to button. This is triggering key event in button. If we update to suggested condition, it wouldn't pass as we didn't use return/tab key, thereby causing visible focus to fail in spite being the active element. Also mentioning that onfocus event isn't occurring in this case.

supreethmr:
Closing this PR as changes has been made on terra-popup to return focus back onto terra-button on popup close :  https://github.com/cerner/terra-framework/pull/1641

PR #3778, opened 2023-04-12
Updated package.json scripts

sdadn:
### Summary

This PR updates the scripts in package.json so they are more streamlined.

- `clean` was updated to clean the repo of all generated files and not just the root level.
- `clean:install` was updated to utilize the `time` command to keep track of `npm install` times.
- `clean:lib` was added as a replacement for `compile:clean` for better nomenclature.
- `clean:root` was added for deleting files on the root level.
- `compile` was updated to utilize the new `clean:lib` script
- `postinstall` was updated for better streamlining of the install process. Now, all dependencies are bootstrapped and installed before the packages are compiled.
- `pretest` was deleted as it was already covered by the `test` script.

@cerner/terra


package.json
```diff
@@ -39,11 +39,13 @@
...
+    "clean": "npm run clean:lib && npm run clean:root",
+    "clean:install": "npm run clean:all && time npm install",
+    "clean:lib": "rm -rf packages/**/lib",
+    "clean:root": "$(lerna clean --yes || true) && echo 'removing ./node_modules' && rm -rf node_modules && echo 'removed ./node_modules'",
     "clean:obsolete-snapshots": "npm test -- -u",
     "create-props-markdown": "node scripts/props-table-gen/generatePropsTables.js",
-    "compile": "npm run compile:clean && npm run compile:build",
+    "compile": "npm run clean:lib && npm run compile:build",
     "compile:build": "lerna run compile --parallel",
     "compile:clean": "rm -rf packages/**/lib",
```
sycombs, line 50: Should this also be removed, since it is a duplicate of `clean:lib`?

package.json
```diff
@@ -39,11 +39,13 @@
   },
   "scripts": {
     "bootstrap:hoist": "lerna bootstrap --hoist --nohoist=one-cerner-style-icons",
-    "clean": "lerna clean --yes && rm -rf node_modules || true",
-    "clean:install": "npm run clean && npm install",
+    "clean": "npm run clean:lib && npm run clean:root",
+    "clean:install": "npm run clean:all && time npm install",
```
sycombs, line 43: Is `clean:all` a script?

package.json
sdadn: Removed here: 1867d83b7a899ee4bedaf6a3973a61f36ba6c1ea

sycombs (approved):

artpark (approved):

PR #3796, opened 2023-05-11
[terra-demographics-banner] Follow a11y meaningful sequence in large …

smason0:
…banner

### Summary

**What was changed:**
Large demographics banner elements ordering updated, with css also updated to allow the component to remain visually passive. The identifiers are also now organized as a definition list.

**Why it was changed:**
To allow assistive technology to read the content in the proper [meaningful sequence](https://www.w3.org/WAI/WCAG21/Understanding/meaningful-sequence.html) of content, with patient identifiers immediately following the patient name. The definition list allows each of the identifiers to be read one-by-one when navigating via screen reader. This mitigates a patient safety risk in the clinical space where some identifiers could previously be skipped if navigating quickly.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

Tested with VO in Safari, Chrome, and Firefox, and with JAWS in Edge browser.

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-8981 

---

@cerner/terra


packages/terra-demographics-banner/src/DemographicsBanner.module.scss
```diff
@@ -106,5 +109,10 @@
     .identifiers {
       text-align: right;
     }
+
+    .row {
+      flex: 100%;
```
cm9361, line 114: This does not appear to be a valid value based on the documentation for this property.

packages/terra-demographics-banner/src/DemographicsBanner.module.scss
```diff
@@ -106,5 +109,10 @@
...
+      order: 3;
```
cm9361, line 115: Why was an order of 3 selected?  Should we add a comment so that the reason for using this value is clear?

cm9361:
Are there not existing web driver tests and screenshots that need to be updated based on these changes?

smason0:
> Are there not existing web driver tests and screenshots that need to be updated based on these changes? It also appears that the PR template was not completed.

Ideally there are no visual changes, and it should be passive with existing WDIO tests. I still need to complete the PR template, I meant to open this as a draft to deploy the preview site but opened the PR on accident.

packages/terra-demographics-banner/src/DemographicsBanner.module.scss
smason0: This is a shorthand for the `flex-basis` property supported by modern browsers, but apparently not IE 10. I found an alternative that should work for IE and will include that in my next commit.

packages/terra-demographics-banner/src/DemographicsBanner.module.scss
smason0: Good point, will add a comment. This is needed to "undo" the changing of the order of elements to fix the order read by the screen reader back to its original visual ordering, since the DOM ordering impacts screen readers but not the css. In this case, the 'row' element should be the third item in the flexbox, when it would be the second by default based on the order of elements.

packages/terra-demographics-banner/src/DemographicsBanner.module.scss
```diff
@@ -106,5 +109,11 @@
     .identifiers {
       text-align: right;
     }
+
+    .row {
```
trandrew1023, line 113: Since this is used specifically for the second row containing patient identifiers, could this be named something like patient-identifier-row?

Also never knew about these CSS properties but this is sweet, also wraps nicely when there is overflow
image

trandrew1023 (approved):

sycombs (approved):

packages/terra-demographics-banner/src/DemographicsBanner.module.scss
smason0: Changed to 'identifier-row': https://github.com/cerner/terra-core/pull/3796/commits/f0bd4ce0afcf337a5da63963902f5226070c034a

mjpalazzo:
@smason0 - The Demo Banner is working well with JAWS. I used the down arrow to navigate the text elements. Navigation is now Name, then demo identifiers (age, sex, DOB, etc., then MRN and REA numbers, then additional info (for examples that have it).
VoiceOver is not working well. For the Demographics Banner with Gestational and Post Menstrual Ages example, when I use the Down arrow to navigate to the banner, VO announces, "My Cat image. You are currently on an image. To begin interacting with the contents of this image, press Control Option Shift Down Arrow." This key combination does not do anything and the helper text is not necessary. 
When I press down arrow again, VO announces, Heading level 2 Johnathan Doe John. You are currently on a text element inside of a heading level 2." 
When I press Down arrow, VO announces. "10 Weeks You are currently on a text element."
Down arrow, VO announces, "Date of Birth, DOB October 25, 2018, You are currently on a text element."
Down arrow: VO announces, "GA 33 weeks"
Down arrow: VO announces, "Post Menstrual Age, PMA 43 weeks, You are currently on a text element."
Down arrow: VO announces, "MRN 12343"
Here VO gets stuck. Pressing down arrow does not take you to the next data field. Eventually VO announced "end of text element".

eawww:
There's a renewed conversation in UX about how this should behave with regard to multiple items reading as one long string in JAWS vs being able to navigate to them individually. It looks like it's probably going to override the previous guidance that was given. 😵‍💫

Should this PR go on hold until that's figured out or should there be a new story to separate the items if it comes to that?

smason0:
> There's a renewed conversation in UX about how this should behave with regard to multiple items reading as one long string in JAWS vs being able to navigate to them individually. It looks like it's probably going to override the previous guidance that was given. 😵‍💫
> 
> Should this PR go on hold until that's figured out or should there be a new story to separate the items if it comes to that?

If it could potentially override this change, I'd be in favor of putting it on hold, so we aren't making changes we'll just have to undo.

cm9361 (approved):

sycombs (approved):

trandrew1023 (approved):

chrismichalewicz:
+1 from a product perspective.

eawww (approved):

PR #3797, opened 2023-05-15
[terra-alert] Update alert keyboard navigation when assertive

trandrew1023:
### Summary

**What was changed:**
* Added focus shift to the notification banner's content when the notification is an Alert with an action element
* Updated wording on documentation site to use notification banner instead of alert
* Updated examples to be able to better test focus shift

**Why it was changed:**
* Role of alert in Terra can include an action element and because Alert is interruptive and should be time critical, if there is an action element the user must take immediately then the Alert is focused so user can take action

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

Edge
image

Chrome
image

Firefox
image

Edge + Jaws
Due to screenshotting, the Jaws focus indicator doesn't show but it was around the dotted focus indicator.
Jaws still interrupts and reads "Alert. Alert. Site is down for maintenance. Alert. Site is down for Maintenance. Action Dismiss Alert." Tabbing after focus is shifting takes to the next focusable element which is the action button and Jaws will continue reading the button once tabbed.
image

If an alert is present on the page with an action element, then it is also focused.
image

Chrome + Jaws
"Alert. Alert. Site is down for maintenance. Alert. Site is down for Maintenance. Action. Dismiss."

Firefox + Jaws
"Alert. Alert. Site is down for maintenance. Action. Action. Dismiss. Dismiss."

### Reviews

- [x] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

Big thanks to @sycombs for helping with these code changes and testing!

Added some notes about Jaws here https://gist.github.cerner.com/AT070262/d359c07b096b92072fe877a759d8dbad
There were tradeoffs focusing different things with both Jaws and VO and various combos. Focusing the notification content div 

**This PR resolves:**

UXPLATFORM-8890 

---

@cerner/terra


packages/terra-alert/src/Alert.module.scss
```diff
@@ -90,7 +90,7 @@
     color: var(--terra-alert-custom-text-color, #1c1f21);
   }
 
-  .actions > :first-child {
+  .actions > :nth-child(2) {
```
cm9361, line 93: If you used padding vs margin, could you add the CSS to the action class and not need to tie to a numbered child.  If your layout changes, you are susceptible to a defect being introduced.

mjpalazzo:
@trandrew1023 - The custom prop alert works well with Chrome and Edge and JAWS 2023 with Edge. Nice work! As we agreed on Friday when we met, JAWS announces the assertive alert and banner content twice. Visible focus went to the text item in the banner as we agreed. Tabbing to Dismiss works as expected, JAWS announces, "Dismiss button, Alert, To activate press Enter."
I tested a role=status notification and, as expected, focus stayed on the Trigger Notification Banner button.
I tested the Trigger Replaceable Notification Banner with a role=alert button. With this variant, the Trigger button became disabled and I could not tell where focus went. JAWS announced, Alert, Alert Site is down for maintenance. Replaceable Alert, Dismiss Alert." The Trigger button resumed active state when I clicked the Dismiss button, but I had to Tab around the page to get to it. I will test with Voice Over next.

mjpalazzo:
@trandrew1023 - I tested with VoiceOver on Safari. The first time I triggered a notification banner with role=alert VO announced the alert and content 3 times. Visual focus went to the alert banner text component, but when VO on there was VO's  solid black border on the text item and the dashed border around the group. 
When I tested keyboard navigation without VO on, the alert notification worked as expected and visual focus with the dashed border displayed around the notification content.

packages/terra-alert/src/Alert.module.scss
trandrew1023: This change has been undone with the latest changes for review, for context this was updated since we temporarily added an empty span so the action element no longer had the correct CSS

eawww (approved):

packages/terra-alert/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+  * Added automatic focus to notification banner's with high criticality and included action element.
```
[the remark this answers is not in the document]
trandrew1023: Update changelog here https://github.com/cerner/terra-core/pull/3797/commits/d5da9b78a95d9c86135bde7aa230885e9aa3d96d#diff-b62ab7ae03f56c47f8aba033b108058b187c6bb44ff7dbe93a72c454c1393279

sdadn:
I feel unsure about the change in terminology to notification banner when the component and section is still `Alert`. It feels like it could be confusing two terms for the same component and interchanging them.

chrismichalewicz:
> I feel unsure about the change in terminology to notification banner when the component and section is still `Alert`. It feels like it could be confusing two terms for the same component and interchanging them.

To be clear, we should not be renaming the component, that is outside the scope of our work and should stay focused on a11y.  I agree with Saad's comment, the mix of terminology only adds to the confusion.  We can make a minor doc update indicating to consumers this displays a notification banner and if the role=alert then focus will be moved to it but others I don't think we should be making updates to the name in the code.

trandrew1023:
> I feel unsure about the change in terminology to notification banner when the component and section is still `Alert`. It feels like it could be confusing two terms for the same component and interchanging them.

@sdadn Really good point! We discussed this during scrum a bit after reviewing the updated doc site and it seems like as long as the doc makes it clear Terra Alert is a notification banner then this would be ok. Something like this:

image

Let me know if what you think!

smason0:
+1 for adding some missing WDIO screens: https://github.com/cerner/terra-core/pull/3797/files#diff-d9838db6615282465bac668e2bd8a2aefca50b2c75666b0007d2e83197922f12

trandrew1023:
> @trandrew1023 - I validated the Custom Prop Alert with VoiceOver on Safari. The notification with role=alert worked as expected. I did not observe VO announcing the Alert content multiple times today. Visual focus went to the alert text box and I could navigate with Tab to the Action and Dismiss buttons. Pressing VO +Space on the Dismiss button closed the alert. I tried some of the other notification types. VO announced the content and did not move focus. When I pressed Shift Tab to get the Dismiss button, the dismiss button did not activate. This seems to be a problem with this example.

Thank you for catching that @mjpalazzo! Michael shared the issue with me offline and we found that this was a mistake in the example which been resolved in the latest changes

mjpalazzo:
@trandrew1023 - I tested the Custom Prop Alert with Safari using the keyboard and also with VoiceOver. The issue is resolved. For notification with role=status the Dismiss button can be used to Dismiss the notification banner.

packages/terra-alert/src/Alert.jsx
```diff
@@ -194,6 +203,15 @@ const Alert = ({
...
   );
 
+  useEffect(() => {
+    const isAlert = role === 'alert' || defaultRole === 'alert';
+    if (!disableAlertActionFocus && isAlert && action && alertRef?.current?.focus) {
+      alertRef.current.focus();
+    }
+
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [alertRef.current]);
```
smason0, line 213: It looks like changing the [current property of the ref doesn't trigger a re-render](https://github.com/facebook/react/issues/14387#issuecomment-503616820) meaning this useEffect wouldn't get called, so is this just happening once after mounting when the ref is initially set up?

trandrew1023: Oh yeah that's a good point, not sure why I even added the alertRef cause it shouldn't even change actually. But yeah, we are just wanting this to run after the notification banner is mounted.

Not sure if there is a preference but exhausting the deps in the useEffect could potentially cause a re-render and it could refocus unexpectedly, so for now I just removed all of them to just have the focus occur on initial mount

Updated here https://github.com/cerner/terra-core/pull/3797/commits/292fc6406dbbe795bc23956ea70940228d723749

packages/terra-alert/src/Alert.jsx
```diff
@@ -194,6 +203,15 @@ const Alert = ({
     </div>
   );
 
+  useEffect(() => {
+    const isAlert = role === 'alert' || defaultRole === 'alert';
+    if (!disableAlertActionFocus && isAlert && action && alertRef?.current?.focus) {
```
smason0, line 208: nit: Since the disableAlertActionFocus check is more likely to pass, should we put isAlert first here to break out earlier?

trandrew1023: Good call, I also moved action since we only focus when there's an action element and added comment here https://github.com/cerner/terra-core/pull/3797/commits/217745d6cf1a459d940744060cc3dd15c1828016

smason0 (approved):

sycombs (approved):

eawww:
@trandrew1023  After reviewing this with another a11y SME, it seems that the extra repeat of the information could potentially be avoided by moving focus instead to an empty element positioned so that the focus indicator takes up the same area as the one that's currently being focused. Thoughts?

trandrew1023:
> @trandrew1023 After reviewing this with another a11y SME, it seems that the extra repeat of the information could potentially be avoided by moving focus instead to an empty element positioned so that the focus indicator takes up the same area as the one that's currently being focused. Thoughts?

@eawww Definitely can try that! Do you happen to have an example or guidance on how this is supposed to be done? In the past we got something similar working with empty spans placed beside the content but that had some issues

trandrew1023:
> > @trandrew1023 After reviewing this with another a11y SME, it seems that the extra repeat of the information could potentially be avoided by moving focus instead to an empty element positioned so that the focus indicator takes up the same area as the one that's currently being focused. Thoughts?
> 
> @eawww Definitely can try that! Do you happen to have an example or guidance on how this is supposed to be done? In the past we got something similar working with empty spans placed beside the content but that had some issues

The latest change will now focus an empty div that manually gets the width and height set to match that of the original dimensions of the notification content

This commit adds the new empty body div and changes the focus to it https://github.com/cerner/terra-core/pull/3797/commits/2e0112aa7b2693dbc36a49a570e4782c1fdcf1d2

However, the changes above seemed to have slightly odd dimensions from the alertRef for smaller screens, the solution that seems to work included added new state values and using useLayoutEffect to then set those width and height values here https://github.com/cerner/terra-core/pull/3797/commits/154688a1a6edec8883dd88d20726eecf1f2e3e7e

PR #3806, opened 2023-05-22
[terra-content-container] Added an prop to set focus on content of the container without interactive elements

AC086323:
### Summary

**What was changed:**
Added a prop to allow keyboard users to scroll the content when there are no interactive elements present in the content of the content-container

https://github.com/cerner/terra-core/assets/126561169/76324f4e-da59-4ad1-86b0-be2ed1a67dbf
Lowlighttheme
fusiontheme
defaulttheme

**Why it was changed:**
This change will allow users to scroll up and down the content if there are no interactive elements present in the content of the terra-content-container

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**
UXPLATFORM-8983

---

@cerner/terra


packages/terra-content-container/src/ContentContainer.jsx
```diff
@@ -54,7 +61,7 @@ const ContentContainer = ({
     <div {...customProps} className={contentLayoutClassNames}>
       {header && <div className={cx('header')}>{header}</div>}
       <div className={cx('main')}>
-        <Scroll className={cx('normalizer')} refCallback={scrollRefCallback}>
+        <Scroll className={cx('normalizer')} refCallback={scrollRefCallback} tabIndex={setFocusOnContent ? '0' : '-1'}>
```
sugan2416, line 64: So when there are interactive elements and setFocusOnContent is true, scrollbar would also be part of navigation right?

packages/terra-content-container/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+  * Added scroll capability for keyboard only users
```
supreethmr, line 6:
```suggestion
   * Added `setFocusOnContent` prop to provide scroll capability for keyboard only users. (Requires Jest Snapshot updates)
```

packages/terra-content-container/src/ContentContainer.jsx
ShettyAkarsh: shouldn't it go through the first interactive element directly as discussed when the new prop is set ?

ShettyAkarsh:
@AC086323 can we update the pr title to reflect the recent changes ?

packages/terra-content-container/src/ContentContainer.jsx
```diff
@@ -27,6 +27,11 @@ const propTypes = {
    * Ref callback for the scrollable area of the content container.
    */
   scrollRefCallback: PropTypes.func,
+  /**
+   * ![IMPORTANT](https://badgen.net/badge/UX/Accessibility/blue)
+   * Sets focus on content when set to `true`. Focus on content helps in scrolling  within container when there is no interactive element to focus within container.
+   */
+  setFocusOnContent: PropTypes.bool,
```
ShettyAkarsh, line 34: can we rename this ? this is usefull when there is a scroll over content but there is no interactive element. it can be something like setfocusoncontentcontainer or setfocusoncontainer.

packages/terra-content-container/src/ContentContainer.jsx
AC086323: It will go to first interactive element directly when the prop is not set to true

packages/terra-content-container/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+   * Added `setFocusOnContent` prop to provide scroll capability for keyboard only users. (Requires Jest Snapshot updates)
```
supreethmr, line 6:
```suggestion
   * Added `setFocusOnContainer` prop to provide scroll capability for keyboard only users. (Requires Jest Snapshot updates)
```

AC086323:
> @AC086323 , do we have a WDIO tests that covers this scenario ?

We have wdios based on layout of content container, none of the tests use examples
image

ShettyAkarsh:
> > @AC086323 , do we have a WDIO tests that covers this scenario ?
> 
> We have wdios based on layout of content container, none of the tests use examples image

can we draft a new test case for this scenario ?

supreethmr (approved):

sugan2416 (approved):

packages/terra-content-container/src/ContentContainer.jsx
```diff
@@ -54,7 +61,7 @@ const ContentContainer = ({
     <div {...customProps} className={contentLayoutClassNames}>
       {header && <div className={cx('header')}>{header}</div>}
       <div className={cx('main')}>
-        <Scroll className={cx('normalizer')} refCallback={scrollRefCallback}>
+        <Scroll className={cx('normalizer')} refCallback={scrollRefCallback} tabIndex={setFocusOnContainer ? '0' : '-1'}>
```
supreethmr, line 64: Can we add tab-index to above div instead of adding it on scroll. so that we can add focus styles for class `main` similar to other terra-components like below 
Screenshot 2023-05-26 at 11 31 55 AM

AC086323: Added the role region and added the border style as per discussion

supreethmr (approved):
+1 with approval from UX on Focus styles added for content container

PR #3809, opened 2023-05-30
[terra-scroll] Adding new A11y example for testing purposes.

BenBoersma:
### Summary

**What was changed:**
Added a new A11y example.

**Why it was changed:**
Current examples weren't good enough for testing.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-8578 

---

@cerner/terra

### DIFF
--- a/packages/terra-core-docs/src/terra-dev-site/test/scroll/A11yDefault.test.jsx
+++ b/packages/terra-core-docs/src/terra-dev-site/test/scroll/A11yDefault.test.jsx
@@ -1,0 +1,55 @@
+import React from 'react';
+import classNames from 'classnames/bind';
+import Scroll from 'terra-scroll';
+import styles from './DefaultTest.module.scss';
+
+const cx = classNames.bind(styles);
+
+class DefaultScroll extends React.Component {
+  constructor(props) {
+    super(props);
+    this.setNode = this.setNode.bind(this);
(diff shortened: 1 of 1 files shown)


packages/terra-core-docs/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+ * Added terra-scroll A11y tests
```
sdadn, line 6: nit:
```suggestion
 * Added terra-scroll A11y tests.
```

sdadn (approved):

sycombs (approved):

artpark:
Could we call the new tests just Horizontal and Vertical instead? Do we need to add a11y in front of their names?

artpark:
> > Could we call the new tests just Horizontal and Vertical instead? Do we need to add a11y in front of their names?
> 
> So my main concern was that 4-5 months from now no one would remember why we have these examples that don't have WDIO tests that use them, and since (currently) their sole purpose is for A11y testing I figured naming them as such would help future proof it.

Gotcha that sounds good

eawww (approved):

PR #3820, opened 2023-06-13
Fix Lint failures in list

saket2403:
### Summary

**What was changed:**

**Why it was changed:**

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra

### DIFF
--- a/packages/terra-list/CHANGELOG.md
+++ b/packages/terra-list/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Fixed lint failures.
+
 ## 4.57.0 - (April 27, 2023)
 
 * Changed
--- a/packages/terra-list/CHANGELOG.md
+++ b/packages/terra-list/CHANGELOG.md
@@ -2,9 +2,6 @@
 
 ## Unreleased
 
-* Changed
-  * Fixed lint failures.
-
 ## 4.57.0 - (April 27, 2023)
 
 * Changed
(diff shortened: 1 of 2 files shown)


supreethmr (approved):

packages/terra-list/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Fixed lint failures.
+
```
sdadn, line 7: This change is not relevant for consumers, so we can remove this.

sdadn: It might but we can leave a comment and force merge it e.g.
https://github.com/cerner/terra-framework/pull/1666#issuecomment-1574204451

Dev side changes (e.g. tests or linter changes) result in duplicate releases as these have no impact on consumers.

supreethmr: instead of removing let's update the changeLog here. because consumer has been noticing this error on their builds. so if we update the change log it might help consumers to know this version has a fix for the error `The attribute aria-description is not supported by the role list. This role is implicit on the element ul`

sdadn: I'm confused; linters are limited to our dev environment and should have no affect on consumers, as terra will only exist inside their node_modules folder. They should never have been getting this error to begin with. Am, I missing something?

PR #3829, opened 2023-06-23
[terra-form-select] Native Select allow initial values

MadanKumarGovindaswamy:
### Summary

**What was changed:**

Changes made to allow initial  null value.

**Why it was changed:**
NativeSelect does not allow updating the value prop in subsequent renders

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9184 

---

@cerner/terra


packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -167,9 +168,19 @@ const NativeSelect = ({
...
+
+  useEffect(() => {
+    // added this to allow initial null values
+    if (value) {
+      setrefIsControlled(true);
+      setUncontrolledValue('');
+    } else {
+      setrefIsControlled(false);
+    }
+  }, [value]);
```
supreethmr, line 183: I don't think we need state to identify controlled version here !! the issue is due to use of `useRef` to store `isControlled` boolean value, because of useRef `refIsControlled` is initialized only once and ignored for the re-renders that happens with prop updates. 

So i think just removing useRef from line No.170 and keeping normal variable would fix this issue of `not updating value prop on re-renders`.

MadanKumarGovindaswamy: Updated the cod with out usage of state variable. Thanks

packages/terra-form-select/CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 # Changelog
 
 ## Unreleased
+* Fixed
+  * Fixed `Native Select` to allow updating the value prop in subsequent renders.
+
+* Added
+  * Added WDIO screenshots for `Native Select` with initial value as null and without null.
```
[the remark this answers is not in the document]
MadanKumarGovindaswamy: changes made Thanks.

packages/terra-form-select/tests/wdio/native-select-spec.js
```diff
@@ -187,4 +187,18 @@ Terra.describeViewports('Native Select', ['tiny'], () => {
...
+  describe('Controlled Select', () => {
+    before(() => browser.url('/raw/tests/cerner-terra-core-docs/form-select/native-select/controlled-select'));
+
+    it('native select with null value', () => {
+      Terra.validates.element('controlled-select-null');
+    });
+
+    it('should set Mercedes value on click', () => {
+      $('#test-button-id').click();
+
```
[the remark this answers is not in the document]
MadanKumarGovindaswamy: Added this check. Thanks

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -167,9 +168,16 @@ const NativeSelect = ({
   ...customProps
 }) => {
   const [uncontrolledValue, setUncontrolledValue] = useState(!isValuePresent(value) ? defaultValue || getFirstValue(options, isFilterStyle) : undefined);
-  const refIsControlled = useRef(isValuePresent(value));
   const refSelect = useRef();
   const theme = React.useContext(ThemeContext);
+  const refIsControlled = isValuePresent(value);
+
+  useEffect(() => {
+    // added this to allow initial null values
```
supreethmr, line 176: I think this condition was not added to allow initial null values !!

supreethmr (approved):

ShettyAkarsh (approved):

PR #3830, opened 2023-06-23
[terra-show-hide] Shift keyboard focus to newly revealed content with focus indicator

adoroshk:
### Summary
When Show Hide component expands the content, the screen reader users have to navigate back in order to read that content. In order to fix that issue the focus moves to the newly revealed content so that screen readers read that content out without user having to back navigate. The focus indicator has been added to the content when focused.

**What was changed:**
The current implementation doesn't allow programmatically detect what portion of the hidden content was previously displayed in preview. The hidden content is implemented via <Toggle> component and while in general it's possible to detect the first child node of the hidden content to assign focus, it poses a threat of changes to <Toggle> component breaking the logic of assigning focus to the first child node. The safest way is to wrap a <Toggle> element into <div> container and assign the focus to it. That would allow the screen reader to read the content and get a visual focus indicator around the content while on focus. 
The cons: the whole content container is on focus rather than it's first child node. 
The pros: changes to <Toggle> component won't affect the ShowHide component content 

**Why it was changed:**
To fix the accessibility issue for the screen reader users.

### Testing
This change was tested using:

- [ ] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

**This PR resolves:**

UXPLATFORM-8542


packages/terra-show-hide/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 ## Unreleased
 
+* Changed
+  * Updated WDIO Screenshots due to added focus.
```
smason0, line 6: I feel like this item isn't relevant to consumers and can be removed

packages/terra-show-hide/src/ShowHide.jsx
```diff
@@ -74,13 +84,24 @@ const ShowHide = (props) => {
...
 
   return (
     <div {...customProps}>
       {!isOpen && preview}
-      <Toggle isOpen={isOpen}>
-        {children}
-      </Toggle>
+      <div
+        className={cx(['show-hide', 'content-container', theme.className])}
+        tabIndex={-1}
```
smason0, line 97: nit: In other cases in the code, I'm seeing this as a numeric string when assigning the tabIndex

smason0 (approved):
Functionally seems to work as expected with VO. Nice job! Just left a couple comments

packages/terra-show-hide/src/ShowHide.jsx
```diff
@@ -58,6 +59,15 @@ const ShowHide = (props) => {
     ...customProps
   } = props;
 
+  const theme = React.useContext(ThemeContext);
+  const ref = React.useRef(null);
+
+  React.useEffect(() => {
+    if (isOpen && ref?.current) {
+      ref.current.focus();
```
sdadn, line 67: Can we add a test for this?
e.g.

https://github.com/cerner/terra-framework/blob/13f7f4c7e2bc71f5a650646bd887d99c730bca48/packages/terra-slide-panel/tests/wdio/slide-panel-spec.js#L185C4-L195

```js
$('#test-slide [aria-label="Panel content area"][aria-hidden="false"]').waitForExist();
browser.keys(['Tab']);
expect($('#close-panel-button').isFocused()).toBeTruthy();
```

trandrew1023 (approved):

packages/terra-show-hide/src/ShowHide.jsx
adoroshk: Good catch, fixed in [this commit](https://github.com/cerner/terra-core/pull/3830/commits/9e8aff2a397aa02be6fda129d3f049612e78d1e0)

packages/terra-show-hide/CHANGELOG.md
adoroshk: Good catch, removed in [this commit](https://github.com/cerner/terra-core/pull/3830/commits/9e8aff2a397aa02be6fda129d3f049612e78d1e0)

packages/terra-show-hide/src/ShowHide.jsx
adoroshk: Added [here](https://github.com/cerner/terra-core/blob/9e8aff2a397aa02be6fda129d3f049612e78d1e0/packages/terra-show-hide/tests/wdio/show-hide-spec.js#L32-L42)

eawww (approved):
Works nicely. Will need to make sure to document this behavior as part of the documentation changes on UXPLATFORM-9213

mjpalazzo:
@adoroshk - Good start on this component. As I went through the examples, I observed some odd screen reader behavior that is probably an issue with the examples and not the component itself. Here is how I tested with JAWS and the Edge browser:
**Default Show/Hide**
Used the H key to navigate to the header of the example text.  👍 
Pressed Down Arrow. 👍 
JAWS announced the text visually displayed in the example. There was no visual focus indicator (which I expected since the text is not interactable). 👍 
Pressed Down Arrow to navigate to the Show/Hide control. 👍 
JAWS announced "Show/Hide button, collapsed. To activate press Enter." 👍 
Pressed Enter to activate the control. 👍 
Full text displayed with dashed border visual focus indicator around it. 👍 
JAWS announced the text from the beginning. 👍 
When JAWS got to the end of the example text, I pressed Down Arrow again. 👍 
JAWS began announcing the text again, but began at the phrase "...less compared to a hospital outpatient department." This is an issue that needs to be fixed. 👎 
Pressed Tab and focus went to the Hide control. 👍 
Pressed Enter and the initial truncated text displayed. 👍 
**Custom Button Text ShowHide**
Used the H key to navigate to the header of the example text.  👍 
Pressed Down Arrow. 👍 
JAWS announced the text visually displayed in the example. There was no visual focus indicator (which I expected since the text is not interactable). 👍 
Pressed Down Arrow and JAWS announced the second line of text. 👍 
Pressed Down Arrow to navigate to the Show/Hide control. 👍 
Pressed Enter to activate the control. 👍 
Full text displayed with dashed border visual focus indicator around it. 👍 
JAWS announced "Show 2 more sentences button, collapsed. To activate press Enter." 👍 
Pressed Enter to activate the control. 👍 
Full text displayed with dashed border visual focus indicator around it. 👍 
Pressed Down Arrow; JAWS announced header of the example text. 👍 
Pressed Down Arrow; JAWS announced next line of text. Repeated until all lines were read. 👍 
JAWS announced "Hide 2 sentences button, expanded. To activate press Enter." 👍 
Pressed Enter to activate the control. 👍 
Full text displayed with dashed border visual focus indicator around it. 👍 
Pressed Tab. JAWS announced "Show 2 more sentences button, collapsed. To activate press Enter." 👍 
This example worked as expected.
**Initially Open ShowHide**
There were points where JAWS stopped announcing. Pressing Down Arrow caused JAWS to continue reading. This may be a problem because a blind user may not know that there is more text to read.
Used the H key to navigate to the header of the example text.  👍 
Pressed Down Arrow. 👍 
JAWS announced the text visually displayed in the example. There was no visual focus indicator (which I expected since the text is not interactable). 👍 
Pressed Down Arrow to navigate to the "Hide" control. 👍 
Pressed Enter to activate the control. JAWS announced "collapsed" and button label changed to "Show More" 👍 
Pressed Enter to activate the control. 👍 
Full text displayed with dashed border visual focus indicator around it. 👍 
JAWS announced the full text. 👍 
When it got to the end, Pressed Down arrow.
Here is where the example got strange. JAWS announced the text beginning at the phrase, "and follow-up regimens. 👎 
I tested the other examples and got similar results.

eawww:
@mjpalazzo 

> JAWS began announcing the text again, but began at the phrase "...less compared to a hospital outpatient department." This is an issue that needs to be fixed.

I think this is a result of JAWS's up/down arrow keys being for "Read next line". This doesn't happen in VoiceOver where that behavior doesn't exist. 

Focusing the entire content is only going to be fallback behavior after UXPLATFORM-9213 is implemented, which should resolve this issue. Since this is an issue with AT and one that won't exist when consumers have properly consumed the component, it's probably okay.

Interested to get @brooksienoodlesoup 's take though.

adoroshk (commented):
@mjpalazzo Thank you so much for such a detailed comment, it helped me a lot to understand how the component will be used with JAWS and VO. I looked into the problem, in both cases the problem represents itself within a content passed by user as a <p> element. As @eawww mentioned above, we are about to change the functionality of the show-hide component in cases like that, reserving existing behavior as a fallback for cases when user passes many children, like in example for Custom Button Text ShowHide (which is the example that worked as expected). Based on that, I think we should keep it as is and ensure we don't have that problem when implementing the changes in UXPLATFORM-9213.

brooksienoodlesoup:
Hi @eawww and @mjpalazzo ,
I just tested the Default Show/Hide using:

- Chrome 114.0.5735.106 (Official Build) (arm64)) + VO (Ventura 13.2.1) and
- Edge 114.0.1823.58 (Official build) (64-bit) + JAWS 2023.2303.144

In Chrome+VO (Ctrl+Opt+Right arrow):

1. After I activated the Show More button, focus shifted to the full text and Chrome+VO read it correctly.
2. When I tried to move forward to the Hide button, Chrome+VO re-read the full text again.
3. When I tried to move forward to the Hide button for the second time, Chrome+VO announced "end of," then re-read the full text for a third time.
4. On my third attempt to move forward to the Hide button, I was able to navigate to it successfully without a repeat of the full text.

In Edge+Jaws (arrowing down in Jaws Virtual PC Cursor mode with red focus indicator):

1. After I activated the Show More button, focus shifted to the full text and Edge+Jaws read it correctly.
2. When I tried to move forward to the HIde button, Edge+Jaws started re-reading the full text block, but it started to read at the words "less compared to a hospital outpatient department" and continued to read the rest of the full block as I arrowed down.
3. When I tried to move forward to the Hide button for a second time, I was able to navigate to the it successfully without a repeat of the full text.

mjpalazzo:
@adoroshk - Since there is more work to do with component, I will not add the accessibility approved label at this time.

adoroshk:
The problem with Voice Over + Right Arrow was fixed and tested to work as expected in Safari, Chrome, Edge and Firefox.
As this PR's scope is intended as a first step to prepare for UXPLATFORM-9213 and does not resolve the accessibility issue in it's entirety (UXPLATFORM-9213 would), the Accessibility Review and UX review labels will be removed.

PR #3839, opened 2023-07-10
[terra-form-select] Fix for voiceover reading index twice

ST063655:
### Summary

**What was changed:**

Remove the index being read twice by the Screen Reader.

**Why it was changed:**
Screen Reader was reading the index of the expanded form single select twice.

Issue: Voice Over reads the expanded single select index twice.{}

Chrome with Voice-Over:

Screenshot 2023-07-10 at 9 39 51 PM

Safari with Voice-Over:

Screenshot 2023-07-06 at 4 22 26 PM

Post Implementation:

Screenshot 2023-07-10 at 9 57 19 PM

Screenshot 2023-07-10 at 9 55 38 PM

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9292 

---

@cerner/terra

### DIFF
--- a/packages/terra-form-select/CHANGELOG.md
+++ b/packages/terra-form-select/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 * Fixed
   * Fixed `Native Select` to allow updating the value prop in subsequent renders.
+  * Removed index to avoid screen reader reading index twice in `terra form single select`.
 
 ## 6.42.0 - (June 28, 2023)
 
--- a/packages/terra-form-select/CHANGELOG.md
+++ b/packages/terra-form-select/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 * Fixed
   * Fixed `Native Select` to allow updating the value prop in subsequent renders.
-  * Removed index to avoid screen reader reading index twice in `terra form single select`.
+  * Fixed VoiceOver reading the index twice in `terra-form-single-select`.
 
 ## 6.42.0 - (June 28, 2023)
 
--- a/packages/terra-form-select/src/shared/_Option.jsx
+++ b/packages/terra-form-select/src/shared/_Option.jsx
@@ -118,12 +118,7 @@ const Option = ({
     }
   }
 
-  let label = display;
-  // Allows VO to announce index of items
-  if (SharedUtil.isMac() && index && totalOptions) {
-    label = `${display}`;
-  }
-  const itemLabel = isSelected || index === 1 ? `${expandedStateText} ${label}` : label;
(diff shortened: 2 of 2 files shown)


packages/terra-form-select/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 * Fixed
   * Fixed `Native Select` to allow updating the value prop in subsequent renders.
+  * Removed index to avoid screen reader reading index twice in `terra form single select`.
```
sdadn, line 6:
```suggestion
  * Fixed screen reader reading the index twice in `terra form single select`.
```

sdadn (changes requested):

packages/terra-form-select/src/shared/_Option.jsx
```diff
@@ -121,7 +121,7 @@ const Option = ({
   let label = display;
   // Allows VO to announce index of items
   if (SharedUtil.isMac() && index && totalOptions) {
-    label = `${display} (${index} ${ofText} ${totalOptions})`;
+    label = `${display}`;
```
sdadn, line 124: Isn't this affecting all terra-form-select variants since it's updating a shared file?

packages/terra-form-select/CHANGELOG.md
supreethmr: this change will affect only single select because values for `totalOptions` and `index` is set only on [clone()](https://github.com/cerner/terra-core/blob/a066d1f08dc384e15f4351631e468f4b96ab12ac/packages/terra-form-select/src/single/Menu.jsx#L395-L396) function of single-select/menu.jsx 
remaining variants seems to be not using `totalCount` and `index` due to which below code block was skipped for other variants.
```
  if (SharedUtil.isMac() && index && totalOptions) {
    label = `${display} (${index} ${ofText} ${totalOptions})`;
  }
```

packages/terra-form-select/src/shared/_Option.jsx
supreethmr, line 124: I think we don't need this `if()` block and variable label to update `itemLabel` with latest changes. instead we can directly use display as an value for `itemLabel`

```js
const itemLabel = isSelected || index === 1 ? `${expandedStateText} ${display}` : display;
```

supreethmr (approved):

packages/terra-form-select/CHANGELOG.md
sdadn: That makes sense. Thanks!

sdadn (approved):

PR #3849, opened 2023-07-14
[terra-search-field] Group SearchField component and optionally add Label

nikhitasharma:
### Summary

**What was changed:**
* All of the elements of the search field component are programmatically grouped together. 
* Consumers are able to define group name
* Optional prop (yes/no) to include a visible label is added. The label will match what is defined for the group name. The label will display above the search field
* Examples are added demonstrating label usage involving groupName prop as well. Prop list docs updated. 
* Added New WDIO tests for label
* Internationalized the `Search` word as aria-label. 

**Additional Accessibility updates:**
* Updated the default search delay time from 250ms to 2500ms 
* Updated component description in docs
* Removed the default search length message from various examples to simplify and better focus on the functionality the examples explain. 

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

image

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-8109 

---

@cerner/terra


packages/terra-search-field/CHANGELOG.md
```diff
@@ -2,9 +2,11 @@
 
 ## Unreleased
 
+* Changed
+  * Group Terra Search Field components together and add optional Label
```
sdadn, line 6: Can you update the CHANGELOG entry to be similar to this:

https://github.com/cerner/terra-framework/blob/main/packages/terra-worklist-data-grid/CHANGELOG.md

basically a separate entry for every prop added?

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -95,8 +104,10 @@ const propTypes = {
 const defaultProps = {
   defaultValue: undefined,
   disableAutoSearch: false,
+  groupName: 'Search',
```
trandrew1023, line 107: Should 'Search' be internationalized?

nikhitasharma: Internationalization makes sense to me.
@eawww  What do you think?

cm9361: If it will be read by the screen reader or displayed to a user, it must be internationalized.

packages/terra-icon/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Fixed lint errors
+
```
sdadn, line 7: Linter changes aren't relevant to consumers so we can skip adding this entry. If there is a build error, we can force merge it with a comment e.g:

https://github.com/cerner/terra-framework/pull/1666#issuecomment-1574204451

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -284,34 +297,44 @@ class SearchField extends React.Component {
...
-            ref={this.setInputRef}
-          />
-          {clearButton}
+      <div className={cx('search-container')}>
+        {isLabelVisible
+        && (
+        <label className={cx('label')}>{groupName}</label>
+        )}
+        <div className={cx('search-container')} role="search" aria-label="Search">
+          <div className={cx('search-container')} role="group" aria-label={groupName}>
```
cm9361, line 306: It seems incorrect for a child div to have the same class name as its parent container.  Also, do we really need both of these containers?

packages/terra-core-docs/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Added
+  * Added Search field example with optional custom Label.
+
 * Added
   * Added email validation for `terra-form-field`.
```
sdadn, line 9: These `Added` sections can be combined.

packages/terra-search-field/src/SearchField.jsx
cm9361, line 306: If we already have a label.  Can't we use aria-labelledby?  Do we technically need another aria-label if the label will be read by the screen reader?  If the label is visible, is the value read twice?

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -284,34 +297,44 @@ class SearchField extends React.Component {
...
-            onChange={this.handleTextChange}
-            disabled={isDisabled}
-            aria-label={inputText}
-            aria-disabled={isDisabled}
-            onKeyDown={this.handleKeyDown}
-            onInput={this.handleInput}
-            ref={this.setInputRef}
-          />
-          {clearButton}
+      <div className={cx('search-container')}>
```
cm9361, line 300: It looks like we have 3 separate containers in this hierarchy with the same "search-container" class name.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -284,34 +297,44 @@ class SearchField extends React.Component {
...
+        {isLabelVisible
+        && (
+        <label className={cx('label')}>{groupName}</label>
+        )}
+        <div className={cx('search-container')} role="search" aria-label="Search">
+          <div className={cx('search-container')} role="group" aria-label={groupName}>
+            <div {...customProps} className={searchFieldClassNames}>
```
cm9361, line 307: Can these two div elements not be combined into one container.  It doesn't seem like we should need both.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -284,34 +297,44 @@ class SearchField extends React.Component {
...
-            aria-disabled={isDisabled}
-            onKeyDown={this.handleKeyDown}
-            onInput={this.handleInput}
-            ref={this.setInputRef}
-          />
-          {clearButton}
+      <div className={cx('search-container')}>
+        {isLabelVisible
+        && (
+        <label className={cx('label')}>{groupName}</label>
```
cm9361, line 303: Would it be better to wrap the input element inside the label?

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -284,34 +297,44 @@ class SearchField extends React.Component {
...
+            <div {...customProps} className={searchFieldClassNames}>
+              <div className={cx('input-group')}>
+                <input
+                  {...additionalInputAttributes}
+                  className={inputClass}
+                  type="search"
+                  placeholder={placeholder}
+                  onChange={this.handleTextChange}
+                  disabled={isDisabled}
+                  aria-label={inputText}
```
cm9361, line 316: Is this correct for accessibility?

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -24,6 +24,11 @@ const propTypes = {
    */
   disableAutoSearch: PropTypes.bool,
 
+  /**
+   * Group name value for search group elements. This value is also used as Label name.
+   */
+  groupName: PropTypes.string,
```
cm9361, line 30: Is this really a group name?  It seems more like the label for the component.

cm9361:
You should have new WDIO tests, correct?  You added new properties to show a label, but there are no test updates.

packages/terra-search-field/src/SearchField.jsx
nikhitasharma: This will be the default label name if isLabelVisible is set to true. Will work on internationalizing this.

cm9361:
There does not appear to be a new test in the dev site for your new isLabelVisible property or the group name.

packages/terra-search-field/src/SearchField.jsx
nikhitasharma: Yes we need both the role divs per requirements and preferences on the Jira. And I tested that not having a display:inline styling for the child div makes the content display in a new line, which I am trying to avoid. 
Screenshot with and without the child div styling:
image

image

cm9361: I am not talking about the styling.  You are referencing a class name of search container for 3 separate containers.  That seems incorrect.

cm9361: What is the requirement specifically that forces both divs to be necessary?  How should it read to the end user.  On VoiceOver, I hear both search and group.  Is that expected?

nikhitasharma: About having two divs, I am referring to the requirements mentioned on [UXPLATFORM-8109 ](https://jira2.cerner.com/browse/UXPLATFORM-8109)'s description under `Additional info`.

 I can look at combining one of the divs with existing ones if that doesn't modify the way the component renders.

packages/terra-search-field/src/SearchField.module.scss
```diff
@@ -95,6 +95,19 @@
     }
   }
 
+  .label {
+    color: var(--terra-search-field-label-color, #1c1f21);
```
trandrew1023, line 99: I think we could also add this var in dark theme (I think it would be #b2b5b6, Orion Fusion has the same value so not sure if we need to specify anyways there)
https://github.com/cerner/terra-core/blob/71c926b2d981ab5ffbe8c62e9b8ea3bb5529be88/packages/terra-alert/src/clinical-lowlight-theme/Alert.module.scss#L16C46-L16C46

nikhitasharma:
> There does not appear to be a new test in the dev site for your new isLabelVisible property or the group name.

You are right, working on adding them.

mjpalazzo:
@nikhitasharma - the search field updates look good. I have some concerns about the documentation. There are many examples but no information about that properties the examples are illustrating and there is no accessibility documentation.
I also have a few questions for @eawww and @scottwilmarth about accessibility of certain properties and attributes of the search component.
1. The searchDelay property refers to automatic search. Does an automatic search comply with WCAG 3.2.2 On Input?
2. The isLabelVisible property and example - WCAG 3.3.2 requires visible labels. Should this property be deprecated.
3. The placeholder property and example - My recollection from other input components is that placeholders are not accessible. Should this one be deprecated as well? If we are keeping it for passivity, should we have a comment that tells consumers that it is not accessible and should not be used for new UIs?

eawww:
@mjpalazzo 
> The searchDelay property refers to automatic search. Does an automatic search comply with WCAG 3.2.2 On Input?

This is a very good question. While this doesn't make the component itself inaccessible, it might facilitate/encourage some inaccessible patterns. I believe an auto-search or suggestions pattern could be made accessible by ensuring screen readers read the number of results and/or that a search was performed. Of course there are no details about this included in the documentation. I'm interested to get @scottwilmarth 's thoughts.

> The isLabelVisible property and example - WCAG 3.3.2 requires visible labels. Should this property be deprecated?

The `isLabelVisible` prop was added for this PR to optionally make the component more accessible. The no-label state is allowed by Sufficient Technique [ARIA14: Using aria-label to provide an invisible label where a visible label cannot be used](https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA14) because the visible label is implied by the context and the use of the 'magnifying glass' button.

> The placeholder property and example - My recollection from other input components is that placeholders are not accessible. Should this one be deprecated as well? If we are keeping it for passivity, should we have a comment that tells consumers that it is not accessible and should not be used for new UIs?

The big no-no on placeholder text is trying to use it instead of a label. Due to the temporary nature of placeholder text, it can cause problems for people with cognitive/memory issues or just anyone who leaves and returns to a partially completed form. If the content used as placeholder text is non-essential or duplicative of information given elsewhere, it's usually okay. It certainly couldn't hurt to include this information in the documentation.

eawww:
@mjpalazzo I chatted with Scott in a dev SME meeting and we agreed that we should ideally have an example showing how to accessibly implement a delayed search/suggestion pattern. The current example isn't that useful.

mjpalazzo:
> 

Thanks, @eawww - how do we communicate the need for a new example to the team? Jira comment?

packages/terra-search-field/src/SearchField.jsx
nikhitasharma: Logged a different Jira for internationalization. [UXPLATFORM-9454](https://jira2.cerner.com/browse/UXPLATFORM-9454)

packages/terra-search-field/src/SearchField.jsx
nikhitasharma: Verified with Eric that having aria-label at these levels is expected. Looks like aria-label being read twice in case label is visible is not a problem.

packages/terra-search-field/src/SearchField.jsx
nikhitasharma: Was able to combine these two divs.

packages/terra-search-field/src/SearchField.jsx
nikhitasharma: The `inputText` variable was named poorly. It it updated now to `inputAriaLabelText` which is a custom aria-label passed in by consumer.

packages/terra-search-field/src/SearchField.jsx
nikhitasharma: I think it is better to have the label separate from the input field. The idea is to group together the search components which would be the input field, the clear button and the search button. I feel the way it is now is also more readable.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -250,8 +263,13 @@ class SearchField extends React.Component {
       customProps.className,
     );
 
-    const inputText = inputAttributes && Object.prototype.hasOwnProperty.call(inputAttributes, 'aria-label') ? inputAttributes['aria-label'] : intl.formatMessage({ id: 'Terra.searchField.search' });
+    const labelClassNames = cx([
+      'label',
+      theme.className,
```
cm9361, line 268: We should be adding the theme at the top container level, not the label.

packages/terra-search-field/src/SearchField.module.scss
```diff
@@ -95,6 +95,23 @@
...
+    display: block;
+    font-size: var(--terra-search-field-label-font-size, 1rem);
+    font-weight: var(--terra-search-field-label-font-weight, bold);
+    margin-bottom: var(--terra-search-field-label-margin-bottom, 0.357em);
+    text-align: left;
+  }
```
cm9361, line 105: Not a comment for your logic, but it is interesting that we have to customize the styles for a label instead of having a Terra Label component.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -250,8 +263,13 @@ class SearchField extends React.Component {
...
+    ]);
 
+    const inputAriaLabelText = inputAttributes && Object.prototype.hasOwnProperty.call(inputAttributes, 'aria-label') ? inputAttributes['aria-label'] : intl.formatMessage({ id: 'Terra.searchField.search' });
```
cm9361, line 271: Do we need a hasOwnProperty method call?  Won't accessing the property just return undefined and give us the desired behavior?

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -250,8 +263,13 @@ class SearchField extends React.Component {
...
+    const groupNameValue = groupName === 'Search' ? intl.formatMessage({ id: 'Terra.searchField.search' }) : groupName;
```
cm9361, line 272: Can't search be internationalized?  The groupName is a prop to be displayed to the user.  I would expect that we would only use our message if groupName is undefined.  However, that could just be the default prop in that case.

packages/terra-search-field/src/SearchField.jsx
```diff
@@ -250,8 +263,13 @@ class SearchField extends React.Component {
       customProps.className,
     );
 
-    const inputText = inputAttributes && Object.prototype.hasOwnProperty.call(inputAttributes, 'aria-label') ? inputAttributes['aria-label'] : intl.formatMessage({ id: 'Terra.searchField.search' });
+    const labelClassNames = cx([
```
cm9361, line 266: You don't really need a separate variable for this.  I would provide the classNames inline.

packages/terra-search-field/src/SearchField.jsx
nikhitasharma: What I am doing here is internationalizing that default prop value which is `Search`. I do not have access to intl object when defining default props, hence this code here. 
Basically, If there is a custom user defined groupName, that would take precedence. Otherwise, it would be the default internationalized value. 
Does that make sense?

packages/terra-core-docs/src/terra-dev-site/doc/search-field/example/SearchFieldBlock.jsx
```diff
@@ -1,18 +1,11 @@
 import React, { useState } from 'react';
 import SearchField from 'terra-search-field';
 
-const INVALID_MESSAGE = 'The default minimum search length is 2.';
-
 const SearchFieldBlock = () => {
-  const [searchText, setSearchText] = useState('');
-
-  const message = searchText.length >= 2 ? `Search text: ${searchText}` : INVALID_MESSAGE;
+  const [setSearchText] = useState('');
```
sdadn, line 5: what is the purpose of this state variable?

sycombs, line 5: Is this still needed if we are removing the `searchText` state variable?

packages/terra-search-field/CHANGELOG.md
```diff
@@ -15,8 +15,10 @@
 ## 3.84.0 - (August 1, 2023)
 
 * Changed
+  * Add optional `groupName` prop to group Search Field components.
+  * Add optional `isLabelVisible` prop to add Label to Search Field.
   * Updated focus indicators to be a standard dashed line.
-
+
```
sycombs, line 21: nit: extra whitespace
```suggestion

```

packages/terra-search-field/tests/jest/SearchField.test.jsx
```diff
@@ -221,7 +225,7 @@ describe('Auto Search', () => {
     searchField.find('.input').simulate('change', { target: {} });
 
     for (let i = 0; i < 3; i += 1) {
-      jest.advanceTimersByTime(250);
+      jest.advanceTimersByTime(2500);
       expect(onSearch).not.toBeCalled();
     }
```
sdadn, line 230: Loops are not recommended in unit tests. Since this is executed 3 times, can we break it out?

packages/terra-search-field/tests/wdio/search-field-spec.js
```diff
@@ -156,6 +190,8 @@ Terra.describeViewports('Search Field', ['medium'], () => {
     });
 
     it('should display Search Field with long enough text', () => {
+      // wait for the default search delay time for callback text to update
+      browser.pause(2500);
```
sdadn, line 194: Is this delay needed? there doesn't seem to be logic executing before this.

packages/terra-search-field/tests/wdio/search-field-spec.js
```diff
@@ -26,6 +26,40 @@ Terra.describeViewports('Search Field', ['medium'], () => {
...
+
+    it('should enter a search term', () => {
+      $('input').setValue('Lorem');
+    });
+
+    it('should display Search Field label as groupname with text', () => {
+      expect($('input[type="search"]').isFocused()).toBeTruthy();
+      Terra.validates.element('default search with label as groupname and search text');
+    });
+  });
```
sdadn, line 61: +1 on the testing setup

packages/terra-core-docs/src/terra-dev-site/doc/search-field/example/SearchFieldBlock.jsx
nikhitasharma: This is still needed for the search text updates to work and removing this will affect functionality of the examples

nikhitasharma: Answered on Sydney's comment: https://github.com/cerner/terra-core/pull/3849#discussion_r1291623111

mjpalazzo:
@nikhitasharma - great work on the Search Field component. Keyboard navigation is working as required. Overall, JAWS is working well, but I observed an issue that I think should be corrected. When I Tab to the component, JAWS makes some repetitive announcements, "Search group, Search Search region, Search group, Search edit, Type in text." Is it possible to reduce the amount of repetitiveness?
A few other notes: Some examples do not include the clear button. Is that intentional? 
The tooltips and helper text look good.

packages/terra-search-field/tests/wdio/search-field-spec.js
sdadn: Is this needed for every test or can we just do it once in the before block?

mjpalazzo:
@nikhitasharma - Great work on the Search Field! Keyboard navigation works as required. JAWS is working well. I did observe one behavior that needs to be addressed. When I Tab to the Search Field, JAWS makes some repetitive announcements: "Search group, Search Search region, Search group, Search edit, Type in text." VoiceOver announces, "Search text field, Search, Search, group." Can some of these repetitions be reduced.
The tooltip displays with Edge and with JAWS running on Edge. With Chrome and Safari, the tooltips do not display with keyboard focus. They do with mouse hover.

packages/terra-search-field/tests/wdio/search-field-spec.js
nikhitasharma: Well, it is needed only when we are checking for search text to be updated/reflected in some tests. It is not required before all the tests since not all tests are checking for search text updates. I did not add in a beforeAll because didn't want to delay tests unnecessarily. 
Here are some example snapshots we validate where we need to pause for search text updates to reflect:
https://github.com/cerner/terra-core/blob/main/packages/terra-search-field/tests/wdio/__snapshots__/reference/terra-default-theme/en/chrome_medium/search-field-spec/callback_with_too_short_text.png 
https://github.com/cerner/terra-core/blob/main/packages/terra-search-field/tests/wdio/__snapshots__/reference/terra-default-theme/en/chrome_medium/search-field-spec/callback_with_long_enough_text.png

kenk2: Concurring with @sdadn and a question: Is it actually possible to avoid doing things like this? https://webdriver.io/docs/api/element/isFocused/

I'm seeing in an example that we do:
```
const loginInput = await $('[name="login"]');
console.log(await loginInput.isFocused()); // outputs: false
```

Could we attempt to do something similar here?

sdadn: @kenk2 async methods were introduced in v8 I believe. We're using v7.

https://v7.webdriver.io/docs/api/element/isFocused

packages/terra-core-docs/src/terra-dev-site/doc/search-field/example/SearchFieldBlock.jsx
sycombs: Doesn't this just change `setSearchText` to be the state variable instead of the set function? Based on the useState doc: https://react.dev/reference/react/useState#returns

To me it seems like this is just passing the state variable `setSearchText` with value '' to the `onSearch` and `onInvalidSearch` props and the example would be equally functional by passing dummy functions to those props like `() => {}`.

packages/terra-search-field/tests/wdio/search-field-spec.js
sdadn: @nikhitasharma thanks for answering my question. That's fair and is fine by me. 👍

sdadn (approved):

nikhitasharma:
> VoiceOver announces, "Search text field, Search, Search, group." Can some of these repetitions be reduced.

The repetitions are because of the below two components having aria-label, but those values could be different depending on consumer's custom props. In the case of examples, it repeats `Search` since the default is that.  
-  the new div role="search":
```<div role="search" className={cx('search-role-container')} aria-label={intl.formatMessage({ id: 'Terra.searchField.search' })}>```
- the `<input>` component having its own aria-label (which is either the default value `Search` or can be passed in a custom value by consumer through `inputAttributes` prop)
- Another thing to notice is these aria-labels will be different from other `Search` announcements when internationalized. 

Will discuss this with @eawww  as well on Monday and see if we can remove any of these without affecting existing custom functionalities.

kenk2 (approved):

packages/terra-search-field/src/SearchField.jsx
cm9361: Even though you logged a JIRA for translations, you should still have the proper logic for i18n in this PR to use the intl object.  That should not be part of a separate PR.

packages/terra-search-field/src/SearchField.jsx
cm9361: There are ways to associate the input to the label.  You can either wrap or there is an aria attribute that can be used.  However, we don't appear to be leveraging either.

packages/terra-core-docs/src/terra-dev-site/doc/search-field/example/SearchFieldBlock.jsx
nikhitasharma: You are right, thanks for pointing this out. I also saw that the `onSearch` and `onInvalidSearch` are optional props which are not needed in the example after removing the minimum search length logic. So I have removed them [here](https://github.com/cerner/terra-core/pull/3849/commits/053b9ac552ec6d90f39535255fbc7a6914884a07).

eawww (approved):

eawww (approved):

PR #3850, opened 2023-07-17
Fix: Enable custom aria-label for dropdown button

SG067724:
### Summary

Updated the aria-label announced for the dropdown button to have custom aria-label added by the user.

**Why it was changed:**
When the user adds aria-label to drop-down it is added to the wrapping div and hence its not announced when the user tabs on the button. This change is made to enable the user to announce custom aria label for dropdown button.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9194 

---

@cerner/terra


packages/terra-dropdown-button/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Updated
```
supreethmr, line 5:
```suggestion
* Added
```

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -47,13 +47,17 @@ const propTypes = {
    * The intl object to be injected for translations.
    */
   intl: PropTypes.shape({ formatMessage: PropTypes.func }).isRequired,
+  // eslint-disable-next-line react/forbid-prop-types
+  inputAttrs: PropTypes.object,
```
supreethmr, line 51: this prop is missing prop usage information. `inputAttrs` does not sound relative to button 
should be renamed like `buttonAttrs` OR equivalent text.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -169,6 +174,8 @@ class DropdownButton extends React.Component {
       theme.className,
     );
 
+    const buttonAriaLabel = (inputAttrs && inputAttrs['aria-label']) ? inputAttrs['aria-label'] : '';
```
supreethmr, line 177: these can be achieved even without `inputAttrs` with `customProps`. so when ever consumer pass aria-label we can do 
`const buttonAriaLabel = (customProps['aria-label']) ? customProps['aria-label'] : '';`

but the actual reason to introduce attrs object is to support consumer passing attributes to button.  In this case we need add new `attrs` prop to button element by removing aria-label from it since button has already got aria-label.
```js
 let buttonAriaLabel;
 if (inputAttrs && inputAttrs['aria-label']) {
      buttonAriaLabel = inputAttrs['aria-label'];
      delete inputAttrs['aria-label'];
 }

<button
{...inputAttrs}
```

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -193,7 +200,7 @@ class DropdownButton extends React.Component {
           aria-disabled={isDisabled}
           ref={this.setButtonNode}
           aria-expanded={isOpen}
-          aria-label={selectText ? `${selectText}, ${selectedLabel}` : ''}
+          aria-label={selectText ? `${selectText}, ${selectedLabel}, ${label}, ${buttonAriaLabel}` : `${label}, ${buttonAriaLabel}`}
```
supreethmr, line 203: what value does `label` has here ?? looks like it was not present on previous implementation

SG067724: Label has the button name. Earlier it was getting announced but with new changes its getting overridden. That's why I have concatenated in aria-label.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -47,6 +47,7 @@ const propTypes = {
    * The intl object to be injected for translations.
    */
   intl: PropTypes.shape({ formatMessage: PropTypes.func }).isRequired,
+
```
sdadn, line 50: nit: Unnecessary extra space

sdadn, line 50: nit: unnecessary extra line

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -169,6 +170,12 @@ class DropdownButton extends React.Component {
       theme.className,
     );
 
+    let buttonAriaLabel;
+    if (customProps && customProps['aria-label']) {
+      buttonAriaLabel = customProps['aria-label'];
+      delete customProps['aria-label'];
```
sdadn, line 176: I'm confused what is the purpose of this.

sycombs:
@SG067724 Can you please fill out the PR description template and update the PR title to be Terra-centric? "Study options" does not seem to be a concept within terra-dropdown-button.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -169,6 +170,12 @@ class DropdownButton extends React.Component {
...
+    }
```
sdadn, line 177: Is there a reason we're doing this instead of creating an `ariaLabel` prop like in other components?

```suggestion
    let buttonAriaLabel;
    if (customProps && customProps['aria-label']) {
      buttonAriaLabel = customProps['aria-label'];
      delete customProps['aria-label'];
    }
```

SG067724: Making this change so that the aria-label added by the user applies to the button not the wrapping div

saket2403 (approved):

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -169,6 +176,13 @@ class DropdownButton extends React.Component {
       theme.className,
     );
 
+    let buttonAriaLabel = '';
+    const modifiedButtonAttrs = { ...buttonAttrs };
+    if (modifiedButtonAttrs && modifiedButtonAttrs['aria-label']) {
+      buttonAriaLabel = modifiedButtonAttrs['aria-label'];
+      delete modifiedButtonAttrs['aria-label'];
+    }
```
supreethmr, line 184: does using buttonAttrs like below throws linter error..??
no need to check for (buttonAttrs) since`buttonAttrs` is initialised with defaultValue. (buttonAtrrs) will always evaluate to `true`.

```suggestion
    if (buttonAttrs['aria-label']) {
      buttonAriaLabel = buttonAttrs['aria-label'];
      delete buttonAttrs['aria-label'];
    }
```

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -183,6 +197,7 @@ class DropdownButton extends React.Component {
         getSelectedOptionText={this.getSelectedOptionText}
       >
         <button
+          {...modifiedButtonAttrs}
```
supreethmr, line 200:
```suggestion
          {...buttonAttrs}
```

packages/terra-dropdown-button/src/DropdownButton.jsx
supreethmr: have we verified with VO and jaws to ensure button label is not announced twice ?

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -169,6 +176,13 @@ class DropdownButton extends React.Component {
       theme.className,
     );
 
+    let buttonAriaLabel = '';
+    const modifiedButtonAttrs = { ...buttonAttrs };
+    if (modifiedButtonAttrs && modifiedButtonAttrs['aria-label']) {
+      buttonAriaLabel = modifiedButtonAttrs['aria-label'];
+      delete modifiedButtonAttrs['aria-label'];
+    }
+
```
supreethmr, line 185: Nit: 
`buttonAriaLabel` can be constructed here instead of line no. 211. for better readablity
```js
const customLabel =  (selectText) ? `${selectText}, ${selectedLabel}, ${label}` :  label;
buttonAriaLabel = customLabel + buttonAriaLabel;
```
since buttonAttrs is initialised with '' no need for check before using it.

packages/terra-dropdown-button/src/DropdownButton.jsx
```diff
@@ -193,7 +208,7 @@ class DropdownButton extends React.Component {
           aria-disabled={isDisabled}
           ref={this.setButtonNode}
           aria-expanded={isOpen}
-          aria-label={selectText ? `${selectText}, ${selectedLabel}` : ''}
+          aria-label={selectText ? `${selectText}, ${selectedLabel}, ${label} ${buttonAriaLabel ? `,${buttonAriaLabel}` : ''}` : `${label} ${buttonAriaLabel ? `,${buttonAriaLabel}` : ''}`}
```
supreethmr, line 211:
```suggestion
          aria-label={buttonAriaLabel}
```

packages/terra-dropdown-button/src/DropdownButton.jsx
SG067724: Yes

packages/terra-dropdown-button/src/DropdownButton.jsx
SG067724: let buttonAriaLabel = '';
if (buttonAttrs['aria-label']) {
//Now since this condition is true it will enter the if condition
and buttonAttrs['aria-label'] is deleted
But it won't enter this condition the second time which is why buttonAriaLabel will be empty and when the user tabs on the button second time it won't announce the correct aria-label.

That's why I used object spread syntax ({ ...buttonAttrs }) to create a shallow copy of the buttonAttrs object and then modify the copied object. This ensures that the original buttonAttrs object remains unchanged.

ShettyAkarsh (approved):

PR #3852, opened 2023-07-21
[terra-form-field] Updated Doc-site examples of Email validation 

KV106606Viswanath:
### Summary
Added E-mail validation and submit button to check the format of email entered is being correct or not

**What was changed:**
Previously there was no submit button to check whether the email address entered is correct or not. Now it has been implemented along with submit button.

**Why it was changed:**
To check whether the email entered is correct or not.

### Additional Details

**This PR resolves:**

UXPLATFORM-7513

---

@cerner/terra

### DIFF
--- a/packages/terra-core-docs/src/terra-dev-site/doc/form-field/example/FieldExamples.jsx
+++ b/packages/terra-core-docs/src/terra-dev-site/doc/form-field/example/FieldExamples.jsx
@@ -7,6 +7,26 @@ const cx = classNames.bind(styles);
 
 const FieldExamples = () => {
   const [isInvalid, setIsInvalid] = useState(false);
+  const [email, setEmail] = useState('');
+  const [message, setMessage] = useState('');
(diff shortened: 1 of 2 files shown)


packages/terra-core-docs/src/terra-dev-site/doc/form-field/example/FieldExamples.jsx
```diff
@@ -56,17 +76,19 @@ const FieldExamples = () => {
...
-      </Field>
-
+      <form onSubmit={emailValidation}>
+        <p> Accessibility Hooks Example Field Label </p>
+        <Field
+          label="E-mail Label"
+          htmlFor="input_id"
+          help="Please enter a valid e-mail address (abc@example.com)."
+          error={message}
+          isInvalid={emailInvalid}
```
supreethmr, line 86: when does `emailInvalid` will be set to true !!

saket2403 (approved):

PR #3854, opened 2023-07-21
[terra-tag] Create tag-list subcomponent

swetasingh9714:
### Summary
Created a new component TagList which will have tags grouped together. A Subcomponent, Tag, was created for this main component.

**What was changed:** Created a new component `terra-tag-list`.

**Why it was changed:** To be able to group all tags together and add keyboard navigation using left and right arrow keys to navigate the list.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-7629 

---

@cerner/terra


JessieRandle:
Why was a new tag component created instead of using the existing terra-tag package?

swetasingh9714:
> Why was a new tag component created instead of using the existing terra-tag package?

I created this PR only so I could get the functionality reviewed. I am still unclear whether a new package needs to be created in which case we would keep both terra-tag and terra-tag-list like it's done for terra-button or we need to have just one package like the one we have for terra-pills. If it's the former we can keep this structure else we can leverage the existing terra-tag package and update it.

supreethmr:
> > Why was a new tag component created instead of using the existing terra-tag package?
> 
> I created this PR only so I could get the functionality reviewed. I am still unclear whether a new package needs to be created in which case we would keep both terra-tag and terra-tag-list like it's done for terra-button or we need to have just one package like the one we have for terra-pills. If it's the former we can keep this structure else we can leverage the existing terra-tag package and update it.

I think design was to create a new subcomponent ( basically container which would have prop children of type terra-tag ) within terra-tag component to support list of terra-tag components. Can you revert these changes and add new subcomponent for tag-list like terra-dropdown-button-split-button.

packages/terra-tag/src/Tag.jsx
```diff
@@ -59,6 +59,10 @@ class Tag extends React.Component {
     // Apply focus styles for keyboard navigation
     if (event.nativeEvent.keyCode === KeyCode.KEY_TAB) {
       this.setState({ focused: true });
+    }else if (event.nativeEvent.keyCode === KeyCode.KEY_RIGHT) {
+      this.setState({ focused: true });
+    }else if (event.nativeEvent.keyCode === KeyCode.KEY_LEFT) {
+      this.setState({ focused: true });
```
supreethmr, line 65:
```suggestion
    }else if (event.nativeEvent.keyCode === KeyCode.KEY_RIGHT) {
      this.setState({ focused: true });
    }else if (event.nativeEvent.keyCode === KeyCode.KEY_LEFT) {
      this.setState({ focused: true });
```
Can we move this to list component as tag will support arrow keys only when they are used in group and not when they are used as individual element

swetasingh9714: there were some issues seen with the control flow when I had tried this before. If it's a necessity, will try and implement it. Should I?

sugan2416 (approved):

packages/terra-tag/src/RollupTag.jsx
```diff
@@ -0,0 +1,99 @@
+import React, {
+  useRef,
+} from 'react';
+import {
+  KEY_SPACE, KEY_RETURN,
+} from 'keycode-js';
```
sdadn, line 6: Multiple lines are unnecessary for less than 3 items.
```suggestion
import React, { useRef } from 'react';
import { KEY_SPACE, KEY_RETURN } from 'keycode-js';
```

packages/terra-tag/package.json
```diff
@@ -23,14 +23,18 @@
...
     "classnames": "^2.2.5",
     "keycode-js": "^3.1.0",
     "prop-types": "^15.5.8",
-    "terra-icon": "^3.56.0",
-    "terra-theme-context": "^1.0.0"
+    "resize-observer-polyfill": "^1.4.1",
+    "terra-icon": "^3.54.0",
+    "terra-theme-context": "^1.0.0",
+    "terra-visually-hidden-text": "^2.36.0",
+    "uuid": "^9.0.0"
```
sdadn, line 37: We are using uuid v3.4.0
```suggestion
    "uuid": "3.4.0"
```

packages/terra-tag/package.json
```diff
@@ -23,14 +23,18 @@
   "homepage": "https://github.com/cerner/terra-core#readme",
   "peerDependencies": {
     "react": "^16.8.5",
-    "react-dom": "^16.8.5"
+    "react-dom": "^16.8.5",
+    "react-intl": ">=2.8.0 <6.0.0"
```
sdadn, line 27: For better readability
```suggestion
    "react-intl": "^2.8.0 || 3 || 4 || 5"
```

packages/terra-tag/src/Tag.jsx
```diff
@@ -33,6 +33,10 @@ const propTypes = {
    * Callback function triggered when key is released.
    */
   onKeyUp: PropTypes.func,
+  /**
+   * Callback function triggered when tag is clicked with mouse.
+   */
+  onTagClick: PropTypes.func,
```
sdadn, line 39: this should be mentioned in the CHANGELOG

packages/terra-core-docs/src/terra-dev-site/doc/tag/TagList.2.doc.mdx
```diff
@@ -0,0 +1,45 @@
...
+| Peer Dependency | Version |
+|-|-|
+| react | ^16.8.5 |
+| react-dom | ^16.8.5 |
+
+<!-- AUTO-GENERATED-CONTENT:END -->
+
+## Usage
+```jsx
+import TagList from 'terra-tag-list';
```
sdadn, line 32: this import looks incorrect

packages/terra-tag/src/TagList.jsx
sdadn, line 1: This needs a unit test file

packages/terra-tag/src/TagsUtils.jsx
sdadn, line 1: Same here

packages/terra-core-docs/src/terra-dev-site/test/tag/TagListTags.test.jsx
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable no-alert */
+import React from 'react';
+import IconBookmark from 'terra-icon/lib/icon/IconTile';
+import IconTag from 'terra-icon/lib/icon/IconTag';
+import Tag from 'terra-tag';
+import TagList from 'terra-tag/src/TagList';
```
sdadn, line 6: We should _NOT_ use an internal path to import in a file (I know some Terra components use this pattern but it is a bad one).

We should create an index.js file that exports the subcomponent and then we can import it as:

```js
import { TagList } from 'terra-tag';
```

packages/terra-tag/src/RollupTag.jsx
sdadn, line 1: this needs a unit test file

sdadn (changes requested):

packages/terra-core-docs/src/terra-dev-site/doc/tag/TagList.2.doc.mdx
```diff
@@ -0,0 +1,49 @@
+import { Badge } from 'terra-tag/package.json?dev-site-package';
+
+import TagListCollapsed from './example/TagListCollapsed?dev-site-example';
+
+import TagListSmall from './example/TagListSmall?dev-site-example';
+
+import TagPropsTable from 'terra-tag/src/TagList?dev-site-props-table';
```
sdadn, line 7: We should use `lib` instead of `src`
```suggestion
import TagPropsTable from 'terra-tag/lib/TagList?dev-site-props-table';
```

packages/terra-tag/src/TagList.jsx
```diff
@@ -0,0 +1,348 @@
...
+import PropTypes from 'prop-types';
+import ResizeObserver from 'resize-observer-polyfill';
+import ThemeContext from 'terra-theme-context';
+import { injectIntl } from 'react-intl';
+import { v4 as uuidv4 } from 'uuid';
+import VisuallyHiddenText from 'terra-visually-hidden-text';
+import TagsUtils from './TagsUtils';
+import RollUpTag from './RollupTag';
+import styles from './Tag.module.scss';
+import SharedUtil from './_SharedUtil';
```
sdadn, line 21: For organization & readability
```suggestion
import React, { useCallback, useLayoutEffect, useRef, useState } from 'react';
import { KEY_LEFT, KEY_RIGHT, KEY_HOME, KEY_END } from 'keycode-js';

import { injectIntl } from 'react-intl';
import classNames from 'classnames';
import classNamesBind from 'classnames/bind';
import PropTypes from 'prop-types';
import ResizeObserver from 'resize-observer-polyfill';

import { v4 as uuidv4 } from 'uuid';

import VisuallyHiddenText from 'terra-visually-hidden-text';
import ThemeContext from 'terra-theme-context';

import TagsUtils from './TagsUtils';
import RollUpTag from './RollupTag';
import SharedUtil from './_SharedUtil';
import styles from './Tag.module.scss';
```

packages/terra-tag/src/RollupTag.jsx
```diff
@@ -0,0 +1,95 @@
+import React, { useRef } from 'react';
+import { KEY_SPACE, KEY_RETURN } from 'keycode-js';
+import { injectIntl } from 'react-intl';
+import classNamesBind from 'classnames/bind';
+import PropTypes from 'prop-types';
+import ThemeContext from 'terra-theme-context';
+import { v4 as uuidv4 } from 'uuid';
+import VisuallyHiddenText from 'terra-visually-hidden-text';
+import styles from './Tag.module.scss';
```
sdadn, line 9:
```suggestion
import React, { useRef } from 'react';
import { injectIntl } from 'react-intl';
import { KEY_SPACE, KEY_RETURN } from 'keycode-js';
import { v4 as uuidv4 } from 'uuid';
import classNamesBind from 'classnames/bind';
import PropTypes from 'prop-types';

import ThemeContext from 'terra-theme-context';
import VisuallyHiddenText from 'terra-visually-hidden-text';

import styles from './Tag.module.scss';
```

sdadn (changes requested):
If there are multiple components being exported from this package then there needs to be an `index.js` file to export them.

packages/terra-core-docs/src/terra-dev-site/doc/tag/example/HrefTag.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import IconPerson from 'terra-icon/lib/icon/IconPerson';
-import Tag from 'terra-tag';
+import { Tag } from 'terra-tag';
```
sdadn, line 3: This is a breaking change. Tag should remain a default export for passivity.

packages/terra-tag/src/index.js
```diff
@@ -0,0 +1,7 @@
+import TagList from './TagList';
+import RollupTag from './RollupTag';
+import TagsUtils from './TagsUtils';
+import Tag from './Tag';
+
+/* eslint-disable object-curly-newline */
+export { TagList, RollupTag, TagsUtils, Tag };
```
sdadn, line 7: `Tag` should be a default export, otherwise this is a breaking change.
```suggestion
export default Tag;
export { TagList, RollupTag, TagsUtils};
```

sdadn, line 7: `Tag` should be a default export, otherwise this is a _breaking change_!
```suggestion
export default Tag;
export { TagList, RollupTag, TagsUtils};
```

packages/terra-tag/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Added
+  * Added a subcomponent TagList.
```
sdadn, line 6:
```suggestion
  * Added a new subcomponent, TagList.
```

packages/terra-tag/src/Tag.jsx
```diff
@@ -59,13 +64,31 @@ class Tag extends React.Component {
     // Apply focus styles for keyboard navigation
     if (event.nativeEvent.keyCode === KeyCode.KEY_TAB) {
       this.setState({ focused: true });
+    } else if (event.nativeEvent.keyCode === KeyCode.KEY_RIGHT) {
+      this.setState({ focused: true });
+    } else if (event.nativeEvent.keyCode === KeyCode.KEY_LEFT) {
+      this.setState({ focused: true });
+    } else if (event.nativeEvent.keyCode === KeyCode.KEY_RETURN) {
+      this.setState({ focused: true });
     }
```
sdadn, line 73: These seem like a bunch of redundant checks for the same outcome. We can optimize this using a single `if` block and a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/Set).

packages/terra-tag/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
...
+  * Prop `onTagClick` in TagList, triggered when tag is clicked with mouse.
```
sdadn, line 7:
```suggestion
  * Added prop `onTagClick` in TagList which takes a callback function that is triggered when the tag is clicked.
```

packages/terra-core-docs/src/terra-dev-site/test/tag/TagListCollapsedTags.test.jsx
```diff
@@ -0,0 +1,47 @@
+/* eslint-disable no-alert */
+import React from 'react';
+import IconPerson from 'terra-icon/lib/icon/IconPerson';
+import IconBookmark from 'terra-icon/lib/icon/IconTile';
```
sdadn, line 4: With the recent changes in #3915 , Icons can now be imported as:

```suggestion
import { IconBookmark, IconPerson } from 'terra-icon';
```

sdadn (changes requested):

packages/terra-tag/src/Tag.jsx
MadanKumarGovindaswamy: changes done. Thanks

packages/terra-tag/CHANGELOG.md
MadanKumarGovindaswamy: changes done. Thanks

packages/terra-core-docs/src/terra-dev-site/test/tag/TagListCollapsedTags.test.jsx
MadanKumarGovindaswamy: changes done. Thanks

packages/terra-tag/src/index.js
MadanKumarGovindaswamy: changes done. Thanks

packages/terra-tag/src/Tag.jsx
```diff
@@ -97,16 +116,19 @@ class Tag extends React.Component {
...
         {...customProps}
         className={tagClasses}
+        role={tagRole}
         onKeyUp={this.handleKeyUp}
         onBlur={this.handleOnBlur}
-        onClick={onClick}
+        onClick={this.handleClick}
         onFocus={onFocus}
         href={href}
+        data-terra-tag
```
sdadn, line 131: Is this a typo? This does not look like a JavaScript variable.

packages/terra-tag/tests/jest/RollUpTag.test.jsx
sdadn, line 1: For these unit tests, we should also explicitly test for values and not just rely on snapshot testing.

packages/terra-tag/tests/jest/RollUpTag.test.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+/* eslint-disable-next-line import/no-extraneous-dependencies */
+import { mountWithIntl } from 'terra-enzyme-intl';
+import { v4 as uuidv4 } from 'uuid';
+/* eslint-disable import/no-extraneous-dependencies */
+import { RollupTag } from 'terra-tag';
```
sdadn, line 6: Does this use the local source code?

sdadn (changes requested):

packages/terra-tag/tests/jest/RollUpTag.test.jsx
MadanKumarGovindaswamy: Updated the import path points to the local. Thanks

packages/terra-tag/src/Tag.jsx
MadanKumarGovindaswamy: Added for selectors here
Screenshot 2023-09-25 at 7 07 49 PM

packages/terra-tag/tests/jest/RollUpTag.test.jsx
MadanKumarGovindaswamy: changes done. Thanks

packages/terra-tag/src/index.js
```diff
@@ -0,0 +1,7 @@
+import TagList from './TagList';
+import RollupTag from './RollupTag';
+import TagsUtils from './TagsUtils';
+import Tag from './Tag';
+
+export default Tag;
+export { TagList, RollupTag, TagsUtils };
```
sdadn, line 7: I thought only `TagList` was exported. Are `RollupTag` and `TagsUtils` also meant for consumers?
cc: @supreethmr

MadanKumarGovindaswamy: Changes Done. Thanks

sdadn: +1

packages/terra-tag/src/TagsUtils.jsx
sdadn, line 1: if this is an internal file, then we should rename this to `_TagUtils.jsx`. Same with `_RollupTag.jsx`.

packages/terra-tag/src/RollupTag.jsx
```diff
@@ -0,0 +1,94 @@
+import React, { useRef } from 'react';
+import { injectIntl } from 'react-intl';
+import { KEY_SPACE, KEY_RETURN } from 'keycode-js';
+import { v4 as uuidv4 } from 'uuid';
+import classNamesBind from 'classnames/bind';
+import PropTypes from 'prop-types';
+import ThemeContext from 'terra-theme-context';
+import VisuallyHiddenText from 'terra-visually-hidden-text';
+import styles from './Tag.module.scss';
```
cm9361, line 9: This new RollupTag component should have its own style module.

packages/terra-tag/src/RollupTag.jsx
```diff
@@ -0,0 +1,94 @@
...
+    >
+      <button
+        id={rollUpTagId}
+        className={cx(['rollup-tag', theme.className])}
+        onClick={handleOnSelectRollUp}
+        onKeyDown={handleRollUpTagKeyDown}
+        onMouseDown={handleRollUpTagMouseDown}
+        onBlur={handleOnBlur}
+        ref={rollUpTagRef}
+        tabIndex="0"
```
cm9361, line 80: You should not need to set the tabIndex property to 0 on a button element.  The tabIndex will be 0 by default.

packages/terra-tag/src/TagList.jsx
```diff
@@ -0,0 +1,342 @@
...
+    }
+  }, [children, isCollapsed]);
+
+  // Modifies the tabindex of the tag
+  const setTabIndex = (val) => {
+    const currentNode = currentTag.current ? filterTagsRef.current.querySelector(`#${currentTag.current}`) : null;
+    if (currentNode) {
+      currentNode.setAttribute('tabIndex', val);
+    }
+  };
```
cm9361, line 99: We should create an appropriate state variable and update the tabIndex when the value changes.  We should not need to query the DOM since we are rendering the components, correct?

packages/terra-tag/src/TagList.jsx
```diff
@@ -0,0 +1,342 @@
...
+
+  // Sets focus to the current tag with tabindex 0
+  const focusCurrentNode = () => {
+    const currentNode = currentTag.current ? filterTagsRef.current.querySelector(`#${currentTag.current}`) : null;
+    if (currentNode) {
+      if (SharedUtil.isSafari()) {
+        // Safari Browser
+        setTimeout(() => {
```
cm9361, line 107: Why do we need a setTimeout function?  I am not aware of the focus method needing to be called inside a setTimeout for Safari.

packages/terra-tag/src/RollupTag.jsx
```diff
@@ -0,0 +1,94 @@
...
+
+const RollUpTag = (props) => {
+  const {
+    isCollapsed, onSelectRollUp, intl, rollupCount,
+  } = props;
+  const rollUpTagRef = useRef();
+  const theme = React.useContext(ThemeContext);
+
+  const handleOnSelectRollUp = (event) => {
+    rollUpTagRef.current.setAttribute('data-terra-rollup-tag-show-focus-styles', 'true');
```
cm9361, line 41: Why do we need to set a data attribute here?

packages/terra-tag/src/RollupTag.jsx
```diff
@@ -0,0 +1,94 @@
...
+    event.preventDefault();
+    event.stopPropagation();
+    onSelectRollUp(event);
```
cm9361, line 44: This method is not required.  You need to add a check to ensure that it is defined to avoid an exception.

packages/terra-tag/src/RollupTag.jsx
```diff
@@ -0,0 +1,94 @@
...
+  const {
+    isCollapsed, onSelectRollUp, intl, rollupCount,
+  } = props;
+  const rollUpTagRef = useRef();
+  const theme = React.useContext(ThemeContext);
+
+  const handleOnSelectRollUp = (event) => {
+    rollUpTagRef.current.setAttribute('data-terra-rollup-tag-show-focus-styles', 'true');
+    event.preventDefault();
+    event.stopPropagation();
```
cm9361, line 43: Do we need to stop the event propagation and default behavior?  If so, why?  We should probably add a comment to ensure it is clear and not broken by developers in the future.

packages/terra-tag/src/RollupTag.jsx
```diff
@@ -0,0 +1,94 @@
...
+    onSelectRollUp(event);
+  };
+
+  const handleRollUpTagKeyDown = (event) => {
+    rollUpTagRef.current.setAttribute('data-terra-rollup-tag-show-focus-styles', 'true');
```
cm9361, line 48: In a React component, we should not be dependent upon data attributes.  You are interacting directly with the DOM instead of taking advantage of the virtual DOM.

packages/terra-tag/src/RollupTag.jsx
```diff
@@ -0,0 +1,94 @@
...
+    if (event.keyCode === KEY_RETURN || event.keyCode === KEY_SPACE) {
+      event.preventDefault();
+      event.stopPropagation();
+      onSelectRollUp(event);
```
cm9361, line 52: This method is not required.  You need to ensure it exists before calling it.

packages/terra-tag/src/RollupTag.jsx
```diff
@@ -0,0 +1,94 @@
...
+    }
+  };
+
+  const handleRollUpTagMouseDown = () => {
+    rollUpTagRef.current.setAttribute('data-terra-rollup-tag-show-focus-styles', 'false');
+  };
+
+  const handleOnBlur = () => {
+    rollUpTagRef.current.setAttribute('data-terra-rollup-tag-show-focus-styles', 'true');
+  };
```
cm9361, line 62: You should not be using the DOM.  You could use a ref variable, if necessary.

packages/terra-tag/src/RollupTag.jsx
```diff
@@ -0,0 +1,94 @@
...
+        ref={rollUpTagRef}
+        tabIndex="0"
+        type="button"
+        aria-describedby={visuallyHiddenContentId}
+        data-terra-rollup-tag-show-focus-styles
+        data-terra-rollup-tag
+      >
+        <span className={cx('rollup-tag-label')}>{rollUpLabel}</span>
+      </button>
+      <VisuallyHiddenText id={visuallyHiddenContentId} text={rollUpHint} aria-hidden="true" />
```
cm9361, line 88: Why do we need both VisuallyHiddenText and an aria-describedby attribute?  Instead of dealing with identifiers, could we not just use the aria-label for the button?  I believe that would remove the uuid dependency, correct?

packages/terra-tag/src/RollupTag.jsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const rollUpLabel = isCollapsed && rollupCount > 0 ? intl.formatMessage({ id: 'Terra.tags.label.rollupTag' }, { tagsNotVisibleCount: rollupCount }) : intl.formatMessage({ id: 'Terra.tags.label.showLess' });
+  const rollUpHint = isCollapsed ? intl.formatMessage({ id: 'Terra.tags.hint.rollupTag' }, { tagsNotVisibleCount: rollupCount }) : intl.formatMessage({ id: 'Terra.tags.hint.showLess' });
+  const rollUpTagId = `terra-rollup-tag-${uuidv4()}`;
+  const visuallyHiddenContentId = `terra-rollup-tag-${uuidv4()}`;
+  return (
+    <div
+      className={cx('tag-list-item')}
+    >
+      <button
```
cm9361, line 72: Should we be using a Terra Button control for our button?

packages/terra-tag/src/RollupTag.jsx
```diff
@@ -0,0 +1,94 @@
...
+        onClick={handleOnSelectRollUp}
+        onKeyDown={handleRollUpTagKeyDown}
+        onMouseDown={handleRollUpTagMouseDown}
+        onBlur={handleOnBlur}
+        ref={rollUpTagRef}
+        tabIndex="0"
+        type="button"
+        aria-describedby={visuallyHiddenContentId}
+        data-terra-rollup-tag-show-focus-styles
+        data-terra-rollup-tag
```
cm9361, line 84: I don't think we should need these data attributes.

packages/terra-tag/src/Tag.jsx
cm9361, line 39: Is clicking the only way to activate this?  If so, ss there a specific reason we only care about the click event as it would not be accessible?

packages/terra-tag/src/Tag.jsx
```diff
@@ -66,6 +72,18 @@ class Tag extends React.Component {
...
+  handleClick(event) {
+    if (this.props.onTagClick) {
+      this.props.onTagClick(event.target.id);
+    }
+    if (this.props.onClick) {
+      this.props.onClick();
+    }
+    setTimeout(() => {
+      this.setState({ focused: true });
+    }, 100);
```
cm9361, line 84: Why do we need a setTimeout?

packages/terra-tag/src/Tag.jsx
```diff
@@ -97,16 +116,19 @@ class Tag extends React.Component {
     const isClickable = href || onClick;
     const clickableComponent = href ? 'a' : 'button';
     const ComponentType = isClickable ? clickableComponent : 'span';
+    const tagRole = clickableComponent === 'a' ? 'link' : 'button';
```
cm9361, line 119: The role should be handled by the element being an anchor or button.  This logic does not seem necessary.

packages/terra-tag/translations/en-CA.json
```diff
@@ -0,0 +1,15 @@
+{
+  "Terra.tags.label.delete": "Remove {tagLabelName}",
+  "Terra.tags.label.rollupTag": "{tagsNotVisibleCount} more...",
```
cm9361, line 3: Why does it end with an ellipsis in the screen reader text?

packages/terra-tag/translations/en-CA.json
```diff
@@ -0,0 +1,15 @@
...
+  "Terra.tags.label.showLess": "Show Less",
+  "Terra.tags.hint.removable": "item is removable via Delete or Back Space key",
+  "Terra.tags.hint.selectable": "for additional information, button is selectable via Space Bar or Enter key",
+  "Terra.tags.hint.selectableAndRemovable": "item is removable via Delete or Back Space key, for additional information button is selectable via Space Bar or Enter key",
```
cm9361, line 7: It should be a period (.) after "Back space key", not a comma.

packages/terra-tag/src/_SharedUtil.js
```diff
@@ -0,0 +1,22 @@
...
+
+    return false;
+  }
+
+  /**
+   * Util to determine if the user agent indicates that it is macOS
+   * @return {boolean}
+   */
+
+  static isMac = () => navigator.userAgent.indexOf('Mac') !== -1 && navigator.userAgent.indexOf('Win') === -1;
```
cm9361, line 19: I cannot find where isMac is used.  I am also curious why we need this file.  What logic is really specific only to Safari?

packages/terra-tag/src/TagsUtils.jsx
```diff
@@ -0,0 +1,67 @@
...
+  const rollupTagWidth = rollUpTag.parentElement.offsetWidth + parseFloat(window.getComputedStyle(rollUpTag.parentElement, null).getPropertyValue('margin-right'), 10);
+  return rollupTagWidth;
+};
+
+/**
+ * Takes tag container ref as input and returns the index value from which the tags need to be hidden/rolled up.
+ */
+const getRollUpIndex = (tagListRef) => {
+  const rollUpTag = tagListRef.current.querySelector('[data-terra-rollup-tag]');
+  const tags = tagListRef.current.querySelectorAll('[data-terra-tag]');
```
cm9361, line 22: Why do you need to query the DOM?  Don't you have the reference data in React props?

packages/terra-tag/src/TagsUtils.jsx
```diff
@@ -0,0 +1,67 @@
...
+    tags[i].setAttribute('tabindex', val);
+  }
+};
+
+/**
+ * Takes 'rollUpTag' element and tabindex 'val' as inputs and sets the rollUpTag tabindex.
+ */
+const setRollUpTagTabIndex = (rollUpTag, val) => {
+  rollUpTag.setAttribute('tabindex', val);
+};
```
cm9361, line 59: Couldn't this be handled by the re-render?

packages/terra-tag/src/TagsUtils.jsx
MadanKumarGovindaswamy: Changes Done Thanks.

packages/terra-tag/src/TagList.jsx
MadanKumarGovindaswamy: Removed setTimeout as well as isSafari Check. Thanks

packages/terra-tag/src/_SharedUtil.js
MadanKumarGovindaswamy: Removed IsMac as no references found. Thanks

packages/terra-tag/src/RollupTag.jsx
MadanKumarGovindaswamy: Added check to avoid exception. Thanks

packages/terra-tag/src/RollupTag.jsx
MadanKumarGovindaswamy: Added check to avoid exception. Thanks

packages/terra-tag/translations/en-CA.json
MadanKumarGovindaswamy: Changes Made. Thanks

packages/terra-tag/src/RollupTag.jsx
MadanKumarGovindaswamy: Removed TabIndex set to 0. Thanks

packages/terra-tag/src/RollupTag.jsx
MadanKumarGovindaswamy: changes Done. Thanks

packages/terra-tag/src/Tag.jsx
MadanKumarGovindaswamy: Removed this lines. Thanks

packages/terra-tag/src/RollupTag.jsx
MadanKumarGovindaswamy: We use VisuallyHiddenText for announcing the description. if we use aria label we need to concate both label as well as description. But still in the file uuid dependency is used for button Id. Thanks

packages/terra-tag/src/Tag.jsx
MadanKumarGovindaswamy: Added additional description for this prop.
This is an Internal prop used to focus to current tag if the tag is clicked with mouse. And there are other props exposed to consumers to manage the events.
Thanks

packages/terra-tag/src/RollupTag.jsx
MadanKumarGovindaswamy: Changed Done .Thanks

packages/terra-tag/src/Tag.jsx
MadanKumarGovindaswamy: I guess this was used for focus issue. Have removed this. Thanks

packages/terra-tag/src/TagsUtils.jsx
MadanKumarGovindaswamy: This resets all tag nodes tabindex to -1, except for the one tag that receives focus. Thanks

packages/terra-tag/src/TagsUtils.jsx
MadanKumarGovindaswamy: tagListRef is a ref item prop, which contains all tag nodes plus rollup tag node. For differentiating we use data prop here. Thanks

packages/terra-tag/src/RollupTag.jsx
MadanKumarGovindaswamy: We use this data attributes for css focus.
Screenshot 2023-10-06 at 9 29 04 PM

packages/terra-tag/src/TagList.jsx
MadanKumarGovindaswamy: Since we invoke this setTabIndex method on every keyboard navigation to toggle the focused tag tabindex using a state will re-render for every navigation we have used the ref here to updated the tabindex of focused tab. Thanks

packages/terra-tag/src/RollupTag.jsx
MadanKumarGovindaswamy: Screenshot 2023-10-06 at 9 29 04 PM
we have used the dotted border for the focused tag item. Css change

packages/terra-tag/src/RollupTag.jsx
MadanKumarGovindaswamy: Screenshot 2023-10-06 at 9 57 06 PM
rollUpTagRef is a ref here. we have used this for css identifying the focused tag item. Thanks

packages/terra-tag/src/RollupTag.jsx
MadanKumarGovindaswamy: Since this TagList is replica of Terra-pills we have used most the same logic here for setting the focus of selected tag item.
Thanks

packages/terra-tag/src/RollupTag.jsx
MadanKumarGovindaswamy: Replaced html button with terra-button. Thanks

packages/terra-tag/src/RollupTag.module.scss
saket2403, line 1: Should'nt we add values for other two themes as well for all the themed style variables?

packages/terra-tag/src/Tag.module.scss
```diff
@@ -39,7 +39,11 @@
     }
 
     &:focus {
-      outline: none;
+      border-color: var(--terra-tags-rollup-tag-focus-border-color, #868a8c);
+      border-width: var(--terra-tags-rollup-tag-focus-border-width, 1px);
+      box-shadow: var(--terra-tags-rollup-tag-focus-box-shadow, none);
+      outline: var(--terra-tags-rollup-tag-focus-outline, 2px dashed #000);
+      outline-offset: var(--terra-tags-rollup-tag-focus-outline-offset, 2px);
```
saket2403, line 46: Please add values for fusion and lowlight for relevant theme variables

packages/terra-tag/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Added
+  * Added a new subcomponent, TagList.
+  * Added prop `onTagClick` in TagList which takes a callback function that is triggered when the tag is clicked.
```
saket2403, line 7: Since you are adding a new subcomponent as mentioned above, no need to mention a specific private prop in changelog.

MadanKumarGovindaswamy: Changes Added. Thanks

packages/terra-tag/src/Tag.module.scss
MadanKumarGovindaswamy: Changes Done. Thanks

packages/terra-tag/src/RollupTag.module.scss
MadanKumarGovindaswamy: Added. Thanks

packages/terra-tag/src/RollupTag.jsx
cm9361: We should not replicate bad practices in a new component.  This is not the proper way to handle this in React.

packages/terra-tag/src/RollupTag.jsx
cm9361: You should be using state variables and updating in the render of the component.  We should not be setting attributes in the DOM.

packages/terra-tag/src/RollupTag.jsx
cm9361: Wouldn't you use classes instead of data attributes?

packages/terra-tag/src/Tag.jsx
cm9361: What happens if the current tag and focus and the user presses enter?  Also, if it is internal, why does it need to be a prop?

packages/terra-tag/src/TagsUtils.jsx
cm9361: It still could be handled by a rerender.

packages/terra-tag/src/TagsUtils.jsx
cm9361: We should not have to access the DOM for this information though.  Somehow we populated the DOM from the data passed to the component.  React should rely on the virtual DOM, when possible.

packages/terra-tag/src/Tag.jsx
```diff
@@ -57,7 +64,8 @@ class Tag extends React.Component {
 
   handleKeyUp(event) {
     // Apply focus styles for keyboard navigation
-    if (event.nativeEvent.keyCode === KeyCode.KEY_TAB) {
+    const eventKeyCode = new Set([KeyCode.KEY_TAB, KeyCode.KEY_LEFT, KeyCode.KEY_RIGHT, KeyCode.KEY_RETURN]);
```
cm9361, line 67: This code does not seem correct to me.  You are not actually moving the focus so why are we making a state change?

packages/terra-tag/src/Tag.jsx
```diff
@@ -66,6 +74,16 @@ class Tag extends React.Component {
     }
   }
 
+  handleClick(event) {
+    if (this.props.onTagClick) {
+      this.props.onTagClick(event.target.id);
```
cm9361, line 79: We should not be using event.target.  You are not guaranteed to get the item that you desire.  It might be working because your component is simple.  Since we are using React, the callback should be able to provide the necessary information without trying to rely on event.target.

packages/terra-tag/src/TagList.jsx
```diff
@@ -0,0 +1,331 @@
...
+  const setTabIndex = (val) => {
+    const currentNode = currentTag.current ? filterTagsRef.current.querySelector(`#${currentTag.current}`) : null;
+    if (currentNode) {
+      currentNode.setAttribute('tabIndex', val);
+    }
+  };
+
+  // Sets focus to the current tag with tabindex 0
+  const focusCurrentNode = () => {
+    const currentNode = currentTag.current ? filterTagsRef.current.querySelector(`#${currentTag.current}`) : null;
```
cm9361, line 102: It looks like you are using querySelector to find an element by id.  There is a document.getElementById method for that.

packages/terra-tag/src/TagList.jsx
```diff
@@ -0,0 +1,331 @@
...
+        break;
+      case KEY_LEFT:
+        event.preventDefault();
+        focusPreviousNode(tags, rollUpTag);
+        break;
+      case KEY_HOME:
+        event.preventDefault();
+        setTabIndex('-1');
+        focusNode.current = 0;
+        currentTag.current = tags[focusNode.current].id;
```
cm9361, line 221: Should we have refs on the Tags component elements instead of the element ids?

packages/terra-tag/src/_RollupTag.jsx
```diff
@@ -0,0 +1,92 @@
...
+    rollUpTagRef.current.setAttribute('data-terra-rollup-tag-show-focus-styles', 'true');
+  };
+
+  const rollUpLabel = isCollapsed && rollupCount > 0 ? intl.formatMessage({ id: 'Terra.tags.label.rollupTag' }, { tagsNotVisibleCount: rollupCount }) : intl.formatMessage({ id: 'Terra.tags.label.showLess' });
+  const rollUpHint = isCollapsed ? intl.formatMessage({ id: 'Terra.tags.hint.rollupTag' }, { tagsNotVisibleCount: rollupCount }) : intl.formatMessage({ id: 'Terra.tags.hint.showLess' });
+  const rollUpTagId = `terra-rollup-tag-${uuidv4()}`;
+
+  return (
+    <div
+      className={cx('tag-list-item')}
```
cm9361, line 68: Why do we need an extra div element around each single button?

packages/terra-tag/src/index.js
```diff
@@ -0,0 +1,5 @@
+import TagList from './TagList';
+import Tag from './Tag';
+
+export default Tag;
```
cm9361, line 4: Is this correct to make the default, Tag, when there are multiple components exported for this package?

packages/terra-tag/tests/wdio/__snapshots__/reference/clinical-lowlight-theme/en/chrome_tiny/tag-list-spec/focus_second_tag_of_list_1.png
cm9361, line 1: Are we sure the colors are correct for 11 more here?  You can barely see the text.

cm9361, line 1: Why is this rolling up so early and leaving so much additional whitespace in the container?

sdadn:
Marking it as blocked since it seems like we'll need to have a design discussion around this PR. 

cc: @supreethmr

packages/terra-tag/tests/wdio/__snapshots__/reference/clinical-lowlight-theme/en/chrome_tiny/tag-list-spec/focus_second_tag_of_list_1.png
MadanKumarGovindaswamy: Screenshot 2023-10-11 at 4 36 59 PM
Screenshot 2023-10-11 at 4 37 50 PM
This is happening because of the long text and icon present in the tag working in a responsive manner. Thanks

packages/terra-tag/src/index.js
sdadn: Before this PR, there was only one export, `Tag`. I suggested to make it the default export to maintain passivity.

packages/terra-tag/src/Tag.jsx
MadanKumarGovindaswamy: changes done. Thanks

packages/terra-tag/src/Tag.jsx
MadanKumarGovindaswamy: Updated. Thanks

packages/terra-tag/src/TagList.jsx
MadanKumarGovindaswamy: We can still use that, But in worst case if multiple ID'S present with same name in the document could create a problem. Here filterTagsRef variable is a ref which hold all tag elements.
Thanks

packages/terra-tag/tests/wdio/__snapshots__/reference/clinical-lowlight-theme/en/chrome_tiny/tag-list-spec/focus_second_tag_of_list_1.png
MadanKumarGovindaswamy: Updated.
 Thanks

packages/terra-tag/src/TagList.jsx
```diff
@@ -0,0 +1,331 @@
...
+    observer.observe(filterTagsRef.current.parentNode);
+
+    return () => {
+      observer.disconnect();
+      observer = null;
+    };
+  }, [children, handleResize]);
+
+  const focusNextNode = (tags, rollUpTag) => {
+    const rollUpTagId = rollUpTag ? rollUpTag.getAttribute('id') : null;
```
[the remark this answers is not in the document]
MadanKumarGovindaswamy: Updated.  we were using element id to find the tag for setting the focus. Now have removed the support of ID.
Thanks

packages/terra-tag/src/TagList.jsx
MadanKumarGovindaswamy: Updated. Now have removed the support of ID.
Thanks

MadanKumarGovindaswamy:
> Marking it as blocked since it seems like we'll need to have a design discussion around this PR.
> 
> cc: @supreethmr

The Design Changes for Terra-TagList was reviewed with UX team and got an approval from Libby.
cc. @Herald Samuel Sujith, @supreethmr

packages/terra-tag/src/TagList.jsx
```diff
@@ -0,0 +1,331 @@
...
+      setUpdatedCount(startIndex);
+      setRollUpCount(React.Children.count(children) - startIndex);
+    } else if (startIndex === React.Children.count(children)) {
+      setRollUpCount(0);
+    }
+  }, [children, isCollapsed]);
+
+  // Modifies the tabindex of the tag
+  const setTabIndex = (val) => {
+    const currentNode = currentTag.current ? filterTagsRef.current.querySelector(`#${currentTag.current}`) : null;
```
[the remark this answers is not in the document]
MadanKumarGovindaswamy: Hi,
Similar approach is being used in the other terra-components to handle keyboard navigation. we have also validated with screen reader mode and it works as expected. if we use the state it will cause re-render of component which is not required for this tabIndex attribute. Introducing state variable will be help full only when we are updating content of the component.
Thanks

packages/terra-tag/src/_RollupTag.jsx
MadanKumarGovindaswamy: This div was used as wrapper for rollup button. Updated this.
Thanks

sdadn:
> > Marking it as blocked since it seems like we'll need to have a design discussion around this PR.
> > cc: @supreethmr
> 
> The Design Changes for Terra-TagList was reviewed and got an approval from UX Team. You can find the UX approval document on this Jira: --- UXPLATFORM-7629 cc. @Herald Samuel Sujith, @supreethmr

There are still concerns around the implementation of the code changes. We can discuss this in detail offline.

packages/terra-core-docs/src/terra-dev-site/doc/tag/TagList.2.doc.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+<!-- AUTO-GENERATED-CONTENT:START Peer Dependencies -->
+## Peer Dependencies
+
+This component requires the following peer dependencies be installed in your app for the component to properly function.
+
+| Peer Dependency | Version |
+|-|-|
+| react | ^16.8.5 |
+| react-dom | ^16.8.5 |
```
supreethmr, line 28: Can we add React-Intl as a peer dependency in both Tag About page and here as well

packages/terra-core-docs/src/terra-dev-site/doc/tag/TagList.2.doc.mdx
```diff
@@ -0,0 +1,49 @@
...
+
+<!-- AUTO-GENERATED-CONTENT:END -->
```
supreethmr, line 30: we might need to add implementation details on both tag and taglist about page similar to this  https://github.com/cerner/terra-core/blob/ceabc03e855b802114d93f40b95594068b7c0503/packages/terra-core-docs/src/terra-dev-site/doc/badge/About.1.doc.mdx#L35

packages/terra-core-docs/src/terra-dev-site/doc/tag/TagList.2.doc.mdx
supreethmr: Updated :) https://github.com/cerner/terra-core/pull/3854/commits/8ded50f47258067e34fd8a01eed5eb3b47b92fd7#diff-43d71167fbce146e41d65538d63d237642369f5d98c53c55d8294d764b82ddfb

packages/terra-core-docs/src/terra-dev-site/doc/tag/TagList.2.doc.mdx
supreethmr: https://github.com/cerner/terra-core/pull/3854/commits/8ded50f47258067e34fd8a01eed5eb3b47b92fd7#diff-a9cc2ccc1dba9de522f4dfa33689424aee764e7c6f91af3b23e597e3b897afd3

packages/terra-tag/src/TagList.jsx
supreethmr: It has been added with commit : https://github.com/cerner/terra-core/pull/3854/files#diff-c795f28e1f11a92cd988ce46427df277c1f37df1557058cbad5feec737d88840

supreethmr (approved):

rahulkumar8cs:
+1

packages/terra-tag/src/TagList.jsx
```diff
@@ -0,0 +1,329 @@
...
+      setUpdatedCount(startIndex);
+      setRollUpCount(React.Children.count(children) - startIndex);
+    } else if (startIndex === React.Children.count(children)) {
+      setRollUpCount(0);
+    }
+  }, [children, isCollapsed]);
+
+  // Modifies the tabindex of the tag
+  const setTabIndex = (val) => {
+    const currentNode = currentTag.current ? currentTag.current : null;
```
cm9361, line 94: This line of code seems unnecessary.  You are using the value if defined or setting to null.  However, you just do an if check on the next line that nullifies the need for this line.

cm9361:
> > This is not the correct way to handle this in React. You should update the state variable and update in the rerender.
> 
> @cm9361 I took a closer look on the code after our discussion yesterday, the one of the reason why we cannot use state to update tabIndex is because here we are trying to update the tabIndex of the child element which is tag. In order to update tabIndex with state we might need to pass down this value as prop for tag, this would require us to maintain an array of states for both tabIndex and position in order to update tabIndex that are appropriate to individual tag (-1 or 0 depending on it's position). Maintaining state will be helpful when we are making changes on same component OR if we are passing down same value for all child components. In this case we are updating tabIndex based on the interactions like mouse click, key press and on rollup and collapse which makes it difficult to pass down appropriate tabIndex for Tag from Taglist

It is based on the way the logic has been desired.  This could be handled in React but would require architectural design changes.  I can review it in its current state due to time constraints, but I am not sure this is the correct design.

packages/terra-tag/src/Tag.jsx
```diff
@@ -42,30 +41,29 @@ const propTypes = {
...
     if (this.props.onKeyUp) {
       this.props.onKeyUp(event);
     }
   }
 
+  handleClick(event) {
+    if (this.props.onClick) {
+      this.props.onClick(event);
+    }
+  }
```
cm9361, line 65: Do you not have a requirement for this to also be triggered Return and/or Space?

packages/terra-tag/src/Tag.module.scss
cm9361, line 46: Why do we need a border, box shadow, and outline?  It seems like we are using three different approaches in one implementation.

packages/terra-tag/src/RollupTag.module.scss
```diff
@@ -0,0 +1,100 @@
...
+    line-height: var(--terra-tags-rollup-tag-line-height, 1.16667);
+    margin: 0;
+    margin-bottom: var(--terra-tags-tag-list-item-margin-bottom, 0.28571rem);
+    margin-right: var(--terra-tags-tag-list-item-margin-right, 0.42857rem);
+    margin-top: var(--terra-tags-tag-list-item-margin-top, 0.28571rem);
+    max-width: 100%; // Needed for IE10 truncation
+    min-width: 0;
+    outline: none;
+    overflow: hidden;
+    padding: 0;
```
cm9361, line 53: Should we be setting the padding to 0 and then to values afterwards?  Do the following lines cancel this style?

packages/terra-tag/src/RollupTag.module.scss
```diff
@@ -0,0 +1,100 @@
...
+    border-radius: var(--terra-tags-rollup-tag-border-radius, 3px);
+    border-style: solid;
+    border-width: var(--terra-tags-rollup-tag-border-width, 1px);
+    box-shadow: var(--terra-tags-rollup-tag-box-shadow, 0 1px 2px 0 rgba(0, 0, 0, 0.2));
+    box-sizing: border-box;
+    color: var(--terra-tags-rollup-tag-color, #595f62);
+    cursor: pointer;
+    display: inline-block;
+    flex: 1 1 auto;
+    font-family: inherit;
```
cm9361, line 41: Is this style necessary?  Isn't this the default behavior?

packages/terra-tag/src/RollupTag.module.scss
```diff
@@ -0,0 +1,100 @@
...
+    vertical-align: top;
+    white-space: nowrap;
+
+    &[data-terra-rollup-tag-show-focus-styles='true']:focus {
+      background-color: var(--terra-tag-focus-background-color, #fff);
+      border-color: var(--terra-tag-focus-border-color, rgba(76, 178, 233, 0.5));
+      border-width: var(--terra-tags-rollup-tag-focus-border-width, 1px);
+      box-shadow: var(--terra-tags-rollup-tag-focus-box-shadow, none);
+      outline: var(--terra-tags-rollup-tag-focus-outline, 2px dashed #000);
+      outline-offset: var(--terra-tags-rollup-tag-focus-outline-offset, 2px);
```
cm9361, line 69: We should not need to set the border, box shadow, and outline.  We should be able to use one style to show the focus indication.  We use outline and outline-offset in other Terra implementations.

packages/terra-tag/src/Tag.jsx
MadanKumarGovindaswamy: Terra-Tag can be either button or links based on props provided.  "handleClick" handler gets invoked on both space and return key for button by default. For hyperlink "handleClick" handler will be triggered by only return key.

packages/terra-tag/src/TagList.jsx
MadanKumarGovindaswamy: Changes made Thanks.

packages/terra-tag/src/RollupTag.module.scss
MadanKumarGovindaswamy: Updated. Thanks
commit: b6dcb40f236e4f0e9b0fde34475a6946d4a2e6ba

packages/terra-tag/src/Tag.module.scss
MadanKumarGovindaswamy: Updated Thanks
commit: e43a8de12501cf88bde82bee516607eed4736a70

packages/terra-tag/src/RollupTag.module.scss
MadanKumarGovindaswamy: No we have lint error setup should be in a order and also does not made any differences after render. Thanks

packages/terra-tag/src/Tag.module.scss
```diff
@@ -5,7 +5,7 @@
 :local {
   .tag {
     align-items: center;
-    background-color: var(--terra-tag-background-color, rgba(28, 31, 33, 0.04));
+    background-color: var(--terra-tag-background-color, rgba(240, 243, 245, 0.04));
```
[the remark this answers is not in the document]
MadanKumarGovindaswamy: Did not find any UX requirement change for this, reverting back this change. Thanks
aae5ba1fa9acbaac420e4798c9e3d6acebcd0321

MadanKumarGovindaswamy:
> It is based on the way the logic has been desired. This could be handled in React but would require architectural design changes. I can review it in its current state due to time constraints, but I am not sure this is the correct design.

@cm9361  We have addressed the PR review comments can you have a look into that.

We updated Terra-TagList to use state variables to update Tabindex and we found that there was not much difference on page load time compared to the existing way of updating the tabIndex instead this approach adding state requires more code changes to the existing implementation of TagList. 

Adding snapshots of performance loading:
>> Below screenshot was taken while performing keyboard navigation which updates tabIndex through setAttribute method
Navigation without state with multiple tags

Below screenshot was taken while performing keyboard navigation which updates tabIndex through state variable
Navigation after state with multiple tags

The existing approach has been validated with multiple components and scenarios and works as expected. 
New implementation requires some of the validations to be performed again which conflicts with the current time constraints.

Thanks

packages/terra-tag/src/RollupTag.module.scss
cm9361: I think the comment might have been misunderstood.  You are setting a style here that you are overriding in the next 4 lines.  I am saying that this line is not necesary.

MadanKumarGovindaswamy: Updated Thanks
commit Link: 2228c2c9c32ef975b0f1d078f314d436dca9fbe5

cm9361 (approved):

sdadn (approved):

MadanKumarGovindaswamy:
Hi @scottwilmarth . Could you please review this PR. It's been pending for the UXApproval. Thanks

supreethmr:
@MadanKumarGovindaswamy Can you check the cause for build failures

MadanKumarGovindaswamy:
> @MadanKumarGovindaswamy Can you check the cause for build failures

some lint error after merging with main branch with recent changes. Even local is getting failed.

rahulkumar8cs:
+1 for accessibility testing and Approved
Validated with Windows 11 cloud desktop, Edge Version 116.0.1938.62 (Official build) (64-bit), JAWS 2023.2307.37
When VPC cursor is on:
When user navigates to the "11 more" button, then the hidden list items appears on the page. Then when user further navigates to the "Show Less" button and trigger it, then list items are collapsed and the button instruction "11 items are hidden, select to display hidden items" is exposed twice by JAWS.

MadanKumarGovindaswamy:
Created a jira for this issue found in accessibility testing.
JIRA Link: https://jira2.cerner.com/browse/UXPLATFORM-9812
Thanks

PR #3855, opened 2023-07-25
[Terra-Form-Select] Fix for Form Single Select Voice Over Reader Issue.

ST063655:
### Summary

**What was changed:**
1. If the index is equal to 1 and the item is not selected, the String was added to read as Not Selected.
2. Removed role as radio for single selection.
3. On the Down Arrow Key, Item from the last item, the index is set to first option.
4. On the Up Arrow Key, Item from the last item, the index is set to option. length.

**Why it was changed:**
1. Safari+VO =>When the user expands a combo box, VO reads as a radio button.
2. When the user expands the combo box, the user is unable to traverse between the items in the dropdown in cyclic and in reverse manner = Cyclic behavior is missing a feature in the form single select.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9307 

---
**Existing Implementation**

image

**Proposed Implementation:**

image

@cerner/terra


packages/terra-icon/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+*Changed
+  * Fixed lint error.
+
```
sdadn, line 7: CHANGELOG entries for linter changes are not necessary.
```suggestion
```

packages/terra-icon/scripts/src/generate-icons/parseSvg.js
sdadn, line 1: These changes are already in main.

packages/terra-icon/scripts/src/generate-icons/renderJsx.js
sdadn, line 1: Same here

sdadn (changes requested):

sdadn (changes requested):

packages/terra-form-select/src/shared/_MenuUtil.js
```diff
@@ -194,7 +194,10 @@ class MenuUtil {
    */
   static findNext(object, value) {
     const options = MenuUtil.flatten(object, true);
-    const index = options.findIndex(({ props }) => MenuUtil.isEqual(props.value, value));
+    let index = options.findIndex(({ props }) => MenuUtil.isEqual(props.value, value));
+    if (options.length - 1 === index) {
+      index = -1;
+    }
     return index === -1 ? null : options[Math.min(index + 1, options.length - 1)].props.value;
   }
```
saket2403, line 202:
```suggestion
    if (options.length - 1 === index) {
      index = 0;
    }
    return index
  }
```

packages/terra-form-select/tests/jest/MenuUtil.test.jsx
```diff
@@ -302,7 +302,7 @@ describe('MenuUtil', () => {
         <OptGroup label="2">{options.slice(2, 5)}</OptGroup>,
       ];
 
-      expect(MenuUtil.findNext(group, '5')).toEqual('5');
+      expect(MenuUtil.findNext(group, '5')).toEqual(null);
```
saket2403, line 305: findNext for last element should return 1st element not null.

packages/terra-icon/CHANGELOG.md
ST063655: Fixed.

packages/terra-icon/scripts/src/generate-icons/renderJsx.js
ST063655: Removed the changes.

packages/terra-form-select/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 * Changed
   * Minor dependency version bump.
+  * Fixed for broken cyclic navigation in `terra-form-single-select`.
+  * Fixed VoiceOver reading as radio button in `terra-form-single-select`.
```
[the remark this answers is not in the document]
ST063655: Added in the correct section.

packages/terra-form-select/src/shared/_MenuUtil.js
ST063655: Fixed.

packages/terra-form-select/tests/jest/MenuUtil.test.jsx
ST063655: Changed as suggested.

packages/terra-icon/scripts/src/generate-icons/parseSvg.js
ST063655: Removed the Changes.

packages/terra-form-select/src/shared/_Option.jsx
```diff
@@ -105,21 +112,20 @@ const Option = ({
...
-   * Switching to role="radio" and role="checkbox" mitigates this behavior.
-   */
-  if (SharedUtil.isSafari() && !('ontouchstart' in window)) {
-    role = 'radio';
+  let itemLabel = display;
+  if (index === 1 || isSelected) {
+    itemLabel = `${expandedStateText} ${display} `;
+    if (!isSelected && SharedUtil.isMac() && SharedUtil.isSafari() && variant !== 'tag' && variant !== 'multiple') {
+      itemLabel += intl.formatMessage({ id: 'Terra.form.select.notselected' });
+    }
```
cm9361, line 120: This does not seem to be the proper way to internationalize this string.  You should have a "not selected" message with a placeholder for the display.  You would pass the display as a value to the formatMessage method.  Then the message can be altered by the globalization team as necessary for the language.  We should not just concatenate.

packages/terra-form-select/src/shared/_MenuUtil.js
```diff
@@ -195,6 +195,9 @@ class MenuUtil {
   static findNext(object, value) {
     const options = MenuUtil.flatten(object, true);
     const index = options.findIndex(({ props }) => MenuUtil.isEqual(props.value, value));
+    if (options.length - 1 === index) {
```
cm9361, line 198: Should this be
`if (index + 1 >== options.length - 1)`

It seems like you are trying to wrap the selection.

packages/terra-form-select/src/shared/_MenuUtil.js
```diff
@@ -195,6 +195,9 @@ class MenuUtil {
...
+      return options[0].props.value;
+    }
     return index === -1 ? null : options[Math.min(index + 1, options.length - 1)].props.value;
```
cm9361, line 201: Instead of having two return statements, we should be able to simply change the value of the index variable and then return the options value at that index.  You would no longer need the Math.min function because you are already checking for the boundaries.

packages/terra-form-select/src/shared/_MenuUtil.js
```diff
@@ -218,7 +221,10 @@ class MenuUtil {
    */
   static findPrevious(object, value) {
     const options = MenuUtil.flatten(object, true);
-    const index = options.findIndex(({ props }) => MenuUtil.isEqual(props.value, value));
+    let index = options.findIndex(({ props }) => MenuUtil.isEqual(props.value, value));
+    if (index === 0) {
+      index = options.length;
+    }
     return index === -1 ? null : options[Math.max(index - 1, 0)].props.value;
```
cm9361, line 228: If we are checking the index in the preceding if block, we should no longer require the Math.max function, correct?

PK106552: Removed Math.max function

packages/terra-form-select/src/shared/_MenuUtil.js
```diff
@@ -218,7 +221,10 @@ class MenuUtil {
    */
   static findPrevious(object, value) {
     const options = MenuUtil.flatten(object, true);
-    const index = options.findIndex(({ props }) => MenuUtil.isEqual(props.value, value));
+    let index = options.findIndex(({ props }) => MenuUtil.isEqual(props.value, value));
+    if (index === 0) {
```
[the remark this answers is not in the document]
PK106552: Yes, we are checking index - 1

packages/terra-form-select/src/shared/_MenuUtil.js
PK106552: If we add this check focus is skipping last item in dropdown while navigate using arrow key

packages/terra-form-select/src/shared/_Option.jsx
PK106552: Removed concatenate and passed the display as a value to the formatMessage method.

packages/terra-form-select/src/shared/_MenuUtil.js
PK106552: Removed Math.max function.
If we change the index value to -1 then null value will be return from another return statement

saket2403 (approved):

sugan2416 (approved):

packages/terra-form-select/translations/fr.json
```diff
@@ -15,6 +15,7 @@
   "Terra.form.select.dimmed": "Grisé.",
   "Terra.form.select.expanded": "Zone combinée développée",
   "Terra.form.select.collapsed": "Zone combinée réduite",
+  "Terra.form.select.notselected": "{text} Non sélectionné",
```
supreethmr, line 18: what about other locales ..? I see translations missing for below locales 
en-CA
en-US
fi-FI
fr-FR
es-ES
es-US
nl-BE
pt-BR
sv-SE

samuelsujith:
Functional_Review done.

packages/terra-form-select/src/shared/_Option.jsx
```diff
@@ -105,21 +112,20 @@ const Option = ({
...
-   * Switching to role="radio" and role="checkbox" mitigates this behavior.
-   */
-  if (SharedUtil.isSafari() && !('ontouchstart' in window)) {
-    role = 'radio';
+  let itemLabel = display;
+  if (index === 1 || isSelected) {
+    itemLabel = `${expandedStateText} ${display} `;
+    if (!isSelected && SharedUtil.isMac() && SharedUtil.isSafari() && variant !== 'tag' && variant !== 'multiple') {
+      itemLabel = intl.formatMessage({ id: 'Terra.form.select.notselected' }, { text: itemLabel });
+    }
```
supreethmr, line 120:
```suggestion
    if (!isSelected && SharedUtil.isMac() && SharedUtil.isSafari() && variant !== 'tag' && variant !== 'multiple') {
      itemLabel = intl.formatMessage({ id: 'Terra.form.select.notselected' }, { text: itemLabel });
    }
```
move this if block outside of the above if block. so that translation will be available for all the items instead of just being available for 1st item

KV106606Viswanath: Updated

packages/terra-form-select/translations/fr.json
KV106606Viswanath: Updated

packages/terra-form-select/CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 ## 6.46.1 - (September 19, 2023)
 
 * Fixed
+  * Fixed for broken cyclic navigation in `terra-form-single-select`.
+  * Fixed VoiceOver reading as radio button in `terra-form-single-select`.
```
sdadn, line 14: We should write these 
```suggestion
  * Fixed bug where cyclic navigation in `terra-form-single-select` was not working.
  * Fixed bug where VoiceOver was reading `terra-form-single-select` as a radio button .
```

KV106606Viswanath: Updated.

saket2403 (approved):

sdadn (approved):

rahulkumar8cs:
+1 for the cyclic behaviour of form single select and expand/collapse state announcement.
We need to fix few issues which is related with screenreader.
1. Safari+VO =>When the user expands a combo box, VO reads as a radio button.
2. Reading label twice when we are selecting item from the single select.
3. JAWS is announcing leaving right paren and left paren for the option.

PR #3856, opened 2023-07-26
[terra-form-field] Updated Regular expression of Email

KV106606Viswanath:
### Summary
Previously email field was accepting n number of characters after dot( . ). Now it has been updated to accept minimum 2 and maximum 63 characters after the dot ( . ) in the email field.

**Why it was changed:**
Because the email field was accepting n number of characters which is not a valid email address.

### Additional Details

**This PR resolves:**

UXPLATFORM-9384

---

@cerner/terra

### DIFF
--- a/packages/terra-core-docs/CHANGELOG.md
+++ b/packages/terra-core-docs/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 * Updated
   * Updated an example for `terra-dropdown-button`.
+  * Updated email field validation for `terra-form-field`.
 
 * Added
   * Added documentation updates for `terra-form-input`.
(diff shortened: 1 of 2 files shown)


sycombs:
I believe there are domains with more than 3 characters after the dot, i.e. `.info`, etc. Is this a restriction we really want to make?

supreethmr:
> I believe there are domains with more than 3 characters after the dot, i.e. `.info`, etc. Is this a restriction we really want to make?
> 
> example: https://www.identity.digital

+1

KV106606Viswanath:
> I believe there are domains with more than 3 characters after the dot, i.e. `.info`, etc. Is this a restriction we really want to make?
> 
> example: https://www.identity.digital



> I believe there are domains with more than 3 characters after the dot, i.e. `.info`, etc. Is this a restriction we really want to make?
> 
> example: https://www.identity.digital

We have updated the email address field that will accept upto 63 characters after the dot ( . )  For this, we have referred the below link. 
https://www.mindbaz.com/en/email-deliverability/what-is-the-maximum-size-of-an-mail-address/
MicrosoftTeams-image (3)

supreethmr (approved):

sdadn (approved):

PR #3857, opened 2023-07-27
[Terra-Button-Group] Fixes accessibility issues of button-group 

gt106551:
**Summary**
  Terra Button Group component has some accessibility issues related to keyboard navigation and focus management. When using JAWS screen reader, pressing the up and down arrow keys to navigate the group does not change the visual focus. Additionally, for keyboard-only navigation, the arrow keys do not move focus to the buttons within the group, requiring the user to rely on the TAB key.

**What was changed**:

 I have implemented Cyclic Behavior for terra button group examples and fixed visual focus When using JAWS screen reader, pressing the up and down arrow keys to navigate to button group.

**Why it was changed**:

 The arrow keys do not move focus to the buttons within the group, requiring the user to rely on the TAB key with JAWS Screen reader.

### Testing

This change was tested using:
- [x] WDIO
- [ ] Jest
- [ ]  Visual testing (please attach a screenshot or recording)
- [ ] No tests are needed

### **Reviews**
- [ ]  UX review
- [x]  Accessibility review
- [ ]  Functional review

### **Additional Details**
This PR resolves:

UXPLATFORM-9386

@cerner/terra


supreethmr:
@gt106551 I think summary of this PR is not clear it looks like copy pasted message. Can you re-format summary and just add what are the issues being addressed by this PR and also add Jira Key (Jira Number)

packages/terra-button-group/CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
 ## Unreleased
+*Fixed
+  *Fixed Terra- Button Group Accessiblity Issues.
```
supreethmr, line 5: Needs a correction ! Alignment is not matching with existing logs

```suggestion

* Fixed
  * Fixed Terra- Button Group Accessibility Issues.
```

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -60,17 +62,26 @@ class ButtonGroup extends React.Component {
   handleKeyDown(event, idx) {
     const allBtns = this.btnGrpRef.querySelectorAll('[data-terra-button-group-button]');
     let key = idx;
-
-    if (event.keyCode === KEY_RIGHT && allBtns[key + 1]) {
-      key += 1;
-      while (allBtns[key] && allBtns[key].hasAttribute('disabled')) {
+    if ((event.keyCode === KEY_RIGHT || event.keyCode === KEY_DOWN)) {
```
supreethmr, line 65: we need to include `allBtns[key + 1]` check as well in order to prevent an error of accessing length when there is no data on array `allBtns`

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -123,17 +134,17 @@ class ButtonGroup extends React.Component {
...
     const allButtons = children ? [] : undefined;
     // eslint-disable-next-line no-nested-ternary
-    const btnRole = onChange ? isMultiSelect ? 'checkbox' : 'radio' : 'button';
+    const btnRole = onChange ? isMultiSelect ? 'checkbox' : 'radio' : '';
 
     React.Children.forEach(children, (child, index) => {
       const isSelected = selectedKeys.indexOf(child.key) > -1;
       const cloneChild = React.cloneElement(child, {
-        role: btnRole,
+        [btnRole !== '' && 'role']: btnRole,
```
supreethmr, line 142: why was this change made ??

supreethmr:
We need a either WDIO / Jest test to validate cyclic behavior of button-group

packages/terra-button-group/src/ButtonGroup.jsx
gt106551: Because of role button it is reading as null.

packages/terra-button-group/src/ButtonGroup.jsx
gt106551: We won't get any errors, if it is last element, it'll come to first element focus as we implement Cyclic Behavior.

artpark:
Should the page be locked while focus is maintained in the button group? Pressing up and down moves the focus as well as the page currently.

packages/terra-button-group/src/ButtonGroup.jsx
cm9361: So are you just trying to only create the role btnRole variable has a value?  I also am not clear on why you are trying to accomplish here.  You also mentioned JAWS on Edge.  Do we know if there was any impact to VoiceOver or JAWS on other browsers by this change?  Could you not just use something like
 role: btnRole || undefined

packages/terra-button-group/src/ButtonGroupButton.jsx
```diff
@@ -88,7 +88,7 @@ class ButtonGroupButton extends React.Component {
   handleKeyUp(event) {
     // Apply focus styles for keyboard navigation.
     // The onFocus event doesn't get triggered in some browsers, hence, the focus state needs to be managed here.
-    if (event.nativeEvent.keyCode === KeyCode.KEY_TAB || event.nativeEvent.keyCode === KeyCode.KEY_LEFT || event.nativeEvent.keyCode === KeyCode.KEY_RIGHT) {
+    if (event.nativeEvent.keyCode === KeyCode.KEY_TAB || event.nativeEvent.keyCode === KeyCode.KEY_LEFT || event.nativeEvent.keyCode === KeyCode.KEY_RIGHT || event.nativeEvent.keyCode === KeyCode.KEY_UP || event.nativeEvent.keyCode === KeyCode.KEY_DOWN) {
```
cm9361, line 91: We should not have to access the underlying native event to obtain the key code.

packages/terra-button-group/CHANGELOG.md
gt106551: I have made change as per Suggestion.

packages/terra-button-group/src/ButtonGroup.jsx
gt106551: I have made changes as per Suggestion.

packages/terra-button-group/src/ButtonGroupButton.jsx
gt106551: I have made change as per Suggestion.

gt106551:
> Should the page be locked while focus is maintained in the button group? Pressing up and down moves the focus as well as the page currently -- which does not seem intended.

I have made change as per Suggestion.

packages/terra-button-group/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+
+*Fixed
+  *Fixed Terra- Button Group Accessiblity Issues.
```
sdadn, line 7: This is too vague. What ate the accessibility issues you fixed?

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -123,17 +137,17 @@ class ButtonGroup extends React.Component {
...
     const allButtons = children ? [] : undefined;
     // eslint-disable-next-line no-nested-ternary
-    const btnRole = onChange ? isMultiSelect ? 'checkbox' : 'radio' : 'button';
+    const btnRole = onChange ? isMultiSelect ? 'checkbox' : 'radio' : '';
 
     React.Children.forEach(children, (child, index) => {
       const isSelected = selectedKeys.indexOf(child.key) > -1;
       const cloneChild = React.cloneElement(child, {
-        role: btnRole,
+        role: (btnRole || undefined),
```
sdadn, line 145: the parentheses are unnecessary

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -123,17 +137,17 @@ class ButtonGroup extends React.Component {
 
     const allButtons = children ? [] : undefined;
     // eslint-disable-next-line no-nested-ternary
-    const btnRole = onChange ? isMultiSelect ? 'checkbox' : 'radio' : 'button';
+    const btnRole = onChange ? isMultiSelect ? 'checkbox' : 'radio' : '';
```
sdadn, line 140: Let's add parentheses here for clarity:
```suggestion
    const btnRole = onChange ? (isMultiSelect ? 'checkbox' : 'radio') : '';
```

packages/terra-button-group/src/ButtonGroup.jsx
sdadn: This behavior needs to be showcased through a unit or wdio test.

sdadn (changes requested):

packages/terra-button-group/src/ButtonGroup.jsx
PK106552: Removed parentheses

packages/terra-button-group/src/ButtonGroup.jsx
PK106552: Added parentheses

packages/terra-button-group/CHANGELOG.md
PK106552: Updated doc file

packages/terra-button-group/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+
+*Fixed
+  *Fixed visual focus When using SR and added Cyclic Behavior for terra button group.
+
```
sdadn, line 8: Corrections:
- Since you're adding a new feature, that should be a separate entry under `Added`.
- Screenreader should be fully spelled out
- Markdown syntax requires a space after `*` for bullet points.
- "Cyclic Behavior" is not a proper noun and does not need to be capitalized.
- This is a dedicated CHANGELOG file for terra button group so we don't need to explicitly state that.

```suggestion

* Fixed
  * Fixed visual focus when using screenreaders.
  
 * Added
   * Added cyclic behavior when navigating through buttons with a keyboard.

```

PK106552: Updated doc file - [1b80e6a](https://github.com/cerner/terra-core/pull/3857/commits/1b80e6a33f8584d5fb55dfb752f2bf6c005407f5)

packages/terra-button-group/tests/wdio/button-group-spec.js
```diff
@@ -120,6 +120,26 @@ Terra.describeViewports('Button Group', ['huge'], () => {
       browser.keys('ArrowLeft');
       Terra.validates.element('Button-1 focused');
     });
+
+    it('Navigates from first button to last button using left arrow', () => {
+      browser.keys('ArrowLeft');
+      Terra.validates.element('Last button focused');
```
sdadn, line 126: can we also programmatically check for focus? e.g. 
https://github.com/cerner/terra-core/blob/b635ece6c935b6004548684c03a881c9336c317a/packages/terra-search-field/tests/wdio/search-field-spec.js#L293

PK106552: Added programmatically check for focus

saket2403:
If the first button is disabled the cyclic navigation is not working.

packages/terra-button-group/src/ButtonGroupButton.jsx
```diff
@@ -88,7 +88,7 @@ class ButtonGroupButton extends React.Component {
   handleKeyUp(event) {
     // Apply focus styles for keyboard navigation.
     // The onFocus event doesn't get triggered in some browsers, hence, the focus state needs to be managed here.
-    if (event.nativeEvent.keyCode === KeyCode.KEY_TAB || event.nativeEvent.keyCode === KeyCode.KEY_LEFT || event.nativeEvent.keyCode === KeyCode.KEY_RIGHT) {
+    if (event.keyCode === KeyCode.KEY_TAB || event.keyCode === KeyCode.KEY_LEFT || event.keyCode === KeyCode.KEY_RIGHT || event.keyCode === KeyCode.KEY_UP || event.keyCode === KeyCode.KEY_DOWN) {
```
sdadn, line 91: This seems to be a lot of conditions for a single if statement. For example, the keydown event will only be true after 4 failures. We can optimize this by using a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set).

```js
const validKeys = new Set([key1, key2, ...]);

if(validKeys.has(event.keyCode))
   ....

```

PK106552: Updated condition using Set - [2696d6d](https://github.com/cerner/terra-core/pull/3857/commits/2696d6d45a49ed12b5690bdf2c1326b0e1a357a4)

packages/terra-button-group/src/ButtonGroupButton.jsx
```diff
@@ -88,7 +88,8 @@ class ButtonGroupButton extends React.Component {
   handleKeyUp(event) {
     // Apply focus styles for keyboard navigation.
     // The onFocus event doesn't get triggered in some browsers, hence, the focus state needs to be managed here.
-    if (event.nativeEvent.keyCode === KeyCode.KEY_TAB || event.nativeEvent.keyCode === KeyCode.KEY_LEFT || event.nativeEvent.keyCode === KeyCode.KEY_RIGHT) {
+    const validKeys = new Set([KeyCode.KEY_TAB, KeyCode.KEY_LEFT, KeyCode.KEY_RIGHT, KeyCode.KEY_UP, KeyCode.KEY_DOWN]);
```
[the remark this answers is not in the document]
PK106552: Updated - [16f411e](https://github.com/cerner/terra-core/pull/3857/commits/16f411e795bea61d182dc84fcdca923483c869a1)

packages/terra-button-group/src/ButtonGroup.jsx
gt106551: wdio test cases updated.

gt106551:
> Should the page be locked while focus is maintained in the button group? Pressing up and down moves the focus as well as the page currently -- which does not seem intended.

Updated

sdadn (approved):

supreethmr (approved):

PR #3860, opened 2023-08-01
[terra-form-native-select] Added aria-label  to announce selected option

AH106586Harika:
### Summary

**What was changed:**
1.Added Margin for buttons to provide space between buttons.
2.Updated aria-label to announce selected state when traversed through options for all the terra form native select examples.

**Why it was changed:**
1.Added Margin for buttons as the border is truncated for clear and set Mercedes buttons.
2.Updated aria-label as it is not announcing the selection state when we navigate through options and clicked Enter.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details
Added Margin for buttons.
Updated aria-label to announce selected option.

**This PR resolves:**

UXPLATFORM-9316

---

@cerner/terra


packages/terra-form-select/CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 # Changelog
 
 ## Unreleased
-
+* Added
+  * Added aria-required property to avoid displaying browser default tooltip.
```
sdadn, line 5: Let's give this some breathing room :)
```suggestion

* Added
  * Added aria-required property to avoid displaying browser default tooltip.
  
```

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -216,14 +216,14 @@ const NativeSelect = ({
...
+  const selectedLabel = intl.formatMessage({ id: 'Terra.form.select.selected' });
   const selectAttrs = {
     'aria-describedby': ariaDescribedBy,
-    'aria-label': ariaLabel,
+    'aria-label': (uncontrolledValue || value !== undefined) ? `${ariaLabel}, ${selectedLabel}` : ariaLabel,
     id,
     disabled,
     'aria-invalid': isInvalid || undefined,
-    required,
+    'aria-required': required,
```
artpark, line 226: Would removing the actual `required` attribute allow the select tag to submit? Shouldn't the original `required` attribute still be here alongside `aria-required`?

supreethmr: I think we should revert this change and keep the `required` attribute as it servers both purpose of preventing form submit and helps screen reader to identify field as required

packages/terra-form-select/CHANGELOG.md
supreethmr: @AH106586Harika this change log is missing details about aria-label changes please include.

AH106586Harika: Added aria-label details on  CHANGELOG

packages/terra-form-select/src/native-select/NativeSelect.jsx
AH106586Harika: reverted as suggested.Thanks

packages/terra-core-docs/src/terra-dev-site/doc/form-select/example/native-select/Controlled.jsx
```diff
@@ -22,8 +22,8 @@ const ControlledExample = () => {
         value={value}
         className={cx('form-select')}
       />
-      <button type="button" onClick={() => setValue()}>Clear</button>
-      <button type="button" onClick={() => setValue('mercedes')}>Set Mercedes</button>
+      <button type="button" className={cx('controlled-button')} onClick={() => setValue(undefined)}>Clear</button>
+      <button type="button" className={cx('controlled-button')} onClick={() => setValue('mercedes')}>Set Mercedes</button>
```
cm9361, line 26: I would suggest that we do not use anonymous functions in our examples.  This would be a bad practice that we would not want propagated as it creates a new function for each click.  We would want to use a defined function for the click event.

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -216,10 +216,10 @@ const NativeSelect = ({
     currentValue = getFirstValue(options, isFilterStyle);
     currentDisplay = getDisplay(currentValue, options, isFilterStyle, intl);
   }
-
+  const selectedLabel = intl.formatMessage({ id: 'Terra.form.select.selected' });
   const selectAttrs = {
     'aria-describedby': ariaDescribedBy,
-    'aria-label': ariaLabel,
+    'aria-label': (uncontrolledValue || value !== undefined) ? `${ariaLabel}, ${selectedLabel}` : ariaLabel,
```
cm9361, line 222: This does not seem to be accounting for internationalization properly.  Are we certain that in every language it makes sense to just add a comma between these values.  We are taking away the flexibility for the translators here.  What message are you trying to create?

packages/terra-core-docs/src/terra-dev-site/doc/form-select/example/native-select/Controlled.jsx
PK106552: Added named function instead of anonymous function - [e5e0ac0](https://github.com/cerner/terra-core/pull/3860/commits/e5e0ac0037181db80a83493913e9bd04e58e66df)

packages/terra-form-select/src/native-select/NativeSelect.jsx
PK106552: We are including selected instruction in aria-label attribute and already we have translated selected text so we are combining selected native select value as well as selected instruction, now i have removed comma between native select value and selected text. - [e5e0ac0](https://github.com/cerner/terra-core/pull/3860/commits/e5e0ac0037181db80a83493913e9bd04e58e66df)

packages/terra-form-select/src/native-select/NativeSelect.jsx
```diff
@@ -216,10 +216,10 @@ const NativeSelect = ({
     currentValue = getFirstValue(options, isFilterStyle);
     currentDisplay = getDisplay(currentValue, options, isFilterStyle, intl);
   }
-
+  const selectedText = intl.formatMessage({ id: 'Terra.form.select.selected' });
   const selectAttrs = {
     'aria-describedby': ariaDescribedBy,
-    'aria-label': ariaLabel,
+    'aria-label': (uncontrolledValue || value !== undefined) ? `${ariaLabel} ${selectedText}` : ariaLabel,
```
supreethmr, line 222: (,) separator between arialabel and selected text will enhance the screen reader response. 
(value) is same as (value !== undefined) ??
```suggestion
    'aria-label': (uncontrolledValue || value ) ? `${ariaLabel}, ${selectedText}` : ariaLabel,
```

packages/terra-form-select/src/native-select/NativeSelect.module.scss
```diff
@@ -34,7 +34,9 @@
       background-color: var(--terra-form-select-native-focus-background-color, #fff);
       background-size: var(--terra-form-select-native-focus-background-size, 0 0);
       border-color: var(--terra-form-select-native-focus-border-color, #4cb2e9);
+      outline: var(--terra-form-select-native-focus-keyboard-outline, 2px dashed #000);
```
supreethmr, line 37:
```suggestion
      outline: var(--terra-form-select-native-focus-mouse-outline, 2px dashed #000);
```
you need to create these variables in other theme files as well

packages/terra-form-select/src/native-select/NativeSelect.jsx
supreethmr: we have included separator on other terra components as well and has verified with Translation team. we can use the separators to enhance the screen reader response ( whenever screen reader encounters `,` it gives natural pause before reading following phrase ).

packages/terra-form-select/src/native-select/NativeSelect.jsx
PK106552: Updated check condition and added separator - [865c9cc](https://github.com/cerner/terra-core/pull/3860/commits/865c9cc8226bedc26815dad21a509bbf06e15b34)

packages/terra-form-select/src/native-select/NativeSelect.module.scss
KV106606Viswanath: Updated.

supreethmr (approved):

packages/terra-form-select/src/native-select/NativeSelect.module.scss
```diff
@@ -34,7 +34,10 @@
       background-color: var(--terra-form-select-native-focus-background-color, #fff);
```
cm9361, line 34: Do we still need the &[data-focus-interaction='keyboard'] styles above?  The only difference now appears to be the box-shadow.  Should that be part of this selector as well, or is it no longer applicable.  Also, based on the selector being used, could this actually be replaced with the ":focus" pseudo class selector?

supreethmr: +1

KV106606Viswanath: 1) Removed duplicate [data-focus-interaction='keyboard'] selector as suggested. 
2) Box-shadow property is required in default theme since we are applying box-shadow in Fusion theme. 
3) We can't replace [data-focus-interaction='keyboard'], [data-focus-interaction='mouse'] selectors with ":focus" pseudo class selector because these styles has been added for div element and not for select element, and we are getting the above keyboard and mouse attributes on the same div element. Based on the attributes of div element we are applying the focus styles.
Screenshot 2023-08-10 at 5 58 29 PM

sdadn (approved):

packages/terra-form-select/src/native-select/NativeSelect.module.scss
supreethmr: we can still remove mouse and keyboard values from data attrs. updated with commit [9f570493](https://github.com/cerner/terra-core/pull/3860/commits/65689d4dd67ae4293cff83ad933a83289f570493)

saket2403:
The black focus border should appear only for keyboard navigation and not while using mouse.

supreethmr (approved):

saket2403 (approved):

rahulkumar8cs:
+1 for screenreader and keyboard accessibility.

PR #3866, opened 2023-08-04
[terra-button]: Convert Terra Button from Class Component to Functional Component

kenk2:
### Summary

**What was changed:**
Implementation of Terra Button has changed from a Class Component to Functional Component.

**Why it was changed:**
To keep up with modern React standards.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [x] Functional review

### Additional Details
---

@cerner/terra


packages/terra-button/src/Button.jsx
```diff
@@ -115,221 +115,194 @@ const defaultProps = {
...
+  onMouseDown,
+  onBlur,
+  onFocus,
+  onKeyDown,
+  onKeyUp,
+  refCallback,
+  title,
+  ...customProps
+}) => {
+  const theme = React.useContext(ThemeContext);
```
cm9361, line 139: You can import useContext from react just as you did useState and useEffect.

packages/terra-button/src/Button.jsx
```diff
@@ -115,221 +115,194 @@ const defaultProps = {
...
+  href,
+  onClick,
+  onMouseDown,
+  onBlur,
+  onFocus,
+  onKeyDown,
+  onKeyUp,
+  refCallback,
+  title,
+  ...customProps
```
cm9361, line 137: I don't think we need to allow custom props to be passed.  This was not an option before, right?

packages/terra-button/src/Button.jsx
```diff
@@ -115,221 +115,194 @@ const defaultProps = {
...
+}) => {
+  const theme = React.useContext(ThemeContext);
+  const [focused, setFocused] = useState(false);
+  const [active, setActive] = useState(false);
```
cm9361, line 141: We should use isFocused and isActive for our getter methods since these are Boolean state variables?

packages/terra-button/src/Button.jsx
```diff
@@ -115,221 +115,194 @@ const defaultProps = {
...
+    if (onClick) {
+      onClick(event);
     }
-  }
+  };
 
-  handleKeyDown(event) {
+  const handleKeyDown = (event) => {
     // Add active state to FF browsers
     if (event.nativeEvent.keyCode === KeyCode.KEY_SPACE) {
```
cm9361, line 173: I know that this is not your code, but we should not need the nativeEvent to access the keyCode.

packages/terra-button/src/Button.jsx
```diff
@@ -115,221 +115,194 @@ const defaultProps = {
...
+  const [active, setActive] = useState(false);
+
+  useEffect(() => {
+    setActive(false);
+    if (isDisabled) {
+      setFocused(false);
     }
-    return null;
-  }
+  }, [isDisabled]);
```
cm9361, line 148: Are we sure that this useEffect block is correct?  I could be wrong, but it does not seem correct that the control will be inactive and setFocused can be false regardless of the value of isDisabled.  If a consumer rerenders this component with a different value for isDisabled, it does not seem like the Button component would handle it properly with this logic.

packages/terra-button/src/Button.jsx
```diff
@@ -115,221 +115,194 @@ const defaultProps = {
...
-  }
+  };
 
-  handleFocus(event) {
-    if (this.shouldShowFocus) {
-      this.setState({ focused: true });
-    }
+  const handleFocus = (event) => {
+    setFocused(true);
+    event.currentTarget.focus();
```
cm9361, line 210: Why do we need to call the focus() method in the onFocus event handler?  Wouldn't the component already have focus here?

packages/terra-button/src/Button.jsx
```diff
@@ -115,221 +115,194 @@ const defaultProps = {
...
-    super(props);
-    this.state = { active: false, focused: false };
-    this.handleKeyDown = this.handleKeyDown.bind(this);
-    this.handleKeyUp = this.handleKeyUp.bind(this);
-    this.handleOnBlur = this.handleOnBlur.bind(this);
-    this.handleFocus = this.handleFocus.bind(this);
-    this.handleMouseDown = this.handleMouseDown.bind(this);
-    this.handleClick = this.handleClick.bind(this);
-
-    this.shouldShowFocus = true;
```
cm9361, line 129: How are we accounting for this shouldShowFocus variable in the new logic?

packages/terra-button/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Changed
+  * Changed implementation from Class Component to Functional Component.
```
sdadn, line 6: We can remove this line as this information is not relevant for consumers.

sdadn (commented):
I think you have changes from other PRs. Can you rebase to make sure your PR is up to date?

packages/terra-button/src/Button.jsx
kenk2: Discussed in sync: Un-necessary variable now removed.

packages/terra-button/CHANGELOG.md
kenk2: Done

kenk2:
> Fails
> 🚫	Please include a CHANGELOG entry for each changed package this PR. Looks like a CHANGELOG entry is missing for:
> * terra-button
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [2b891b5](https://github.com/cerner/terra-core/commit/2b891b5eac678e1dfc4dacef5f536fbe0bbb94e3)

Since this change is not consumer-related, no CHANGELOG update is needed.

packages/terra-button/src/Button.jsx
```diff
@@ -115,225 +115,165 @@ const defaultProps = {
...
+  };
 
-    if (this.props.onKeyUp) {
-      this.props.onKeyUp(event);
+  const handleFocus = (event) => {
+    if (onFocus) {
+      onFocus(event);
     }
-  }
+  };
```
cm9361, line 185: You don't need a separate method for this.  Since you are only checking for the existence of the onFocus callback, you can pass that function directly to the onFocus React event.  If it is undefined, the event just won't be created.  Same comment for the other handlers that do not have additional logic.

kenk2:
> Fails
> 🚫	Please include a CHANGELOG entry for each changed package this PR. Looks like a CHANGELOG entry is missing for:
> * terra-button
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [34bd35f](https://github.com/cerner/terra-core/commit/34bd35f257e9e0f9c33acc5e13962579bb214855)

Not consumer-related, no changelog is needed.

kenk2:
@cm9361 We'll have to retain `isFocused` and `isActive` like the original implementation of this component due to us overriding the default accessibility behaviors of clicking and Firefox issues with button active pseudo-class in our project.

packages/terra-button/src/Button.jsx
```diff
@@ -115,225 +115,197 @@ const defaultProps = {
...
+  onKeyDown,
+  onKeyUp,
+  refCallback,
+  title,
+  ...customProps
+}) => {
+  const theme = useContext(ThemeContext);
+
+  // We need to manually control active state for FF compatibility
+  const [isActive, setIsActive] = useState(false);
```
sdadn, line 142: Does this logic have any affect in other browsers? Would it make sense to only execute this logic if the browser is FireFox?

kenk2: Just checked and I think it only affects Firefox like the original comment.

packages/terra-button/src/Button.jsx
```diff
@@ -115,225 +115,197 @@ const defaultProps = {
...
+      aria-label={ariaLabel}
+      onKeyDown={handleKeyDown}
+      onKeyUp={handleKeyUp}
+      onBlur={handleOnBlur}
+      title={buttonTitle}
+      onClick={handleClick}
+      onMouseDown={handleMouseDown}
+      onFocus={handleFocus}
+      href={href}
+      ref={refCallback}
```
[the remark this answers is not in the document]
kenk2: Done!

packages/terra-button/src/Button.jsx
```diff
@@ -115,225 +115,197 @@ const defaultProps = {
...
+  variant,
+  href,
+  onClick,
+  onMouseDown,
+  onBlur,
+  onFocus,
+  onKeyDown,
+  onKeyUp,
+  refCallback,
+  title,
```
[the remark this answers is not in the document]
kenk2: Done!

sdadn (approved):

cm9361 (approved):

PR #3872, opened 2023-08-09
UXPLATFORM-9423 Add Paginator buttons link role

chenyixuan625:
### Summary

**What was changed:**
Add role="link" to the PaginatorButton.
Wrap up Paginator views to an unordered list by using HTML ul tag.

**Why it was changed:**
Now the accessibility tooling recognizes the paginator navigation controllers as "buttons" since they are built on HTML button element. This is not good from the Accessibility perspective. We would like the accessibility tooling to read out the paginator navigation controllers of the paginator as "links". And we also want these "links" to be grouped by a list.

### Testing

VoiceOver:

https://github.com/cerner/terra-core/assets/44185348/a4209235-f7d1-49d2-86ef-3335680439e1

JAWS:

https://github.com/cerner/terra-core/assets/44185348/c3631151-a357-4fef-8021-3cf65559e324

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9423
---

@cerner/terra


packages/terra-paginator/src/Paginator.module.scss
```diff
@@ -145,6 +145,7 @@
     }
 
     &.pageless {
+      // display: flex;
```
vinaybhargavar, line 148: nit: We can remove this comment if it is not needed.

chenyixuan625: good catch. fixed: https://github.com/cerner/terra-core/pull/3872/commits/2b488819c07215061791907e4d16fce30a0cf270#diff-639527009c6ffee66446a77ca5dec0c77f73dec484857a448936050b3f7a2e19L148

RayGunY:
I think this is looking good overall. I'm wondering about those wdio snapshots that updated - what kind of visual difference is causing those to update? It's hard to tell from looking between the two, they appear the same to me. Is is a few pixels shifted or coloring or something like that?

sdadn (approved):

sdadn:
> I think this is looking good overall. I'm wondering about those wdio snapshots that updated - what kind of visual difference is causing those to update? It's hard to tell from looking between the two, they appear the same to me. Is is a few pixels shifted or coloring or something like that?

@RayGunY There seems to be some minor pixel shifts. If you use the swipe tool when comparing on GitHub, you can see the difference.

RayGunY (approved):

chenyixuan625:
> I think this is looking good overall. I'm wondering about those wdio snapshots that updated - what kind of visual difference is causing those to update? It's hard to tell from looking between the two, they appear the same to me. Is is a few pixels shifted or coloring or something like that?
> 
> @chenyixuan625

Was going to say the same thing as Saad mentioned above. Just rebasing took me a bit longer than I expect. Thank you guys for reviewing this!

vinaybhargavar (approved):

scottwilmarth (approved):

eawww (approved):

PR #3884, opened 2023-08-17
[terra-arrange] Removed scroll at 400% zoom  and custom styles in examples

KV106606Viswanath:
### Summary
Custom styles of arrange control examples (Height, width and border) has been removed. We have reverted the styles (max-width and overflow) which were added to fix reflow issue at zoom level 400%. Because, these custom styles are causing issues in other components where arrange has been used. 
To fix the 400% zoom level issue, we have removed Height and Width of all examples. Previously, we are getting fixed height and width for the examples. So that we're not able see the whole content and it's getting cropped in mobile view or 400% zoom level. Since we removed the height and width of the examples we can see the whole content without any cropping.

**What was changed:**
Custom styles of arrange control examples has been removed.

**Why it was changed:**
These styles are causing errors in other controls where arrange has been used.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9475

---

@cerner/terra

### DIFF
--- a/packages/terra-arrange/src/Arrange.module.scss
+++ b/packages/terra-arrange/src/Arrange.module.scss
@@ -7,8 +7,6 @@
   // Vertically aligns arrange content to the top by default
   .arrange {
     display: flex;
-    max-width: 100%;
-    overflow: auto;
   }
 
   // Makes sure fill expands to fill the remaining space of its parent container
--- a/packages/terra-core-docs/src/terra-dev-site/doc/arrange/example/ArrangeAlignAllContainers.jsx
+++ b/packages/terra-core-docs/src/terra-dev-site/doc/arrange/example/ArrangeAlignAllContainers.jsx
@@ -1,18 +1,13 @@
 import React from 'react';
-import classNames from 'classnames/bind';
 import Arrange from 'terra-arrange';
 import {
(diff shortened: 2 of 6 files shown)


saket2403 (approved):
Once this PR is merged. Open another PR to revert the locked version of terra-arrange in framework components and update the relevant jest snapshots as well.

packages/terra-arrange/src/Arrange.module.scss
```diff
@@ -7,8 +7,6 @@
   // Vertically aligns arrange content to the top by default
   .arrange {
     display: flex;
-    max-width: 100%;
```
ShettyAkarsh, line 10: is these styles used in source code ? if yes, have we seen the impact of this removal in all the components of terra framework where this component is referred ?

KV106606Viswanath: Yes, these styles are used in source code. But these styles are added recently to fix the zoom reflow issue in mobile view. I have verified in terra core and terra framework, its working as expected even though we removed these styles.

supreethmr (approved):

PR #3885, opened 2023-08-17
[terra-alert] focus on empty element

adoroshk:
### Summary
Currently the screen readers read the notification banner's content twice when the notification is an Alert with an action element (custom prop alert). That happens as the alert content receives focus and has a role="alert".  

**What was changed:**
The focus has been shifted to an empty div of the same size as the content div for avoiding the content being read twice.

### Testing
This change was tested using:

- [x] WDIO
- [x] Jest
- [x] Visual testing (please attach a screenshot or recording)
- [ ] Other (please describe below)
- [ ] No tests are needed

### Reviews
In addition to engineering reviews, this PR needs:
- [x] UX review
- [x] Accessibility review
- [ ] Functional review

**This PR resolves:**

UXPLATFORM-9144

PR #3889, opened 2023-08-21
[terra-alert] alert focus shift to empty element fix

adoroshk:
**What was changed:**
This PR fixes the terra-alert focus shift change blocking terra-alert content from being interact-able.
The PR for the previous change: https://github.com/cerner/terra-core/pull/3885

### Testing
This change was tested using:

- [X] WDIO
- [X] Jest
- [X] Visual testing (please attach a screenshot or recording)
- [ ] No tests are needed

A screenshot demonstrating the Show More button being accessible (focused):
Screenshot 2023-08-21 at 3 15 55 PM

Tested with VO and JAWS to ensure the change hasn't changed the assistive technologies announcements.

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

** This PR is a follow-up PR for **

UXPLATFORM-9144


packages/terra-alert/tests/wdio/alert-spec.js
```diff
@@ -72,6 +72,12 @@ Terra.describeViewports('Alert', ['tiny', 'large'], () => {
   });
 
   describe('Alert Long Text', () => {
+    it('focuses on show more button', () => {
+      browser.url('/raw/tests/cerner-terra-core-docs/alert/long-text-alert');
+      browser.keys(['Tab']);
+      Terra.validates.element('show more button focused');
```
sdadn, line 78: we should add a programmatic test for focus.

packages/terra-alert/src/Alert.module.scss
```diff
@@ -33,6 +33,16 @@
     margin-top: var(--terra-alert-icon-margin-top);
   }
 
+  .content-container {
+    align-items: flex-start;
+    display: flex;
+    flex: 1 1 auto;
+    flex-direction: row;
+    line-height: 1.2;
```
cm9361, line 41: Should we be specifying a line height or let it be inherited.  Also, shouldn't the line height have units if it is required?

adoroshk: This styling was there before the change, I just changed an HTML layout a bit hence had to move some styling around per new class names, so it looks like a new code here, but it has been there a long time and it's out of the scope of this JIRA to change or keep it.

adoroshk: Here is the source of this styling: https://github.com/cerner/terra-core/blob/5f0141ba64d5ae282d5321629e7fdb0385884b18/packages/terra-alert/src/Alert.module.scss#L41

cm9361: ok.  It just looked new based on the changes.

packages/terra-alert/tests/wdio/alert-spec.js
adoroshk: There is update to the tests: https://github.com/cerner/terra-core/pull/3889/commits/fcfbb3334033048c261b10bc89941bc6578a132d
Is that it, or need some other improvements? If more tests needed, the examples would be very appreciated.

nikhitasharma (approved):

packages/terra-alert/tests/wdio/alert-spec.js
```diff
@@ -72,6 +72,22 @@ Terra.describeViewports('Alert', ['tiny', 'large'], () => {
...
+    });
+
+    it('open show hide upon button click', () => {
+      $('button.button').click();
+      Terra.validates.element('disclosed show hide');
+    });
+
+    it('close show hide', () => {
+      $('button.button').click();
```
trandrew1023, line 88: I think validation that the show hide was closed could be added here as well

adoroshk: Added here: https://github.com/cerner/terra-core/pull/3889/commits/20269f6fbdba9a65a3f4a2ad03501a2a069a200b

packages/terra-alert/tests/wdio/alert-spec.js
sdadn: That looks good!

sdadn (approved):

kenk2 (approved):

PR #3900, opened 2023-08-31
[terra-paginator] Styling update to support anchor attribute gap in F…

chenyixuan625:
…usion

### Summary

**What was changed:**
A new optional boolean prop(false by default) is added to support the anchor style. When the new prop is set(true), the paginator will be rendered in a gray background color with a borderline at the top.

**Why it was changed:**
Currently the paginator renders without any borderline and the background color is white. However, MPage PagiantorFooter(utilizing Progressive Paginator) has an "anchor" style, which has a top borderline and the background color is gray. We will want to add a new optional boolean prop(false by default) to support the anchor style from the Terra perspective.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9527 

---

@cerner/terra


packages/terra-paginator/src/ProgressivePaginator.jsx
```diff
@@ -47,6 +47,14 @@ const propTypes = {
    * Allows user to set custom page label. _(usage note: User must pass translated text)_.
    */
   pageLabel: PropTypes.string,
+  /**
+   * Apply fusion anchor style or not.
```
kolkheang, line 51:
```suggestion
   * Apply Fusion anchor style or not.
```

packages/terra-paginator/tests/wdio/paginator-spec.js
```diff
@@ -19,6 +19,13 @@ Terra.describeViewports('Paginator', ['tiny', 'medium', 'large'], () => {
       Terra.validates.element('progressive paginator');
     });
   });
+
+  describe('ProgressivePaginator with Fusion Anchor style', () => {
+    it('default', () => {
```
kolkheang, line 24: Perhaps update the "default" text with more verbose text?

packages/terra-paginator/src/ProgressivePaginator.jsx
```diff
@@ -47,6 +47,14 @@ const propTypes = {
...
+   */
+  fusionAnchor: PropTypes.bool,
```
kolkheang, line 53: Any consideration to auto apply this prop when theme.className is "orion-fusion-theme"? Do you think this prop is useful to have on its own even if current theme is not "orion-fusion-theme"?

packages/terra-paginator/tests/wdio/__snapshots__/reference/clinical-lowlight-theme/en/chrome_large/paginator-spec/progressive_paginator_fusion_anchor_style.png
kolkheang, line 1: I feel like we should add a left padding for these ones with "Page x of y" text so that there is a bit of spacing.

RayGunY: I was thinking the same thing Kol, I would just match whatever padding value there is on the right side after the 'Last' button.

chenyixuan625: The progressive paginator itself doesn't contain any left padding before. The spacing is coming from the relative position of the paginator within the outer Terra Dialog.
Screenshot 2023-09-07 at 3 49 35 PM

Default snapshots of progressive paginator: https://github.com/cerner/terra-core/blob/main/packages/terra-paginator/tests/wdio/__snapshots__/reference/clinical-lowlight-theme/en/chrome_large/paginator-spec/progressive_paginator.png, which you could notice that there is no left spacing as well.

I get the idea but I just hesitate to do that because this is more like a UX change that applies to both progressive paginator and controlled progressive paginator(since both contain this text on the left of the paginator by default)

packages/terra-paginator/src/ProgressivePaginator.jsx
chenyixuan625: MPage PaginatorFooter has 2 styles: anchor(top borderline and gray background) and canvas(no borderline and no background color). My initial thoughts would be that since we need to support both and we don't know which one is preferred by the consumer, I would leave the consumer to set the prop depending on their usage. 

I feel like this prop would be specifically for MPage's usage since no other consumer teams request this requirement. The reason for adding a prop to control this style is that we need to support both styles within the fusion theme, instead of applying a new theme.

packages/terra-paginator/src/ProgressivePaginator.jsx
chenyixuan625: https://github.com/cerner/terra-core/pull/3900/commits/3fe3671fd8da47f0a3745644ef0cf2210a9bad24#diff-a0973b9ee3044c6c2559965dcdbcc62bf0c707fb0c977ae1f59ca342ac87a7e3R51

packages/terra-paginator/tests/wdio/paginator-spec.js
chenyixuan625: Sure, updated: https://github.com/cerner/terra-core/pull/3900/commits/3fe3671fd8da47f0a3745644ef0cf2210a9bad24#diff-65b614af80147a518e5be5f9679cb2d513196d5503a8cbb0a71f052cbbe10059R24

packages/terra-paginator/tests/wdio/__snapshots__/reference/clinical-lowlight-theme/en/chrome_large/paginator-spec/progressive_paginator_fusion_anchor_style.png
RayGunY: Ah ok I see what you're saying - then yes, I say just keep it as you have it right now.

RayGunY (approved):
Looks good!

chenyixuan625:
@eawww @sdadn @sycombs This PR is ready for review. Thanks!

packages/terra-paginator/src/ProgressivePaginator.jsx
```diff
@@ -47,6 +47,14 @@ const propTypes = {
    * Allows user to set custom page label. _(usage note: User must pass translated text)_.
    */
   pageLabel: PropTypes.string,
+  /**
+   * Apply Fusion anchor style or not.
+   */
+  fusionAnchor: PropTypes.bool,
```
sdadn, line 53: It seems odd to add a prop for a very specific use case. Especially since it's related to theming. I'm confused why this is needed to be an API change and not a theme update or any other alternative solution.

cc: @JessieRandle , @mjpalazzo , @sycombs

sdadn (changes requested):

packages/terra-paginator/src/ProgressivePaginator.jsx
chenyixuan625: For the Orion fusion theme, there are 2 different styles(anchor and canvas) and we need to support both. So applying this new optional prop is to distinguish each other in the same fusion theme.
Screenshot 2023-09-13 at 1 09 41 PM
Any suggestions?

sdadn: If it's styling we need to support then I would say that we should support it for all themes and not add a niche prop that is only useful for a specific theme. 

cc: @eawww , @mjpalazzo , @JessieRandle

chenyixuan625:
Synced with Scott and Terra folks, we believe that the best practice would be that Fusion takes care of its own customized style. Adding a prop for a specific styling usage is very confusing for consumers. Closing this PR out. Will comment on the UXPLATFORM-9527 soon.

PR #3905, opened 2023-09-08
Show hide examples updated

adoroshk:
### Summary
After `ShowHide` component was updated with focusRef prop for accessibility, the examples of Alert component that use that component (long text alert) need to be updated with passing that new prop for consistency. Currently, upon opening the ShowHide it sets the focus to the whole text instead of focusing on the first line of the text from the place it was cut off.

The long text alert example is not consistent with the rest of the Alert examples, as the `ShowHide` component is displayed to the right side from the Alert title vs being displayed right below it in the rest of the examples. Per the investigation it has been discovered, that the reason for that is "display: flex" set to the `<div>` with `show-hide` class. That class is used twice, first time for the content wrapper, second to the button wrapper, but the display:flex property is actually needed only for the button wrapper and has no influence on the ShowHide content.

Screenshot 2023-09-08 at 10 10 28 AM

**What was changed:**

1. The `ShowHide` component button wrapper class name was changed to `show-hide-button` to isolate the `display:flex` property to button container as it's required for the button positioning. That allowed the `ShowHide` content to go under the Alert title.
2. The Long text alert implementation of the The `ShowHide` component was updated to pass a focusRef prop to focus on the first line of hidden text upon opening.
3. The typo in ShowHide documentation in code sections (redundant `this` in `this.focusRef`) was fixed.

Screenshot 2023-09-08 at 1 38 06 PM

### Testing
This change was tested using:

- [x] WDIO
- [x] Jest
- [x] Visual testing (see the screenshot above)
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review


packages/terra-alert/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Updated wdio screenshots due to examples changes.
+
```
kenk2, line 7: I believe wdio/jest tests normally aren't releaseable so we don't need a `Changed` entry for  wdio/test changes? This change also isn't directly associated with the `alert` component itself so it probably doesn't need this changelog. Someone can correct me if I'm wrong though.

kenk2 (commented):
One small comment, otherwise looks good.

packages/terra-alert/CHANGELOG.md
adoroshk: Good catch, removed a changelog for terra-alert: https://github.com/cerner/terra-core/pull/3905/commits/de37ab419d195da266b8a2c3efacbca353448573

trandrew1023 (approved):

BenBoersma (approved):

PR #3907, opened 2023-09-13
[form-single-select] doc site updates of escape key instruction

KV106606Viswanath:
### Summary
Updated doc site of form single select and single select field with escape key observation with JAWS. Which is we have to press escape key twice to close the combobox while using JAWS + VPC ON mode.

**What was changed:**
Updated the doc site of form single select and single select field.

**Why it was changed:**
To inform consumers of the issue with escape key.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9641

---

@cerner/terra

### DIFF
--- a/packages/terra-core-docs/CHANGELOG.md
+++ b/packages/terra-core-docs/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Updated
+  *  Updated doc site for `terra-form-single-select` and `terra-form-single-select-field`.
+  
 * Added
   * Added drag and drop example for `terra-list`.
   * Added instruction note for `terra-form-radio` label hidden example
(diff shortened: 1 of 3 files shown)


sdadn:
Why are we only giving instructions for Edge + Jaws?

KV106606Viswanath:
> Why are we only giving instructions for Edge + Jaws?

This is happening with Edge and Chrome browser as well while using JAWS with VPC mode ON. It is working fine with Voice over and the combobox is closing for one escape key press. I have updated the doc site for the same and now the instruction says "browser" instead of only Edge browser.

supreethmr (approved):

PR #3917, opened 2023-09-20
[terra-table] New Terra table

kenk2:
### Summary

**What was changed:**
Implementing new Terra Table. Terra Outline Table is deleted. 

**Why it was changed:**
Implementing new table for datagrid initiatives. Terra Outline Table is obsolete and not being worked on.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [x] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra


packages/terra-core-docs/package.json
```diff
@@ -84,7 +84,7 @@
     "terra-status": "^4.57.0",
     "terra-status-view": "^4.62.0",
     "terra-switch": "^1.12.0",
-    "terra-table": "^4.36.0",
+    "terra-table": "~/terra-core/packages/terra-table",
```
sdadn, line 87: Not sure if this is temporary or not but FYI, this will cause failures in terra-ui terra-table won't exist at this path.

sdadn (commented):
Looks like I forgot to update the CICD to allow for PR previews on this branch; I'll work on that.

mjpalazzo:
@kenk2 - Is there a deployment you would like me to test? If so, please add a link. Thanks!

packages/terra-table/package.json
```diff
@@ -23,14 +23,17 @@
   "homepage": "https://github.com/cerner/terra-core#readme",
   "peerDependencies": {
     "react": "^16.8.5",
-    "react-dom": "^16.8.5"
+    "react-dom": "^16.8.5",
+    "react-intl": ">=2.8.0, <6.0.0"
```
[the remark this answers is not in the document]
kenk2: Fixed!

kenk2:
> @kenk2 - Is there a deployment you would like me to test? If so, please add a link. Thanks!

Done, should be available at: https://engineering.cerner.com/terra-core/pull/3917/

chrismichalewicz:
@kenk2  For sorting I don't think we'd want the indicator displaying by default unless sorting is displayed. I included a screenshot of what is displayed on initial load. @eawww do we want to support multiple columns sorted?  
image

This is an issue only if we decide to support multiple columns being sorted.  If I have the patient column sorted descending the Location column is ascending which is correct.  when I select the location column to descending sort on location the Patient sort updates to ascending, which is correct. Now if I select the Location column again to chance the sort back to ascending the sort indicator for Patient doesn't update. That same happens when clicking Patient column and the Location column's sort automatically updates (or doesnt).

kenk2:
> @kenk2 For sorting I don't think we'd want the indicator displaying by default unless sorting is displayed. I included a screenshot of what is displayed on initial load. @eawww do we want to support multiple columns sorted? image
> 
> This is an issue only if we decide to support multiple columns being sorted. If I have the patient column sorted descending the Location column is ascending which is correct. when I select the location column to descending sort on location the Patient sort updates to ascending, which is correct. Now if I select the Location column again to chance the sort back to ascending the sort indicator for Patient doesn't update. That same happens when clicking Patient column and the Location column's sort automatically updates (or doesnt).

Correct, as it currently is when we click on one sortable column, in the sorter function (which defines this behavior) it resets all the sort indicators, although if you have a sort function that doesn't reset the other columns it will not happen. As you explained however this would only really be an issue if we needed to support multiple columns.

Otherwise I could probably re-define the sort functions in the example so that it doesn't give the wrong impression to our users.

mjpalazzo:
@kenk2 - Great job on this initial version of Terra Table! I will not do an Accessibility Review of this version because my findings will be addressed in future stories. I did do some keyboard navigation testing and screen reader testing with JAWS. Here are some observations:
Keyboard navigation works well. When I tabbed to the Table it got visual focus. With Chrome, Down and Up Arrows caused the I-beam pointer to move from cell to cell, the scroll bar moved and I could scroll horizontally to disclose the right most content. With Edge I had to use the Right and Left Arrow keys to navigate. The scroll bar moved, and table scrolled but the I-beam pointer was not visible.
Navigation with JAWS worked as expected. I could Tab to the Table and use the Down Arrow and Up Arrow keys to navigate from cell to cell. JAWS announced that the control is a table, but no row and column counts (I assume that is a future story). JAWS announced the content of the column header cells and also the word, "button." We definitely do not want JAWS to announce "button" where none exists. JAWS announced "masked data" for the masked data cells and "blank" for the cells with no data. JAWS announced "End of Table" when I pressed Down Arrow on the last cell. 
I also test the sortable table example. I could Tab to the Table and then the first sortable column header. I pressed Enter to sort. I tested with JAWS and it announced the column header content and sort order. JAWS also announced the cell position "column 1 row 1". When I pressed Enter it sorted, but JAWS did not announce the new sort order. I could navigate to non-interactive cells with the Down and Up Arrow keys.
Overall this is a great start.

cm9361:
You should probably create separate example pages for your Table features such as Pinned Columns and Sortable Columns.  There are examples of this in some of our other projects.

packages/terra-core-docs/src/terra-dev-site/doc/table/example/DefaultTable.jsx
```diff
@@ -0,0 +1,65 @@
...
+        { content: 'Admitting Physician' },
+        { content: '' },
+      ],
+    },
+  ],
+};
+
+const DefaultTable = () => (
+  // eslint-disable-next-line react/forbid-dom-props, jsx-a11y/no-noninteractive-tabindex
+  <div style={{ overflow: 'scroll' }} tabIndex={0}>
```
cm9361, line 56: There should not be a scroll component added around the Terra Table.  The Table component should have its own scrollbar.

packages/terra-table/src/Table.jsx
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    // Remove active index
+    setActiveIndex(null);
+  };
+
+  // -------------------------------------
+
+  return (
+    <div>
```
cm9361, line 248: We should probably have a class associated to this container div.  It will need styles.  For example it should have scroll capabilities.  You should be able to reference the container that was around the grid control.  We would want to encapsulate that capability within our control.

packages/terra-table/src/subcomponents/Cell.jsx
cm9361, line 1: It looks like you removed logic such as the selection of the cell.  We still need that capability for the grid role of the table.  We should not remove this necessary logic in this PR.

packages/terra-table/package.json
```diff
@@ -23,14 +23,17 @@
   "homepage": "https://github.com/cerner/terra-core#readme",
   "peerDependencies": {
     "react": "^16.8.5",
-    "react-dom": "^16.8.5"
+    "react-dom": "^16.8.5",
+    "react-intl": "2 || 3 || 4 || 5"
```
cm9361, line 27: Was this change necessary?  Do we actually support all of these versions?

packages/terra-table/src/proptypes/cellShape.js
cm9361, line 1: You removed props from the cell shape.  We need those props for the grid implementation.

packages/terra-table/src/subcomponents/Row.jsx
cm9361, line 1: You removed necessary logic from the Row component.  We should log separate JIRAs necessary to remove some of the necessary logic so that we can analyze our options appropriately while still allowing the grid components to build off your base component.

packages/terra-table/src/utils/constants.js
```diff
@@ -0,0 +1,7 @@
+const ERRORS = {};
```
cm9361, line 1: Do we really need a separate constant file for these errors?

packages/terra-table/src/Table.jsx
```diff
@@ -0,0 +1,291 @@
...
+      </table>
+      <VisuallyHiddenText aria-live="polite" aria-atomic="true" text={cellAriaLiveMessage} />
+    </div>
+  );
+}
+
+Table.propTypes = propTypes;
+Table.defaultProps = defaultProps;
+
+export default injectIntl(Table);
```
cm9361, line 291: I am wondering if we should memoize this component so that we don't have to worry about unnecessary rerenders.

packages/terra-core-docs/src/terra-dev-site/doc/table/example/DefaultTable.jsx
eawww: @cm9361 I advised this change after a sync with Kenny since: 
- we won't need to manually set scroll position like we do with grid.
- because we're not managing scroll position, any scrolling container will need to be focusable for keyboard control when there's no focusable elements in the table. So rather than having an always focusable scroll area, we can leave that to the discretion of the consumer.

If this change is technically incompatible with sticky headers or pinned columns or there's some other reason to have it in the component, we should discuss.

packages/terra-table/src/subcomponents/Cell.module.scss
```diff
@@ -0,0 +1,42 @@
+/* stylelint-disable terra/custom-property-namespace */
```
cm9361, line 1: You should not be disabling linter rules.  You just need to update the names to resolve appropriately.

packages/terra-core-docs/src/terra-dev-site/doc/table/example/DefaultTable.jsx
cm9361: I think the scrollbar should live within the Terra Table element.  We don't want to rely on consumers providing a scroll container so that functionality such as pinned columns or being able to see data beyond the bounds of the container works.  It also allows us to properly control the tabIndex on the scroll container.  If we leave this up to consumers, we leave ourselves open to issues.

cm9361: Also, this component will be used by the grids as well.

chrismichalewicz:
@kenk2  Thanks for the updated example for column sorting, I think it is much clearer now!  Overall did a quick functional review saw a lot of the same things the @mjpalazzo  already pointed out as far as navigation goes and table behavior goes. I am going to add the functionally reviewed label but with the disclaimer that as this component is still in progress I would expect further functional verifications for the remaining stories or when new functionality is added.  I agree with Charles' comment that I think a page just for examples would be beneficial especially as more functionality starts to get built out.

chrismichalewicz:
@mjpalazzo Below are the items from your post I believe need Jiras. I am thinking 1 Jira to encompass all of the AT issues, we can break it a part later if necessary. 

- I-beam pointer should display when hovering over contents of the table

- Assistive Technology/Screen reader
- [ ] When navigating via screen reader row and column counts should be announced.
- [ ] We should not be announcing the work "button" if one does not exist when navigating column header cells.  Is this related to column sorting? 
- [ ] JAWS should announce the new sort order when pressing enter on a header cell to sort the data. 

@cm9361 You mentioned the following items that might need Jiras
- Review of the Row logic to determine what can be removed while still allowing the grid components to build off of this base. You might need to expand on this more as I am not totally clear on what would need to go into the Jira. 
- Prop to toggle zebra striping on/off. 

If you guys will confirm this list I will start getting Jira's logged for them. Thanks!

packages/terra-table/src/utils/constants.js
sdadn: This file is not only for errors but for any constants that we may use. Having a separate file would mean a single place to update them instead of having to hunt for them through multiple source files.

packages/terra-table/src/subcomponents/index.js
sdadn, line 1: We don't seem to be using this file when consuming these components within `Table.jsx`. I'm leaning towards not having it to avoid confusion with `src/index.js` and since it's only used within this package anyway.

packages/terra-table/src/utils/constants.js
```diff
@@ -0,0 +1,7 @@
...
+ERRORS.ROW_HEADER_INDEX_EXCEEDS_PINNED = 'Prop rowHeaderIndex exceeds the size of pinnedColumns.';
+ERRORS.ROW_HEADER_INDEX_LESS_THAN_ZERO = 'Prop rowHeaderIndex must be a positive integer.';
+ERRORS.ROW_HEADER_INDEX_NOT_AN_INTEGER = 'Prop rowHeaderIndex must be an integer.';
+ERRORS.PINNED_COLUMNS_UNDEFINED = 'to be properly accessible, the row header column should be a pinned column. please set pinned columns';
```
sdadn, line 5:
```suggestion
ERRORS.PINNED_COLUMNS_UNDEFINED = 'To be properly accessible, the row header column should be a pinned column. please set pinned columns';
```

mjpalazzo:
> @mjpalazzo Below are the items from your post I believe need Jiras. I am thinking 1 Jira to encompass all of the AT issues, we can break it a part later if necessary.
> 
> * I-beam pointer should display when hovering over contents of the table
> * Assistive Technology/Screen reader
> * [ ]  When navigating via screen reader row and column counts should be announced.
> * [ ]  We should not be announcing the work "button" if one does not exist when navigating column header cells.  Is this related to column sorting?
> * [ ]  JAWS should announce the new sort order when pressing enter on a header cell to sort the data.
> 
> @cm9361 You mentioned the following items that might need Jiras
> 
> * Review of the Row logic to determine what can be removed while still allowing the grid components to build off of this base. You might need to expand on this more as I am not totally clear on what would need to go into the Jira.
> * Prop to toggle zebra striping on/off.
> 
> If you guys will confirm this list I will start getting Jira's logged for them. Thanks!

@chrismichalewicz - regarding the I beam pointer, I do not know if we need to do anything with this. I was noting a difference between Chrome and Edge, so it is likely browser specific. Safari does not show the pointer either. I defer to the SEs on whether we need to do anything for this.
I did some testing with VO and it is announcing column and row totals. With VO I can enter the table and navigate with control shift arrow. When I arrow through the cells with VO it announces the value, the column header and the cell position (column x row y). This makes me want to retest JAWS and ensure that I enter table mode before trying to navigate.



I agree with the other items listed.

packages/terra-table/src/Table.jsx
```diff
@@ -0,0 +1,291 @@
...
+     * Data for pinned columns. Pinned columns are the stickied leftmost columns of the table.
+     * Columns will be presented in the order given.
+     */
+  pinnedColumns: PropTypes.arrayOf(columnShape),
+
+  /**
+     * Data for overflow columns. Overflow columns are rendered in the table's horizontal overflow.
+     * Columns will be presented in the order given.
+     */
+  overflowColumns: PropTypes.arrayOf(columnShape),
```
cm9361, line 52: This does not have to be addressed in this JIRA, but I wonder if we should update our pattern to just have a single columns array and have consumers pass the pinned column count as a prop.  We really don't need two separate arrays and end up needing to combine them.  I understand for this JIRA that it is existing code.

packages/terra-table/src/proptypes/rowShape.js
cm9361, line 1: This should be the same as the row shape from the existing data-grid package.  Same comment for all other shapes.

packages/terra-core-docs/src/terra-dev-site/doc/table/example/DefaultTable.jsx
eawww: Alrighty. I'm cool with it. There should be a prop to disable it for Table though. We won't always want the scroll container.

eawww (approved):
A few things to add to what @cm9361 and @mjpalazzo said:
- We still need to ensure the click target for the resize handle is only in the header.
- This seems like the PR for terra-table to have a mvb since we're changing a bunch of the API. This should propagate to the grids as well.
- We won't necessarily want to use the same styling for Table as we do for grid. We should have design and restyling stories but those can come later.

Approving since this is just a base component for people to work off of.

cm9361:
@eawww - Is the focus ring correct for the scroll container getting focus?  What style should it have when focus is given?  This is probably another reason we should have control of that container.

cm9361:
> A few things to add to what @cm9361 and @mjpalazzo said:
> 
> * We still need to ensure the click target for the resize handle is only in the header.
> * This seems like the PR for terra-table to have a mvb since we're changing a bunch of the API. This should propagate to the grids as well.
> * We won't necessarily want to use the same styling for Table as we do for grid. We should have design and restyling stories but those can come later.
> 
> Approving since this is just a base component for people to work off of.

I'll set up a meeting to discuss this with the team.  I think we need to dedicate some time to building out these stories so that everyone is on the same page and there is clear direction so that we can complete the work in a timely manner.

mjpalazzo:
@kenk2 and @chrismichalewicz - I retested with JAWS and remembered to use T to navigate to the table examples. This time JAWS announced the number of columns and rows. I used the Down Arrow to navigate to each cell and JAWS announced the cell contents. Unlike VO on Safari, JAWS does not announce the position of the cell (col x row y) and it does not announce the column header. I would love to know why VO does a better job with Tables than JAWS does. I ask Google.

cm9361:
We are missing the ability to tab through the column headers that are sortable.  Is that part of this JIRA or do we need an additional JIRA logged to complete that work?

cm9361:
> @kenk2 and @chrismichalewicz - I retested with JAWS and remembered to use T to navigate to the table examples. This time JAWS announced the number of columns and rows. I used the Down Arrow to navigate to each cell and JAWS announced the cell contents. Unlike VO on Safari, JAWS does not announce the position of the cell (col x row y) and it does not announce the column header. I would love to know why VO does a better job with Tables than JAWS does. I asked Google. According to Freedom Scientific "The default behavior for JAWS is to read only headers that have been properly marked as headers by the author of a table. In many tables, the first row, going from left to right, contains headers for the information in the columns below it. Also, many times the first column, going up and down, contains headers for the information in the rows to the right." Since JAWS is announcing the column headers as "button" it must not be seeing them as as column headers. For some reason, VO does see them as column headers and is functioning as we want it to.

This is interesting and worth investigating.  We have the role of columnheader on the th elements so I would think that would signify to the screen reader that this is a column header.

packages/terra-table/package.json
kenk2: Paging @sdadn for this comment.

packages/terra-table/src/Table.jsx
```diff
@@ -0,0 +1,352 @@
...
+  /**
+   * String that will be used to identify the table. If multiple tables are on the same page, each table should have
+   * a unique id.
+   */
+  id: PropTypes.string.isRequired,
+
+  /**
+   * Prop that determines whether the table's container should be scrollable.
+   */
+  isScrollable: PropTypes.bool,
```
cm9361, line 42: I don't think consumers should have to pass a prop for if the table is scrollable.

packages/terra-table/src/Table.jsx
```diff
@@ -0,0 +1,352 @@
...
+  }, [tableRef]);
+
+  // Make the table container focusable when scrollable.
+  useEffect(() => {
+    const containerRect = tableContainerRef.current.getBoundingClientRect();
+    const tableRect = tableRef.current.getBoundingClientRect();
+    if (tableRect.height > containerRect.height || tableRect.width > containerRect.width) {
+      setIsContainerFocusable(true);
+    }
+  }, [tableContainerRef, tableRef]);
```
cm9361, line 204: This does not seem correct.  We would probably want to use resize observers as the contents of the table can change.

packages/terra-table/src/Table.jsx
```diff
@@ -0,0 +1,352 @@
...
+  };
+
+  // -------------------------------------
+
+  return (
+    <div
+      className={cx('table-container')}
+      ref={tableContainerRef}
+      // eslint-disable-next-line jsx-a11y/no-noninteractive-tabindex
+      tabIndex={isContainerFocusable ? 0 : undefined}
```
cm9361, line 304: We should probably handle the scrolling in a separate JIRA.  We are adding too much to one PR.  However, I am also not sure this container should be focusable when in a grid context.

packages/terra-table/src/Table.jsx
```diff
@@ -0,0 +1,352 @@
...
+    >
+      <table
+        ref={tableRef}
+        id={id}
+        role={gridContext.role}
+        aria-labelledby={ariaLabelledBy}
+        aria-label={ariaLabel}
+        className={cx('table', theme.className)}
+        tabIndex={gridContext.role === 'grid' ? 0 : undefined}
```
cm9361, line 313: I don't believe the Table should have a tabIndex of 0 for the grid context.  The grid will have to add the tabIndex to the container element for its logic.

packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -0,0 +1,222 @@
...
+      scope="col"
+      tabIndex={gridContext.role === 'grid' ? -1 : undefined}
+      aria-sort={sortIndicator}
+      onMouseDown={(isSelectable && onColumnSelect) ? handleMouseDown : undefined}
+      onKeyDown={isSelectable && onColumnSelect ? handleKeyDown : undefined}
+      // eslint-disable-next-line react/forbid-dom-props
+      style={{ width: `${width}px`, height: headerHeight, left: cellLeftEdge }}
+    >
+      {/* eslint-disable jsx-a11y/no-noninteractive-tabindex */}
+      <div className={cx('header-container')} role={displayName && 'button'} tabIndex={isSelectable ? 0 : undefined}>
```
cm9361, line 199: You have the tabIndex at two separate levels.

kenk2: Stripped this out, since we don't want to announce buttons in VO/Jaws anyways.

packages/terra-table/src/Table.jsx
kenk2: Fixed!

packages/terra-table/src/Table.jsx
kenk2: Removed all logic about manually managing container focus. Although we do set the tabIndex to 0 in DataGrid so we'll probably flex it to 0 on grid context here.

packages/terra-table/src/Table.jsx
kenk2: This was something I tried to do and somehow I forgot to remove this. It is now gone.

packages/terra-table/src/Table.jsx
```diff
@@ -0,0 +1,326 @@
+import React, {
+  useState, useContext, useRef, useCallback, useEffect, useMemo,
+} from 'react';
+import PropTypes from 'prop-types';
+import { injectIntl } from 'react-intl';
+import classNames from 'classnames/bind';
+import ResizeObserver from 'resize-observer-polyfill';
+import ThemeContext from 'terra-theme-context';
+import VisuallyHiddenText from 'terra-visually-hidden-text';
```
sdadn, line 9: Can we "group" terra imports separately?
```suggestion
import ResizeObserver from 'resize-observer-polyfill';

import ThemeContext from 'terra-theme-context';
import VisuallyHiddenText from 'terra-visually-hidden-text';
```

packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -0,0 +1,220 @@
+import React, { useContext, useRef, useCallback } from 'react';
+import PropTypes from 'prop-types';
+import { injectIntl } from 'react-intl';
+import * as KeyCode from 'keycode-js';
+import classNames from 'classnames/bind';
+
+import ThemeContext from 'terra-theme-context';
+import IconUp from 'terra-icon/lib/icon/IconUp';
+import IconDown from 'terra-icon/lib/icon/IconDown';
+import IconError from 'terra-icon/lib/icon/IconError';
```
sdadn, line 10: Can we used the updated import pattern? 

```suggestion
import {IconUp,IconDown,IconError } from 'terra-icon';
```

packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+   * String that specifies the default width for columns in the table. Any valid CSS width value is accepted.
+   */
+  width: PropTypes.number.isRequired,
+
+  /**
+   * String that specifies the column height. Any valid CSS height value accepted.
+  */
+  headerHeight: PropTypes.string.isRequired,
```
sdadn, line 75: Can we put all required tops at the top?

sdadn (commented):
I know these are stylistic nitpicks but I would like to update these to since we are adding these lines at this time and to serve as a reference for future work.

packages/terra-table/src/Table.jsx
cm9361: The DataGrid is setting it to 0 to control focus as it is a composite widget.  The Terra Table will not be a composite widget so we don't need to control focus in the same way.

packages/terra-table/src/Table.jsx
```diff
@@ -0,0 +1,326 @@
...
+    width: column.width || defaultColumnWidth,
+    minimumWidth: column.minimumWidth || defaultColumnMinimumWidth,
+    maximumWidth: column.maximumWidth || defaultColumnMaximumWidth,
+  });
+
+  const displayedColumns = [
+    ...pinnedColumns,
+    ...overflowColumns,
+  ];
+  const [renderedColumns, setRenderedColumns] = useState(displayedColumns.map((column) => initializeColumn(column)));
```
cm9361, line 147: Why do we need both displayed columns and rendered columns?  I would expect that we need only one like in the DataGrid implementation.  This is confusing have so many variables for our columns.

packages/terra-table/src/Table.jsx
```diff
@@ -0,0 +1,326 @@
...
+    setActiveIndex(null);
+  };
+
+  // -------------------------------------
+
+  return (
+    <div
+      className={cx('table-container')}
+      ref={tableContainerRef}
+      tabIndex={gridContext.role === 'grid' ? 0 : -1}
```
cm9361, line 279: The tabIndex for the table container would not be 0 for a grid control.  Also, why would we set the value to -1 for a table role.  Who is writing code to set focus to this container?

packages/terra-table/src/Table.jsx
```diff
@@ -0,0 +1,326 @@
...
+
+  // Manage column resize
+  const [tableHeight, setTableHeight] = useState(0);
+  const [activeIndex, setActiveIndex] = useState(null);
+
+  const activeColumnPageX = useRef(0);
+  const activeColumnWidth = useRef(200);
+  const tableWidth = useRef(0);
+  const tableRef = useRef();
+  const tableContainerRef = useRef();
```
cm9361, line 157: It does not seem like this ref variable is necessary with your current logic.

cm9361:
It still looks like you are missing a lot of the logic from the subcomponents from the DataGrid project.  We are not looking to remove the logic in this PR.  All of the functionality should be there to ensure the success of the initial integration.  We can always log JIRAs as necessary to break out logic.  However, that should be designed out.

packages/terra-table/src/subcomponents/Row.jsx
```diff
@@ -0,0 +1,108 @@
...
+          columnId={displayedColumns[cellColumnIndex].id}
+          rowIndex={rowIndex}
+          columnIndex={cellColumnIndex}
+          key={`${id}_${displayedColumns[cellColumnIndex].id}`}
+          isMasked={cellData.isMasked}
+          isRowHeader={cellColumnIndex === rowHeaderIndex}
+          height={height}
+          isHighlighted={isSelected}
+          onCellSelect={onCellSelect}
+          isSelectable={cellData.isSelectable || gridContext.role === 'grid'}
```
cm9361, line 96: Shouldn't this be an AND operator instead of an OR operator?  We only want the cell to be selectable when in the grid context.  I would make the grid check first as well so that it reads better.

packages/terra-table/src/subcomponents/Row.jsx
```diff
@@ -0,0 +1,108 @@
...
+          rowId={id}
+          columnId={displayedColumns[cellColumnIndex].id}
+          rowIndex={rowIndex}
+          columnIndex={cellColumnIndex}
+          key={`${id}_${displayedColumns[cellColumnIndex].id}`}
+          isMasked={cellData.isMasked}
+          isRowHeader={cellColumnIndex === rowHeaderIndex}
+          height={height}
+          isHighlighted={isSelected}
+          onCellSelect={onCellSelect}
```
cm9361, line 95: Are we allowing cell selection for the table role?  I don't think that should be possible.

packages/terra-core-docs/src/terra-dev-site/doc/table/About.1.doc.mdx
sdadn, line 1: Thinking about it now, we won't be fully releasing this right away for consumers and this MVB is a big change (technically, it's a new component with the same name). Can we have a `legacy table` page for the older table that basically lists the props and indicates that version of table is deprecated and no longer recommended?

cc: @cm9361 , @eawww

packages/terra-table/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Changed
+  * Re-implementing Terra Table in preparation of being consumed by DataGrid package.
+
```
sdadn, line 7: A few things here:

1. Since we are releasing this component for consumers, we should use consumer-oriented language. Yes, the motivation is to be consumed by DataGrid for _now_ but we still releasing this for consumers so we should describe the changes from a consumer's perspective
2. These are breaking changes so we should put them under the `Breaking Changes` header.

packages/terra-table/package.json
sdadn: We support react-intl 2 through 5. The original syntax was `>= 2, < 6` but this makes it more clearer.

packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
kenk2: So it turns out that I can't do this way because the way this package is set up. However we can do something like:

```import { IconUp, IconDown, IconError } from 'terra-icon/lib/IconExports';```

File: packages/terra-table/CHANGELOG.md (hunk 24 above)
Comment (kenk2) [reply packages/terra-table/CHANGELOG.md]: Done!

File: packages/terra-core-docs/src/terra-dev-site/doc/table/About.1.doc.mdx
Comment (kenk2) [reply packages/terra-core-docs/src/terra-dev-site/doc/table/About.1.doc.mdx:1]: Done!

File: packages/terra-table/src/subcomponents/Row.jsx (hunk 22 above)
Comment (kenk2) [reply packages/terra-table/src/subcomponents/Row.jsx]: Fixed: in Grid mode, it is always selectable unless the cell has `isSelectable: false` as a value. In table this will always be disabled.

File: packages/terra-table/src/Table.jsx (hunk 21 above)
Comment (kenk2) [reply packages/terra-table/src/Table.jsx]: Removed.

File: packages/terra-table/src/Table.jsx (hunk 20 above)
Comment (kenk2) [reply packages/terra-table/src/Table.jsx]: Removed as discussed per sync.

File: packages/terra-table/src/Table.jsx (hunk 19 above)
Comment (kenk2) [reply packages/terra-table/src/Table.jsx]: I attempted to re-factor this but it seems to be a non-trivial change so I reverted this back to how it used to be in DataGrid. I agree though that this is not an ideal way to manage the table columns.

File: packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx (hunk 18 above)
Comment (kenk2) [reply packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx]: Done

File: packages/terra-table/src/Table.jsx (hunk 16 above)
Comment (kenk2) [reply packages/terra-table/src/Table.jsx]: Sorted and grouped imports in all files.

File: packages/terra-table/src/Table.jsx [hunk 25]
```diff
@@ -0,0 +1,353 @@
...
+  hasSelectableRows: PropTypes.bool,
+};
+
+const defaultProps = {
+  rowHeaderIndex: 0,
+  defaultColumnWidth: 200,
+  columnHeaderHeight: '2.5rem',
+  rowHeight: '2.5rem',
+  pinnedColumns: [],
+  overflowColumns: [],
```
smason0, line 118: Should we add a default empty array for `rows` as well?

kenk2: Right now I'm requiring `rows` to be supplied (but it can be empty) since in DataGrid this is actually a bug right now (component will not render otherwise). But I could also see a case where we might want to default to `[]` if `rows` is not specified.

packages/terra-table/src/subcomponents/Cell.jsx
```diff
@@ -0,0 +1,259 @@
...
+};
+
+const defaultProps = {
+  isRowHeader: false,
+  isSelected: false,
+  isSelectable: false,
+  isMasked: false,
+};
+
+function Cell(props) {
```
smason0, line 102: So just wanting to make sure, this package will be where all these subcomponents will live instead of in data-grid? So we plan on removing them from data-grid when consuming Table? I have some in-flight changes to both Row and Cell as part of Flowsheet that, depending on the PR review, I would need to make sure don't get lost in the process of migrating these.

kenk2: In the future, yes these components will all live inside this Table package. If you finish first, then I'll be adding your changes into this PR. If I finish first with this then we'll have to make another PR to add in-fight work to this PR when it happens. But most of it should be copy-paste and not require significant rework.

packages/terra-table/src/Table.jsx
```diff
@@ -0,0 +1,354 @@
...
+    // eslint-disable-next-line no-console
+    console.warn(ERRORS.PINNED_COLUMNS_UNDEFINED);
+  }
+
+  // Manage column resize
+  const [tableHeight, setTableHeight] = useState(0);
+  const [activeIndex, setActiveIndex] = useState(null);
+  const [cellAriaLiveMessage, setCellAriaLiveMessage] = useState(null);
+
+  const [pinnedColumnOffsets, setPinnedColumnOffsets] = useState([0]);
```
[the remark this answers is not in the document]
kenk2: At some point I'd love to re-factor this and move the code around but like the table columns logic this is non-trivial to change since AriaLiveMessage is used pretty much everywhere for the screen reader. We could probably log another JIRA to do this or flesh out the re-factor in another story.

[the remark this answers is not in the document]
sdadn: Makes sense. That's fine then.

packages/terra-table/src/subcomponents/Cell.jsx
smason0: Disregard, I'm changing my code to no longer affect Cell or Row.

packages/terra-table/src/Table.jsx
cm9361: Did we consider this comment?

packages/terra-table/src/proptypes/cellShape.js
```diff
@@ -0,0 +1,14 @@
...
+
+const cellShape = PropTypes.shape({
+  /**
+   * Content that will be rendered within the Cell.
+   */
+  content: PropTypes.node,
+  /**
+   *  Boolean indicating if cell contents are masked.
+   */
+  isMasked: PropTypes.bool,
```
cm9361, line 11: The cell shape still does not have all of its props from the DataGrid package implementation.

packages/terra-table/src/proptypes/rowShape.js
```diff
@@ -0,0 +1,22 @@
...
+
+  /**
+   * Data to be displayed in the cells of the row. Cells will be rendered in the row in the order given.
+   */
+  cells: PropTypes.arrayOf(cellShape),
+  /**
+   * A string identifier used to describe the row contents. This value will be used to construct additional labels
+   * for internal controls (e.g. row selection cells).
+   */
+  ariaLabel: PropTypes.string,
```
cm9361, line 18: The rowShape object does not have all of its props from the DataGrid implementation.

packages/terra-table/src/subcomponents/ColumnHeader.jsx
```diff
@@ -0,0 +1,77 @@
...
+            id={column.id}
+            rowIndex={0}
+            columnIndex={columnIndex}
+            displayName={column.displayName}
+            width={column.width}
+            minimumWidth={column.minimumWidth}
+            maximumWidth={column.maximumWidth}
+            headerHeight={headerHeight}
+            isResizable={column.isResizable}
+            isSelectable={isGridContext ? column.isSelectable !== false : column.isSelectable}
```
cm9361, line 63: The check for column.isSelectable != false is not a present on the current ColumnHeader.

packages/terra-table/src/subcomponents/Row.jsx
```diff
@@ -0,0 +1,149 @@
...
+      {cells.map((cellData, cellColumnIndex) => (
+        <Cell
+          rowId={id}
+          columnId={displayedColumns[cellColumnIndex].id}
+          rowIndex={rowIndex}
+          columnIndex={cellColumnIndex + columnIndexOffSet}
+          key={`${id}_${displayedColumns[cellColumnIndex].id}`}
+          isSelected={!hasRowSelection && cellData.isSelected}
+          isMasked={cellData.isMasked}
+          isSelectable={isGridContext ? cellData.isSelectable !== false : false}
```
cm9361, line 133: You don't need a ternary operator here.  You could just check 
isGridContext && cellData.isSelectable !== false

packages/terra-table/src/utils/_elementPolyfill.js
```diff
@@ -0,0 +1,19 @@
+// Source: https://www.npmjs.com/package/element-closest-polyfill
```
cm9361, line 1: I cannot find where this polyfill file is being used?

packages/terra-table/src/utils/GridContext.jsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+const GridContext = React.createContext({
+  role: 'table',
+  cellAriaLiveMessage: '',
```
cm9361, line 5: This property should not be required.  Nobody is reading the value.  The table just needs the ability to call the callback.

packages/terra-table/src/subcomponents/Cell.jsx
```diff
@@ -0,0 +1,263 @@
...
+  }, theme.className);
+
+  const cellLeftEdge = (columnIndex < columnContext.pinnedColumnOffsets.length) ? columnContext.pinnedColumnOffsets[columnIndex] : null;
+
+  const CellTag = isRowHeader ? 'th' : 'td';
+
+  return (
+    <CellTag
+      ref={cellRef}
+      aria-selected={isGridContext ? isSelected : undefined}
```
cm9361, line 236: We should not have the isGridContext check here.  We can just set it based on the selection state.

packages/terra-table/src/subcomponents/ColumnResizeHandle.jsx
```diff
@@ -0,0 +1,109 @@
...
+  };
+
+  return (
+    /* eslint-disable react/forbid-dom-props */
+    <div
+      ref={resizeHandleRef}
+      draggable
+      role="slider"
+      tabIndex={isGridContext ? -1 : 0}
+      aria-hidden={isGridContext ? !isActive : undefined}
```
cm9361, line 93: Does this really depend on the grid context?

packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
```diff
@@ -0,0 +1,230 @@
...
+    onResizeMouseDown,
+    rowIndex,
+  } = props;
+
+  const columnContext = useContext(ColumnContext);
+  const gridContext = useContext(GridContext);
+  const theme = useContext(ThemeContext);
+  const columnHeaderCell = useRef();
+
+  const isGridContext = gridContext.role === GridConstants.GRID;
```
cm9361, line 137: Are we able to just made this a function in the GridContext that can be called instead of having to duplicate the logic in multiple files?

packages/terra-table/src/subcomponents/ColumnResizeHandle.jsx
kenk2: I was facing an issue earlier where I was getting accessibility errors, but this works now. Done.

packages/terra-table/src/subcomponents/ColumnHeaderCell.jsx
kenk2: So it turns out that React contexts are read-only and we can't reference the values of a context inside itself. So This will have to do for now.

packages/terra-table/src/subcomponents/Cell.jsx
kenk2: There is an edge case where `aria-selected={false}` in tables creates an accessibility violation, but since since we're saying that cell selection is a grid functionality (which setting the value is allowed) then we can just honor the value from the cell data. Done.

packages/terra-table/src/utils/GridContext.jsx
kenk2: Done.

cm9361 (approved):

mjpalazzo:
@kenk2 - Great work on Table and the examples! I have a few observations:
Keyboard navigation - there are no keyboard requirements for the Table pattern. I was able to tab to the table. For examples with scroll bars I was able use the down arrow keys to scroll through the table content. This provides sighted users with access to off screen content, so that is a valid behavior.
The Pinned Column table navigation with the Down Arrow key worked well. When I Down Arrowed to the right most cell and then pressed Down again, the cells hiddenb behind the pinned columns were visible again. However, when using the Up Arrow, I could not get the columns hidden behind the pinned columns completely visible.
JAWS Screen reader - From the Default Table header I used down arrow to go to the table. JAWS announced "Table with 10 columns and 3 rows." I used Down arrow to navigate the cells. For the table header cells, JAWS announced the role "button" even though there was no sorting enabled. For example, "Patient button." I think JAWS should announce "column header" @eawww - what are your thoughts on the header cell announcements?
Once I was in the body of the table, JAWS announced the cell content correctly. When I pressed Down Arrow from the last cell, JAWS announced, "Table end"
For the Sorted Table example:
APG recommends the following:
"To help screen reader users understand the purpose of the buttons in the column headers, an off-screen description of the sort functionality of the buttons is appended to the caption text. The description is added to the caption instead of to each button to prevent repetitious verbosity that could interfere with understanding of the column titles." I suggest that we add this suggested description to our example.
when I pressed Down Arrow to navigate to the first header cell, JAWS announced "Patient column header sorted ascending column 1" I pressed Enter which changed the sort order. JAWS did not announce the sort order. I tabbed to the column 2 header. JAWS announced "Location column header, column 2" The sort icon did not display (it should) and JAWS did not announce that the column is sortable or the sort order. JAWS should announce that the cell is interactable and how to interact with it.  I pressed Enter and the sort order changed, but JAWS only announced "Enter" (Note that on the APG Sortable Table example page, JAWS did not announce the change in sort order either.)
That's all for now. I want to get these observations posted. Let me know if you would like to discuss.
I will continue testing the examples with focusable elements and also test with VoiceOver

PR #3920, opened 2023-09-21
[terra-button] VoiceOver in Safari is highlighting the Terra Button in a frame bigger than the button size, when the button is included to a modal dialog.

MadalinaStelea:
### Summary
Fixed issue - "VoiceOver in Safari is highlighting the Terra Button in a frame bigger than the button size, when the button is included to a modal dialog."

**What was changed:**
Removed 'position: relative', 'left: 0' and 'top: 0' from the .css when on Mac.

**Why it was changed:**
The 'position: relative' was creating the issue. The position, along with the left and top attributes were set specifically for IE mode, so they were not necessary on Mac anyway.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest

   Button before changes:
   image
   
   Button after changes:
   image

- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9645 

---

@cerner/terra


package.json
```diff
@@ -77,7 +77,7 @@
     "wdio-default": "terra wdio",
     "wdio-lowlight": "terra wdio --themes clinical-lowlight-theme",
     "wdio-fusion": "terra wdio --themes orion-fusion-theme",
-    "wdio": "terra wdio --themes terra-default-theme clinical-lowlight-theme orion-fusion-theme",
+    "wdio": "terra wdio --themes terra-default-theme clinical-lowlight-theme orion-fusion-theme --spec packages/terra-button/tests/wdio/button-spec.js --browsers chrome --gridUrl grid314.test.devcernerpowerchart.com",
```
razvanluca, line 80: I think you want to discard this change

supreethmr (approved):

package.json
MadalinaStelea: Right, I've added it by mistake. Thank you!

packages/terra-button/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fixed
+  * VoiceOver in Safari was highlighting the Terra Button in a frame bigger than the button size, when the button was included to a modal dialog.
```
sdadn, line 6:
```suggestion
  * Fixed bug where VoiceOver in Safari highlighting Terra Button within a modal dialog with a frame bigger than the button size.
```

packages/terra-button/src/Button.jsx
```diff
@@ -243,6 +243,8 @@ class Button extends React.Component {
     } = this.props;
 
     const theme = this.context;
+    const isMac = () => navigator.userAgent.indexOf('Mac') !== -1 && navigator.userAgent.indexOf('Win') === -1;
```
sdadn, line 246: Why do we need to test for Win AND Mac? Shouldn't the first condition be sufficient?

sdadn (changes requested):
We should write a unit test where we mock the `isMac` and test if the correct class is applied.

packages/terra-button/src/Button.jsx
MadalinaStelea: The first condition should be sufficient indeed. I saw that both terra-dropdown-button and terra-form-select have both conditions and I thought to keep the "pattern", but the second condition is probably redundant, so I'm removing it.

MadalinaStelea:
> We should add a unit test where we mock the `isMac` and test if the correct class is applied.

Added the jest tests.

packages/terra-button/tests/jest/Button.test.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import ThemeContextProvider from 'terra-theme-context/lib/ThemeContextProvider';
+import { clear, mockUserAgent } from 'jest-useragent-mock';
```
sdadn, line 3: I would rather not pull in another dependency for this, especially when it's something easily achievable on our end.

We can simply create a spy for navigator e.g.
https://stackoverflow.com/questions/52868727/how-to-mock-window-navigator-language-using-jest

Some more tips:
- Use `beforeAll` and `afterAll` blocks to set and [restore](https://jestjs.io/docs/mock-function-api#mockfnmockrestore) the spy.
- Use the `mockReturnValueOnce()` method to set the appropriate value for the spy.

MadalinaStelea: Made the change. Thank you!

packages/terra-button/tests/jest/Button.test.jsx
sdadn, line 1: +1 looks good!

PR #3924, opened 2023-09-22
[Terra-Paginator] Added `hidePageCount` prop to hide default page count provided by Paginator

supreethmr:
### Summary

Added `hidePageCount` prop to hide default page count provided by Paginator. so that user can set custom page count.
 
**What was changed:**
Progressive Paginator and Controlled Progressive Paginator has been updated to hide the default page count ( highlighted on below screenshot )

Screenshot 2023-09-22 at 10 26 53 PM

**Why it was changed:**
To support passivity for fusion pass through of paginator and To allow user to pass in the count in desired format along with customPageLabel when required.

### Testing

This change was tested using:

- [X] WDIO
- [ ] Jest
- [ ] No tests are needed

Updated `terra-core --> tests ---> Paginator ---> Progressive Paginator With Custom Label` to hide default Page count when hidePageCount is set to true

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9662 

---

@cerner/terra


packages/terra-paginator/src/ControlledProgressivePaginator.jsx
```diff
@@ -47,6 +47,12 @@ const propTypes = {
    * Allows user to set custom page label. _(usage note: User must pass translated text)_.
    */
   pageLabel: PropTypes.string,
+  /**
+   * ![IMPORTANT](https://badgen.net/badge/UX/Accessibility/blue)
+   * When specified allows user to set custom page count. User should provide custom page count as part `pageLabel` for best accessibility practices.
+   * _(usage note: when `pageLabel` is not provided page count will not be hidden and default page count is displayed for best accessibility practices)_.
+   */
+  hidePageCount: PropTypes.bool,
```
chenyixuan625, line 55: Should we make this new prop private since it's only for Fusion usage?

packages/terra-paginator/src/ControlledProgressivePaginator.jsx
```diff
@@ -113,14 +120,11 @@ class ControlledProgressivePaginator extends React.Component {
     const renderFirstandLastButton = totalCount > 0;
 
     const { messageId, messageAttributes } = getPageLabel(pageLabel, selectedPage, totalPages);
+    const pageDetails = (hidePageCount && pageLabel) ? <div>{pageLabel}</div> : <div>{intl.formatMessage({ id: messageId }, messageAttributes)}</div>;
```
chenyixuan625, line 123: I think we will need to use the translation for the pure _pageLabel_ text as well to meet internationalization criteria.

packages/terra-paginator/tests/wdio/paginator-spec.js
```diff
@@ -148,5 +148,15 @@ Terra.describeViewports('Controlled Paginator', ['large'], () => {
...
+
+    it('should display custom page count with custom page label when hidePageCount is set to true', () => {
+      $('#hide-count-button').click();
+      Terra.validates.element('custom page label with custom page count');
+    });
+
+    it('should display progressive paginator with default page count when page label is not provided', () => {
+      $('#clear-page-lable-button').click();
+      Terra.validates.element('paginator with default page count');
+    });
```
chenyixuan625, line 160: Other than wdio tests, could we also add some jest tests to reflect this change?

chenyixuan625 (commented):
One thing that I want to call out is that, with this change, when the consumer sets a static pageLabel string on the ProgressivePaginator https://github.com/cerner/terra-core/pull/3924/files#diff-5e4d0230ddc94d1702b4ce988ef0f63246ba4a209192779171626a877ee85097R19, this displayed text would **not** change by any page navigation. For example, if they have the pageLabel as "Page 1", it will be still "Page 1" after they click the Next button because this pageLabel won't change. Just want to confirm that it's expected from the Fusion usage perspective.

packages/terra-paginator/src/ControlledProgressivePaginator.jsx
supreethmr: I think we decided to make this available for other consumers as well. hence keeping it available for all ( not just for fusion).

packages/terra-paginator/src/ControlledProgressivePaginator.jsx
supreethmr: Translations will be provided by user. Since Page Label is custom string provided user we cannot translate them, Instead user should always provide translated string instead of normal string.

packages/terra-paginator/tests/wdio/paginator-spec.js
supreethmr: Initially I thought of JEST but I was able to think of any value or additional validation that Jest can bring in this scenario as the change made here is more visually oriented. Please share your inputs on how Jest can add extra validation will include them if they really required.

supreethmr:
> One thing that I want to call out is that, with this change, when the consumer sets a static pageLabel string on the ProgressivePaginator https://github.com/cerner/terra-core/pull/3924/files#diff-5e4d0230ddc94d1702b4ce988ef0f63246ba4a209192779171626a877ee85097R19, this displayed text would **not** change by any page navigation. For example, if they have the pageLabel as "Page 1", it will be still "Page 1" after they click the Next button because this pageLabel won't change. Just want to confirm that it's expected from the Fusion usage perspective.

That's the expectation of custom page label. user will see value exactly how he provides without any intervention of terra-paginator. ( There might be other use cases where user may require page count in different format than the one that terra-provides. This changes will help user to make those customization )

chenyixuan625:
Looks like there are a few wdio test failures, could we fix them?

supreethmr:
> Looks like there are a few wdio test failures, could we fix them?

fixed the screenshot failures

chenyixuan625 (approved):

packages/terra-paginator/src/ControlledProgressivePaginator.jsx
```diff
@@ -50,6 +50,12 @@ const propTypes = {
    * Allows user to set custom page label. _(usage note: User must pass translated text)_. It should not contain page number and total count details, which will lead to conflicts with built-in accessibility ARIA contexts.
    */
   pageLabel: PropTypes.string,
+  /**
+   * ![IMPORTANT](https://badgen.net/badge/UX/Accessibility/blue)
+   * When specified allows user to set custom page count. User should provide custom page count as part `pageLabel` for best accessibility practices.
+   * _(usage note: when `pageLabel` is not provided page count will not be hidden and default page count is displayed for best accessibility practices)_.
+   */
+  hidePageCount: PropTypes.bool,
```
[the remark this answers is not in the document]
supreethmr: Added : https://github.com/cerner/terra-core/pull/3924/commits/5fd4d938be53bd4b52d50ecbe610acc124b70b52

PR #3935, opened 2023-10-12
[terra-list]Fix for terra-menu console error 

ashishkcerner:
### Summary

**What was changed:**
Implemented the condition before focus to handle undefined value condition

**Why it was changed:**
User is getting Errors when navigating using arrow keys on terra-menu.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9738 

---

Pre implementation:- 

https://github.com/cerner/terra-core/assets/126875089/f6a0180f-be70-4017-9114-945f7369dbef

Post implementation:- 

https://github.com/cerner/terra-core/assets/126875089/a50e93e3-e70d-4c65-b123-729241adee70

@cerner/terra


packages/terra-list/src/List.jsx
```diff
@@ -162,13 +162,17 @@ const List = ({
       case KeyCode.KEY_UP: {
         event.preventDefault();
         const previousIndex = currentIndex > 0 ? currentIndex - 1 : lastIndex;
-        listItems[previousIndex].focus();
+        if (listItems[previousIndex]) {
+          listItems[previousIndex].focus();
+        }
```
sdadn, line 167: These changes should be tested

packages/terra-list/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Fixed
+  * Fixed the console error appearing in the terra-menu examples while keyboard navigation.
```
sdadn, line 6: What are the errors?

ashishkcerner: Following is the error getting while navigating through arrow key:- 
 Uncaught TypeError: Cannot read properties of undefined (reading 'focus')

packages/terra-list/src/List.jsx
ashishkcerner: Yes have tested by creating the tar file with this change on terra-menu . For reference please check the screen recording attached in PR description for pre and post implementation .

sdadn: Sorry, bad wording on my part. I trust that this was tested before submitting.

What I meant was that we should write tests for these to test the condition and any potential edge cases. It can be a unit test or a wdio test.

packages/terra-list/CHANGELOG.md
sdadn: We can rephrase this to:

`Fixed _property is undefined_ error while navigating with a keyboard.`

supreethmr (approved):

packages/terra-list/src/List.jsx
ashishkcerner: For this particular scenario not being able to create test :- These functionality can be hard because not being able to assert the focus invocation on null object ,which is making hard to write the test case to justify the changes made on List.jsx.

sdadn: Hmm I thought you could test the condition but that makes sense. If it's not possible, then the screen recording should be good enough. Thanks for trying!
+1

sdadn (approved):

PR #3942, opened 2023-10-17
[terra-tag] Combobox throws error when integrated with Fusion Combo Box

MadanKumarGovindaswamy:
### Summary

Terra Form Select Combobox component tries to access a DOM element using a hardcoded classname string in method. This component works fine when consumed normally. But in fusion, the compilation step transpiles these sass classes coming from terra into some dynamic string. So, when consuming Combox in fusion, the component is not able to find the element using the logic mentioned above.

**What was changed:**

Added data attribute instead of querying class Name.

**Why it was changed:**

When Combo box is used in Mpage Application all the class names are transpiled. And when we try to query using the class name component gets crashed.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9759 

---

@cerner/terra


packages/terra-form-select/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Changed
+  * Fixed replacing hardcoded classname with data attribute in form select combo box.
```
supreethmr, line 6:
```suggestion
* Changed
  * Changed `class` selector used for accessing input text with data-attribute in form select `combobox` and `tag`.
```

MadanKumarGovindaswamy: Updated. Thanks

supreethmr (approved):

PR #3946, opened 2023-10-18
CI/CD-NPM_auditbuild scripts

udaymattam:
### Summary

Creating "npm audit" during the ci/cd build

**What was changed:**

**Why it was changed:**

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-XXXX 

---

@cerner/terra

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -71,7 +71,10 @@ jobs:
         npm audit --json > audit.json || true
         if [ -s audit.json ]; then
           echo 'NPM audit vulnerabilities found:'
-          cat audit.json
+          npm audit --registry=https://registry.npmjs.org --json > audit-full.json
+          cat audit-full.json  # Display the contents of audit-full.json
+        else
+          echo 'No NPM audit vulnerabilities found.'
         fi
   release-and-deploy:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -59,7 +59,20 @@ jobs:
         if: matrix.theme == 'fusion'
       - name: Run WDIO Tests for form factor ${{ matrix.form-factor }} and ${{ matrix.theme }}
         run: SITE=build/${{ matrix.theme }} FORM_FACTOR=${{ matrix.form-factor }} npm run wdio-${{ matrix.theme }}
+  NPM_Audit:
+    runs-on: ubuntu-latest
 
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - name: Install dependencies and run npm audit
+        run: |
+        npm i --package-lock-only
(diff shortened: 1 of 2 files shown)


sdadn:
What is the purpose for this change?

udaymattam:
> What is the purpose for this change?

https://jira2.cerner.com/browse/UXPLATFORM-8278 - this is a Pr audit will run when a PR builds. as of now it will give audit report & log all the vulnerabilities. if we want to fail we can fail for critical & high vulnerabilities that will be feature enhancement.

sycombs:
> > What is the purpose for this change?
> 
> https://jira2.cerner.com/browse/UXPLATFORM-8278 - this is a Pr audit will run when a PR builds. as of now it will give audit report & log all the vulnerabilities. if we want to fail we can fail for critical & high vulnerabilities that will be feature enhancement.

Do we really want to fail builds for vulnerabilities or should we just run the audit script separately from the CI/CD periodically and log Jiras based on that?

sdadn:
> Do we really want to fail builds for vulnerabilities or should we just run the audit script separately from the CI/CD periodically and log Jiras based on that?

Yeah I don't think we should fail builds, this will add potential blockers to merging PRs & doing releases. I also have concerns about inflating build times if we're only going to throw out warnings.

udaymattam:
> > Do we really want to fail builds for vulnerabilities or should we just run the audit script separately from the CI/CD periodically and log Jiras based on that?
> 
> Yeah I don't think we should fail builds, this will add potential blockers to merging PRs & doing releases. I also have concerns about inflating build times if we're only going to throw out warnings.



> > > What is the purpose for this change?
> > 
> > 
> > https://jira2.cerner.com/browse/UXPLATFORM-8278 - this is a Pr audit will run when a PR builds. as of now it will give audit report & log all the vulnerabilities. if we want to fail we can fail for critical & high vulnerabilities that will be feature enhancement.
> 
> Do we really want to fail builds for vulnerabilities or should we just run the audit script separately from the CI/CD periodically and log Jiras based on that?

we are not failing the build for vulnerabilities. And compared the build time and it doesn't take much time(< 201 325ms ) . moreover it is used to see how many CVES (vulnerability's) are monitoring.

sdadn:
If it's simply a matter of monitoring CVE's then those are output from `npm install` anyway e.g.

CleanShot 2023-10-23 at 07 57 07

`npm audit` simply gives a detailed report with potential fixes. I do not see any advantage of adding this command to the CICD when it should be run locally.

kenk2:
Agreed with @sdadn. If we introduce this into the CI, we'll either ignore it if it doesn't fail the build or it will become a needless blocker of our work if it does.

PR #3637, opened 2022-06-27
[terra-action-header] Removed default heading level.

saket2403:
### Summary
This PR removes the default heading level of action header and also changes the title prop to text prop.

Closes #

### Deployment Link

https://terra-core-deployed-pr-#.herokuapp.com/

### Additional Details

Thank you for contributing to Terra.
@cerner/terra

PR #3956, opened 2023-10-30
[terra-section-header] SectionHeader a11y updates

cm9361:
### Summary

**What was changed:**
The SectionHeader component was updated to meet accessibility standards.

**Why it was changed:**
The change was made so that the component was compliant with our accessibility strategy.

### Testing

This change was tested using:

- [X] WDIO
- [X] Jest
- [ ] No tests are needed

### Reviews

- [X] UX review
- [X] Accessibility review
- [ ] Functional review

### Additional Details
N/A

**This PR resolves:**

UXPLATFORM-9794 

---

@cerner/terra


packages/terra-section-header/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+  * Breaking Change
```
supreethmr, line 5: Since this is Breaking Change we would need Upgrade guide to be updated as well

supreethmr:
Since this PR has opened for Major version release we can remove some of the todo's that we had planned to do in next MVB release of terra-section-header 

1. We need to remove the `title` prop from section-header : https://github.com/cerner/terra-core/blob/376977c875123802ada4bc34bca6c606755e65b6/packages/terra-section-header/src/SectionHeader.jsx#L22
2. We need to remove default value for `level` prop : https://github.com/cerner/terra-core/blob/376977c875123802ada4bc34bca6c606755e65b6/packages/terra-section-header/src/SectionHeader.jsx#L147

packages/terra-core-docs/src/terra-dev-site/doc/section-header/example/FocusSectionHeader.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+const FocusSectionHeader = () => {
+  const sectionHeader = useRef();
+
+  const sectionHeaderRef = (header) => {
+    sectionHeader.current = header;
+  };
+
+  const handleClick = () => {
+    sectionHeader.current.setFocus();
```
supreethmr, line 12:
```js
sectionHeader.current.focus();
``` 
We can also set focus on `sectionHeader` element like above code why do we need a handler to set Focus on component.?

supreethmr:
I noticed an error on Axe tool ( **which is not related to the changes of this PR** ). It looks like we have added heading element `h2` on section-header example template which is causing the above Axe violation. While updating the prop name from `title` to `text` on examples template was not updated to use new prop due to which empty h2 being added for each of the section-header example. 

Screenshot 2023-10-31 at 6 13 38 PM

I am thinking instead of updating the example template with new prop name `text`. we can remove the [h2 element](https://github.com/cerner/terra-core/blob/376977c875123802ada4bc34bca6c606755e65b6/packages/terra-core-docs/src/terra-dev-site/doc/section-header/example/SectionHeaderExampleTemplate.jsx#L19) from template since we already have an heading that is being provided from terra-dev-site.

Screenshot 2023-10-31 at 6 05 52 PM

mjpalazzo:
I validated with Edge/JAWS and Safari/VoiceOver. Keyboard and AT interactions worked as required.

packages/terra-section-header/src/SectionHeader.jsx
```diff
@@ -135,35 +123,37 @@ class SectionHeader extends React.Component {
...
-    if (!level) {
-      // eslint-disable-next-line no-console
-      console.warn('Default heading level may not appropriate has it would fail to convey context of heading in a site / application where it is used. Heading level should be set explicitly depending on the position of header in site / application to allow screen readers to identify headers consistently.'); // to be removed on next major version release.
-    }
-
-    if (title) {
-      // eslint-disable-next-line no-console
-      console.warn('`title` prop has been renamed to `text`. please update all the refernces of `title` prop to use prop `text`.'); // to be removed on next major version release.
-    }
     const Element = `h${level || 2}`;
```
supreethmr, line 126:
```suggestion
    const Element = `h${level}`;
```
We need to remove default value, As we need to make `level` as required prop, `level` should be specified explicitly to allow screen readers to identify headers consistently.

supreethmr:
Can we update the prop description of `level` prop. `level` prop is required prop from next MVB 
```jsx
  /**
   * ![IMPORTANT](https://badgen.net/badge/UX/Accessibility/blue)
   * Sets the heading level. One of `1`, `2`, `3`, `4`, `5`, `6`.
   * `level` should be specified explicitly to allow screen readers to identify headers consistently.
   */
  level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
```

packages/terra-section-header/src/SectionHeader.jsx
cm9361: @supreethmr - I don't understand why the prop needs to now be required.  The level will have a value a read the same way in the screen reader.  If the consumer does not want the default value of 2, they can pass the desired value.

cm9361:
> Can we update the prop description of `level` prop. `level` prop is required prop from next MVB
> 
> ```js
>   /**
>    * IMPORTANT
>    * Sets the heading level. One of `1`, `2`, `3`, `4`, `5`, `6`.
>    * `level` should be specified explicitly to allow screen readers to identify headers consistently.
>    */
>   level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
> ```

I don't understand why this prop would need to be required.

packages/terra-section-header/src/SectionHeader.jsx
artpark: Could we move the default value to the defaultProps constant? That way it will show up in the documentation as well

sdadn:
> Can we update the prop description of `level` prop. `level` prop is required prop from next MVB
> 
> ```js
>   /**
>    * IMPORTANT
>    * Sets the heading level. One of `1`, `2`, `3`, `4`, `5`, `6`.
>    * `level` should be specified explicitly to allow screen readers to identify headers consistently.
>    */
>   level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
> ```

@supreethmr we should log a Jira to do an audit and make all the MVB changes before release e.g. remove deprecation warnings etc.

artpark (approved):

eawww:
There's been confusion on this in the past, including myself, but Fusion theme should still get the silly blue glow border.
Otherwise, looks great!

packages/terra-core-docs/src/terra-dev-site/doc/section-header/example/LongTitleSectionHeader.jsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import SectionHeaderExampleTemplate from './SectionHeaderExampleTemplate';
 
-const LongTitleSectionHeader = () => <SectionHeaderExampleTemplate text="Long Title" exampleProps={{ text: 'This is a header with long text .This is an element provides a customizable header layout with a prominent title options for the elements.This component saves time and effort in designing a header from scratch and ensures consistency across a websites pages.' }} />;
+const LongTitleSectionHeader = () => <SectionHeaderExampleTemplate exampleProps={{ text: 'This is a header with long text .This is an element provides a customizable header layout with a prominent title options for the elements.This component saves time and effort in designing a header from scratch and ensures consistency across a websites pages.' }} />;
```
adoroshk, line 4: This is not exactly related to the change, but as this text is user facing, it might be a good idea to fix typos such as no white space after a dot, extra white space before a dot, and "This is an element provides" needs rephrasing.

adoroshk (approved):

packages/terra-section-header/src/SectionHeader.jsx
supreethmr: @cm9361  The reason why we need to make the `level` prop as required is because As part of A11y audit UX team had reported that having default heading level will create inconsistency in heading order. Header elements should always be used with appropriate heading level depending on it's position in a page. Chances application using header components with incorrect heading level is High when we have default value for `level` prop. Hence we had a conclusion to remove default value and make the prop as required. So that consumer will have to provide appropriate heading level to use terra-header components.

This is one of the approach which we have taken for all the terra header components which had heading level. To make this transition of using explicit heading level we had removed value from `defaultProps` and added up it here for passivity concerns.

cc: @scottwilmarth

supreethmr:
> > Can we update the prop description of `level` prop. `level` prop is required prop from next MVB
> > ```js
> >   /**
> >    * IMPORTANT
> >    * Sets the heading level. One of `1`, `2`, `3`, `4`, `5`, `6`.
> >    * `level` should be specified explicitly to allow screen readers to identify headers consistently.
> >    */
> >   level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
> > ```
> 
> I don't understand why this prop would need to be required.

Because to ensure header components are always provided with valid heading levels we need this prop to be required as explained here : https://engineering.cerner.com/terra-core/components/cerner-terra-core-docs/action-header/accessibility-guide#heading-level-order-and-arrangement-always

packages/terra-section-header/src/SectionHeader.jsx
```diff
@@ -55,6 +51,14 @@ class SectionHeader extends React.Component {
     this.state = { isActive: false };
     this.wrapOnKeyDown = this.wrapOnKeyDown.bind(this);
     this.wrapOnKeyUp = this.wrapOnKeyUp.bind(this);
+    this.setFocus = this.setFocus.bind(this);
+    this.focusableElementRef = React.createRef();
+  }
+
+  setFocus() {
```
supreethmr, line 58: @cm9361  instead of having focus handler written on component we can passback the ref of section-header `this.focusableElementRef` (similar to [refCallback](https://github.com/cerner/terra-core/blob/97be1c221e0f977646be81235a1597e6ef95dd45/packages/terra-button/src/Button.jsx#L84) of button) to consumer they can call the focus() or any required native events on application end. 

So that consumer can write ` sectionHeader.current.focus();` instead of `sectionHeader.current.setFocus();`

supreethmr:
> > Can we update the prop description of `level` prop. `level` prop is required prop from next MVB
> > ```js
> >   /**
> >    * IMPORTANT
> >    * Sets the heading level. One of `1`, `2`, `3`, `4`, `5`, `6`.
> >    * `level` should be specified explicitly to allow screen readers to identify headers consistently.
> >    */
> >   level: PropTypes.oneOf([1, 2, 3, 4, 5, 6]),
> > ```
> 
> @supreethmr we should log a Jira to do an audit and make all the MVB changes before release e.g. remove deprecation warnings etc.

I have gone through the MVB PR's which we had for header components. There were only two MVB changes that were done 1. Replacing `Title` Prop with `Text` prop .
2. Removal of Default value.
https://github.com/cerner/terra-core/pull/3637/files#diff-730a23e7b4aa5e890677524012b92eac20af22dd6bdf26203ac82387daed394c

packages/terra-section-header/src/SectionHeader.jsx
```diff
@@ -44,9 +44,10 @@ const defaultProps = {
   onClick: undefined,
   isOpen: false,
   isTransparent: false,
+  level: 2,
```
sdadn, line 47: Is setting the default level correct?

sdadn: nvm I see what you did here.

eawww (approved):

sdadn (approved):

PR #3966, opened 2023-11-07
[terra-form-select] Combobox and Search select Accessibility Issues

MadanKumarGovindaswamy:
### Summary

**What was changed:**

Accessibility and Functional issues for combobox and search select
1. Screen reader will read as combobox for Terra Combobox and search select component.
2. Combobox and search select should announce expanded or collapsed state.
3. When text is typed in search bar first ESC should close the dropdown and then pressing ESC again should clear out the text.

**Why it was changed:**

1. Terra Combobox component should have a role of combobox and it should be announced by the screen reader. 
2. When there is some user typed text in search bar and dropdown of available options is open, first ESC should close the dropdown and then pressing ESC again should clear out the text.
3. Combobox should announce expanded or collapsed state based on whether dropdown of available options is open or closed.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [x] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9798 

---

@cerner/terra


packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -536,8 +535,7 @@ class Frame extends React.Component {
    */
   role() {
     const { disabled } = this.props;
-    // role="application" needed to allow JAWS to work correctly with the select and use our key event listeners
```
supreethmr, line 539: Looks like the role `application` was added for JAWS. Does it have any impact on JAWS response.

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -536,8 +535,7 @@ class Frame extends React.Component {
...
-    return disabled ? undefined : 'application';
+    return disabled ? undefined : 'combobox';
```
supreethmr, line 538: If we are removing the role `application` and using `combobox` then we can use it has default role without need of a function

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -396,11 +402,13 @@ class Frame extends React.Component {
       'aria-describedby': ariaDescribedBy,
       'aria-disabled': disabled,
       'aria-owns': this.state.isOpen ? id : undefined,
+      'aria-expanded': !disabled && this.state.isOpen,
       type: 'text',
       className: cx('search-input'),
       required,
       'aria-required': required,
       id: inputId,
+      role: this.role(),
```
supreethmr, line 411: May I know the reason for adding role `combobox` for input ..?

MadanKumarGovindaswamy: Hi @supreethmr . 
Now combo-box component reads as application which we need to change to combobox role. And as per W3C example we updated this input field role to combo box.
W3C Example -- https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-both/
Thanks

packages/terra-form-select/src/combobox/Frame.jsx
MadanKumarGovindaswamy: Updated. Thanks.

packages/terra-form-select/src/combobox/Frame.jsx
MadanKumarGovindaswamy: Hi @supreethmr . We did not find any difference in JAWS after updating the role to combobox

supreethmr:
As per the https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-both/ on second esc key press even the selected value should get cleared. But it looks like in deployment selected values are not being cleared out on second esc key press.

https://github.com/cerner/terra-core/assets/12869809/8fe28e45-762d-43a1-a505-fe776dcef1a6

Instead of clearing it is being reset to previously selected value 

https://github.com/cerner/terra-core/assets/12869809/5e29be76-680a-448a-909c-921736e81701

PK106552:
> As per the https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-both/ on second esc key press even the selected value should get cleared. But it looks like in deployment selected values are not being cleared out on second esc key press.
> 
>  notclearingEsc.mov 
> Instead of clearing it is being reset to previously selected value
> 
>  insteadofclearingresetstopreviousselected.mov

Fixed

Uploading Screen Recording 2023-11-09 at 7.45.35 PM.mov…

PK106552:
> > As per the https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-both/ on second esc key press even the selected value should get cleared. But it looks like in deployment selected values are not being cleared out on second esc key press.
> > notclearingEsc.mov
> > Instead of clearing it is being reset to previously selected value
> > insteadofclearingresetstopreviousselected.mov
> 
> Fixed
> 

https://github.com/cerner/terra-core/assets/127083404/924a1847-51cc-419c-b8e2-0dc6c1a559ed

packages/terra-form-select/src/search/Frame.jsx
```diff
@@ -410,9 +417,14 @@ class Frame extends React.Component {
...
       isPositioned: false,
-      hasSearchChanged: false,
-      searchValue: '',
     });
+    if (!this.hasEscPressed) {
+      this.setState({
+        hasSearchChanged: false,
+        searchValue: '',
+      });
+      this.props.onSelect('');
```
supreethmr, line 426:
```suggestion
      this.props.onSelect('');
```
why  `onSelect` when we are clearing the value

MadanKumarGovindaswamy: This change is added to clear the selected value from the dropdown when we press the esc key. This onSelect handler clears the state binded to input field. Thanks

supreethmr: OnSelect()  is callback provided from consuming application I don't think we need to trigger this for clearing the value

MadanKumarGovindaswamy: Updated. Thanks
cb3824e9ed2ce3f1ed0e175cd02c7d61318c7629

packages/terra-form-select/src/combobox/Frame.jsx
```diff
@@ -420,21 +427,13 @@ class Frame extends React.Component {
       isFocused: document.activeElement === this.input || document.activeElement === this.select,
       isOpen: false,
       isPositioned: false,
-      hasSearchChanged: false,
-      searchValue: '',
     });
-
-    // 'Tag' and 'Combobox' variants select the current search value when the component loses focus.
-    const { searchValue } = this.state;
-    if (Frame.shouldAddOptionOnBlur(this.props, this.state)) {
```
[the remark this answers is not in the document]
MadanKumarGovindaswamy: This line actually adds new option to dropdown list. As soon as dropdown is closed the entered free text is getting added to the dropdown list. Now the entered free text will be added only on click event.
Thanks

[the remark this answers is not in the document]
supreethmr: That's fine but it clears out free text on press Tab key ( basically when input loses its focus ) which should not happen !! free-text should stay on input 

Before this PR :

https://github.com/cerner/terra-core/assets/12869809/b5d0a684-d796-4e84-bce8-c00158f29294

After the code changes made on this PR : 

https://github.com/cerner/terra-core/assets/12869809/6bb37731-ea20-437c-95ed-e6e20376d1c5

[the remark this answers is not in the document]
MadanKumarGovindaswamy: Updated. Thanks
43d5927ba60a5d62f3db7753422d66012c6e20f4

packages/terra-form-select/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Changed
+  * Change role from application to combobox in `terra-form-select-combobox`.
+  * Ability to close dropdown and clear text by ESC key in `terra-form-select-combobox`.
```
supreethmr, line 7:
```suggestion
* Changed
  * Changed role from application to combobox in `terra-form-select-combobox`.
  * Updated keydown handler to clear the text on `ESC` key press in `terra-form-select-combobox`.
```

PK106552: Updated - https://github.com/cerner/terra-core/pull/3966/commits/ba254493cd46d9da320dfc4c84c42c892250c24e

supreethmr (approved):

sugan2416 (approved):

rahulkumar8cs:
+1 for the issue mentioned  in PR.
Observation :- For add text, SR is not reading in top to bottom format. Currently the SR is reading the the add text followed by no results message. Also for the add text, user is not aware what will happen when we click on add text, so appropriate instruction as per its name, role and value needs to be announced.

PR #3978, opened 2023-11-17
[Terra-Tag] SR announces instructions twice with VPC ON Mode 

gt106551:
### Summary
Fixed-SR announces instructions twice with VPC ON Mode

**What was changed:**
Removed the `aria-live` property to prevent JAWS (screen reader) from announcing twice in the Edge browser.
**Why it was changed:**
When VPC  is in ON mode, screen reader announces instruction "11 items are hidden" twice .

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

[UXPLATFORM-9812](https://jira2.cerner.com/browse/UXPLATFORM-9812)
---

@cerner/terra

### DIFF
--- a/packages/terra-tag/src/_RollupTag.jsx
+++ b/packages/terra-tag/src/_RollupTag.jsx
@@ -74,7 +74,7 @@ const RollUpTag = (props) => {
       onBlur={handleOnBlur}
       refCallback={(ref) => { rollUpTagRef.current = ref; }}
       isCompact
-      aria-live={(isCollapsed) ? 'polite' : 'off'}
+      aria-live={(navigator.userAgent.indexOf('Edg') === -1 && isCollapsed) ? 'polite' : 'off'}
       aria-label={rollUpHint}
       aria-describedby={rollUpHint}
       aria-atomic="true"
--- a/packages/terra-tag/src/_RollupTag.jsx
+++ b/packages/terra-tag/src/_RollupTag.jsx
@@ -75,7 +75,6 @@ const RollUpTag = (props) => {
       refCallback={(ref) => { rollUpTagRef.current = ref; }}
       isCompact
       aria-label={rollUpHint}
-      aria-describedby={rollUpHint}
       aria-atomic="true"
       data-terra-rollup-tag-show-focus-styles
       data-terra-rollup-tag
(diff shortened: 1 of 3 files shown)


packages/terra-tag/src/_RollupTag.jsx
```diff
@@ -74,7 +74,7 @@ const RollUpTag = (props) => {
       onBlur={handleOnBlur}
       refCallback={(ref) => { rollUpTagRef.current = ref; }}
       isCompact
-      aria-live={(isCollapsed) ? 'polite' : 'off'}
+      aria-live={(navigator.userAgent.indexOf('Edg') === -1 && isCollapsed && !/^(?=.*\bSafari\b)(?!.*\bChrome\b).*/i.test(navigator.userAgent)) ? 'polite' : 'off'}
```
sugan2416, line 77: aria-live is now used only when its either safari or chrome ? Can the expression be simplified ?

supreethmr: @gt106551  What does this expression validate to ..??

gt106551: I have tried to simplify in different ways but getting issues in safari browser

gt106551: > @gt106551 What does this expression validate to ..??

it checks if the user agent string contains 'Safari' but does not contain 'Chrome'.

sugan2416: If you need only safari, refer : https://github.com/cerner/terra-core/blob/8220010acd80830e9bd7368ebbf7bc694d731e8a/packages/terra-form-select/src/shared/_SharedUtil.js#L7
Include edge as well in this

supreethmr: I think you're trying to set aria-live to `polite` for edge and JAWS. In this case you can refer to some of the existing regEx for edge browser instead of adding new expression from web.

gt106551: I Have removed  aria-live attribute and tested in all browsers it is working as expected.

supreethmr (approved):

sugan2416 (approved):

PR #4015, opened 2024-01-16
Update About.1.doc.mdx for Terra Status

mjpalazzo:
Added deprecation notice and badge to Terra Status About page

### Summary

**What was changed:**
Added deprecation notice and badge to Terra Status About page

**Why it was changed:**

Component is no longer supported and will not be included in Redwood theme project

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review
- [ ] Tech writer review

### Additional Details

**This PR resolves:**

UXPLATFORM-9930 

---

@cerner/terra

### DIFF
--- a/packages/terra-core-docs/src/terra-dev-site/doc/status/About.1.doc.mdx
+++ b/packages/terra-core-docs/src/terra-dev-site/doc/status/About.1.doc.mdx
@@ -16,7 +16,7 @@ import StatusPropsTable from 'terra-status/lib/Status?dev-site-props-table';
 
 <Notice variant="deprecation">
 
-`terra-status` does not comply with accessibility guidelines and is no longer supported. If you are using `terra-status`, consider replacing it with `terra-icon` or `terra-image`.
+`terra-status` does not comply with accessibility guidelines and is no longer supported. If you are using `terra-status`, consider replacing it with descriptive text or an icon.
 
 </Notice>
 
--- a/packages/terra-core-docs/src/terra-dev-site/doc/status/About.1.doc.mdx
+++ b/packages/terra-core-docs/src/terra-dev-site/doc/status/About.1.doc.mdx
@@ -1,3 +1,4 @@
+import { Notice } from "@cerner/terra-docs";
 import { Badge } from 'terra-status/package.json?dev-site-package';
 
 import StatusDefault from './example/StatusDefault?dev-site-example';
@@ -9,7 +10,15 @@ import StatusPropsTable from 'terra-status/lib/Status?dev-site-props-table';
 
 <Badge />
 
+[![Deprecated](https://badgen.net/badge/status/Deprecated/grey)](https://github.com/cerner/terra-core/tree/main/packages/terra-status#readme)
+
 # Terra Status
+
+<Notice variant="deprecation">
+
+`terra-status` is no longer supported. If you are using `terra-status`, consider replacing it with a supported component.
+
+</Notice>
 
 The status component provides a customizable color indictor to signify a specific condition.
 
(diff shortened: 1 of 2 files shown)


sycombs (approved):

JessieRandle:
What component are we recommending consumers use in place of terra-status.  If we don't have an alternative, why is terra-status being deprecated?

packages/terra-core-docs/src/terra-dev-site/doc/status/About.1.doc.mdx
```diff
@@ -9,8 +10,16 @@ import StatusPropsTable from 'terra-status/lib/Status?dev-site-props-table';
 
 <Badge />
 
+[![Deprecated](https://badgen.net/badge/status/Deprecated/grey)](https://github.com/cerner/terra-core/tree/main/packages/terra-status#readme)
+
 # Terra Status
 
+<Notice variant="deprecation">
+
+`terra-status` is no longer supported. If you are using `terra-status`, consider replacing it with a supported component.
```
sdadn, line 19: What is the supported alternative? We should probably link it here.

RachaelMarso: I agree with adding a supported component.

RachaelMarso (approved):
Reviewed and approved per OHAIUAD

packages/terra-core-docs/src/terra-dev-site/doc/status/About.1.doc.mdx
mjpalazzo: Thanks for the feedback. This component does not meet accessibility criteria, and I am not aware of any alternative that we can point to. @eawww - do you have any suggestions for replacing terra-status? Is there an alternative or does the consumer need to work with UX on a different approach?

eawww: @mjpalazzo @sdadn yeah this is an unfortunately common pattern in Cerner stuff that needs to go away. The alternative will largely depend on context but a nice plug-n-play replacement could often be iconography where meaning is conveyed by shape (and color if appropriate) and accessible alt-text.

I'd say it would usually be best to work with UX on a per-use-case basis.

mjpalazzo: Thanks @eawww  So would terra-icon and terra-image be appropriate alternatives? These components have accessibility guides which provide guidance about alt-text, etc.

mjpalazzo: Updated notice with reason, terra status does not comply with accessibility guidelines and alternatives.

packages/terra-core-docs/src/terra-dev-site/doc/status/About.1.doc.mdx
```diff
@@ -9,8 +10,16 @@ import StatusPropsTable from 'terra-status/lib/Status?dev-site-props-table';
 
 <Badge />
 
+[![Deprecated](https://badgen.net/badge/status/Deprecated/grey)](https://github.com/cerner/terra-core/tree/main/packages/terra-status#readme)
+
 # Terra Status
 
+<Notice variant="deprecation">
+
+`terra-status` does not comply with accessibility guidelines and is no longer supported. If you are using `terra-status`, consider replacing it with `terra-icon` or `terra-image`.
```
cm9361, line 19: You would not replace status with a terra-icon or terra-image.  That would be the child of the component.  Consumers would have to create their own div to replace the terra-status component.

adoroshk (approved):
Needs Changelog update, otherwise LGTM. 
Screenshot 2024-01-22 at 4 56 24 PM

JessieRandle (approved):

sdadn (approved):

eawww (approved):

adoroshk (approved):

PR #4022, opened 2024-02-06
[terra-button-group] Updated tabIndex on button group for single select example 

KV106606Viswanath:
### Summary
Previously, we are unable to tab (using tab key) between buttons on the button group in single select example. With this fix it has been resolved and we can move focus to the button even it is not selected on single select example. 

**What was changed:**
Updated tabIndex on button group for single select example. 

**Why it was changed:**
User unable to tab between buttons on button group.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10171

---

@cerner/terra


saket2403:
Keyboard navigation for single select buttons should be similar to radio button fields ie: Tab should put focus on the first button and subsequent tabbbing should move the focus out of the group to the next focusable element. Only arrow keys should work inside the group for single select buttons.Removing the entire tabIndex logic means that functioanlity is breaking.

KV106606Viswanath:
> Keyboard navigation for single select buttons should be similar to radio button fields ie: Tab should put focus on the first button and subsequent tabbbing should move the focus out of the group to the next focusable element. Only arrow keys should work inside the group for single select buttons.Removing the entire tabIndex logic means that functioanlity is breaking.

Updated.

saket2403:
Focus should come onto the selected item (when present) initially but as per these changes, it will always come onto the first button.

KV106606Viswanath:
> Focus should come onto the selected item (when present) initially but as per these changes, it will always come onto the first button.

Updated

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -170,7 +170,10 @@ class ButtonGroup extends React.Component {
 
     React.Children.forEach(children, (child, index) => {
       const isSelected = selectedKeys.indexOf(child.key) > -1;
-      const tabAttr = (this.props.onChange && !isMultiSelect) ? { tabIndex: (isSelected) ? '0' : '-1' } : null;
+      // Checking whehter any option is selected or not
+      const isAnyOptionSelected = children && Array.isArray(children) ? children.findIndex((e) => this.props.selectedKeys.includes(e?.key)) : -1;
```
saket2403, line 174:
```suggestion
      const isAnyOptionSelected = children && Array.isArray(children) ? children.findIndex((child) => this.props.selectedKeys.includes(child?.key)) : -1;
```

saket2403:
Please add a wdio test to verify that focus moves to the first button when no buttons are selected initially in a single select button group.

KV106606Viswanath:
> Please add a wdio test to verify that focus moves to the first button when no buttons are selected initially in a single select button group.

Added.

packages/terra-button-group/src/ButtonGroup.jsx
KV106606Viswanath: Updated

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -170,7 +170,10 @@ class ButtonGroup extends React.Component {
 
     React.Children.forEach(children, (child, index) => {
       const isSelected = selectedKeys.indexOf(child.key) > -1;
-      const tabAttr = (this.props.onChange && !isMultiSelect) ? { tabIndex: (isSelected) ? '0' : '-1' } : null;
+      // Checking whehter any option is selected or not
```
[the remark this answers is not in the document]
KV106606Viswanath: Updated

packages/terra-button-group/src/ButtonGroup.jsx
```diff
@@ -170,7 +170,11 @@ class ButtonGroup extends React.Component {
 
     React.Children.forEach(children, (child, index) => {
       const isSelected = selectedKeys.indexOf(child.key) > -1;
-      const tabAttr = (this.props.onChange && !isMultiSelect) ? { tabIndex: (isSelected) ? '0' : '-1' } : null;
+      // Checking whether any option is selected or not
+      // eslint-disable-next-line no-shadow
```
saket2403, line 174: Why is this eslint rule disabled?

saket2403 (approved):

packages/terra-button-group/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+* Removed
+  * Removed `tabAttr` variable to fix tabbing issue on the button group.
```
sugan2416, line 6:
```suggestion
* Fixed
  * Fixed tabbing issue for button-group without initial focus.
```

packages/terra-core-docs/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Added
+  * Added a test example to validate focus when none of the buttons are selected.
```
sugan2416, line 6:
```suggestion
  * Added a test example for `terra-button-group` to validate focus when none of the buttons are selected.
```

KV106606Viswanath: Updated

packages/terra-button-group/CHANGELOG.md
KV106606Viswanath: Updated

sugan2416 (approved):

rahulkumar8cs:
+1

PR #4024, opened 2024-02-07
Tag select fixes

PK106552:
### Summary

**What was changed:**

1. Added error message and instruction text to validate invalid example.
2. Screen reader will read as combobox for Terra tag select.
3. When text is typed in search bar first ESC should close the dropdown and then pressing ESC again should clear out the text.
4. Added placeholder text to tag select aria-label.

**Why it was changed:**

1. There was no error message for invalid example.
2. Terra Combobox component should have a role of combobox and it should be announced by the screen reader.
3. When there is some user typed text in search bar and dropdown of available options is open, first ESC should close the dropdown and then pressing ESC again should clear out the text.
4. Placeholder text was not included to tag select aria-label.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-9715 

---

@cerner/terra


packages/terra-form-select/src/shared/_MenuUtil.js
```diff
@@ -331,6 +331,8 @@ class MenuUtil {
     });
   }
 
+  static isMac () { return navigator.userAgent.indexOf('Mac') !== -1 && navigator.userAgent.indexOf('Win') === -1 };
```
sdadn, line 334: Is the second condition needed? I would assume that if the OS is Mac then it's automatically not windows.

sdadn (approved):

PR #4043, opened 2024-02-25
[terra-section-header] Fixes header title width for sticky header

sugan2416:
### Summary

**What was changed:**

Sets width of header title element when section headers are bounded by width and have sticky header. 

**Why it was changed:**

This fixes issue when auto width was set to collapsible section headers causing the onclick action to fail due to insufficient size of clickable area of section-header.
Sticky headers which are bounded have width set to bounded width, else set to auto.
Non-sticky headers have width as 100%

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10215 

---

@cerner/terra

### DIFF
--- a/packages/terra-section-header/src/SectionHeader.jsx
+++ b/packages/terra-section-header/src/SectionHeader.jsx
@@ -154,15 +154,15 @@ class SectionHeader extends React.Component {
       buttonAttributes.onKeyDown = this.wrapOnKeyDown(headerAttributes.onKeyDown);
       buttonAttributes.onKeyUp = this.wrapOnKeyUp(headerAttributes.onKeyUp);
       buttonAttributes.onClick = onClick;
-      if (headerAttributes.isTitleSticky && headerAttributes.boundedWidth) {
-        buttonAttributes.style = { width: `${headerAttributes.boundedWidth}px` };
+      if (customProps.isTitleSticky && customProps.boundedWidth) {
+        buttonAttributes.style = { width: `${customProps.boundedWidth}px` };
       }
 
       // Specify button element for header content
       ArrangeWrapper = 'button';
     } else {
       headerAttributes.ref = refCallback;
-      if (headerAttributes.isTitleSticky) {
+      if (customProps.isTitleSticky) {
         buttonAttributes.style = { width: 'auto' };
       }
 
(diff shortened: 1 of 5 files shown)


packages/terra-section-header/src/SectionHeader.module.scss
```diff
@@ -161,7 +161,6 @@
     &.title-sticky {
       position: sticky;
       left: 12px;
-      width: auto;
```
ShettyAkarsh, line 164: why are we removing this here ? are we expecting all the consumers of section-header to pass isTitleSticky as customprops ?

packages/terra-section-header/src/SectionHeader.jsx
```diff
@@ -154,19 +154,28 @@ class SectionHeader extends React.Component {
       buttonAttributes.onKeyDown = this.wrapOnKeyDown(headerAttributes.onKeyDown);
       buttonAttributes.onKeyUp = this.wrapOnKeyUp(headerAttributes.onKeyUp);
       buttonAttributes.onClick = onClick;
+      if (customProps.isTitleSticky && customProps.boundedWidth) {
```
ShettyAkarsh, line 157: is it mandator to pass boundedwidth along with isTitleSticky prop always ?

packages/terra-section-header/src/SectionHeader.module.scss
sugan2416: **Bounded grids**

- When title is sticky and is a non-actionable header, we set to auto width: https://github.com/cerner/terra-core/pull/4043/files#diff-9bb4e4b8bfb491074cf911831bc7c9ec606b84abb69895cfe86dc2c758e8e270R166

- When title is sticky and actionable, we use boundedWidth.

**Non-bounded grids** 
We set width to 100%

isTitleSticky is passed only if grid is bounded

saket2403:
Can you check the focus borders, they are getting cut off in the test you added.

sugan2416:
> Can you check the focus borders, they are getting cut off in the test you added.

This is expected behaviour since header is under horizontal scroll

saket2403 (approved):

sugan2416:
@eawww @ShettyAkarsh @saket2403  UX suggestions (if any) will be addressed in a separate PR.

PR #4068, opened 2024-03-20
Update wdio config

sdadn:
### Summary

This PR simplifies the wdio config and does so miscellaneous code cleanup.

### Testing

This change was tested using:

- [x] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

---

@cerner/terra


wdio.conf.js
```diff
@@ -1,12 +1,6 @@
-const path = require('path');
 // eslint-disable-next-line import/no-extraneous-dependencies
-const defaultWdioConfig = require('@cerner/terra-functional-testing/lib/config/wdio.conf');
+const { config } = require('@cerner/terra-functional-testing');
 
-const wdioConfig = defaultWdioConfig.config;
+config.specs = ['./packages/**/wdio/**/*-spec.js'];
```
kenk2, line 4: Will this break on windows? We're not using `path.join` anymore to normalize between POSIX vs non-POSIX.

Do we also still need to strip out the `@cerner` stuff from package names?

sdadn: All the other scripts are also unix based and we use devcontainers for windows development anyway so this shouldn't affect it. 

The `@cerner/` stripping is to convert the package name to the folder name, which I am now doing directly using a glob.

cm9361, line 4: Do we no longer need "tests" in the path?  It appears to be there in the wdioConfig.specs path on line #9

kenk2 (approved):

wdio.conf.js
cm9361: It would still work because the packages/** would cover any directory structure in between.  It just would not limit to the specific path that was in place before.

cm9361 (approved):

PR #4069, opened 2024-03-20
[terra-form-select] Update focus indicators for Fusion theme

cm9361:
### Summary

**What was changed:**
The focus indicators were updated for the terra-form-select controls for the orion-fusion-theme.

**Why it was changed:**
The change was made so that the focus indicators would be consistent with others in the theme.

### Testing

This change was tested using:

- [X] WDIO
- [ ] Jest
- [ ] No tests are needed

### Reviews

- [X] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10318 

---

@cerner/terra


artpark (approved):

sdadn (approved):

nikhitasharma (approved):

sycombs (approved):

eawww:
It doesn't look like the new focus style was applied to the invalid variant:
image

Noting that the big fuzzy focus indicator looks a tad strange while navigating the items but the borderless light blue background that would otherwise be used would be much further from meeting contrast requirements.
image

cm9361:
@eawww - I fixed the invalid select control.  Is there any change you would like me to make on the selection focus indicator?

eawww:
Looking at this again, I don't think the giant glow on the secondary indicator is going to go over well. While I think that indicator _just_ had the light blue background before,
terra select component with options expanded and a faint blue background indicating the focused option
that has a 1.1:1 contrast ratio which will actually disappear on some displays regardless of vision ability. So I'm not comfortable signing off on that.

After chatting with a few of my UX compatriots, I think just removing the outer glow from the indicator on the options will have to do for now. `box-shadow: rgba(76, 178, 233, 0.5) 0px 0px 1px 3px inset;`
terra select component with options expanded and a faint blue background with a slightly deeper blue inset box shadow indicating the focused option

eawww (approved):

PR #4072, opened 2024-03-21
[terra-hyperlink] Hyperlink ellipses

kenk2:
### Summary

**What was changed:**
A hyperlink will now truncate with an ellipses if its content should overflow.

**Why it was changed:**
Feature request from a consuming team

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10303 

---

@cerner/terra


packages/terra-core-docs/src/terra-dev-site/test/hyperlink/Test.module.scss
```diff
@@ -0,0 +1,8 @@
+:local {
+    .truncated {
+        width: 150px;
+        height: 50px;
+        border-style: dashed;
+        margin-bottom: 10px;
+    }
+  }
```
sdadn, line 8: If this is only used for a single file, can we use inline styles for this? e.g. 
https://css-tricks.com/different-ways-to-write-css-in-react/#aa-write-inline-styles

packages/terra-hyperlink/src/Hyperlink.module.scss
```diff
@@ -8,11 +8,14 @@
     background-color: transparent; // Remove the gray background on active links in IE 10
     color: var(--terra-hyperlink-color, #006fc3);
     cursor: pointer;
-    display: inline-block; // prevents variant icon from breaking onto its own line
     outline: none;
     text-decoration: var(--terra-hyperlink-text-decoration, underline);
     touch-action: manipulation; // Enable fast tap interaction in webkit browsers; see https://webkit.org/blog/5610/more-responsive-tapping-on-ios/
     vertical-align: baseline;
+    display: inline-flex;
```
cm9361, line 15: Why are we changing the style from inline-block to inline-flex?  I would not expect this change to be necessary.

packages/terra-hyperlink/src/Hyperlink.jsx
```diff
@@ -245,7 +245,9 @@ class Hyperlink extends React.Component {
           ref={this.linkRef}
         >
           <span className={cx('button-inner')}>
-            {text || children}
+            <span className={cx('inner-text')}>
```
cm9361, line 248: Shouldn't we be able to accomplish this with one span?

kenk2: As per the requirements, we need to make only the text have an ellipsis while the icon stays at the same level. So we need to use 2 spans for this. Otherwise it will start wrapping around if we try to implement the ellipses without having an additional inner span.

packages/terra-hyperlink/src/Hyperlink.module.scss
kenk2: Since we're needing to have 2 spans with 1 being an ellipses but another being on the same level, we'll need to go with this.

packages/terra-core-docs/src/terra-dev-site/test/hyperlink/Test.module.scss
kenk2: Done.

packages/terra-hyperlink/src/Hyperlink.module.scss
BenBoersma: Also inline-block wasn't allowing us to properly truncate without the hyperlink "clickable" area extending past the "visual" area of the hyperlink. Using inline-flex, we were able to properly have the "text" portion of hyperlink resize itself so it would truncate without issue while allowing the icon to always render at full size.

sdadn (approved):

adoroshk (approved):
LGTMM, works as expected.

PR #4089, opened 2024-04-17
[terra-form-select] Fixed focus issue in MultiSelect

adavijit:
### Summary

**What was changed:**
- Added `aria-label` attribute in `Tag` component
- Modified `role` attribute logic depending on the `isInputFocused` prop.
- Consumed  `inputRef` as callback in `MultiSelect` component and passed down to `Tag` component.
- Consumed `disabled` and `isInputFocused` props in `Tag` component
- Added test for testing disabled `MultiSelect` component
- Updated wdio tests
- Updated translations files

**Why it was changed:**
Keyboard TAB focus also not moving to the added list of options. The User is not able to access the previous selected options that appears above the multi-select field.

### Testing

This change was tested using:

- [x] WDIO
- [x] Jest
- [ ] No tests are needed

### Reviews

- [ ] UX review
- [x] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10191 

---

@cerner/terra

### DIFF
--- a/packages/terra-alert/tests/wdio/alert-spec.js
+++ b/packages/terra-alert/tests/wdio/alert-spec.js
@@ -65,7 +65,7 @@ Terra.describeViewports('Alert', ['tiny', 'large'], () => {
     it('alert content is focused when rendered with an action element', () => {
       browser.url('/raw/tests/cerner-terra-core-docs/alert/custom-prop-alert');
 
-      browser.keys(['Tab', 'Tab', 'Tab', 'Tab', 'Enter']);
+      browser.keys(['Tab', 'Tab', 'Tab', 'Tab', 'Tab', 'Tab', 'Enter']);
 
       Terra.validates.element('alert focused');
     });
--- a/packages/terra-form-select/src/shared/_Tag.jsx
+++ b/packages/terra-form-select/src/shared/_Tag.jsx
@@ -38,12 +38,12 @@ const Tag = ({
   const theme = React.useContext(ThemeContext);
   const tagRef = useRef(null);
 
-  const handleEnterKeyPress = (event) => {
+  const handleKeyPress = (event) => {
     if ((event.key === 'Enter' || event.key === 'Backspace') && !disabled) {
       event.stopPropagation();
       onDeselect(value);
       const previousLi = tagRef.current.previousElementSibling;
-
+      const selectInput = tagRef.current.closest('ul').parentElement.parentElement.children[1].children[0];
       if (previousLi) {
         const deselectElement = previousLi.children[1];
         if (deselectElement) {
@@ -51,16 +51,15 @@ const Tag = ({
         }
       } else {
         const nextLi = tagRef.current.nextElementSibling;
-        const parentUl = tagRef.current.closest('ul');
         if (nextLi) {
           const deselectElement = nextLi.children[1];
           if (deselectElement) {
             deselectElement.focus();
           } else {
-            parentUl.click();
+            selectInput.focus();
           }
         } else {
-          parentUl.click();
+          selectInput.focus();
         }
       }
     }
@@ -72,7 +71,7 @@ const Tag = ({
       </span>
       <span
         id={`terra-tag-deselect-${value}`}
-        onKeyDown={handleEnterKeyPress}
+        onKeyDown={handleKeyPress}
         className={cx('deselect')}
         onClick={() => { if (!disabled) onDeselect(value); }}
         tabIndex={!disabled ? 0 : -1}
--- a/packages/terra-form-select/src/shared/_Tag.jsx
+++ b/packages/terra-form-select/src/shared/_Tag.jsx
@@ -76,8 +76,8 @@ const Tag = ({
         className={cx('deselect')}
         onClick={() => { if (!disabled) onDeselect(value); }}
         tabIndex={!disabled ? 0 : -1}
-        role={!disabled ? 'button' : null}
-        aria-label={!disabled ? intl.formatMessage({ id: 'Terra.form.select.deselect' }, { text: children }) : null}
+        role="button"
+        aria-label={intl.formatMessage({ id: 'Terra.form.select.deselect' }, { text: children })}
       >
         <span className={cx('icon')} />
       </span>
--- a/packages/terra-form-select/src/shared/_Tag.jsx
+++ b/packages/terra-form-select/src/shared/_Tag.jsx
@@ -82,7 +82,7 @@ const Tag = ({
         onKeyDown={handleKeyPress}
         className={cx('deselect')}
         onClick={() => { if (!disabled) onDeselect(value); }}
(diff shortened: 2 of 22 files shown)


packages/terra-form-select/src/shared/_Tag.jsx
```diff
@@ -19,17 +20,64 @@ const propTypes = {
...
+}) => {
   const theme = React.useContext(ThemeContext);
+  const tagRef = useRef(null);
+
+  const handleKeyPress = (event) => {
+    if ((event.key === 'Enter' || event.key === 'Backspace') && !disabled) {
+      event.stopPropagation();
+      onDeselect(value);
+      const previousLi = tagRef.current.previousElementSibling;
+      const selectInput = tagRef.current.closest('ul').parentElement.parentElement.children[1].children[0];
```
supreethmr, line 46: This way of accessing input control might result in unexpected failure when dom changes are made to component. I would suggest you to create a callback ref for input to set focus.

packages/terra-form-select/src/shared/_Tag.jsx
```diff
@@ -19,17 +20,64 @@ const propTypes = {
...
+        if (nextLi) {
+          const deselectElement = nextLi.children[1];
+          if (deselectElement) {
+            deselectElement.focus();
+          } else {
+            selectInput.focus();
+          }
+        } else {
+          selectInput.focus();
+        }
```
supreethmr, line 63: Nit: 
```suggestion
        if (nextLi && nextLi.children[1]) {
            nextLi.children[1].focus();
        } else {
          selectInput.focus();
        }
```

packages/terra-form-select/src/shared/_Tag.jsx
```diff
@@ -19,17 +20,64 @@ const propTypes = {
...
+  const tagRef = useRef(null);
+
+  const handleKeyPress = (event) => {
+    if ((event.key === 'Enter' || event.key === 'Backspace') && !disabled) {
+      event.stopPropagation();
+      onDeselect(value);
+      const previousLi = tagRef.current.previousElementSibling;
+      const selectInput = tagRef.current.closest('ul').parentElement.parentElement.children[1].children[0];
+      if (previousLi) {
+        const deselectElement = previousLi.children[1];
```
supreethmr, line 48: why is the index being hard-coded to 1 here ?

packages/terra-form-select/translations/de.json
```diff
@@ -27,5 +27,6 @@
   "Terra.form.select.option": "Optionen",
   "Terra.form.select.optGroup": "Gruppe {text}",
   "Terra.form.select.defaultComboboxDisplay": "Auswählen oder eingeben",
-  "Terra.form.select.resultsText":	"Ergebnisse mit '{text}'"
+  "Terra.form.select.resultsText":	"Ergebnisse mit '{text}'",
```
supreethmr, line 30:
```suggestion
  "Terra.form.select.resultsText": "Ergebnisse mit '{text}'",
```

packages/terra-form-select/translations/es.json
```diff
@@ -27,5 +27,6 @@
   "Terra.form.select.option": "Opciones",
   "Terra.form.select.optGroup": "Grupo {text}",
   "Terra.form.select.defaultComboboxDisplay":	"Seleccionar o escribir",
-  "Terra.form.select.resultsText":"Resultados que contienen \"{text}\""
+  "Terra.form.select.resultsText":"Resultados que contienen \"{text}\"",
```
supreethmr, line 30:
```suggestion
  "Terra.form.select.resultsText": "Resultados que contienen \"{text}\"",
```

packages/terra-form-select/tests/wdio/select-spec.js
```diff
@@ -2376,6 +2376,17 @@ Terra.describeViewports('Select', ['tiny'], () => {
       it('should display selected option', () => {
         Terra.validates.element('tag controlled selected option');
       });
+
+      it('should focus on pressing tab key', () => {
```
supreethmr, line 2380: Does this test validates disabled tag navigations as well?

packages/terra-alert/tests/wdio/alert-spec.js
```diff
@@ -65,7 +65,7 @@ Terra.describeViewports('Alert', ['tiny', 'large'], () => {
     it('alert content is focused when rendered with an action element', () => {
       browser.url('/raw/tests/cerner-terra-core-docs/alert/custom-prop-alert');
 
-      browser.keys(['Tab', 'Tab', 'Tab', 'Tab', 'Enter']);
+      browser.keys(['Tab', 'Tab', 'Tab', 'Tab', 'Tab', 'Tab', 'Enter']);
```
supreethmr, line 68: why are the additional tab key is required to bring focus to alert content..?

adavijit: @supreethmr Earlier pressing tab key was not focusing inside the input field, that is the deselect option. Now since there are two options in the list, to get out side of the input box we need to press `tab` two more times

packages/terra-form-select/tests/wdio/select-spec.js
adavijit: @supreethmr I have added tests to validate disabled tag navigations

adavijit:
> +1 Keyboard focus behaviour of multi-select is working as expected. However there are some screen reader issues still existing.

@SwethaM03 , Screen Reader issue is now fixed with this PR

packages/terra-form-select/src/shared/_Tag.jsx
adavijit: I have now used `inputRef` from a callback to set focus on the input element.

packages/terra-form-select/src/shared/_Tag.jsx
adavijit: I have removed the hard coded index. focused input element using `inputRef`.

supreethmr (approved):
+1 with axe error fix

adavijit:
> There is a Axe Violation on disabled multi-select I think this will be same for other variants of select also. disabled frame does not get any value for role attribute :
> 
> https://github.com/cerner/terra-core/blob/778bfbc21edb4212e5d692d51d2f8f12ede5d267/packages/terra-form-select/src/multiple/Frame.jsx#L562
> 
> 
> Hence it's resulting in below axe voilation. It would be better to remove the `aria-expanded` attribute for the disabled frame since it has no role. OR we can add role back but it looks like `role='application' is not consistent with screen readers for disabled elements.


@supreethmr  Since this is not related to this PR, logged a new JIRA here.
[UXPLATFORM-10391](https://jira2.cerner.com/browse/UXPLATFORM-10391)

rbsree:
+1, both keyboard and screen reader focus is managed as expected. And the focus is also managed when options are deselected.

sugan2416 (approved):

PR #4091, opened 2024-04-22
[terra-arrange] Revert change for small viewport flex css properties

artpark:
### Summary

**What was changed:**
Removed the flex property that was only available during a viewport smaller than 320px x 256px. After testing more thoroughly, I found that these did not have the intended behavior and was causing issues in IE.

**Why it was changed:**
To fix the IE overlapping elements bug.

### Testing

This change was tested using:

- [ ] WDIO
- [ ] Jest
- [x] No tests are needed

### Reviews

- [ ] UX review
- [ ] Accessibility review
- [ ] Functional review

### Additional Details

**This PR resolves:**

UXPLATFORM-10305 

---

@cerner/terra


packages/terra-arrange/src/Arrange.module.scss
```diff
@@ -42,12 +38,4 @@
   .default {
     align-self: flex-start;
   }
-
-  @media (max-width: 320px), (max-height: 256px) {
```
supreethmr, line 46: Is Mpage Fusion components that consumes terra-arrange are tested with these changes ? Can add the screenshots on Jira as well before merging this pr

JessieRandle (approved):

adoroshk (approved):

sycombs (approved):

packages/terra-arrange/src/Arrange.module.scss
artpark: @supreethmr Yup, the MPages Fusion component, ContentDecorator, has been tested with these changes. Here is the before and after when the window is resized to 320 x 256px.

Before:
Screenshot 2024-04-23 at 4 49 49 PM

After:
Screenshot 2024-04-23 at 4 51 51 PM

Interestingly, the `flex` property doesn't do anything in the example. The parent div has a style of `display: block`. So `flex` has no effect on it.

If the before change's styling of `flex: 0 0` were actually to be applied, it would look like this:
Screenshot 2024-04-23 at 4 48 15 PM

PR #4094, opened 2024-04-25
Release for April 25th, 2024

artpark:
The following packages will be released:

  - terra-arrange@3.56.1
  - @cerner/terra-core-docs@1.73.0
  - terra-heading@4.58.0
  - terra-section-header@2.70.0
  - terra-status@4.66.0
  - terra-text@4.59.0
  - terra-toggle-section-header@2.81.0

This is the official release for terra-arrange@3.56.1 after the alpha version, 3.56.1-alpha.1, was confirmed to fix the IE reflow issue.


artpark:
No changelog update needed as there is no code change from the previous release

sycombs:
Shouldn't the packages that have terra-arrange as a dependency also be updated in this PR? for example terra-core-docs should have the version updated, shouldn't it? https://github.com/cerner/terra-core/blob/release_04_25_2024/packages/terra-core-docs/package.json#L42

adoroshk (approved):

artpark:
> Shouldn't the packages that have terra-arrange as a dependency also be updated in this PR? for example terra-core-docs should have the version updated, shouldn't it? https://github.com/cerner/terra-core/blob/release_04_25_2024/packages/terra-core-docs/package.json#L42
> 
> Or does it not matter because nothing has changed between those two releases?

Oh right, my bad I completely forgot

sycombs (approved):

packages/terra-arrange/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Unreleased
 
+## 3.56.1 - (April 25, 2024)
+
+* Changed
+  * Minor dependency version bump.
```
sdadn, line 8: Can we change this to something like
```suggestion
  * Stable release for v3.56.1.
```

sdadn (approved):

